PR #11, opened 2020-01-28
adding black linter to commit pushes

Bento007:


maniarathi:
Oops.. Matt and Bruce actually decided that we should use Black instead. https://github.com/chanzuckerberg/dcp-prototype/issues/10

PR #29, opened 2020-01-30
Adding Makefile to run linter and formatter

Bento007:
This is for running the linter and formatter locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL:=/bin/bash
 
-black:
+beautify:
 	black .
 
 lint:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,7 @@
+SHELL:=/bin/bash
+
+black:
```
maniarathi: Nit: could rename to something like `format` which more clearly conveys the effect of the command?

Bento007: not to be confused with formatting your system :0. I'll think of a generic name

Bento007: Also a "nit" shouldn't block a PR.

PR #32, opened 2020-01-31
initial terraform setup and rds setup

parthshahva:


infra/Makefile
```diff
@@ -0,0 +1,10 @@
+default: plan
+
+init:
+	cd envs/$(DEPLOYMENT_STAGE) && make init
```
Bento007: Nit: I like this syntax more.
```suggestion
	make -C envs/$(DEPLOYMENT_STAGE) init
```

infra/Makefile
```diff
@@ -0,0 +1,10 @@
+default: plan
```
Bento007: why not just reuse this https://github.com/HumanCellAtlas/fusillade/blob/b5ec9ed8fa3a45fc0f449ed01b828c2bbafc80a5/infra/Makefile#L31-L48

parthshahva: What is missing in my version of the makefile?

infra/Makefile
parthshahva: Leaving as is. Feel free to change on another PR if you have strong feelings

PR #47, opened 2020-02-03
setup database secrets in secrets manager with config class

parthshahva:
setup database secrets in secrets manager with config class


dcp-prototype/backend/ledger/requirements.txt
```diff
@@ -0,0 +1 @@
+dcplib==3.12.0
```
mckinsel: I don't know if you plan on using `dcp-cli` at all, but it wants `< 3`.

Issue #25, opened 2020-01-29
Create a new endpoint or extend the project mock API endpoint to serve information on an individual project


ambrosejcarr:
In addition to serving information on the matrix file(s) and related metadata within a project, the browser wishes to populate the information on the project programmatically. See this [example](https://data.humancellatlas.org/explore/projects/005d611a-14d5-4fbf-846e-571a1f874f70). Once we know which fields are important to send to users (#24), we should create a mock endpoint (or extend #17) that the browser can develop against.

**AC:** 
---
- [x] Create a mock endpoint whose payload contains all the information needed to populate a basic project page.

Bento007:
Blocked by #24

Bento007:
mock complete

Issue #19, opened 2020-01-29
Design the project page


ambrosejcarr:
Create a few options for how individual project pages could look.

**AC:** 
---
- [x] design mocks for organization of individual project pages
- [x] mocks explore how the user experience would differ if we choose to: 
  - Bundle categories of files together into zip/tar archives that take the same values of the metadata variables we display on the browser
  - Keep files unbundled. 
- [x] obtain user feedback and decide on the correct design.

liaprins-czi:
Here's a list of questions (for SMEs + users) I have about the projects page, the answers to which will help to narrow down the design:
https://docs.google.com/document/d/1lTQTE-KcFy35FTdDWaT2Ob5fjWDXT-6v7CFoSsMcMMA/edit#heading=h.ttd8dx9oe8jj

It includes links to specific Figma prototypes that exemplify the options under consider for the design.

I've met with a few people internally to get their feedback, and will be typing up my notes from those sessions, here:
https://drive.google.com/open?id=1MwdmetBgF6_TJqMgmFICOypiIV6SdVYa

liaprins-czi:
I have added action items to the synthesized notes from the internal feedback I've solicited. These are the next steps for solidifying the browser designs:
https://docs.google.com/document/d/1B7BNDvoxhnkU7vFSSc4kKPQWCp51dqpdzkHTVysMbLA/edit?usp=sharing

ambrosejcarr:
As discussed on Friday, our goal with this initial prototype is to make sure we capture the user facing functionality of the existing DCP software. As part of this, I made a quick assumption early on that this would include the "checkout" capability of the browser. 

However, design discussions with @liaprins-czi and @colinmegill prompted me to revisit browser-based download of multi-files. I am no longer convinced it makes sense and would like to pause this feature to give some time for user research. You can see my reasoning [here](https://docs.google.com/document/d/1wzOMYFzgZsbLOvcPaPo0D4a3qdGnysf1eXJLSSjkGIA/edit#heading=h.bwuly3wdh661).

In the meantime, I would like to downscope on download. The browser need only: 
1. Enable downloading of matrix files and metadata.
2. Communicate to users how they can use the command line to download the rest of the files (sync with @Bento007 and @calvinnhieu on this).

and the back-end needs to: 
1. Create a very simple CLI interface to enable download of multiple files. I'd be fine with aws s3 sync for now. 

I will propagate this feature downscoping through the requirements and backlog.

[...]

PR #53, opened 2020-02-04
Define OpenAPI Project endpoints

Bento007:
This is a rough sketch of the project endpoint needed to support the browser frontend.

Here is the rendered swagger https://app.swaggerhub.com/apis/Bento007/dcp-consumer-backend/1.0.0

## Defines the following endpoints:
- /projects
- /project/{project_uuid}
- /project/{project_uuid}/files/{file_uuid}

fyi: @calvinnhieu 
related to #25


dcp-prototype/frontend/browser/openapi.yml
```diff
@@ -0,0 +1,91 @@
...
+          in: path
+          description: The uuid of the project
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
+        '400':
+          description: Bad Request
```
mweiden: Is there a need for a `HTTP 404` response for each of these endpoints? What happens when a project with a given `uuid` does not exist?

Bento007: Added 404 response

ambrosejcarr:
Per the comment in #19, I this only needs to contemplate count matrix files and related metadata for now.

Bento007:
Here is the rendered swagger https://app.swaggerhub.com/apis/Bento007/dcp-consumer-backend/1.0.0

dcp-prototype/frontend/browser/openapi.yml
```diff
@@ -0,0 +1,108 @@
...
+  /projects:
+    get:
+      tags:
+        - project
+      summary: Retrieve filterable information on all projects
+      operationId: getProjects
+      responses:
+        '200':
+          description: OK
+        '400':
```
mckinsel: How could you make a bad get request to this endpoint?

Bento007: presently, none

dcp-prototype/frontend/browser/openapi.yml
```diff
@@ -0,0 +1,106 @@
...
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Return a download URL. 
+        '400':
+          description: Bad Request
+        '404':
+          description: File Not Found
+  /files:
```
calvinnhieu: Looking ahead to the 2/21 demo, should we exclude this endpoint until we implement multi-file/packaged downloads?

dcp-prototype/frontend/browser/openapi.yml
```diff
@@ -0,0 +1,106 @@
...
+  /projects/{projectUUID}:
+    get:
+      tags:
+        - project
+      summary: |
+        Retrieve detailed information about a project. This includes a list of 
+        downloadable files.
+      operationId: getProjectsUuid
+      parameters: 
+        - name: projectUUID
```
calvinnhieu: Can we use `snake_case` for API parameters?

dcp-prototype/frontend/browser/openapi.yml
```diff
@@ -0,0 +1,106 @@
...
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
+        '400':
+          description: Bad Request
+        '404':
+          description: File Not Found
+  /projects/{projectUUID}/files/{fileUUID}:
```
calvinnhieu: What are your thoughts on revising this to `/projects/<project_uuid>/files` for which returns a list of files, their metadata and S3 identifier?

The advantage of this would be to reduce the number of API calls to render the [Project page and files](https://www.figma.com/proto/LCBTk5cTkoOtW9IC5GfpeF/2-Browser?node-id=49%3A3&scaling=min-zoom) (2 calls in the proposed design, and `num_files` calls for the current design). This assumes the number of files in a project is manageable.

dcp-prototype/frontend/browser/openapi.yml
Bento007: yes, in I'll remove it

dcp-prototype/frontend/browser/openapi.yml
Bento007: My assumption was that the files in the project would be included in the `/projects/<project_uuid>` endpoint

calvinnhieu: I mainly question here the value of an endpoint that returns a single field about a file i.e., `s3_location`. What I'm proposing alternatively is `s3_location` is returned along with all of the other file metadata fields. @colinmegill as our client, do you have an opinion here?

mckinsel: I think it will have to get a signed URL for every file, so you might not want to do that for all files in a project if you don't need to.

dcp-prototype/frontend/browser/openapi.yml
mckinsel: Also it seems there's no way this could be a synchronous call, right?

dcp-prototype/frontend/browser/openapi.yml
mckinsel: Though I'm also not sure what you get out of nesting `files` under project here. Since you are identifying files by their own uuid, isn't the project uuid unneeded? Seems like you could move it to just `/files/{file_uuid}`

Bento007: @mckinsel it's nested under projects since the unit of access control should be a project.

dcp-prototype/frontend/browser/openapi.yml
Bento007: sure

dcp-prototype/frontend/browser/openapi.yml
Bento007: correct it would not be sync. This was the whole zipping downloads feature that isn't happening.

dcp-prototype/frontend/browser/openapi.yml
calvinnhieu: @Bento007 This detail can be handled on the backend (e.g. `project_uuid` will be a field in the `files` table) and I would opt for keeping the endpoint as minimal as possible

ambrosejcarr: > @mckinsel it's nested under projects since the unit of access control should be a project.

Could you say more about this @Bento007?

Bento007: @calvinnhieu so a file id will be unique within a project but not unique relative to all other files? According to @ambrosejcarr the same file can be in multiple projects. I guess it comes down to how we decided to store and group the data and what level of access control we want in the future.

ambrosejcarr: > According to @ambrosejcarr the same file can be in multiple projects.

This is a somewhat distant future that I don't believe we need to design towards right now. For the time being I would assume each file is associated with one project. Could you point me at the place I wrote this down? I'd like to correct the language to make it clearer.

The use case is that a tissue network such as the kidney submits a series of projects. At this point, all files are associated with only one project. Then, they go do a meta analysis and that combines the count matrices from all experiments into one large count matrix. The count matrix is a new file, associated only with the meta-analysis. However, we may want to link the raw fastq and bam files that were used in the meta-analysis, which would mean each of those files is now linked to two experiments. There are alternatives though, which might be to simply nest the projects. 

>> @mckinsel it's nested under projects since the unit of access control should be a project.
>
> Could you say more about this @Bento007?

My question above was with respect to "unit of access control".

Bento007: @calvinnhieu according to @ambrosejcarr a file can be in multiple projects. How would you differentiate which files are in which project? Maybe an additional table of file to projects. If you're ok with handling it in the backend then sure.

Bento007: @ambrosejcarr as far as I see it we have three main tiers of access granularity we can go with, at the artifact, project, or file level. Projects I see as being a logical unit of access since it is a natural grouping of files and is the thing uses will want to see and share. Putting the files under projects is my way of suggesting that this should be how we manage access control of the files.

mckinsel: You'd need map files to project (or whatever the access control unit is) in the backend anyway, right? You couldn't just rely on the project uuid in the path.

Bento007: If this is a hard blocker then I will change it. It can always be improved later.

Bento007:
Moved /projects/{project_uuid/files/{file_uuid} to /files/{file_uuid}

dcp_prototype/backend/browser/browser-api.yml
```diff
@@ -0,0 +1,71 @@
...
+        '404':
+          description: File Not Found
+  /files/{file_uuid}:
+    post:
+      tags:
+        - project
+        - download
+      summary: Request to download a file
+      parameters:
+        - name: project_uuid
```
calvinnhieu: `project_uuid` is no longer a parameter

dcp_prototype/backend/browser/browser-api.yml
```diff
@@ -0,0 +1,71 @@
+openapi: 3.0.0
+info:
+  description: |
+    DCP 2.0 Consumer backend API
```
calvinnhieu: suggest: DCP 2.0 Browser API

dcp_prototype/backend/browser/browser-api.yml
```diff
@@ -0,0 +1,71 @@
...
+  version: "1.0.0"
+  title: DCP 2.0 Consumer
```
calvinnhieu: suggest: DCP 2.0 Browser API

Bento007: done

PR #65, opened 2020-02-10
Script to squish down data into a single directory.

maniarathi:


dcp_prototype/backend/wrangling/migrations/scripts/squish_files.py
```diff
@@ -0,0 +1,193 @@
...
+
+        source_file_path = os.path.join(source_directory, filename)
+        if os.path.isdir(source_file_path):
+            print(f"Processing directory {source_file_path}")
+            _copy_files(
+                source_file_path, target_directory, count_by_file_type, checksums
+            )
+        elif os.path.isfile(source_file_path):
+            with open(source_file_path, "rb") as file_reader:
+                contents = file_reader.read()
```
mckinsel: This will kind of blow up for bams and fastqs won't it?

parthshahva: Yea we need to stream here. I'll work with Arathi on taking this files script work to completion. She has plenty on her plate with the metadata

maniarathi: I'm curious. I've been running this just fine both on my laptop as well as a small-ish EC2 instance that Marcus set up. What is the blow up expected here?

parthshahva: You're only working off of the pancreas dataset, which I'm assuming doesn't have that large of a fastq file. Other projects have fastq files that range from 20-54gb in size. You will run out of memory on most machines if you try to load this rather than streaming.

If you remember, this is the exact issue you encountered in the checksumming feature added to the hca cli 🙂

maniarathi:
Oh right. You did mention this offline too sorry about that! 🤣

dcp_prototype/backend/wrangling/migrations/scripts/squish_files.py
MDunitz: The dcplib extract method lets you specify content type, I'm currently just loading in the metadata. I think for the data files it makes sense to copy directly between the s3 buckets rather than streaming to an ec2 instance/laptop and then uploading to s3.

parthshahva:
@MDunitz is this PR up to date?

maniarathi:
Friendly ping :) 

@MDunitz does your PR build on top of this PR? If yes, I will rebase and @parthshahva could you PTAL? If not, I will close it :)

MDunitz:
@maniarathi yes my branch is built on top of this branch

parthshahva:
@maniarathi can we close this PR and review the full script in Madison's PR? I think this would be easier than approving this then reviewing just the diff set

PR #71, opened 2020-02-12
updating to newest version of schema as of 2/11

parthshahva:


dcp_prototype/backend/ledger/code/common/ledger_orm.py
```diff
@@ -24,161 +24,232 @@ def session(self, **kwargs):
...
+
+
+class ExpressionFile(Base):
+    __tablename__ = "expression_file"
+
+    id = Column(String, primary_key=True)
+    filename = Column(String)
+    file_format = Column(String)
+    file_size = Column(String)
+    created_at = Column(DateTime(True), nullable=False, server_default=text("now()"))
```
maniarathi: Probably also needs an s3_uri. 
Also are you assuming that `quantification_process_id` is the same as a `quantification_protocol_id`?

maniarathi: re: Last comment: I see `quantification_process_id` in the list of fields of the ExpressionFile table in the diagram

dcp_prototype/backend/ledger/code/common/ledger_orm.py
```diff
@@ -24,161 +24,232 @@ def session(self, **kwargs):
...
-        DateTime, default=datetime.utcnow, nullable=False, onupdate=datetime.utcnow
+
+    id = Column(String, primary_key=True)
+    paired_end_sequencing = Column(String)
+    instrument_manufacturer_model = Column(String)
+    created_at = Column(DateTime(True), nullable=False, server_default=text("now()"))
+    updated_at = Column(DateTime(True), nullable=False, server_default=text("now()"))
+
+
+class AnalysisFileQuantificationProtocolExpressionFileProcessJ(Base):
```
maniarathi: I think the last letter was supposed to be Join?

dcp_prototype/backend/ledger/code/common/ledger_orm.py
```diff
@@ -24,161 +24,232 @@ def session(self, **kwargs):
...
+    updated_at = Column(DateTime(True), nullable=False, server_default=text("now()"))
+
+    analysis_file = relationship("AnalysisFile")
+
+
+class Library(Base):
+    __tablename__ = "library"
+
+    id = Column(String, primary_key=True)
+    project_id = Column(ForeignKey("project.id"), nullable=False)
```
maniarathi: Huh? this only contains a project_id.

dcp_prototype/backend/ledger/code/common/ledger_orm.py
```diff
@@ -24,161 +24,232 @@ def session(self, **kwargs):
...
-    libraries = relationship(
-        "Library",
-        order_by=Library.id,
-        back_populates="sequencing_protocol",
-        cascade="all, delete, delete-orphan",
+
+class ProjectContributorJoin(Base):
+    __tablename__ = "project_contributor_join"
+
+    id = Column(String, primary_key=True)
```
maniarathi: General question about the id on the join tables. Based on a few examples I saw in sqlalchemy modeling many-to-many relationships, it seems like what the thing to do is have the things that are being joined, jointly be the primary key. For example, for this table, it would be `contributor_id` and `project_id` together. Then no separate `id` is specified. 

This begs the question about which join tables in the metadata schema actually have biological value and which are simply there to connect the entities because that should influence whether the table gets a unique id or not and influenced whether I create a unique HCA-accession for it or not.

maniarathi: @mckinsel ^^

dcp_prototype/backend/ledger/code/common/ledger_orm.py
parthshahva: s3 uri added. Yes, I assumed quantification_process_id is the same as quantification_protocol_id. Is that a unique field? (seems odd, sounds like a fk with no matching table)

dcp_prototype/backend/ledger/code/common/ledger_orm.py
parthshahva: fixed

dcp_prototype/backend/ledger/code/common/ledger_orm.py
parthshahva: That is all the library entity contains in the schema. I don't understand its point but have told thats the way it is represented with potentially additional fields coming later

dcp_prototype/backend/ledger/code/common/ledger_orm.py
parthshahva: There is no "right" way to handle ids in join tables. They can be an auto incrementing integer. They can be a join of the two ids. They can be a separate accession id. I don't have an opinion here and can change it to whatever makes most sense from the data wrangling point of view

parthshahva:
@maniarathi I have changed the two mentioned fields to type ARRAY(String). Let me know if you encounter any others. I have also made fixes in response to your other comments. Please re-review

dcp_prototype/backend/ledger/code/common/ledger_orm.py
maniarathi: I'm checking with Jason now.

maniarathi:
OK, Jenny said that was a typo so LGTM.

PR #74, opened 2020-02-13
Script to migrate DCP 1.0 data to DCP 2.0 as well as populate the ledger

maniarathi:
The current application of this script is to read from an S3 bucket that contains all the data from an existing project (that has been flattened, so no nested directories) and transforms the data into DCP 2.0 Metadata Schema v1.1.0, then inputs it into the ledger.


parthshahva:
Why are the classes labeled "Old..."? Wouldn't it be cleaner to just change the initialization method to reflect the new schema rather than hiding the abstract in the convert to new entity functions?

maniarathi:
> Why are the classes labeled "Old..."? Wouldn't it be cleaner to just change the initialization method to reflect the new schema rather than hiding the abstraction in the convert to new entity functions?

I tried the method of directly converting and it didn't work out. This specifically didn't work because you need to process the donor organism, specimen from organism and the cell suspension (all dcp 1.0 entities) before you can form the BiosamplePrep entity. Furthermore, you need to process the links.json file which tell you how everything is linked. So I ended up with a method where I simply model the existing data using the existing schema and then do the migration. That way it also allows me to check whether I've imported the data correctly.

parthshahva:
LGTM. Lets get this in since there is a follow up PR already making changes to this code

PR #102, opened 2020-02-14
[EASY] generate main.tf with the correct backend state file

Bento007:
This could save us from accidentally overwriting the statefile in the future.


mweiden:
Can use `terraform init -backend-config=...` instead of envsubst

Bento007:
good point. I will change and ask for you to approve @mweiden

PR #113, opened 2020-02-19
Script to pull DCP 1.0  metadata and data

### LINKED ISSUE #84: Parallelize script to read in bundles and populate an S3 bucket with de-duped files.
As part of the effort to migrate DCP 1.0 data into DCP 2.0 and convert the metadata schema into the DCP 2.0 metadata schema, the end-to-end conversion occurs in two steps:

1) First, the full set of bundles for a single project is downloaded to an EC2 instance following which a script reads, each file in the bundle, determines whether it has been previously copied to an S3 bucket via checksum, and if not, copies the file over to an S3 bucket in the DCP 2.0 AWS account. The files are "flattened" which means that it no longer only exists in its bundle directory so that duplicates may be removed. This process also allows us to retain a lossless copy of all DCP 1.0 data.

2) Once all the data for a project has been copied and de-duped into a single folder in an S3 bucket, a script reads each file, and if it is related to constructing the metadata graph for the project (these are all the .json files), reads the entity into a Python object. Once all entities have been read, all `links.json` files are read to construct the edges between the Python objects that have been created. Following this, the metadata graph is transformed into entities that can be directly inputted into the Ledger using the DCP 2.0 metadata schema.

This ticket refers specifically to part 1 of the above process. This script has no parallelization and is disconnected from the "download_bundle" portion so the goal of this ticket is to speed-up the script as well as connect it to the downloading of the bundles.

MDunitz:
Closes #84

Parallelize script to read in bundles and populate an S3 bucket with de-duped files. #84

Instructions for running the scripts included in the readme


dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,154 @@
...
+
+def list_files_in_bucket(project: str, continuation_token: str = None, keys: typing.List[str] = []) -> list:
+    """
+    Lists all files in a bucket (hard-coded to dunitz-prod-copy) with the prefix {project}/blobs
+    :param project: Name of project files belong to
+    :param continuation_token: s3 will return a max of 1000 files, pass in the continuation token to get the next 1000 for a project
+    :param keys: list of all of the file blob names
+    :return: list of all of the file blob names
+    """
+    s3 = boto3.client("s3")
```
maniarathi: Can this be moved out to another location, perhaps just a global variable in the class so we don't instantiate a new client every time the function is called?

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,154 @@
...
+    Copy file between s3 buckets
+    :param key: File name
+    :param project: Project name
+    :param max_concurrency: Max concurrent threads for s3 bucket, boto3 default is 10
+    """
+    config = TransferConfig(max_concurrency=max_concurrency)
+
+    s3 = boto3.resource('s3')
+    copy_source = {
+        'Bucket': 'org-hca-dss-prod',
```
maniarathi: Can we move the constants to the top of the file as global variables so that it's more clear what constants' values are that it being used it code. Alternatively you could pass in these arguments (in ArgumentParser) as part of calling the script with defaults set.

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,154 @@
...
+    files_moved = 0
+    already_there = 0
+    dispatch_executor_class = concurrent.futures.ThreadPoolExecutor
+    with dispatch_executor_class(max_workers=20) as executor:
+        print(f"{len(project_data_files)} files to copy over for {project}")
+        futures = []
+        for file_info in project_data_files:
+            blob_name = compose_blob_key(file_info)
+            key = f"{project}/{blob_name}"
+            if key in precopied_files:
```
maniarathi: Question: how do you guarantee that the threads aren't copying over files that may be exactly the same? For example if thread_1 is copying over `project_0.json` from one bundle and thread_2 is copying over `project_0.json` from another bundle, the file contents will be the same but since neither copy has fully finished yet, wouldn't you end up with two copies of `project_0.json`?

MDunitz: I think you are right. I refactored to create a set of blob_names to avoid the issue. Thanks!

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
MDunitz: 👍

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
MDunitz: 👍

dcp_prototype/backend/wrangling/migrations/scripts/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+Concrete use
+```bash
+python3 ./dcp_prototype/backend/wrangling/migrations/scripts/get_project_metadata.py -p ./dcp_prototype/backend/wrangling/migrations/projects.json --threads 20
+```
+If --thread not passed in, defaults to 1
+
+### migrations/scripts/load_project_metadata.py
+
+This script will load all files (without repeats within a project) and will rename files to {uuid}.{version}.json (with exception of links.json which will be renamed to its checksum)
```
maniarathi: What exactly do you mean by "will load all files"?

dcp_prototype/backend/wrangling/migrations/scripts/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+Concrete Use
+```bash
+python3 ./dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py -i ./hold_data
+```
+Buckets are currently hard coded:
+SOURCE_BUCKET = 'org-hca-dss-prod'
+DESTINATION_BUCKET = 'dunitz-prod-copy'
+
+
```
maniarathi: Could you write a short blurb on how to run this end-to-end? Like from downloading the bundles all the way to populating the S3 bucket? Actually could maybe stick it at the top of the README.

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,133 @@
...
+import os
+import typing
+
+from argparse import ArgumentParser
+
+from boto3.s3.transfer import TransferConfig
+sys.path.insert(0, "")  # noqa
+from dcp_prototype.backend.wrangling.migrations.utils.util import list_files_in_bucket
+
+s3 = boto3.resource('s3')
```
maniarathi: Capitalize S3 since global variable

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,133 @@
...
+    CRC32C = "crc32c"
+    S3_ETAG = "s3_etag"
+    SHA1 = "sha1"
+    SHA256 = "sha256"
+
+
+def compose_blob_key(file_info: typing.Dict[str, str]) -> str:
+    """
+    Create the key for a blob, given the file metadata.
+    :param file_info: This can either be an object that contains the four keys (SHA256, SHA1, S3_ETAG, and CRC32C) in
```
maniarathi: Incomplete sentence "this can either be ... " without the "or"?

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,133 @@
...
+
+from argparse import ArgumentParser
+
+from boto3.s3.transfer import TransferConfig
+sys.path.insert(0, "")  # noqa
+from dcp_prototype.backend.wrangling.migrations.utils.util import list_files_in_bucket
+
+s3 = boto3.resource('s3')
+SOURCE_BUCKET = 'org-hca-dss-prod'
+DESTINATION_BUCKET = 'dunitz-prod-copy'
```
maniarathi: If possible, could we change the default to be a bucket that doesn't contain your username? Probably best to keep it repeatable by others.

dcp_prototype/backend/wrangling/migrations/utils/util.py
```diff
@@ -12,3 +18,37 @@ def merge_dictionary_into(dict1: Dict, dict2: Dict):
         else:
             merged_dictionary[key] = value if type(value) is list else [value]
     return merged_dictionary
+
+
+def list_files_in_bucket(bucket: str, key: str, continuation_token: str = None, files: typing.Dict[str, str] = {}, checksums: bool=False) -> Dict[str, str]:
+    """
+    Lists all files in a bucket with the prefix key
+    :param bucket: Name of AWS 3s bucket
```
maniarathi: s/3s/S3

dcp_prototype/backend/wrangling/migrations/utils/util.py
```diff
@@ -12,3 +18,37 @@ def merge_dictionary_into(dict1: Dict, dict2: Dict):
         else:
             merged_dictionary[key] = value if type(value) is list else [value]
     return merged_dictionary
+
+
+def list_files_in_bucket(bucket: str, key: str, continuation_token: str = None, files: typing.Dict[str, str] = {}, checksums: bool=False) -> Dict[str, str]:
```
maniarathi: Could you please write unit tests for this function? There is a respective unit test file in tests/

dcp_prototype/backend/wrangling/migrations/utils/util.py
```diff
@@ -1,5 +1,11 @@
+import json
+import typing
 from typing import Dict
 
+import boto3
+
+s3 = boto3.client("s3")
```
maniarathi: This isn't necessary for the entire utility file, so could we keep this variable local please?

dcp_prototype/backend/wrangling/migrations/utils/util.py
maniarathi: Dict is already imported from typing so s/`typing.Dict`/`Dict`.

dcp_prototype/backend/wrangling/migrations/utils/util.py
```diff
@@ -12,3 +18,37 @@ def merge_dictionary_into(dict1: Dict, dict2: Dict):
...
+    if response['KeyCount'] > 0:
+        for x in response['Contents']:
+            if checksums:
+                checksum = json.loads(x['ETag'])
+                files[checksum] = x
+            else:
+
+                files[x['Key']] = x
+    if response['IsTruncated'] is True:
+        list_files_in_bucket(bucket, key, response['NextContinuationToken'], files, checksums)
```
maniarathi: I don't think there's any benefit to making this recursive; the iterative method is more intuitive and is probably more slightly safer for high numbers of recursive calls.

dcp_prototype/backend/wrangling/migrations/scripts/get_project_metadata.py
```diff
@@ -0,0 +1,40 @@
+import json
+from argparse import ArgumentParser
+from dcplib.etl import DSSExtractor
+
+
+def get_project_metadata(project_file, max_workers):
```
maniarathi: pydoc please?

dcp_prototype/backend/wrangling/migrations/scripts/copy_data_files.py
```diff
@@ -0,0 +1,133 @@
...
+    with dispatch_executor_class(max_workers=50) as executor:
+        print(f"{len(project_data_files)} files to copy over for {project}")
+        futures = []
+        for blob_name in project_data_files:
+            key = f"{project}/{blob_name}"
+            try:
+                precopied_files[key]
+                already_there += 1
+                continue
+            # if its not  already copied
```
maniarathi: Extra space

dcp_prototype/backend/wrangling/migrations/scripts/load_project_metadata.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser()
+    parser.add_argument(
+        "-i",
+        "--input_directory",
+        nargs="+",
+        required=True,
+        help="A data directory containing the metadata files for one or more projects"
```
maniarathi: This help statement isn't correct now right?

dcp_prototype/backend/wrangling/migrations/scripts/README.md
MDunitz: updated

dcp_prototype/backend/wrangling/migrations/utils/util.py
MDunitz: discussed offline

dcp_prototype/backend/wrangling/migrations/utils/util.py
MDunitz: discussed offline

maniarathi:
Probably needs a rebase, but LGTM. Thanks!

Issue #58, opened 2020-02-10
Browser File Download


ambrosejcarr:
## Background
The Data Browser will enable users to download bundles of files via a URL that they can curl

## User Stories

_As a computational OR biological user, I need to be able to download files of interest to me so that I can use them in my research_

_As a computational OR biological user, I want the HCA to provide an interface that enables me to download the files I want but does not require me to learn new skills so that I can quickly obtain the information I need for my work_

## Acceptance Criteria
- [ ] Users can download bundles containing all files that correspond to a (project, file type, assay, organ, species)
- [ ] Users can download bundles containing all files that correspond to a (project, file type)

ambrosejcarr:
Estimates: Week on back-end & week on front end. Probably not worth parallelizing with mocks.

ambrosejcarr:
@Bento007 @calvinnhieu @liaprins-czi 

As discussed in the planning meeting yesterday, we're only pursuing the second acceptance criteria (Users can download bundles containing all files that correspond to a (project, file type) for the Mar 20th demo.

Issue #23, opened 2020-01-29
Create a minimal project browser that enables browsing of real projects


ambrosejcarr:
Create a project browser whose project pages are generated from real metadata, so we know it will suit the requirements of the DCP data corpus, but that serves mocked data while we get the back-end set up.

ambrosejcarr:
Moved download to a new epic #131

ambrosejcarr:
Lia's estimate: 2 weeks #19 @ambrosejcarr to sync with @liaprins-czi 
Colin's estimate: 1 day for #72

PR #137, opened 2020-02-21
Project page design

liaprins-czi:
Fixes https://github.com/HumanCellAtlas/dcp-prototype/issues/19

I am trying out a new way to do design reviews, thanks to lots of help from @mweiden ! The plan is to do them like we review code branches: using PRs as the underlying formal platform/structure. But instead of reviewing how the code functions and whether it solves the GH issue, you will look at my Figma prototype to decide that, then leave your review with feedback for me to incorporate if needed.

This PR is for the design of the project page.

Here's the prototype:
https://www.figma.com/proto/ZlGIwcvDzn4HBr50cYsBRV/5-Browser?node-id=2%3A0&viewport=-402%2C629%2C0.125&scaling=min-zoom

And here's the Figma file:
https://www.figma.com/file/ZlGIwcvDzn4HBr50cYsBRV/5-Browser?node-id=0%3A1

Thanks!


jahilton:
Just to be clear, HCA 1-5 are out of our control, right? So nothing to review there?

Technicality, most (all?) organs are multi-tissue. So I'd change that distinction to 'multi-organ'.

I find the left side bar of the Project page with "Project Information" & "Project Data" to be a bit hidden. I'd rather just have one page with an obvious "Download Data" button somewhere on that page.

I am loving the "Project Data" checkbox options. Will the options respond interactively (I select Brain, but this project only has brain data for human so I would like for Mus musculus to grey out when I select Brain)?

Gene metadata download is out-of-place because that's about references, nothing specific to the chosen project.

The "Metadata" header should be more description -- "Project Metadata"?

jahilton:
...and you can see our first draft of the properties that should populate the project table & project page in our [core schema docs](https://docs.google.com/spreadsheets/d/1wAPWTk0rVTeEPE_ndKK7NvYNjO_99D-1gOmcKD6xq1U/edit#gid=1892162826)
In the schema, each of these fields will have a property (currently called user_friendly) that will give the human-readable title to use for each field.

liaprins-czi:
Thanks for this helpful feedback @jahilton ! 

Yes, the first several screens were just to show how a user would access the portal (and also that we are removing the original landing page and taking them straight to the data when they navigate there from the HCA).

I have made a new prototype here:
https://www.figma.com/proto/1lxR6YHtgNLyhPlB4aydoM/7-Browser?node-id=311%3A5292&viewport=-805%2C397%2C0.125&scaling=min-zoom

This addresses the way the checkboxes respond reactively (ideally) if a user selects different ones resulting in combinations that may or may not have data available (pink text explains in more detail).

Still to explore:
- [ ] Making download more noticeable (iterate design of left sidebar and/or have "Download" button as Jason suggested above)
- [ ] Figure out what to do about gene metadata (so far I have done nothing different); @jahilton where would you expect it to be? @ambrosejcarr can you weigh in on Jason's comment above that _"Gene metadata download is out-of-place because that's about references, nothing specific to the chosen project."_? Thank you!

ambrosejcarr:
> Gene metadata download is out-of-place because that's about references, nothing specific to the chosen project.

I agree that the gene information we share are not specific to the project. However, gene metadata are also the column labels that users would expect to see associated with the count matrix, so I think they are needed here. 

Do you think there's a more expressive name for gene/cell metadata that reflects that these information associate with the matrix? 

> I find the left side bar of the Project page with "Project Information" & "Project Data" to be a bit hidden. I'd rather just have one page with an obvious "Download Data" button somewhere on that page.

@liaprins-czi if someone scrolls down past the project information, is the project data loaded below, or do users need to click the link to find the data? So long as a user can scroll down to find the data, I like the sidebar links.

ambrosejcarr:
@liaprins-czi would it be helpful to stage this review a bit more? My sense is that when you've resolved @jahilton 's comments on the pre-download portion of the prototype, everything before the download is complete and can be closed. 

Based on our conversations, it feels like there's some active iteration happening on the download feature and it might benefit from a bit more time to work on it before putting it up for broad review. Is that a fair assessment? 

If this makes sense, I'd suggest: 
1. decide if any changes need to be made to the projects page or list of projects page based on @jahilton 's comments. If there are required changes, make tickets and assign them to #23 
2. make a new ticket to reflect the download prototyping and assigning it to #58 
3. Clarify which pages of the prototype are being worked on in the download ticket, and whether those pages are ready for review.

liaprins-czi:
@ambrosejcarr Yes, if user scrolls down they also get to the downloadable data section fo the project page. I am just planning to make the left sidebar active vs inactive state more clear as a follow up to Jason's feedback.

Also, yes I think we should separate out the design of the "Project Information" section of the project page from the "Project Data" section. 

I will start by comparing @jahilton 's Core Schema Docs to the current Project Information fields shown, and update the design as needed, and make a ticket under epic #23 , then make the other tickets, too.

jahilton:
For the Gene metadata, I think a good first step would be to call it "Reference gene metadata"

ambrosejcarr:
Thinking about this more -- what about "feature metadata"? Features
probably *won't* be the same across all datasets in H2, and won't be gene
data exclusively (CITE-seq, protein data, imaging).

jahilton:
I think "Reference" should be included to clearly communicate that the metadata does not describe anything specific to the project or else a user will assume it informs of 'features with significant data' in that project

liaprins-czi:
Closing because I have now broken down the parts and pieces of the project page into discrete tasks for the Project Information section and the download flow section.

PR #158, opened 2020-02-25
[WIP] v0 Artifact JSON schema

parthshahva:
Things left
1) Add descriptions to each field
2) Add string descriptions to empty array properties
3) Confirm whether artifact should have ontology values or terms
4) Add metadata to files section
5) Add enums where necessary


dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
```diff
@@ -0,0 +1,408 @@
...
+                     "type":"string",
+                     "examples":[
+                        "Diseased",
+                        "Normal"
+                     ]
+                  }
+               },
+               "selected_cell_types":{
+                  "$id":"#/properties/projects/items/properties/selected_cell_types",
+                  "type":"array",
```
bmccandless: Remove the comma at the end of the line.
There are a few other places that have this mistake which you can find by running the scheme through a json reader.

dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
```diff
@@ -0,0 +1,408 @@
...
+                  "type":"integer",
+                  "default":0,
+                  "examples":[
+                     2500
+                  ]
+               }
+            }
+         }
+      },
+      "files":{
```
bmccandless: Question about files.   files are currently separated out from projects.   It makes more sense to me to have files be a required item inside the project.  In this way you determine which files are associated with which project.

dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
```diff
@@ -0,0 +1,408 @@
...
+   "definitions":{
+
+   },
+   "version":"v0.1.0",
+   "$schema":"http://json-schema.org/draft-07/schema#",
+   "$id":"http://data.humancellatlas.org.com/artifact.json",
+   "type":"object",
+   "title":"The Artifact Schema",
+   "required":[
+      "projects",
```
bmccandless: Two more things that would be useful: 

scheme_version:  equal to the version of the schema.  It would be useful to have this in the file so that the code that processes it can handle more than one version.   

schema_name: equal to something like "DCP Artifact Schema", so that the json file is more self describing.  

Other useful metadata:
   version (or maybe date): version (or date) of the current contents of the file.  Useful to know something about the when the data was assembled.
  title : some information about what data is actually here.  E.g. "Demo Dataset", "Production Dataset", "Test Dataset".

jahilton:
@mckinsel is your comment about 'more sources for artifacts' and 'auxiliary fields' referring to the data resource revolving around Seed Network data (and thus, our DataOps team) or is it referring to more general use cases that the CZI Single Cell Program may pursue?

mckinsel:
@jahilton more the latter.

dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
mckinsel: Or we at least need project to be a property of a file, right?

dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
mckinsel: Fixed 👍

dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
mckinsel: Added a `meta` section with this information.

dcp_prototype/backend/artifact_manager/schemas/v0.0.0.json
mckinsel: I added `project_id` to the file objects. I'm not sure that one project will necessarily _contain_ every file, so I think we can leave it separate for right now.

PR #159, opened 2020-02-26
Schema updates - round one

jahilton:
A first step towards #4 

Forked the existing DCP schema [HumanCellAtlas/metadata-schema/tree/master/json_schema] and made a first round of updates
Updates have been focused on core schema, currently defined as properties needed to properly display data on the browser table of projects, filter data on the browser table, populate individual project pages, and cell metadata accompanying matrix files.

Hopefully, this can initially guide developers around the above implementation and metadata migration. Many many more schema updates to follow.


mckinsel:
Hey Jason, this looks like a pretty wholesale copying of the 1.0 schema into 2.0. I don't think that fits with the approach that we've been taking thus far, which has focused on building up what's needed and trying to keep things lightweight. And I may be misinterpreting something, but this seems to be pretty at odds with the spreadsheets and diagrams that we've all been using to communicate about the schema so far.

What would you think of json-ifying the current metadata 2.0 diagram and merging that in instead? If you're going to continue doing a major review of the 1.0 schema, we could keep track of that in a feature branch and move stuff into master as it's ready.

jahilton:
The spreadsheets and diagrams delivered thus far have been limited to the _core_ schema, defined as the schema pertinent to the software. The deliverables were limited to this scope in order to deliver quickly and give the engineers a first draft to build around.
This PR should include what we've laid out for the core schema (@jychien to review explicitly for that), but includes additional fields. I will emphasize that this is **far** from finalized, but again, trying to deliver a first pass quickly.
I have not yet included the join tables as I am still understanding their implementation and how those should be represented in the schema.
I only submitted the PR to deliver what the devs asked for so I may have misinterpreted. In that case, if preferred not to merge at this point, I'm happy to continue working on dataops-metadata-schema branch.

PR #190, opened 2020-03-11
Cleaning up dead code

maniarathi:


dcp_prototype/backend/wrangling/migrations/utils/migration_utils.py
```diff
@@ -91,23 +92,22 @@ class DatasetMetadata(object):
...
             schema_data = open(schema_file).read()
             schema = json.loads(schema_data)
         except FileNotFoundError:
             # FIXME, the schema is not yet committed to main.
-            # validate should fail if the data cannot be validated.
-            print(f"Warning, the schema could not be validates: {schema_file}")
-            return True
+            # Validate should fail if the data cannot be validated.
+            print(f"Warning, the schema could not be validated: {schema_file}")
+            return False
```
bmccandless: I agree this is the behavior we want.  But since the schema is not merged into master yet, this change will cause the create_artifact.py script to fail.  You may want to wait to merge until the schema is in place.

maniarathi: Oops; sorry, I totally forgot that the schema hadn't been pushed in yet. I reverted this change since the primary motivation for this PR was to remove old code.

PR #191, opened 2020-03-12
Adding a subset of tests for the migration scripts

maniarathi:


dcp_prototype/backend/wrangling/migrations/common/gather_dcp_one_data.py
```diff
@@ -176,7 +176,6 @@ def generate_metadata_structure_from_dir(input_dir, num_threads, dataset_metadat
 
 
 def generate_metadata_structure_from_targz(input_source, dataset_metadata):
-    index = 0
```
bmccandless: index is used in line  194,195.  It's used to print the number of files accessed.

dcp_prototype/backend/wrangling/migrations/common/metadata.py
```diff
@@ -1,15 +1,9 @@
 """This module contains simple python classes that correspond to the artifact schema"""
 
-import json
-
 
 class MetadataBase(object):
     """Base class for the metadata objects"""
 
-    def to_json(self):
```
bmccandless: This function is used in migrate_dcp_one_to_dcp_two and create_artifact

dcp_prototype/backend/wrangling/migrations/common/gather_dcp_one_data.py
maniarathi: This variable is defined/incremented in the for loop in 188 though so this is a redundant initiation.

dcp_prototype/backend/wrangling/migrations/common/metadata.py
maniarathi: I couldn't find any uses of it actually in either file; could you point me to the references please?

bmccandless: My bad, I had already refactored that part.  It's not needed.

dcp_prototype/backend/wrangling/migrations/common/gather_dcp_one_data.py
bmccandless: Yes, you're right.

Issue #176, opened 2020-03-03
Incorporate DCP/1 datasets into Artifact


brianraymor:
[**March 20 Demo Action Items**](https://docs.google.com/presentation/d/1nY4PXzOEs8euFEAlp2dHxB5dMZTZLSuinXzOo_eR7Ig/edit#slide=id.p4)

- [x]  Marcus to write script to update cXg metadata in matrices (Loom format)
- [x]  Brian M  to write  DCP/1 project metadata  [bypassing Ledger]
---

bmccandless:
The v.0.0.0 artifact is ready to be consumed.  
It contains only the project metadata, and not files information.

There may be changes to make to the artifact.  In particular the flattening of all the fields loses some potentially useful information.   Jason commented on this in the DCP-2 slack channel.
https://chanzuckerbergteam.slack.com/archives/CS66X37CZ/p1583430918047400

jahilton:
Summarizing my comments from Slack: The artifact schema flattens all metadata within a project. This may suffice for the upcoming demo, but is insufficient for actual consumption use cases that will need to understand the metadata at a finer resolution, as to not be constrained by how the submitter grouped data together. I suggested flattening within a library, which is the general idea behind the biosample_preparation object.

Specific issues:
- what the artifact labels `biosample_categories` is actually what is intended for `biosample_name`. See the [biosample_preparation doc](https://docs.google.com/document/d/175TbHFZ1CUIK_CT8t4G05FDQRe9LlRP7qIeG8MiWzCg/edit) for details.
- `"id": "091cf39b-01bc-42e5-9437-f419a66c8a45"` has no `donor_species`
- no projects have `selected_cell_markers`. I would expect some to have that (from enrichment_protocol.markers)

bmccandless:
Hi Jason

Thanks for the review.  I am working on fixes now.

1.   I'm not sure how to get biosample_category.   From the document it should be a value like 'primary tissue', 'organoid', 'cell line'.   What's the best way to deduce that from DCP1 metadata?


2.  The project with missing donor_species. (Human-Hematopoietic-Profiling)
The script is expecting to read from ontology_label, which is missing:  e..g

 "genus_species": [
  {
   "text": "Homo sapiens",
   "ontology": "NCBITaxon:9606"
  }
 ],

I could add in some special cases to handle this.

3.  selected_cell_markers.  I have it fix.  It was a bug in the script.

jahilton:
[...]

PR #193, opened 2020-03-12
Updates and fixes to the dcp1 migration script

bmccandless:
1. fix biosample_names and biosample_categories
2. fix bug in handing of enrichment_protocol
3. workaround for missing donor species in one of the datasets

The biosample_categories may not be correct.  This is the part
that chooses between primary, cell line and organoid.

Issue #176

### DIFF
--- a/dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
+++ b/dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
@@ -50,6 +50,18 @@ class DatasetMetadata:
         self.process_dissociation_protocol()
         self.process_library_preparation_protocol()
         self.process_enrichment_protocols()
+        self.process_cell_suspension()
+        self.process_organoid()
+        self.process_cell_line()
+
+        # NOTE: this may not be the proper way to get the biosample_category.
+        # This need to be reviewed and checked.
+        if self.entity_data.get("organoid"):
+            self.project.biosample_categories.append("organoid")
+        elif self.entity_data.get("cell_line"):
+            self.project.biosample_categories.append("cell line")
+        elif self.entity_data.get("specimen_from_organism"):
+            self.project.biosample_categories.append("primary tissue")
 
     def validate(self, data):
         """Validate the data with the current artifact schema"""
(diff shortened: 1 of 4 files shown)


dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
```diff
@@ -50,6 +50,18 @@ def process(self):
         self.process_dissociation_protocol()
         self.process_library_preparation_protocol()
         self.process_enrichment_protocols()
+        self.process_cell_suspension()
+        self.process_organoid()
+        self.process_cell_line()
+
+        # NOTE: this may not be the proper way to get the biosample_category.
+        # This need to be reviewed and checked.
```
maniarathi: I think you can use this request: 

http GET https://service.explore.data.humancellatlas.org/repository/projects?size=50 --body | jq -r '.hits[] | "\(.entryId)\t\"\(.projects[].projectShortname)\"\t\(.samples[].sampleEntityType | join(","))"'

Let me know if that doesn't work!

maniarathi: The other option might be just to hard code it because (at least for now) this code is only used to convert the DCP 1.0 data. In case you decide that option, you could probably re-create the `constants.py` file in `common/` and add the information in based on this spreadsheet:

https://docs.google.com/spreadsheets/d/1u6T37eij_Qm0cY2eyKbuAvJhZfGsIQUHDMFhczDEwLA/edit#gid=0

dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
```diff
@@ -50,6 +50,18 @@ def process(self):
...
+        if self.entity_data.get("organoid"):
+            self.project.biosample_categories.append("organoid")
+        elif self.entity_data.get("cell_line"):
+            self.project.biosample_categories.append("cell line")
+        elif self.entity_data.get("specimen_from_organism"):
+            self.project.biosample_categories.append("primary tissue")
```
maniarathi: I think the category might be "primary specimen" instead of "primary tissue".

dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
```diff
@@ -117,12 +129,21 @@ def process_donor_organism(self):
         for data in self.entity_data.get("donor_organism", []):
             self.process_mappings("donor_organism", data, self.project, mapping, self.missing)
 
+        # This is a special case for one project that lacks an ontology_label for genus_species.
+        # An alternative approach to solve this problem is to provide a list of source mappings for
+        # each destination mapping, in order of preference.  Then iterate through each source mapping
+        # until a match is found.  However, since this only impacts one attribute for one project,
+        # it's simpler to make a special case.
+        if not self.project.donor_species:
```
maniarathi: Not really a comment on your code, but out of curiosity, is the `ontology_text` field populated for this project? I think I remember that there was a project where neither ontology nor ontology_label was populated, but ontology_text was.

dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
bmccandless: I updated to the the rest api.

dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
bmccandless: The biosample prep object document lists it as "primary tissue".

dcp_prototype/backend/wrangling/migrations/common/dataset_metadata.py
bmccandless: This one has "text" and "ontology" but no "ontology_text".

PR #194, opened 2020-03-13
Add a couple scripts used to prepare demo matrices

mckinsel:
enloom.py converts the GEO GSE60361 matrix to a loom with dcp/2 metadata
convert_metadata.py does the same for dcp/1 matrices from the matrix service.

Partially cłösës #176.

### DIFF
--- a/dcp_prototype/backend/scripts/convert_metadata.py
+++ b/dcp_prototype/backend/scripts/convert_metadata.py
@@ -1,0 +1,100 @@
+"""
+Simple script to convert a DCP/1 matrix service loom to the DCP/2 cell
+metadata.
+
+Modifies the loom file in place and ends up dropping quite a few of the
+fields from DCP/1.
+"""
+
+import loompy
+import sys
+
+METADATA_MAP = {
+    "CellID": "CellID",
+    "barcode": None,
+    "bundle_uuid": None,
+    "bundle_version": None,
+    "cell_suspension.provenance.document_id": None,
+    "cell_suspension.genus_species.ontology": "biosample_preparation.donor_species",
+    "cell_suspension.genus_species.ontology_label": None,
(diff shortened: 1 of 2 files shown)


dcp_prototype/backend/scripts/convert_metadata.py
```diff
@@ -0,0 +1,100 @@
+"""
+Simple script to convert a DCP/1 matrix service loom to the DCP/2 cell
+metadata.
+
+Modifies the loom file in place and ends up dropping quite a few of the
+fields from DCP/1.
+"""
+
```
bmccandless: Looks good.  
I have some suggestions, which you can take or leave depending on what the use case for this script ends up being.

1.   make the bulk of this script a function that can be called by other scripts.    e.g  write a function alled something like convert_loom_metadata. If we had that, then we could reuse the code more easily.

2. What's the benefit to overwriting the files?   Maybe may an option to supply a command line arg for you the output file.

3.  Perhaps use argparse or something similar to parse the command line args.  right now there is no help message or error checking.

4. add ds.close() after done modifying.

dcp_prototype/backend/scripts/enloom.py
```diff
@@ -0,0 +1,119 @@
+"""
+Create a loom file from a GEO matrix.
+
+This is more of an ad hoc POC, but outlines the steps.
```
bmccandless: I think this is fine as a POC.

dcp_prototype/backend/scripts/convert_metadata.py
mckinsel: Thanks, I fixed everything except for (2). I'm kind of reluctant to make a copy because the script is modifying a tiny part of what could be a rather large file.

PR #207, opened 2020-03-17
Update frontend to enable file download

calvinnhieu:
These changes contain a minimal frontend that enables E2E file download. Styling and polish will be captured in a separate PR


dcp_prototype/backend/browser/code/common/s3_utils.py
```diff
@@ -0,0 +1,25 @@
+import os
+
+import boto3
+from botocore.exceptions import ClientError
+
+
+def generate_file_url(file_prefix, expiration=3600):
```
calvinnhieu: @Bento007 When a user clicks "Download matrix", a download URL with 24hr expiration will be generated and returned. Is this sufficient for our registered access requirements?

Bento007: Yep, that's fine for the time being

PR #178, opened 2020-03-04
Adding Auth0 authentication to Project page

Bento007:
If the user is not authenticated and tries to visit a project page, they will be directed to login before they can proceed. 

This code is based off these tutorials:
- https://auth0.com/blog/securing-gatsby-with-auth0/?_ga=2.87703896.1785923316.1582573428-1657564689.1581035098
- https://www.gatsbyjs.org/tutorial/authentication-tutorial/
- https://auth0.com/blog/authenticating-your-first-react-app/

## Issue Links
- Closes #61, 
- Closes #62 
- Closes #150, 
- Closes #155,
- Closes #153, 
- Closes #170

## Instructions
1. run `gatsby develop`
1. open browser to localhost 
Screen Shot 2020-03-13 at 3 40 42 PM
1. view project 
Screen Shot 2020-03-13 at 3 48 04 PM
1. See download section
Screen Shot 2020-03-16 at 9 43 21 AM

1. select Login/Register 
Screen Shot 2020-03-13 at 3 41 17 PM
1. login 
Screen Shot 2020-03-13 at 3 41 59 PM

## I need help completing these tasks bellow
- DONE: Send the authorization token in the request for project info.
- DONE: Return to original page after login.
- DONE: Add a login/logout/register button

## Future Work
DONE: Use auth0-spa-js instead of auth0-js for authentication
Future TODO: Enforce authorization header in requests to backend.

PR #212, opened 2020-03-20
Gatsby Site Deployment With Authentication

Bento007:
- Adding terraform to host the single-cell browser from s3 behind cloudfront and over https. This will allow us to securely deploy the site using CZI-INFRA best practices. 
- This PR also adds the gastby env.production file that will be will be uploaded to the gatsby site bucket and downloaded and used during the deployment of the site. This allows us to store the configuration of the each deployment environment easier.
- There were conflict with the existing dcp site bucket name so I created a new bucket with appends the aws_account_id to the end of the bucket.
- Changed the AUTH0_CLIENT_ID for env.develop. This will allow us to continue to test gatsby locally while the old AUTH0_CLIENT_ID will be used for the host site. There is no security risk in reusing the client ids.

### DIFF
--- a/infra/envs/dev/main.tf
+++ b/infra/envs/dev/main.tf
@@ -32,10 +32,36 @@ provider "aws" {
 //  preferred_maintenance_window = "${var.preferred_maintenance_window}"
 //}
 
-module "browser_frontend" {
-  source = "../../modules/frontend/browser"
+module "browser_site_cert" {
+  source = "github.com/chanzuckerberg/cztack//aws-acm-cert?ref=v0.29.0"
 
-  deployment_stage = "${var.deployment_stage}"
+  # the cert domain name
+  cert_domain_name = "browser.dev.single-cell.czi.technology"
+  cert_subject_alternative_names = {"www.browser.dev.single-cell.czi.technology"= "Z3GU3D81Z7R7CN"}
+
+  # the route53 zone for validating the `cert_domain_name`
+  aws_route53_zone_id = "Z3GU3D81Z7R7CN"
+
+  # variables for tags
+  env     = "${var.deployment_stage}"
+  project = "single-cell"
+  service = "browser"
+  owner   = "czi-single-cell"
+}
+
+module "browser_frontend" {
+  source = "github.com/chanzuckerberg/cztack//aws-single-page-static-site?ref=v0.29.0"
+
+  aws_route53_zone_id            = "Z3GU3D81Z7R7CN"
+  aws_acm_cert_arn = "${module.browser_site_cert.arn}"
+  bucket_name = "dcp-static-site-${var.deployment_stage}-${data.aws_caller_identity.current.account_id}"
+  subdomain = "browser"
+
+  # Variables used for tagging
+  env     = "${var.deployment_stage}"
+  project = "single-cell"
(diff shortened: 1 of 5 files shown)


infra/envs/test/main.tf
```diff
@@ -10,36 +10,68 @@ terraform {
...
-  db_username                  = "${var.db_username}"
-  db_password                  = "${var.db_password}"
-  db_instance_count            = "${var.db_instance_count}"
-  preferred_maintenance_window = "${var.preferred_maintenance_window}"
+module "siteCert" {
+  source = "github.com/chanzuckerberg/cztack//aws-acm-cert?ref=v0.29.0"
+
+  # the cert domain name
+  cert_domain_name = "browser.dev.single-cell.czi.technology"
+
```
edulop91: I _think_ what is going on is that the static site is also trying to set up the www. hostname. 

Try adding 
```suggestion
cert_subject_alternative_names = {"www.browser.dev.single-cell.czi.technology": "Z0921546EDJ5WWGRFFKB"}
```

Bento007: Thank you this fixed it

infra/envs/dev/main.tf
```diff
@@ -32,10 +32,36 @@ provider "aws" {
 //  preferred_maintenance_window = "${var.preferred_maintenance_window}"
 //}
 
+module "browser_site_cert" {
+  source = "github.com/chanzuckerberg/cztack//aws-acm-cert?ref=v0.29.0"
+
+  # the cert domain name
+  cert_domain_name = "browser.dev.single-cell.czi.technology"
```
calvinnhieu: How was this domain decided on?

infra/envs/test/main.tf
```diff
@@ -10,36 +10,69 @@ terraform {
   backend "s3" {
```
calvinnhieu: Do we still need a `test` environment? When we moved accounts, we decided to drop `test` and [only use dev](https://github.com/chanzuckerberg/dcp-prototype/pull/204#issue-389558030). Let me know what you think!

infra/envs/dev/main.tf
Bento007: d**ev.single-cell.czi.technology** domain already existed for us. I executively decided **browser**, because that's what we have been calling it.

Bento007: If we decided on a better name later we can change it, but we needed a domain with an SSL cert for https for auth0 to work.

infra/envs/test/main.tf
Bento007: I'll continue to use the test environment for the changes I make before make a PR to master. I can removed it from github and use it locally.

infra/envs/dev/main.tf
calvinnhieu: Got it! As long as it won't be a hassle to change it in the future, this sounds good to me.

infra/envs/test/main.tf
calvinnhieu: Sure, keep the test environment then

Bento007:
Related to #178

Bento007:
@calvinnhieu the old deployment has been destroyed

PR #223, opened 2020-03-26
Fix broken cellxgene links

### LINKED ISSUE #218: Prevent users from visiting unsupported cellxgene links
Currently, both matrix and cxg buttons are made available when there is matrix data available. This enables users to visit a broken cxg link when matrix data is available but a cxg object is not.

Update the frontend to enable matrix and cellxgene buttons only when data for that specific action is available.

- [x] matrix
- [x] cellxgene

For cellxgene, HEAD the provided URL and determine data availability from the response code (200 OK)

calvinnhieu:
- determines cxg-supported projects via `cxg_enabled` field to be supplied via Artifact (cc @bmccandless)
- fixed broken S3 paths (matrix and cxg objects are identified by the project's `label` name)
- updated DB schema and API accordingly

Fixes #218

### DIFF
--- a/dcp_prototype/frontend/src/components/exploreData.js
+++ b/dcp_prototype/frontend/src/components/exploreData.js
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { useEffect, useState } from "react"
 import { Flex, Box, Text } from "theme-ui"
 import StyledButton from "./styledButton"
 import LoginSignup from "./login-signup"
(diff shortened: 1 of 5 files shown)


dcp_prototype/frontend/src/components/exploreData.js
```diff
@@ -1,19 +1,20 @@
...
 import { Flex, Box, Text } from "theme-ui"
 import StyledButton from "./styledButton"
 import LoginSignup from "./login-signup"
 
 import { api_prefix } from "../globals"
 
 const ExploreData = ({ project, files, isAuthenticated }) => {
-  const cxg_url = "http://cellxgene-dev.us-west-2.elasticbeanstalk.com/"
-  const disabled = !((files && files.length && isAuthenticated) === true)
+  const cxgUrl= "http://cellxgene-dev.us-west-2.elasticbeanstalk.com/"
```
Bento007: In a future PR we should migrate this value to the [env.production](https://www.gatsbyjs.org/docs/environment-variables/) store in s3 using https://github.com/chanzuckerberg/dcp-prototype/blob/a97a3414490006f0cb60e1506052ac05db298611/dcp_prototype/frontend/Makefile#L28

mweiden: Y'all may want to parameterize these sooner than later. We should remove as many blockers to a production deployment as possible.

calvinnhieu: https://app.zenhub.com/workspaces/dcp2-5e2a191dad828d52cc78b028/issues/chanzuckerberg/dcp-prototype/229

PR #230, opened 2020-03-26
Parameterize frontend environment specific variables

calvinnhieu:
Also updated `.env.production` in the `dev` S3 bucket

Closes #223

### DIFF
--- a/dcp_prototype/frontend/.env.development
+++ b/dcp_prototype/frontend/.env.development
@@ -3,3 +3,6 @@
 AUTH0_DOMAIN=humancellatlasstaging.auth0.com
 AUTH0_CLIENTID=UikiQ2KMnvPgUAcOTRz8h0ELMrXVuSl6
 AUTH0_CALLBACK=http://localhost:8000/
+
+API_URL=https://u5j1wa9u5i.execute-api.us-east-1.amazonaws.com/dev
+CXG_URL=http://cellxgene-dev.us-west-2.elasticbeanstalk.com
--- a/dcp_prototype/frontend/src/components/exploreData.js
+++ b/dcp_prototype/frontend/src/components/exploreData.js
@@ -3,10 +3,9 @@ import { Flex, Box, Text } from "theme-ui"
 import StyledButton from "./styledButton"
 import LoginSignup from "./login-signup"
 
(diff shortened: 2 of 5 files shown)


dcp_prototype/frontend/.env.development
```diff
@@ -3,3 +3,6 @@
 AUTH0_DOMAIN=humancellatlasstaging.auth0.com
 AUTH0_CLIENTID=UikiQ2KMnvPgUAcOTRz8h0ELMrXVuSl6
 AUTH0_CALLBACK=http://localhost:8000/
+
+API_URL=https://u5j1wa9u5i.execute-api.us-east-1.amazonaws.com/dev
+CXG_URL=http://cellxgene-dev.us-west-2.elasticbeanstalk.com
```
mweiden: Why not take this opportunity to move the development secrets to AWS? Seems like the prod environment is [already there](https://github.com/chanzuckerberg/dcp-prototype/blob/master/dcp_prototype/frontend/Makefile#L23-L30).

calvinnhieu: There's a little nuance here that @Bento007 can provide more context on - `.env.production` is the Gatsby file that configures a deployment and is stored in the website deployment S3 bucket. `.env.development` is used to configure a local environment during development. Possibly `.env.local` may be more appropriate  of a name?

mweiden: I think that I understand that. Here's what I'm proposing:

1. Store .env.development in s3
1. Remove it from git, add .env.development and .eng.production to .gitignore
1. Add .env.development to `retrieve-vars` and `upload-vars`
1. Add a bit of documentation in the README.md that walks developers through the setup: `make retrieve-vars` then `gatsby develop`

That way, the frontend directory is generalizable and has no deployment-specific info in it.

That said, this is extracurricular. Feel free to say no. Your other changes LGTM.

calvinnhieu: Got it, that makes sense. My point was mainly that there isn't a logical place to store `.env.development` in s3, and I don't see the value in creating another S3 bucket for it. I think excluding from git is a good idea, and we should include instructions on how to create a `.env.development` via `retrieve-vars` from a deployed environment

PR #234, opened 2020-04-02
Public bucket for host static site

Bento007:
- Modified chanzuckerberg/cztack//aws-single-page-static-site to be public but still restricted to cloudfront using a referer_secret in the request header. It's important to note that communication between cloudfront and s3 host website is over http. AWS does not support https when hosting a static website out of s3.
- Variable referer_secret added to TF variables
- Updated `test` and `dev` environment

Misc:
- Removed interpolation only variables 
- Ran make fmt


infra/envs/dev/main.tf
```diff
@@ -50,12 +50,13 @@ module "browser_site_cert" {
 }
 
 module "browser_frontend" {
-  source = "github.com/chanzuckerberg/cztack//aws-single-page-static-site?ref=v0.29.0"
+  source = "../../modules/frontend/browser"
```
calvinnhieu, line 53: Just confirming, this was incorrect before?

infra/modules/frontend/browser/README.md
```diff
@@ -0,0 +1,78 @@
+**copied from [https://github.com/chanzuckerberg/cztack/tree/v0.30.0/aws-single-page-static-site](https://github.com/chanzuckerberg/cztack/tree/v0.30.0/aws-single-page-static-site)**
```
calvinnhieu, line 1: Is this a full copy? If so, maybe just linking to the up-to-date original would be sufficient

infra/envs/dev/main.tf
calvinnhieu, line 53: What was the reason for moving this module into our repo?

infra/modules/frontend/browser/README.md
calvinnhieu: Oh I see, it includes the `referer_secret`

calvinnhieu:
What was the issue with the private bucket again? Just curious since `browser-testing` and `browser-dev` were live before these changes right?

infra/envs/dev/main.tf
```diff
@@ -50,12 +50,13 @@ module "browser_site_cert" {
...
 
-  aws_route53_zone_id            = "Z3GU3D81Z7R7CN"
-  aws_acm_cert_arn = "${module.browser_site_cert.arn}"
-  bucket_name = "dcp-static-site-${var.deployment_stage}-${data.aws_caller_identity.current.account_id}"
-  subdomain = "browser"
+  aws_route53_zone_id = "Z3GU3D81Z7R7CN"
```
calvinnhieu, line 55: naive q: is this sensitive/should this be stored in `terraform.tfvars`?

calvinnhieu:
What was the issue with the private bucket again? Just curious since `browser-testing` and `browser-dev` were live before these changes right?

infra/envs/dev/main.tf
Bento007: it's not sensitive but should be in terraform.tfvars

Bento007:
I don't imagin cztack adding this since it is less secure than what they have.

Bento007:
@calvinnhieu It's largely due to gatsby magic that I don't fully understand, but I'll do my best. Cloudfront was not able to properly redirect user when they requested anything other than`domain/` from somewhere other than the SPA. This resulted in everything being redirected back the `domain/` when the page loads. This broke deep links and page refreshing. This was because the bucket was private which forces Cloudfront to use the s3 api to access files. Files must be access directly using the s3 api and are not inferred as gatsby expects. 

The reason cloudfront couldn't find the files is because of how gatsby organizes the files when it builds. For example cloudfront expect _project.html_ to exist to access _http://domain/project_, but gatsby creates _s3://bucket/project/index.html_ which cloudfront can't find without the help of an additional aws lambdas to inspect the request and modify it to point to the correct file. This could be a solution in the future, but no a sure fire solution for all of gatsby.

By setting up the bucket as public and using s3 hosted website features the routing is handle by gatsby and s3 static website hosting service.

Issue #237, opened 2020-04-03
Create in-memory test DB for unit tests


calvinnhieu:
Currently the browser backend unit tests mock out all SQL functionality. @mweiden [suggested using an in-memory DB](https://github.com/chanzuckerberg/dcp-prototype/pull/235#discussion_r402701183) to test SQL while preserving sub-second unit test times.

PR #254, opened 2020-04-09
Generate test JSON artifact

calvinnhieu:
- simple script to generate a bite-sized JSON artifact from a real one in S3
- generated artifact stored as a test fixture for unit tests

Relates to #237


calvinnhieu:
The script generates the test artifact fixture. The input is a real artifact stored in S3. Mainly a helper script to simplify updating the fixture

Bento007:
Then why is the name of the generated file "generated_test_fixture.json" and no "test_fixture.json". The person generating is supposed to rename it?

calvinnhieu:
Ah, was being explicit with the output name. Sure will change to `test_fixture.json`

Bento007:
It wasn't a requirements, but if that is the case you should mention is somewhere so a developer would know to rename it after generating it.

PR #258, opened 2020-04-09
Authorization header allowed

Bento007:
- modified chalice application to Allowing Origin "*" using chalice.CORSConfig
- this will allow us to send the auth token to the backend


calvinnhieu:
Do the sensitive endpoints need to be updated to parse the auth token?

Bento007:
Yes the parsing of the token is my next PR

PR #263, opened 2020-04-09
Verify the users token is valid on the backend

### LINKED ISSUE #154: Browser backend authentication verify
The browser backend needs to verify the Authorization header in the requests for file downloads.

# DOD
* Add an authorizer to the browser backend lambda that will authenticate the JWT provided in the request.

Bento007:
* Verify the token using the auth0/userinfo endpoint to retrieve the users info.
* Check if the user's email has been verified before granting access

fixes #154

### DIFF
--- a/dcp_prototype/backend/browser/code/common/authorizer.py
+++ b/dcp_prototype/backend/browser/code/common/authorizer.py
@@ -3,7 +3,7 @@ from functools import lru_cache
 
 import requests
 from chalice import UnauthorizedError
-from jose import jwt
+from jose import jwt, JWTError
 
 USERINFO_ENDPOINT = "https://czi-single-cell.auth0.com/userinfo"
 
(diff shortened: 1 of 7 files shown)


calvinnhieu:
Looks good, but this probably breaks the functional tests in `test_api.py`. Can you add tests to the 2 endpoints requiring auth?

dcp_prototype/backend/browser/code/common/authorizer.py
```diff
@@ -0,0 +1,11 @@
+import requests
+
+token_prefix = "Bearer "
+userinfo_endpoint = "https://czi-single-cell.auth0.com/userinfo"
+
+
+def assert_authorized(header: dict):
+    resp = requests.post(userinfo_endpoint, headers={"Authorization": header["Authorization"]})
```
mweiden, line 8: Just out of curiosity, how does using the the `/userinfo` endpoint differ from the authorization setup [described here](https://auth0.com/docs/quickstart/backend/python#protect-api-endpoints)?

Bento007: That setup is following Oauth2 standard and is only verifying the Access token is valid before proceeding. We are following the ODIC standard which uses the access token to verify the user's identity with the identity provider using /userinfo before proceeding. We don't have the Access token verification step because we are using the opaque token provided by auth0. We will switch to a JWT Access token in a following PR. Using the JWT Access token will allow us to embed scopes into the token in the future.

mweiden: Gotcha. Thank you for the explanation.

Bento007:
making significant changes. Closing and opening a new PR

Issue #249, opened 2020-04-06
Widen Organ and Assay columns


maniarathi:
Currently it very difficult to parse multi-line columns especially in the organ and assay columns because there is no differentiation of gap between a new item or an item that simply has spilled over into the next line. An example:

DNA library
construction
10X v2
sequencing
cDNA library
construction

Can we widen those two columns so that each item can be represented on a single line or add an additional space between items please?

maniarathi:
Oops; reopening because "DNA library construction" and "cDNA library construction" still have the same issue where it's split as 

DNA library
construction
10X v2
sequencing
cDNA library
construction

liaprins-czi:
This is very mysterious... all of the changes I had made previously are missing completely! I will follow up with @calvinnhieu and @Bento007 to see if there is a way to find them. Otherwise I will redo them from scratch.

Issue #246, opened 2020-04-06
Spacing of project page is weird for Diabetic Nephropathy snRNA-seq


maniarathi:
https://browser.dev.single-cell.czi.technology/project/?id=577c946d-6de5-4b55-a854-cd3fde40bff2

Not sure if this is an only me thing but the spacing is weird for the Description and is really narrow.

PR #274, opened 2020-04-13
Fixes for #246 and #249

liaprins-czi:
Better legibility for organ and assay names that run over one line (also increased table max width to make this happen less often) + fixing spacing issues for "Diabetic Nephropathy..." project page. I also added a bit of styling to the project page to differentiate the labels from their values.

### DIFF
--- a/dcp_prototype/frontend/src/components/footer.js
+++ b/dcp_prototype/frontend/src/components/footer.js
@@ -16,7 +16,7 @@ const Header = ({ siteTitle }) => {
           justifyContent: "space-between",
           fontWeight: 700,
           margin: `0 auto`,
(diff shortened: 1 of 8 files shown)


calvinnhieu:
I'll handle the merge conflict

dcp_prototype/frontend/src/components/footer.js
```diff
@@ -16,7 +16,7 @@ const Header = ({ siteTitle }) => {
           justifyContent: "space-between",
           fontWeight: 700,
           margin: `0 auto`,
-          maxWidth: "1100px",
+          maxWidth: "1400px",
```
Bento007, line 19: We should move this value to a reusable vars.

PR #275, opened 2020-04-14
Adding JWT Token verify to browser backend

### LINKED ISSUE #154: Browser backend authentication verify

Bento007:
* The frontend gets a token for the browser-api audience. The new AUDIENCE environment variable for the frontend is used to retrieve a token for the browser-api.
* The JWT is verified by the backend using the flow recommended by auth0 https://auth0.com/docs/quickstart/backend/python
* Testing tokens are generated using a secret stored in AWS dcp/backend/browser/test/auth0-secret. These test secret is only used for the DEPLOYMENT_STAGEs "dev", and "test". Auth tests do not run when DEPLOYMENT_STAGE is not "dev, or "test". The secret used to generate the tokens is not linked to a user, therefore the auth0 userinfo endpoint does not work and is skipped when using the test key.
* The JWT is verified for correctness and the public key from auth0 is used to verify the tokens authenticity. 

fixes #154

depends on #276 and https://github.com/chanzuckerberg/shared-infra/pull/1640

### DIFF
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -25,8 +25,8 @@ jobs:
     - name: Configure AWS Credentials
       uses: aws-actions/configure-aws-credentials@v1
       with:
-        aws-access-key-id: ${{ secrets. AWS_CZI_HCA_DEV_CI_KEY }}
-        aws-secret-access-key: ${{ secrets.AWS_CZI_HCA_DEV_CI_SECRET }}
+        aws-access-key-id: ${{ secrets.GATSBY_AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.GATSBY_AWS_SECRET_ACCESS_KEY }}
         aws-region: us-east-1
     - uses: actions/checkout@v2
     - name: Set up Python 3.7
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -25,8 +25,8 @@ jobs:
     - name: Configure AWS Credentials
       uses: aws-actions/configure-aws-credentials@v1
       with:
-        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
-        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-access-key-id: ${{ secrets.AWS_CZI_HCA_DEV_CI_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_CZI_HCA_DEV_CI_SECRET }}
         aws-region: us-east-1
     - uses: actions/checkout@v2
     - name: Set up Python 3.7
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -22,6 +22,12 @@ jobs:
   unit-test:
     runs-on: ubuntu-latest
     steps:
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets. AWS_CZI_HCA_DEV_CI_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_CZI_HCA_DEV_CI_SECRET }}
+        aws-region: us-east-1
     - uses: actions/checkout@v2
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -25,8 +25,8 @@ jobs:
     - name: Configure AWS Credentials
       uses: aws-actions/configure-aws-credentials@v1
       with:
-        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY}}
-        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY}}
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         aws-region: us-east-1
     - uses: actions/checkout@v2
     - name: Set up Python 3.7
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -25,8 +25,8 @@ jobs:
     - name: Configure AWS Credentials
       uses: aws-actions/configure-aws-credentials@v1
       with:
-        aws-access-key-id: ${{ secrets.GATSBY_AWS_ACCESS_KEY }}
-        aws-secret-access-key: ${{ secrets.GATSBY_AWS_SECRET_ACCESS_KEY }}
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY}}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY}}
         aws-region: us-east-1
     - uses: actions/checkout@v2
     - name: Set up Python 3.7
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -25,8 +25,8 @@ jobs:
     - name: Configure AWS Credentials
       uses: aws-actions/configure-aws-credentials@v1
       with:
-        aws-access-key-id: ${{ secrets.AWS_CZI_HCA_DEV_CI_KEY }}
-        aws-secret-access-key: ${{ secrets.AWS_CZI_HCA_DEV_CI_SECRET }}
+        aws-access-key-id: ${{ secrets.GATSBY_AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.GATSBY_AWS_SECRET_ACCESS_KEY }}
         aws-region: us-east-1
     - uses: actions/checkout@v2
     - name: Set up Python 3.7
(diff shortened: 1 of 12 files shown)


Bento007:
The tests pass locally. I'm working with Infra to connect github actions with AWS so the test pass here as well.

dcp_prototype/backend/browser/api/app.py
```diff
@@ -39,17 +58,18 @@ def get_project(project_id: str):
 
 
 @app.route("/projects/{project_id}/files", cors=cors_config)
-def get_project_files(project_id: str):
+@requires_auth()
+def get_project_files(project_id: str, **kwags):
```
calvinnhieu, line 62: nit: should it be `**kwargs`?

requirements-dev.txt
```diff
@@ -9,3 +9,4 @@ moto
 pytest
 requests>=2.22.0
 sqlalchemy
+chalice
```
calvinnhieu, line 12: You'll wanna remove `chalice` from `dcp_prototype/backend/browser/requirements-dev.txt` then

Bento007: you're right. I was using it previously, but decided to go with a simpler testing approach.

dcp_prototype/backend/browser/api/app.py
Bento007: Will fix.

.github/workflows/push-tests.yml
```diff
@@ -22,6 +22,12 @@ jobs:
   unit-test:
     runs-on: ubuntu-latest
     steps:
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: us-east-1
```
mweiden, line 30: Are we still using us-east-1? I thought we transitioned to using us-west-2

dcp_prototype/backend/browser/code/common/authorizer.py
```diff
@@ -0,0 +1,95 @@
...
+from functools import lru_cache
+
+import requests
+from chalice import UnauthorizedError
+from jose import jwt, JWTError
+
+USERINFO_ENDPOINT = "https://czi-single-cell.auth0.com/userinfo"
+
+AUTH0_DOMAIN = "czi-single-cell.auth0.com"
+API_AUDIENCE = "https://browser-api.dev.single-cell.czi.technology"
```
mweiden, line 11: Shouldn't the environment be parameterized?

tests/dcp-prototype/__init__.py
```diff
@@ -1,3 +1,4 @@
 import os
 
 os.environ["DEPLOYMENT_STAGE"] = "test"
+os.environ["AWS_DEFAULT_REGION"] = "us-east-1"
```
mweiden, line 4: See above.

.github/workflows/push-tests.yml
Bento007: We started using us-east-1 because the AWS Certificates Manager is in us-east-1

Bento007: We started using us-east-1 in order "to use an [ACM Certificate](https://docs.aws.amazon.com/acm/latest/userguide/acm-regions.html) with Amazon CloudFront". We can change back to us-west-2 but will need to keep the certificates in us-east-1. Keeping everything is a single regions seemed easier.

dcp_prototype/backend/browser/code/common/authorizer.py
Bento007: will do

dcp_prototype/frontend/.env.development
```diff
@@ -6,3 +6,5 @@ AUTH0_CALLBACK=http://localhost:8000/
 
 API_URL=https://u5j1wa9u5i.execute-api.us-east-1.amazonaws.com/dev
 CXG_URL=http://cellxgene-dev.us-west-2.elasticbeanstalk.com
+
+AUDIENCE=https://browser-api.dev.single-cell.czi.technology
```
calvinnhieu, line 10: I [updated](https://github.com/chanzuckerberg/dcp-prototype/pull/279/files#diff-acd4877c5205d1b80fcb5cef69e55ddcR7) the local frontend to point to the local backend at http://localhost:8001 . If that makes sense to you, this should point there as well

dcp_prototype/backend/browser/code/common/authorizer.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    :param openid_provider: the openid provider's domain.
+    :return: the openid configuration
+    """
+    res = requests.get("https://{op}/.well-known/openid-configuration".format(op=openid_provider))
+    res.raise_for_status()
+    return res.json()
+
+
+@lru_cache(maxsize=32)
```
calvinnhieu, line 87: Just curious, how slow are these operations without cache?

dcp_prototype/backend/browser/code/common/authorizer.py
```diff
@@ -0,0 +1,95 @@
...
+
+def get_userinfo(token):
+    resp = requests.post(USERINFO_ENDPOINT, headers={"Authorization": token})
+    assert not resp.ok
+    userinfo = resp.json()
+    assert not userinfo["email_verified"]
+    return userinfo
+
+
+@lru_cache(maxsize=32)
```
calvinnhieu, line 76: Any potential issues with a stale cache here?

.github/workflows/push-tests.yml
```diff
@@ -22,6 +22,12 @@ jobs:
   unit-test:
     runs-on: ubuntu-latest
     steps:
+    - name: Configure AWS Credentials
```
calvinnhieu, line 25: This is great! This was one of the blockers to running functional tests during push tests. I'll take a bit of time to look into it again after this gets merged

calvinnhieu:
Functional tests in `test_api.py` should be updated for the 2 endpoints requiring auth

dcp_prototype/frontend/.env.development
Bento007: The audience wont change in this case. 

I made those same changes to connect the frontend to the backend here https://github.com/chanzuckerberg/dcp-prototype/pull/272. Mine includes the ability to connect the python debugger to your local chalice app. Please take a look and merge into your PR.

dcp_prototype/backend/browser/code/common/authorizer.py
Bento007: The openid config does not change frequently or at all, and the cache will only be valid for the life of the lambda.

dcp_prototype/backend/browser/code/common/authorizer.py
Bento007: req < 100ms

dcp_prototype/frontend/.env.development
calvinnhieu: Oh sorry I missed this. Looks great - we'll merge yours first and I'll update my PR to reflect it

tests/functional/dcp-prototype/backend/browser/test_api.py
```diff
@@ -3,6 +3,10 @@
 import unittest
 
 import requests
+from dcplib.aws_secret import AwsSecret
+
+if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
+    os.environ["DEPLOYMENT_STAGE"] = "test"  # noqa
```
calvinnhieu, line 9: Is it possible to use `DEPLOYMENT_STAGE=dev` to test auth? The test deployment is planned to be deprecated

Bento007: isn't `test` going to be used for local tests?

.github/workflows/push-tests.yml
Bento007: You will need to update this policy with the correct permissions https://github.com/chanzuckerberg/dcp-prototype/blob/d0a4c0fd6a4ff45ba86f036812ebe111714f601c/infra/modules/backend/cicd/github_actions_unittests.tf#L14

tests/functional/dcp-prototype/backend/browser/test_api.py
Bento007: This should all still work as expected once the deployed test infra is destroyed and we change the test domain to localhost.

calvinnhieu: The test infra also includes the database, s3 bucket, secret which is controlled by `DEPLOYMENT_STAGE`.

These tests run against deployed environments for the following reasons:
1. Functional tests would run following a deploy
2. We (previously) were not accessing AWS credentials in push tests

This PR resolves (2) and (1) is a decision we can re-visit. It would be great if functional tests ran against a local deployment during push tests on PRs. However, this would require the test infra to remain, specifically to preserve the S3 bucket and database. For simplicity, I'd propose the entire test infra remains, in which case the deployed test API would not be used in functional tests and instead used as a dummy environment during development. Thoughts?

calvinnhieu: Actually, the test infra requirements boil down to the DB credentials in Secrets Manager so that could be pretty simple. Pinged you to chat over VC

PR #277, opened 2020-04-15
updating dependencies

Bento007:
This should resolve the vulnerability warning from github


Bento007:
Relates to https://github.com/chanzuckerberg/dcp-prototype/issues/278

PR #287, opened 2020-04-21
Destroy test env via terraform

calvinnhieu:
- [x] ran terraform
- [x] destroyed `test` Chalice app


infra/envs/dev/Makefile
```diff
@@ -20,7 +20,7 @@ apply: retrieve-vars
 	terraform apply --backup=-
 
 .PHONY: destroy
-destroy: state/rm
```
Bento007, line 23: why was state/rm removed?

calvinnhieu: Ah I removed it to get `destroy` working as `state/rm` doesn't exist right now. I suppose we should leave it and actually create a command for it

calvinnhieu: I assume `state/rm` was intended to delete the state prefix in S3. However, this should not be run until `destroy` completes successfully which may take a couple of tries. For this reason, I've added a `rm-state` target that does is not linked with `destroy`

PR #288, opened 2020-04-22
Github Action to Automated Browser Backend Deployment

Bento007:
- Adding github actions to automate the deployment of the browser backend when the `master` branch changes.
- A new file is uploaded to the infra s3 bucket. This files contains the current configuration for the chalice deployment. This make it easier to reuse the existing deployed chalice app rather than creating a new chalice app in AWS. A deployment file is stored for each deployment stage. 
- The IAM policy for github_actions has been update to allow github to deploy chalice apps. If a new deployment stage is created, the policy will need to be updated.
- Chalice apps for new stage must still be deployed from a local machine first before github can automate the deployment.
- Only the `dev` environment is automatically deployed.

### DIFF
--- a/browser/backend/chalice/Makefile
+++ b/browser/backend/chalice/Makefile
@@ -62,7 +62,7 @@ local-server: package
 retrieve-deployed:
 	if ![[ -d "$(DEPLOYED_PATH)" ]]; then mkdir $(DEPLOYED_PATH); fi
 	aws s3 cp $(S3_DEPLOYMENT_FILE) $(DEPLOYED)
-
+	cat $(DEPLOYED)
 .PHONY: upload-deployed
 upload-deployed:
 	aws s3 cp $(DEPLOYED) $(S3_DEPLOYMENT_FILE)
--- a/browser/backend/chalice/Makefile
+++ b/browser/backend/chalice/Makefile
@@ -60,7 +60,7 @@ local-server: package
 
 .PHONY: retrieve-deployed
 retrieve-deployed:
-	if ![[ -d "$(DEPLOYED_PATH)" ]]; then mkdir $(DEPLOYED_PATH); fi
+	if [[ ! -d "$(DEPLOYED_PATH)" ]]; then mkdir $(DEPLOYED_PATH); fi
 	aws s3 cp $(S3_DEPLOYMENT_FILE) $(DEPLOYED)
 	
 .PHONY: upload-deployed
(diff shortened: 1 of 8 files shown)


browser/backend/chalice/Makefile
```diff
@@ -5,7 +5,9 @@ export ACCOUNT_ID=$(shell aws sts get-caller-identity --query Account --output t
 export APP_NAME=browser-api
 export EXPORT_ENV_VARS_TO_LAMBDA=APP_NAME DEPLOYMENT_STAGE
 export PYTHONHASHSEED=0
-
+export S3_DEPLOYMENT_FILE=s3://org-dcp-infra-$(ACCOUNT_ID)/chalice/envs/$(DEPLOYMENT_STAGE)/deployed.json
+export DEPLOYED_PATH= .chalice/deployed
+export DEPLOYED = $(DEPLOYED_PATH)/$(DEPLOYMENT_STAGE).json
```
calvinnhieu, line 10: Can these be more explicit/consistent, for example: `S3_DEPLOYED_FILE`, `LOCAL_DEPLOYED_PATH` and `LOCAL_DEPLOYED_FILE`?

calvinnhieu: Also consistent spacing please!

browser/backend/requirements-dev.txt
```diff
@@ -1,3 +1,3 @@
 furl
 -r requirements.txt
--r ../../../requirements-dev.txt
+-r ../../requirements-dev.txt
```
calvinnhieu, line 3: ah nice catch

infra/modules/backend/cicd/github_actions_unittests.tf
```diff
@@ -24,6 +24,90 @@ data "aws_iam_policy_document" "policy" {
       aws_secretsmanager_secret.auth0.arn,
     ]
   }
+  statement {
+    sid     = "ChaliceDeployeAPIGateway"
```
calvinnhieu, line 28: typo in Deployed

infra/envs/dev/main.tf
```diff
@@ -92,9 +92,14 @@ module "browser_backend" {
 module "cicd" {
   source = "../../modules/backend/cicd"
 
+  account_id          = data.aws_caller_identity.current.account_id
+  api_gateway_staging = var.api_gateway_staging
+  api_gateway_dev     = var.api_gateway_dev
```
calvinnhieu, line 97: Shouldn't these variables be parameterized by environment?

calvinnhieu: e.g. supplying `api_gateway_id` and `deployment_stage` to `github_actions_unittests.tf ` should be sufficient

Bento007: no, these are the api-gateway ids like for dev we have `u5j1w...`. If we setup chalice with terraform then we could use the API-gateway resource in the future.

calvinnhieu: Hm I am referring to the `api-gateway-id` as well. `github_actions_unittests.tf` is a module that is created per env (`dev`, `staging`), correct? `dev/main.tf` would only need to supply the `u5j1w...` ID right?

calvinnhieu: Ah, I see. The module's only included in the dev env. Since this doesn't follow the convention of the rest of our modules, could you add a comment to `github_actions_unittests.tf`?

PR #291, opened 2020-04-23
Continuous Deploy Frontend

Bento007:
Deploy the frontend to dev whenever master is updated. This will still require us to invalid the cloudfront cache before the changes are seen. This will be automated in the future.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,7 +3,7 @@ name: Deploy
 on:
   push:
     branch:
-      - tsmith/cd-frontend
+      - master
 
 env:
   DEPLOYMENT_STAGE: 'dev'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,3 +29,20 @@ jobs:
         sudo apt-get install moreutils
     - name: deploy backend
       run: make ci-deploy -C browser/backend/chalice
+
+  frontend:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -29,3 +29,20 @@ jobs:
...
+  frontend:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Configure AWS Credentials
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        aws-region: us-east-1
```
calvinnhieu, line 42: Just curious, why doesn't the `single-cell-dev` profile need to be configured to deploy?

calvinnhieu: Oh right, you create a policy in terraform :+1:

Bento007: the single-cell-dev profile is an alias for all of the above configuration. We don't actually require an AWS_PROFILE to be named single-cell-dev anywhere.

calvinnhieu: The above configuration are the account details i.e. stored in `.aws/credentials` . `single-cell-dev` is a profile for an IAM role that I don't believe is reflected here.

Bento007:
@calvinnhieu I think that's a great idea!

Issue #310, opened 2020-05-19
Update OpenAPI Spec with finalized API design


maniarathi:
Now that the [high level architecture design document](https://docs.google.com/document/d/1d8tv2Ub5b3E7Il85adOAUcG8P05N6UBZJ3XbhJSRrFs/edit#) has been completed, tweak the OpenAPI spec that has been developed to reflect the new changes. Any changes that are required based on design doc bugs can be fixed here as well.

brianraymor:
Arathi and I removed the _Publish API to corpora-api.cziscience.com_ item in the original check list, since there is no requirement to publish the API in Milestone 1.

maniarathi:
Hi @Bento007, I'm taking myself off this ticket since the first task item (finalizing the API in the design document) is complete. I trust you've got the reins from here :)

PR #317, opened 2020-05-20
Converting design doc to openapi

Bento007:
Relates to #310

### DIFF
--- a/data_portal_api.yml
+++ b/data_portal_api.yml
@@ -1,12 +1,12 @@
 openapi: 3.0.0
 info:
   version: "1.0.0"
-  title: Copora Data Portal API
+  title: Corpora Data Portal API
   description: >
-    This is the API for the Copora Data Portal
+    This is the API for the Corpora Data Portal
 servers:
   - description: production
-    url: https://copora.czi-single-cell.science
+    url: https://api.corpora.czi-single-cell.science
 paths:
   /v1/submissions:
     get:
@@ -16,15 +16,14 @@ paths:
       description: >-
         This lists all submissions and their UUIDs that currently exist in corpora. If a parameter is specified as a filter, then only projects that meet the status criteria will be outputted.
 
-        If user uuid is provided, submissions from that user are returned. If no user uuid is specified the id of the authenticated user is used.
+        If user_uuid is provided, submissions from that user are returned. If no user uuid is specified the id of the authenticated user is used.
       parameters:
         - name: status
           description: The status of the submissions to list.
           in: query
           schema:
             type: string
-            enum: [all, open, closed]
-            default: all
+            enum: [open, closed]
         - $ref: '#/components/parameters/user_uuid'
       responses:
         '200':
@@ -45,7 +44,7 @@ paths:
         opens a new submission and an empty project for editing. Executing this call will create a new entry in both the Submission and Project tables with empty fields and creates a new folder with the given project uuid in the data portal S3 bucket. If successful, a message will be returned with the project’s UUID.
       responses:
         '201':
-          $ref: '#/components/responses/201'
+          description: A new submissions has been created.
         '400':
           $ref: '#/components/responses/400'
         '401':
@@ -73,7 +72,7 @@ paths:
         '403':
           $ref: '#/components/responses/403'
         '404':
-          $ref: '#/components/responses/404'
+          description: Submission not found
     delete:
       summary: Delete a submissions
       tags:
@@ -106,7 +105,7 @@ paths:
         - $ref: '#/components/parameters/filetype'
       responses:
         '201':
-          $ref: '#/components/responses/201'
+          description: File added to submission.
         '400':
           $ref: '#/components/responses/400'
         '401':
@@ -128,7 +127,7 @@ paths:
         - $ref: '#/components/parameters/filename'
       responses:
         '200':
-          $ref: '#/components/responses/201'
+          $ref: '#/components/responses/200'
         '401':
           $ref: '#/components/responses/401'
         '403':
@@ -141,12 +140,12 @@ paths:
         - submissions
       summary: Validate the Project
       description: >-
-        Validates the project according to the guidelines below in the Validation section and outputs the result, specifying all errors. The result of the validation is also stored in the backend database.
+        Validates the project and outputs the result, specifying all errors.
       parameters:
         - $ref: '#/components/parameters/project_uuid'
       responses:
         '200':
-          $ref: '#/components/responses/201'
+          $ref: '#/components/responses/200'
         '401':
           $ref: '#/components/responses/401'
         '403':
@@ -159,7 +158,7 @@ paths:
         - submissions
       summary: Save the Project
       description: >-
-        If needed, extracts the project-level metadata from the datasets or uses the given project-level metadata imputed via the body parameter and saves it to the database.
+        If needed, extracts the project-level metadata from the datasets or uses the given project-level metadata imputed via the body parameter and saves it.
       parameters:
         - $ref: '#/components/parameters/project_uuid'
       requestBody:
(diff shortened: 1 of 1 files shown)


data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
+openapi: 3.0.0
+info:
+  version: "1.0.0"
+  title: Copora Data Portal API
```
maniarathi, line 4: sp: Copora -> Corpora

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+  description: >
+    This is the API for the Copora Data Portal
```
maniarathi, line 6: sp: Copora -> Corpora

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+servers:
+  - description: production
+    url: https://copora.czi-single-cell.science
```
maniarathi, line 9: sp: copora -> corpora

Also this will likely be something like `api.corpora.cziscience.com`.

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+paths:
+  /v1/submissions:
+    get:
+      summary: List submissions
+      tags:
+        - submissions
+      description: >-
+        This lists all submissions and their UUIDs that currently exist in corpora. If a parameter is specified as a filter, then only projects that meet the status criteria will be outputted.
```
maniarathi, line 17: Capitalize Corpora throughout to keep it looking clean.

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+
+        If user uuid is provided, submissions from that user are returned. If no user uuid is specified the id of the authenticated user is used.
+      parameters:
+        - name: status
+          description: The status of the submissions to list.
+          in: query
+          schema:
+            type: string
+            enum: [all, open, closed]
```
maniarathi, line 26: Does there need to be an "all" here? All is already handled when the parameter is not specified at all.

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+          $ref: '#/components/responses/404'
+    post:
+      summary: Create a new submission
+      tags:
+        - submissions
+      description: >-
+        opens a new submission and an empty project for editing. Executing this call will create a new entry in both the Submission and Project tables with empty fields and creates a new folder with the given project uuid in the data portal S3 bucket. If successful, a message will be returned with the project’s UUID.
+      responses:
+        '201':
+          $ref: '#/components/responses/201'
```
maniarathi, line 48: It seems like we don't need all the responses for all the calls? (Ditto throughout).

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+    post:
+      summary: Create a new submission
+      tags:
+        - submissions
+      description: >-
+        opens a new submission and an empty project for editing. Executing this call will create a new entry in both the Submission and Project tables with empty fields and creates a new folder with the given project uuid in the data portal S3 bucket. If successful, a message will be returned with the project’s UUID.
```
maniarathi, line 45: Something weird is going on with the lines I think? There should be an error outputted since this is over the line length and will line breaks.

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+        '200':
+          $ref: '#/components/responses/201'
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+components:
+  schemas: {}
```
maniarathi, line 275: Can delete this right?

data_portal_api.yml
maniarathi: Also might be helpful to have a description that is more tailored to the API call itself so that we can distinguish between, say, a submission not found 202 and a project not found 404.

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+  /v1/submissions/validate/{project_uuid}:
+    post:
+      tags:
+        - submissions
+      summary: Validate the Project
+      description: >-
+        Validates the project according to the guidelines below in the Validation section and outputs the result, specifying all errors. The result of the validation is also stored in the backend database.
```
maniarathi, line 144: Probably need to update this description so that it doesn't refer to the "Validation section" 😄

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+      responses:
+        '200':
+          $ref: '#/components/responses/201'
+        '400':
+          $ref: '#/components/responses/400'
+  /v1/projects/{project_uuid}:
+    get:
+      tags:
+        - projects
+      summary: Get Poject
```
maniarathi, line 243: sp: Poject -> Project

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+  /v1/projects:
+    get:
+      tags:
+        - projects
+      summary: List Pojects
```
maniarathi, line 214: sp: Pojects -> Projects

data_portal_api.yml
```diff
@@ -0,0 +1,334 @@
...
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+    delete:
+      tags:
+        - projects
+      summary: Delete Poject
```
maniarathi, line 260: sp: Poject -> Project

data_portal_api.yml
Bento007: we can add details in the error response. Adding a different response code isn't the right solution IMO.

data_portal_api.yml
```diff
@@ -0,0 +1,332 @@
...
+        '200':
+          $ref: '#/components/responses/200'
+        '400':
+          $ref: '#/components/responses/400'
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          description: Submission not found
```
maniarathi, line 75: This one has a specific description that doesn't refer to the reference?

data_portal_api.yml
Bento007: the linter job only looks at python and terraform as far as I'm aware. We haven't decided on a standard for yaml files.

data_portal_api.yml
Bento007: I thought you wanted them more specific where applicable. We can keep it generic.

maniarathi: Apologies for being unclear.. I was thinking more on the lines of:

`description: Submission not found
  schema: 
`

data_portal_api.yml
maniarathi: SG

data_portal_api.yml
maniarathi: But no worries - I think you make a good point that more of the error details can be in the message.

data_portal_api.yml
```diff
@@ -0,0 +1,340 @@
...
+        type: string
+    filetype:
+      name: filetype
+      in: query
+      required: true
+      schema:
+        type: string
+        enum: [legal, data]
+    user_uuid:
+      name: userid
```
calvinnhieu, line 317: Are these different for a reason?

Bento007: nope, will fix

maniarathi:
Hey @Bento007 what's the status on this? Are you planning on updating this PR to reflect the new API changes or submit this and then create a new PR to reflect the changes?

Bento007:
I'll submit this first. We can then make the changes in parallel.

PR #323, opened 2020-05-28
Migrating from infra to single-cell-infra

Bento007:
- Renaming dcp to corpora
- Renaming browser to data-portal
- tearing down old dev and staging infra
- deploying new dev and staging infra using single-cell-infra

Depends on https://github.com/chanzuckerberg/single-cell-infra/pull/75


browser/backend/code/common/authorizer.py
```diff
@@ -8,7 +8,7 @@
 USERINFO_ENDPOINT = "https://czi-single-cell.auth0.com/userinfo"
 
 AUTH0_DOMAIN = "czi-single-cell.auth0.com"
-API_AUDIENCE = f"https://browser-api.{os.environ['DEPLOYMENT_STAGE']}.single-cell.czi.technology"
+API_AUDIENCE = f"https://api.corpora.{os.environ['DEPLOYMENT_STAGE']}.single-cell.czi.technology"
```
calvinnhieu, line 11: Do we have plans to use the domain `corpora.cziscience.com`?

maniarathi: Yep, it should be the one that Calvin mentioned.

browser/backend/code/config/db_config.py
```diff
@@ -3,4 +3,4 @@
 
 class BrowserDbConfig(Config):
```
maniarathi, line 4: Rename to DataPortalDbConfig?

browser/backend/code/common/authorizer.py
Bento007: What is the URL for the site, and what is the URL for the api?

maniarathi: corpora.cziscience.com
api.corpora.cziscience.com

Bento007: These domains are for productions only right? It looks like cellxgene is using cellxgene.staging.single-cell.czi.technology for is development stage. Do we want to follow that same naming scheme as cellxgene for our development environements?

maniarathi: You'd tag on the `staging` or `dev` environment name appropriately.

I spoke with Bruce and Madison about the cellxgene URLs and they largely used them due to convenience and bc they were free. We aren't tied to matching the same naming scheme as cellxgene so let's keep ours simple and only use the cziscience.com domain unless there's a strong reason not to.

PR #328, opened 2020-05-31
Massive reorganization of corpora repository.

maniarathi:
- Moving around chalice repository to avoid duplicate code and multiple locations of backend code.
- Removed dependency on dcplib.
- Removed some of the references to artifacts (not all)
- Removed references to DCP.
- Clean up of tests.
- Fixing of any dependencies that were broken by any of the tasks above :)


maniarathi:
@Bento007 I wasn't sure how to fix some of the authorization/secrets unit tests - would you be able to help out here and fix them? Feel free to checkout my branch and push the changes. Thank you in advance!

calvinnhieu:
Nice work. No issues with the code organization decisions made.

All checks should be passing before merging - in particular verifying all deployment workflows still work and deploying a working `dev` environment under the new resource names/code organization. I've taken a look and here's the current state of deployment workflows:
- [x] *API server*: passing
- [ ] *Cloudfront invalidator*: requires code changes, new infra/tf resources
- [ ] *Frontend*: requires new infra/tf resources

I believe verifying the deployment is the best way to verify these changes. Given the infra requirements for some of the deployments, this PR would be blocked on [75](https://github.com/chanzuckerberg/single-cell-infra/pull/75). Do you think this is a reasonable sequence of events @maniarathi ?

maniarathi:
@calvinnhieu Thanks for the breakdown! Yes, this makes sense to me and I agree with you that we should shoot for the green checkmarks. 

@Bento007 do you know what the status is on [75](https://github.com/chanzuckerberg/single-cell-infra/pull/75)? Might need to friendly-ping infra to get it moving (I think their review will be more impactful than mine).

backend/corpora/common/utils/aws_secret.py
```diff
@@ -0,0 +1,90 @@
+from time import sleep
```
Bento007, line 1: isn't this duplicated from dcplib. Is there anything wrong with using dcplib?

backend/corpora/common/utils/s3_utils.py
```diff
@@ -15,7 +15,7 @@ def generate_file_url(file_prefix: str, expiration: int = 3600, s3=None) -> str:
     if not s3:
         s3 = boto3.client("s3")
 
-    bucket_name = f"corpora-data-portal-bucket-{os.environ['DEPLOYMENT_STAGE']}"
+    bucket_name = f"corpora-data-portal-{os.environ['DEPLOYMENT_STAGE']}"
```
Bento007, line 18:
```suggestion
    bucket_name = f"corpora-data-{os.environ['DEPLOYMENT_STAGE']}"
```

tests/functional/backend/corpora/test_api.py
```diff
@@ -0,0 +1,59 @@
...
+
+import requests
+
+from backend.corpora.common.utils.aws_secret import AwsSecret
+
+if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
+    os.environ["DEPLOYMENT_STAGE"] = "dev"  # noqa
+
+API_URL = {
+    "dev": "https://corpora-api.dev.cziscience.com",
```
Bento007, line 13:
```suggestion
    "dev": "https://api.dev.corpora.cziscience.com",
```

maniarathi:
I do not feel comfortable using `dcplib.` While it is technically open source, the purpose of the repository is "contains modules shared across multiple DCP components." not in service of general purpose tooling. I don't know what the future holds for DCP and the corresponding repositories but I would much prefer that we do not get bitten or get stuck into a tight spot if things change in DCP land.

The code that is copied over is very little in the grand scheme of things and we can guarantee stability in those files/tools as well as make modifications if needed later down the line. This would not necessarily be true if we used `dcplib`.

PR #367, opened 2020-06-10
Fixing auth tests

Bento007:
The auth0 test will only run when branches are pushed to master.

### DIFF
--- a/tests/unit/backend/corpora/common/test_authorizer.py
+++ b/tests/unit/backend/corpora/common/test_authorizer.py
@@ -16,7 +16,8 @@ class TestAuthorizer(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        cls.auth0_secret = json.loads(AwsSecret(cls.secret_name).value)
+        secret_name = "corpora/cicd/dev/auth0-secret"
+        cls.auth0_secret = json.loads(AwsSecret(secret_name).value)
         cls.auth0_secret["audience"] = f"https://api.{os.getenv('DEPLOYMENT_STAGE')}.corpora.cziscience.com"
 
     def test_postive(self):
--- a/tests/unit/backend/corpora/common/test_authorizer.py
+++ b/tests/unit/backend/corpora/common/test_authorizer.py
@@ -14,7 +14,7 @@ if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
 secret_name = "corpora/cicd/dev/auth0-secret"
 # using the same secret for all none production stages.
 
-
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE", 'test') == "test", "DEPLOYMENT_STAGE is 'test'")
 class TestAuthorizer(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
--- a/tests/unit/backend/corpora/common/test_authorizer.py
+++ b/tests/unit/backend/corpora/common/test_authorizer.py
@@ -11,11 +11,11 @@ from backend.corpora.common.utils.aws_secret import AwsSecret
 
 @unittest.skipIf(os.getenv("DEPLOYMENT_STAGE"), "DEPLOYMENT_STAGE not set")
 class TestAuthorizer(unittest.TestCase):
-    # using the same secret for all non production stages.
 
     @classmethod
     def setUpClass(cls):
         secret_name = "corpora/cicd/dev/auth0-secret"
+        # using the same secret for all non production stages.
         cls.auth0_secret = json.loads(AwsSecret(secret_name).value)
         cls.auth0_secret["audience"] = f"https://api.{os.getenv('DEPLOYMENT_STAGE')}.corpora.cziscience.com"
 
--- a/tests/unit/backend/corpora/common/test_authorizer.py
+++ b/tests/unit/backend/corpora/common/test_authorizer.py
@@ -13,7 +13,7 @@ from backend.corpora.common.utils.aws_secret import AwsSecret
 class TestAuthorizer(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        secret_name = "corpora/cicd/dev/auth0-secret" # using the same secret for all non production stages.
+        secret_name = "corpora/cicd/dev/auth0-secret"  # using the same secret for all non production stages.
         cls.auth0_secret = json.loads(AwsSecret(secret_name).value)
         cls.auth0_secret["audience"] = f"https://api.{os.getenv('DEPLOYMENT_STAGE')}.corpora.cziscience.com"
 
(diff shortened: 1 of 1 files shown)


tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,27 @@
...
 import requests
 from chalice import UnauthorizedError
 
 from backend.corpora.common.authorizer import assert_authorized
+from backend.corpora.common.utils.aws_secret import AwsSecret
 
 if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
     os.environ["DEPLOYMENT_STAGE"] = "test"  # noqa
 
+secret_name = "corpora/cicd/dev/auth0-secret"
```
maniarathi, line 14: I think the correct paradigm here would be to set it as a class variable in a setUp function instead of as a global variable.

tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,27 @@
...
+# using the same secret for all none production stages.
 
-@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "test", "DEPLOYMENT_STAGE not 'test'")
+
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE", "test") == "test", "DEPLOYMENT_STAGE is 'test'")
```
maniarathi, line 18: I think the second argument of `skipIf` is a justification on why it. was skipped in which case is should be that DEPLOYMENT_STAGE was _not_ set to `test`.

tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,27 @@
...
 import os
 import unittest
 
 import requests
 from chalice import UnauthorizedError
 
 from backend.corpora.common.authorizer import assert_authorized
+from backend.corpora.common.utils.aws_secret import AwsSecret
 
 if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
```
maniarathi, line 11: More of a curiosity question - why does the deployment stage need to be `test` instead of `dev`?

tests/unit/backend/corpora/common/test_authorizer.py
Bento007: moved to class variable

tests/unit/backend/corpora/common/test_authorizer.py
Bento007: I was going to limit the frequency of this test to only run on pushed to master until we find a better way of managing the generated token. This limits the test to only run when deployment is Dev above.

tests/unit/backend/corpora/common/test_authorizer.py
Bento007: should be dev

tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,24 @@
...
 
-if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
-    os.environ["DEPLOYMENT_STAGE"] = "test"  # noqa
 
-
-@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "test", "DEPLOYMENT_STAGE not 'test'")
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE"), "DEPLOYMENT_STAGE not set")
 class TestAuthorizer(unittest.TestCase):
+    secret_name = "corpora/cicd/dev/auth0-secret"
+    # using the same secret for all none production stages.
```
maniarathi, line 15: s/none/non

tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,24 @@
...
+from backend.corpora.common.utils.aws_secret import AwsSecret
 
-if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
-    os.environ["DEPLOYMENT_STAGE"] = "test"  # noqa
 
-
-@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "test", "DEPLOYMENT_STAGE not 'test'")
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE"), "DEPLOYMENT_STAGE not set")
 class TestAuthorizer(unittest.TestCase):
+    secret_name = "corpora/cicd/dev/auth0-secret"
```
maniarathi, line 14: Move to inside `setUpClass` instead of floating?

tests/unit/backend/corpora/common/test_authorizer.py
Bento007:
```suggestion
    # using the same secret for all non production stages.
```

tests/unit/backend/corpora/common/test_authorizer.py
Bento007: Is there a strong reason for that?

maniarathi: Readability. It doesn't make sense to separate this class variable from the others and I'd wonder why.

Bento007: ok ok

maniarathi: Can delete this line now that it has moved to line 19.

tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,24 @@
...
+from backend.corpora.common.utils.aws_secret import AwsSecret
 
-if not os.getenv("DEPLOYMENT_STAGE"):  # noqa
-    os.environ["DEPLOYMENT_STAGE"] = "test"  # noqa
 
-
-@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "test", "DEPLOYMENT_STAGE not 'test'")
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE"), "DEPLOYMENT_STAGE not set")
 class TestAuthorizer(unittest.TestCase):
+    # using the same secret for all non production stages.
```
maniarathi, line 14: Stray comment?

tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -1,23 +1,24 @@
...
-
-@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "test", "DEPLOYMENT_STAGE not 'test'")
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE"), "DEPLOYMENT_STAGE not set")
 class TestAuthorizer(unittest.TestCase):
+
     @classmethod
     def setUpClass(cls):
-        cls.auth0_secret = {}
+        secret_name = "corpora/cicd/dev/auth0-secret"
+        # using the same secret for all non production stages.
```
maniarathi, line 18: For readability, please move this comment to the line before the `secret_name` definition (i.e. swap lines 17 and 18). As it stands today, it seems as though the comment refers to the lines below.

PR #381, opened 2020-06-16
Implement a validator class that verifies that metadata in H5AD is as expected

maniarathi:
Note that this function so far only checks whether the metadata exists or not. Further tickets will look at validation of different file types (i.e. Loom and Seurat) as well as adherence of ontologies to ones that are recommended by the Corpora schema.


backend/corpora/common/dataset_validator.py
```diff
@@ -0,0 +1,119 @@
...
+    def verify_obs(self, data_object: anndata.AnnData):
+        """
+        Validates the observation attribute of an AnnData object. Checks to ensure that all observation IDs are
+        unique and that the observation metadata fields as described by the Corpora Schema exist. If the validation
+        fails in any way, the errors are outputted rather than the validation aborted.
+        """
+
+        # Check to ensure that all IDs are unique
+        observation_ids = data_object.obs.index.tolist()
+        if len(observation_ids) != len(set(observation_ids)):
```
mckinsel, line 68: I think you could do like `data_object.obs.index.duplicated().any()`. Little more succinct.

backend/corpora/common/dataset_validator.py
```diff
@@ -0,0 +1,119 @@
...
+            logging.warning("Each observation is not unique!")
+
+        obs_keys = map(str.upper, data_object.obs_keys())
+        for metadata_field in (
+            CorporaConstants.REQUIRED_OBSERVATION_METADATA_FIELDS
+            + CorporaConstants.REQUIRED_OBSERVATION_ONTOLOGY_METADATA_FIELDS
+        ):
+            if metadata_field.upper() not in obs_keys:
+                self.log_error_message(metadata_field, "obs", type(data_object).__name__)
```
mckinsel, line 77: Do we want the match to be case-insensitive here?

backend/corpora/common/dataset_validator.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def verify_vars(self, data_object: anndata.AnnData):
+        """
+        Validates the variable attribute of the AnnData object to ensure that all variable IDs are unique.
+        """
+
+        variable_ids = data_object.var.index.tolist()
+
+        if len(variable_ids) != len(set(variable_ids)):
```
mckinsel, line 86: I suppose this is where we want to make sure we don't have any "bad" gene names, but maybe that makes sense to break into a separate ticket?

backend/corpora/common/dataset_validator.py
```diff
@@ -0,0 +1,119 @@
...
+        dataset-level and project-level metadata and outputs which metadata fields are missing. Note that no
+        exception is thrown when metadata is found to be missing and rather an informative message is outputted instead.
+        """
+
+        unstructured_metadata_keys = data_object.uns_keys()
+
+        for metadata_field in (
+            CorporaConstants.REQUIRED_DATASET_METADATA_FIELDS
+            + CorporaConstants.REQUIRED_DATASET_PRESENTATION_METADATA_FIELDS
+            + CorporaConstants.REQUIRED_DATASET_PRESENTATION_HINTS_METADATA_FIELDS
```
mckinsel, line 101: I think missing hints probably shouldn't generate error messages.

backend/corpora/common/utils/corpora_constants.py
```diff
@@ -0,0 +1,26 @@
+class CorporaConstants(object):
+
+    # Constants related to the metadata requirements of submitting a dataset to Corpora.
+    REQUIRED_OBSERVATION_METADATA_FIELDS = ["TISSUE", "ASSAY", "DISEASE", "CELL_TYPE", "SEX", "ETHNICITY"]
```
mckinsel, line 4: We have a new one: `development_stage`

backend/corpora/common/dataset_validator.py
maniarathi: Good question. I actually didn't know.. I had assumed that metadata fields would be case-insensitive. Would you prefer it not be?

backend/corpora/common/dataset_validator.py
maniarathi: Removed.

backend/corpora/common/utils/corpora_constants.py
maniarathi: I lumped that into the second validation PR: https://github.com/chanzuckerberg/corpora-data-portal/pull/383.

mckinsel: 👍

backend/corpora/common/dataset_validator.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def validate_h5ad_dataset(self, file_object):
+        """
+        Reads the file contents into an AnnData object. Each attribute of the AnnData object will then be checked to
+        ensure it contains the appropriate metadata.
+        """
+
+        start_time = time.time()
+        logging.info("Reading anndata object...")
+        anndata_object = anndata.read_h5ad(file_object)
```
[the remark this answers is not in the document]
mckinsel: Huh, well okay then. I tried it out locally on a few files too, and it ended up not making much of a difference anyway.

backend/corpora/common/dataset_validator.py
mckinsel: Yeah, I'm pretty sure all the formats/tooling are case-sensitive.

maniarathi: Done

PR #382, opened 2020-06-16
Add curation script for the scv2 dataset

mckinsel:


maniarathi:
Just make sure all the checks are green (I think only the linter is failing), and this LGTM.

PR #383, opened 2020-06-16
Handle loom files for validation and layers validation

maniarathi:
Note that this PR builds off of https://github.com/chanzuckerberg/corpora-data-portal/pull/381 which is still open for review.


backend/corpora/common/dataset_validator.py
```diff
@@ -40,21 +43,85 @@ def validate_dataset_file(self):
...
+        """
+
+        # Check to make sure X data exists
+        data = None
+        if isinstance(data_object.X, DataFrame):
+            data = data_object.X.data
+        elif isinstance(data_object.X, ndarray):
+            data = data_object.X
+        elif isinstance(data_object.X, csr_matrix):
+            data = data_object.X.toarray()
```
mckinsel, line 92: This might blow up the memory, no? There are some big datasets around that are like 1% nonzero.

backend/corpora/common/dataset_validator.py
```diff
@@ -40,21 +43,85 @@ def validate_dataset_file(self):
...
+        else:
+            logging.warning(
+                f"Could not check raw data layer to ensure that it exists. The type is " f"{type(data_object.X)}!"
+            )
+
+        if not data.any():
+            logging.warning("No raw data can be found in the dataset!")
```
mckinsel, line 99: Or I guess all the observations are 0.

backend/corpora/common/dataset_validator.py
```diff
@@ -40,21 +43,85 @@ def validate_dataset_file(self):
...
+        contain an appropriate description.
+        """
+
+        # Check to make sure X data exists
+        data = None
+        if isinstance(data_object.X, DataFrame):
+            data = data_object.X.data
+        elif isinstance(data_object.X, ndarray):
+            data = data_object.X
+        elif isinstance(data_object.X, csr_matrix):
```
mckinsel, line 91: I think it can be any `scipy.sparse.spmatrix`

backend/corpora/common/dataset_validator.py
```diff
@@ -40,21 +43,85 @@ def validate_dataset_file(self):
...
+            logging.warning("No raw data can be found in the dataset!")
+
+        # Ensure that the layer_descriptions metadata key exists in the `uns` field of the anndata object.
+        if (CorporaConstants.LAYER_DESCRIPTIONS not in data_object.uns_keys()) or (
+            not data_object.uns.get(CorporaConstants.LAYER_DESCRIPTIONS)
+        ):
+            logging.warning("Required layers descriptions are missing from uns field to describe data layers!")
+        else:
+            # Check to ensure that there are descriptions for each layer
+            for layer_name, layer_data in data_object.layers.items():
```
mckinsel, line 108: Since you're not doing anything with `layer_data`, maybe just `for layer_name in data_object.layers`? (Might need `.keys()`)

calvinnhieu:
`DatasetValidator` is well organized for handling `AnnData` objects. While I'm not clear on the details for handling `.rds` files (can those even be read in Python?), validation (e.g. `verify_uns`, `verify_obs`, etc) will have different implementations. To support this, I would consider a Factory pattern moving forward. Example:

- `DatasetValidatorFactory` implements the logic in `validate_dataset_file` to create the appropriate `DatasetValidator`
- `DatasetValidator` is an abstract class that may be subclassed into `AnnDataDatasetValidator` and `SeuratDatasetValidator`
    - `DatasetValidator.validate`
    - `AnnDataDatasetValidator.validate` runs `AnnDataDatasetValidator.validate.verify_uns` etc.
    - `SeuratDatasetValidator.validate` implements something different

Usage:
```
validator = DatasetValidatorFactory.create(dataset_file)
validator.validate()
```

Benefits:
- Code re-usability
- Simpler interface
- Code organization

Thoughts?

backend/corpora/common/dataset_validator.py
maniarathi: Oops good call. Done.

backend/corpora/common/dataset_validator.py
maniarathi: Switched to use count_non_zero compared to the number of total elements. Hopefully this is less memory intensive.

maniarathi:
Hey @calvinnhieu 

You make a really good point and I really like the idea of using a Factory. The only thing right now is that I don't have a good sense of how to handle Seurat files since they are usually handled in R at the moment and I have a Spike open to look into it. In case that we decide on a Python validator for Seurat objects I'd be open to moving to the Factory model. There is a chance that we just convert Seurat to H5AD first though in which case it would be an AnnData validation again. Are you open to punting on the decision for now until we have a better sense of handling Seurat files?

backend/corpora/common/dataset_validator.py
```diff
@@ -40,21 +43,88 @@ def validate_dataset_file(self):
...
+        has_data = True
+        if isinstance(data_object.X, DataFrame):
+            has_data = data_object.X.data.any()
+        elif isinstance(data_object.X, ndarray):
+            has_data = data_object.X.any()
+        elif isinstance(data_object.X, spmatrix):
+            has_data = (data_object.X.count_nonzero() == data_object.X.nnz) or data_object.X.nnz == 0
+        else:
+            logging.warning(
+                f"Could not check raw data layer to ensure that it exists. The type is " f"{type(data_object.X)}!"
```
mckinsel, line 98: I believe `raw` means something special here, so maybe use a different word?

calvinnhieu:
@maniarathi Yup for sure. Sounds reasonable to me

backend/corpora/common/dataset_validator.py
maniarathi: Lol, yeah you're right. Don't know what I was thinking - changed to X.

PR #387, opened 2020-06-17
Adding get /v1/project/{project_uuid} 200 response body

### LINKED ISSUE #345: Specify the response schema for return project details
Both `GET /v1/submission/{project_uuid}` and `GET /v1/project/{project_uuid}`return a body with a schema that represents the entire denormalized view of a single project. Currently the response is not populated with the response body in the OpenAPI spec.

The [Data Model section](https://docs.google.com/document/d/1d8tv2Ub5b3E7Il85adOAUcG8P05N6UBZJ3XbhJSRrFs/edit#heading=h.keluekkhba84) in the high level architecture details every piece of information and its cardinality with a Project entity. This needs to be translated and denormalized and placed into the response for both endpoints in order to complete the OpenAPI spec for the Corpora API.

Bento007:
closes #345


corpora-api.yml
```diff
@@ -493,6 +497,131 @@ components:
     visibility:
       type: string
       enum: [public,private]
+    project:
```
maniarathi, line 500: Is there any reason to omit `needs_attestation`, `status`, `tc_uri`, `processing_state` or `validation_state`?

corpora-api.yml
```diff
@@ -493,6 +497,131 @@ components:
...
+                  type: string
+              other:
+                type: array
+                items:
+                  type: string
+              raw_data:
+                type: array
+                items:
+                  type: string
+              attestation:
```
maniarathi, line 533: attestation is a different thing than links - it's directly related to a project, referencing a form that needs to be filled out. There is also only one per project so an array doesn't make sense here.

corpora-api.yml
```diff
@@ -493,6 +497,131 @@ components:
...
+                type: array
+                items:
+                  type: string
+          datasets:
```
maniarathi, line 537: Seems worthwhile to split out `dataset` into its own separate object in case sometime down the line, we need to return the dataset object for an API call instead of only project.

corpora-api.yml
```diff
@@ -493,6 +497,131 @@ components:
...
+                    properties:
+                      id:
+                        type: string
+                      name:
+                        type: string
+                      email:
+                        type: string
+                      institution:
+                        type: string
+                dois:
```
maniarathi, line 612: I think we want two separate DOIs here: a `preprint_doi` and a `publication_doi`.

corpora-api.yml
```diff
@@ -493,6 +497,131 @@ components:
...
+                  items: string
+                    type: object
+                    properties:
+                      id:
+                        type: string
+                      title:
+                        type: string
+                      data:
+                        type: string
+                      journel:
```
maniarathi, line 623: s/journel/journal

corpora-api.yml
```diff
@@ -493,6 +497,131 @@ components:
...
+        properties:
+          description:
+            type: string
+          name:
+            type: string
+          id:
+            $ref: '#/components/schemas/project_uuid'
+          s3_bucket_key:
+            type: string
+          submitter:
```
calvinnhieu, line 511: I've named this `owner` in the ORM to use broader terminology and imply access controls. Does this make sense to you @maniarathi @Bento007 ? Update if so, will update ORM if not.

maniarathi: Hrrrmmm.. `submitter` personally makes more sense to me, but I could be convinced otherwise. What was the rational to making it `owner`?

corpora-api.yml
calvinnhieu, line 537: This is fine to start, though I expect we'll have more nuanced access patterns separating project and dataset details. For example, potentially only returning a list of `dataset_id`s here that can be further introspected on via a GET `/dataset`.

The advantage of this is reducing payload size and organizing the API, though no action required now, we can shrink the response later if it makes sense.

corpora-api.yml
calvinnhieu: I see. `owner` makes sense to me as an internal name since it's a more common term in describing roles/access controls in software, while `submitter` does make more sense contextually/as an external name. The rationale, to be clear, was `owner` implies:
- this is a User in the the system
- this User has all the access permissions
- this User likely submitted the project

So actually, I'm fine either way considering `submitter`s contextual connotation.

corpora-api.yml
Bento007: I wasn't sure if the was need from a get project. I'm happy to add them in.

corpora-api.yml
Bento007: Is it possible for the submitter to not be the owner? If yes, then we should make the distinction. Conceptually it makes sense that the owner could change to someone other than the original submitter. I think we should keep submitter, until there is a need to also track a separate ownership trait. For now the submitter is the only owner.

corpora-api.yml
Bento007: the needs attestations seems like something that would be returned to the submitter, not someone viewing the project, same with status, processing_state, and validations_state. 

Also what is the tc_uri?

corpora-api.yml
Bento007: so it's not a link to a form?

corpora-api.yml
Bento007: I didn't want to prematurely optimize, but it can be broken out.

Bento007: I agree fully, future optimization will likely be needed. For expendiance, I focused on building something simple and expandible.

corpora-api.yml
Bento007: how are they differentiated?

corpora-api.yml
maniarathi: So the response body for `GET /v1/submission/{project_uuid}` is also missing still and I was thinking that you'd use the same project component as a return object for both `GET /v1/submission/{project_uuid}` and `GET /v1/project/{project_uuid}` to keep it simple, in which case you need all of the above pieces of metadata.

Did you have an alternative idea though/were you planning on creating a different object for the  `GET /v1/submission/{project_uuid}` API call?

`tc_uri` is the S3 URI designating the location of the project's terms and conditions form that we need to store.

corpora-api.yml
```diff
@@ -476,10 +480,7 @@ paths:
           $ref: '#/components/responses/403'
         '404':
           $ref: '#/components/responses/404'
-
```
maniarathi, line 479: Please do not delete the spacing here. It helps with readability.

maniarathi: Also likewise with lines 480 and 482.

corpora-api.yml
maniarathi: It's an S3 URI so yes technically a link? I see that you've separated out attestation though so I'm uncertain whether the comment is still applicable?

corpora-api.yml
maniarathi: One project has up to two DOIs; the preprint and publication DOIs are different numbers thus designating different "journals" (technically the preprint will be in bioRxiv).

corpora-api.yml
maniarathi: It's unclear whether in the future, there will be a difference between someone who "owns" a  project versus someone who "submits" a project. @calvinnhieu you make some good points that owner currently refers to a User who has the correct permissions in order to submit a project so I'm leaning towards changing it to `owner` now.

corpora-api.yml
Bento007: I'll recycle it for both.

corpora-api.yml
Bento007: fixed

corpora-api.yml
Bento007: Owner it is.

corpora-api.yml
Bento007: How are they differentiated in the DB?

corpora-api.yml
```diff
@@ -366,7 +366,6 @@ paths:
         '404':
           $ref: '#/components/responses/404'
 
-
```
maniarathi, line 369: Undelete this line?

corpora-api.yml
```diff
@@ -492,7 +494,152 @@ components:
...
+                type: string
+              type:
+                type: string
+                enum: [PROTOCOL, RAW_DATA, OTHER]
+        attestation:
+          type: object
+          properties:
+            needed:
+              type: boolean
+            url:
```
maniarathi, line 540: Given that tc_uri is below, the url isn't needed here. I'm good either way though: either not have the `attestation` object and just have two properties (`attestation_needed` and `tc_uri`) or have an `attestation` object containing a boolean representing whether it's required and the tc_uri.

corpora-api.yml
```diff
@@ -492,7 +494,152 @@ components:
...
+        organism:
+          type: string
+        organism_ontology:
+          type: string
+        title:
+          type: string
+        source_data_location:
+          type: string
+        revision:
+          type: string
```
maniarathi, line 592: revision is an integer type I believe based on the data model.

corpora-api.yml
```diff
@@ -492,7 +496,150 @@ components:
...
+              type: string
+        status:
+          type: string
+          enum: [LIVE, EDIT]
+        processing_state:
+          type: string
+          enum: [NA, IN_VALIDATON, IN_ARTIFACT_CREATION, IN_DEPLOYMENT]
+        validation_state:
+          type: string
+          enum: [NOT_VALIDATION, VALID, INVALID]
```
calvinnhieu, line 526: `NOT_VALIDATED`

Bento007:
```suggestion
          enum: [NOT_VALIDATED, VALID, INVALID]
```

PR #389, opened 2020-06-17
remove deployment stage from auth0-secret

Bento007:


tests/unit/backend/corpora/common/test_authorizer.py
```diff
@@ -13,7 +13,7 @@
 class TestAuthorizer(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
-        secret_name = "corpora/cicd/dev/auth0-secret"  # using the same secret for all non production stages.
+        secret_name = "corpora/cicd/auth0-secret"  # using the same secret for all non production stages.
         cls.auth0_secret = json.loads(AwsSecret(secret_name).value)
```
calvinnhieu, line 17: Just curious, what are the intended use cases for `AwsSecret` and `SecretConfig` utils in our repo? Seems like they serve similar purposes and wondering if we can/should consolidate into 1.

maniarathi:
Hey @Bento007, just trying to understand why the deployment stage needs to be removed from the secret name? The reason why the deployment stage exists is because it adheres to the standard defined in `SecretConfig` which we've been using since DCP 1.0 times. Is there a strong reason to deviate?

tests/unit/backend/corpora/common/test_authorizer.py
maniarathi: `SecretConfig` provides a layer of abstraction over `AwsSecret`. It allows you to think of a configuration file containing a bunch of keys and values for things that you care about (see `corpora_config.py`) while `AwsSecret` is a simple wrapper for handling the interaction with the boto3 `secretsmanager` client. The singleton-ness of the `SecretConfig` class also provides a nice insurance that multiple Configs with differing information doesn't exist. You instantiate once and the class is the same every time (see `test_singletonness` in `test_secret_config.py`).

That being said, yes you could technically subclass AwsSecret into SecretConfig, but it just would clog up the file and reduce readability in my opinion given their different purposes.

calvinnhieu: Thanks for the explanation! Though now I see the classes are well documented and I could have just read those

Bento007: Store a secret value, and retrieve it across our cicd pipeline.

PR #397, opened 2020-06-19
Removed ABC from entity

Bento007:
We decided that using the ABC doesn't give us much, with a lot of added complexity.


Bento007:
We didn't run into any issue, but we aren't really using any of the benefits of the ABC class. Also the Duck typing that we are using is more than enough to  for our use cases.

Bento007:
Mostly for simplicity and readability.

calvinnhieu:
Yup makes sense. Thanks!

PR #398, opened 2020-06-19
Implement Project application-level object

### LINKED ISSUE #392: Implement Project application-level object

calvinnhieu:
Closes #392 

New classes:
- `Entity` is an abstract base class that provides an interface for all application objects reading and writing from the DB. Interface overview:
    - `get`: Public API to retrieve an object
    - `_query`: Sub-routine of `get` - queries the DB
    - ~`_parse`: Sub-routine of `get` - parses the db query result~
    - `_load`: Sub-routine of `get` - instantiates the Entity from query results
    - `list`: Public API to list objects
    - `save`: Public API to write an object to the DB
- `Project` inherits from `Entity`

Testing:
- local Postgres instance running in Docker container
- test database is populated in `tests/unit/backend/corpora/fixtures/database/__init__.py`

### DIFF
--- a/backend/corpora/common/entities/project.py
+++ b/backend/corpora/common/entities/project.py
@@ -7,7 +7,7 @@ from ..corpora_orm import (
     DbProjectDataset,
     DbDatasetContributor,
     DbContributor,
-    ProjectStatus
+    ProjectStatus,
 )
 from .entity import Entity
 from ..utils.exceptions import CorporaException
@@ -15,19 +15,19 @@ from ..utils.exceptions import CorporaException
 
 class Project(Entity):
     def __init__(self, **kwargs):
-        self.id = kwargs['id']
-        self.status = kwargs['status']
+        self.id = kwargs["id"]
+        self.status = kwargs["status"]
         self.name = kwargs.get("name", "")
         self.description = kwargs.get("description", "")
-        self.submitter = kwargs.get('submitter', "")
-        self.s3_bucket = kwargs.get('s3_bucket', "")
-        self.tc_uri = kwargs.get('tc_uri', "")
-        self.needs_attestation = kwargs.get('needs_attestation', "")
-        self.processing_state = kwargs.get('processing_state', "")
-        self.validation_state = kwargs.get('validation_state', "")
-        self.dataset_ids = kwargs.get('dataset_ids', [])
-        self.links = kwargs.get('links', [])
-        self.contributors = kwargs.get('contributors', [])
+        self.submitter = kwargs.get("submitter", "")
+        self.s3_bucket = kwargs.get("s3_bucket", "")
+        self.tc_uri = kwargs.get("tc_uri", "")
+        self.needs_attestation = kwargs.get("needs_attestation", "")
+        self.processing_state = kwargs.get("processing_state", "")
+        self.validation_state = kwargs.get("validation_state", "")
+        self.dataset_ids = kwargs.get("dataset_ids", [])
+        self.links = kwargs.get("links", [])
+        self.contributors = kwargs.get("contributors", [])
 
     @classmethod
     def _query(cls, key: typing.Tuple[str, str]) -> typing.List[Base]:
@@ -58,8 +58,8 @@ class Project(Entity):
                 DbProject.id == DbProjectDataset.project_id,
                 DbProject.status == DbProjectDataset.project_status,
                 DbProjectDataset.dataset_id == DbDatasetContributor.dataset_id,
-                DbContributor.id == DbDatasetContributor.contributor_id
-            ]
+                DbContributor.id == DbDatasetContributor.contributor_id,
+            ],
         )
 
         if not len(result):
@@ -92,25 +92,25 @@ class Project(Entity):
             dataset_ids.add(result.DbProjectDataset.dataset_id)
 
             links[result.DbProjectLink.id] = {
-                'id': result.DbProjectLink.id,
-                'type': result.DbProjectLink.link_type,
-                'url': result.DbProjectLink.link_url
+                "id": result.DbProjectLink.id,
+                "type": result.DbProjectLink.link_type,
+                "url": result.DbProjectLink.link_url,
             }
 
             contributors[result.DbContributor.id] = {
-                'id': result.DbContributor.id,
-                'name': result.DbContributor.name,
-                'institution': result.DbContributor.institution,
-                'email': result.DbContributor.email
+                "id": result.DbContributor.id,
+                "name": result.DbContributor.name,
+                "institution": result.DbContributor.institution,
+                "email": result.DbContributor.email,
             }
 
         # build dict of Project parameters
         project = {}
         for k, v in query_results[0].DbProject.__dict__.items():
             project[k] = v
-        project['dataset_ids'] = list(dataset_ids)
-        project['links'] = list(links.values())
-        project['contributors'] = list(contributors.values())
+        project["dataset_ids"] = list(dataset_ids)
+        project["links"] = list(links.values())
+        project["contributors"] = list(contributors.values())
 
         return project
 
@@ -122,19 +122,21 @@ class Project(Entity):
         :param params: dict of KVPs used to instantiate a Project
         :return: Project
         """
-        return Project(id=params['id'],
-                       status=params['status'],
-                       name=params['name'],
-                       description=params['description'],
-                       submitter=params['owner'],
-                       s3_bucket=params['s3_bucket'],
-                       tc_uri=params['tc_uri'],
-                       needs_attestation=params['needs_attestation'],
-                       processing_state=params['processing_state'],
-                       validation_state=params['validation_state'],
-                       dataset_ids=params['dataset_ids'],
-                       links=params['links'],
-                       contributors=params['contributors'])
+        return Project(
+            id=params["id"],
+            status=params["status"],
+            name=params["name"],
+            description=params["description"],
+            submitter=params["owner"],
+            s3_bucket=params["s3_bucket"],
+            tc_uri=params["tc_uri"],
+            needs_attestation=params["needs_attestation"],
+            processing_state=params["processing_state"],
+            validation_state=params["validation_state"],
+            dataset_ids=params["dataset_ids"],
+            links=params["links"],
+            contributors=params["contributors"],
+        )
 
     @classmethod
     def list(cls):
--- a/backend/corpora/common/entities/project.py
+++ b/backend/corpora/common/entities/project.py
@@ -29,6 +29,15 @@ class Project(Entity):
 
     @classmethod
     def _query(cls, key: typing.Tuple[str, str]) -> typing.List[Base]:
+        """
+        Given a key, queries the database for a project and its relevant entities.
+
+        According to the Entity.get interface, the return value of this function is
+        fed as the input to Project._parse.
+
+        :param key: Composite primary key tuple of the form (project.id, project.status)
+        :return: list of SQLAlchemy query results
+        """
         result = cls.db.query(
             table_args=[DbProject, DbProjectLink, DbProjectDataset, DbDatasetContributor, DbContributor],
             filter_args=[
@@ -45,8 +54,23 @@ class Project(Entity):
         return result
 
     @classmethod
-    def _parse(cls, query_results):
-        # de-dupe and parse project relationships
+    def _parse(cls, query_results: typing.List[Base]) -> typing.Dict:
+        """
+        Parses query result rows produced by Project._query into a dict of KVPs
+        required for entity instantiation.
+        The output of this function is the input to Project._load.
+
+        SQLAlchemy query results are stored in a list in which each item
+        contains Table objects resulting from the query.
+
+        Example query results access:
+        row = query_results[0]
+        project_id = row.DbProject.id
+
+        :param query_results: list of query result rows
+        :return: dict of KVPs
+        """
+        # de-dupe related project entities
         dataset_ids = set()
         links = {}
         contributors = {}
@@ -66,7 +90,7 @@ class Project(Entity):
                 'email': result.DbContributor.email
             }
 
-        # build project params
+        # build dict of Project parameters
         project = {}
         for k, v in query_results[0].DbProject.__dict__.items():
             project[k] = v
@@ -78,6 +102,12 @@ class Project(Entity):
 
     @classmethod
     def _load(cls, params: typing.Dict):
+        """
+        Instantiates and returns a Project instance.
+
+        :param params: dict of KVPs used to instantiate a Project
+        :return: Project
+        """
         return Project(id=params['id'],
                        status=params['status'],
                        name=params['name'],
--- a/backend/corpora/common/entities/project.py
+++ b/backend/corpora/common/entities/project.py
@@ -9,12 +9,12 @@ class Project(Entity):
         # TODO: list and set fields
         self.id = kwargs['id']
         self.status = kwargs['status']
+        self.link = kwargs['link']
 
     @classmethod
-    def get(cls, key: typing.Tuple[str]) -> "Project":
-        # db_project = cls.db._get(DbProject, key)
+    def get(cls, key: typing.Tuple[str, str]) -> "Project":
         # TODO: query multiple tables: DbProjectLink, DbDataset, etc...
-        data = cls.db._query(
+        results = cls.db._query(
             table_args=[DbProject, DbProjectLink],
             filter_args=[
                 DbProject.id == key[0],
@@ -24,8 +24,9 @@ class Project(Entity):
             ]
         )
         # TODO: inst Project object
-        project = Project(id=data.result.DbProject.id,
-                          status=data.result.DbProject.status,
+        project = Project(id=results[0].DbProject.id,
+                          status=results[0].DbProject.status,
+                          link=results[0].DbProjectLink.link_url
                           )
         return project
 
--- a/backend/corpora/common/entities/project.py
+++ b/backend/corpora/common/entities/project.py
@@ -14,22 +14,24 @@ from ..utils.exceptions import CorporaException
 
 
 class Project(Entity):
-    def __init__(self,
-                 id: str,
-                 status: str,
-                 name: str = "",
-                 description: str = "",
-                 owner: str = "",
-                 s3_bucket: str = "",
-                 tc_uri: str = "",
-                 needs_attestation: bool = False,
-                 processing_state: str = "",
-                 validation_state: str = "",
-                 dataset_ids: list = None,
-                 links: list = None,
-                 contributors: list = None,
-                 created_at: str = "",
-                 updated_at: str = ""):
+    def __init__(
+        self,
+        id: str,
+        status: str,
+        name: str = "",
+        description: str = "",
+        owner: str = "",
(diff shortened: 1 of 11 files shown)


backend/corpora/common/utils/db_utils.py
```diff
@@ -39,15 +47,50 @@ def _parse_multivalue(value: str) -> typing.List[str]:
...
+                DbProject.id == DbProjectDataset.project_id,
+                DbProject.status == DbProjectDataset.project_status,
+                DbProjectDataset.dataset_id == DbDatasetContributor.dataset_id,
+                DbContributor.id == DbDatasetContributor.contributor_id
+            ]
+        )
+        return result
+
+    @staticmethod
+    def parse_project(query_results: typing.List[Base]):
```
Bento007, line 67: This function should be moved to `Project`

backend/corpora/common/utils/db_utils.py
```diff
@@ -39,15 +47,50 @@ def _parse_multivalue(value: str) -> typing.List[str]:
...
-        return [
-            {
-                "id": project.id,
-                "status": project.status,
-                "name": project.name,
-                "description": project.description,
-                "processing_state": project.processing_state.name,
-                "validation_state": project.validation_state.name,
+    def get_project(self, key: typing.Tuple[str, str]) -> typing.List[Base]:
+        result = self._query(
```
Bento007, line 51: This code seems project specific, so should be moved to `Project`. What is you reasoning for using it in DB?

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,22 @@
...
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..", "..", "..", ".."))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from backend.corpora.common.entities.project import Project
+
+
+class TestProject(unittest.TestCase):
+    def setUp(self):
+        self.id = "f753497f-f8a5-48b5-b344-de9ad5a4354d"
```
Bento007, line 13: `self.id` is used by unittest.TestCase.

```suggestion
        self.uuid = "f753497f-f8a5-48b5-b344-de9ad5a4354d"
```

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,22 @@
...
+        self.status = "LIVE"
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
+    def test_get(self):
+        # TODO: write tests
+        project = Project.get((self.id, self.status))
```
Bento007, line 19:
```suggestion
        project = Project.get((self.uuid, self.status))
```

backend/corpora/common/utils/db_utils.py
calvinnhieu: Originally, the reason was to keep DB table logic in DbUtils and keep `Entity` child classes "clean".

Sync'd offline w @Bento007 - agreed to move all project specific logic to `Project` and avoid a `DbUtils` superclass.

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,146 @@
...
+        self.links = kwargs.get("links", [])
+        self.contributors = kwargs.get("contributors", [])
+
+    @classmethod
+    def _query(cls, key: typing.Tuple[str, str]) -> typing.List[Base]:
+        """
+        Given a key, queries the database for a project and its relevant entities.
+
+        According to the Entity.get interface, the return value of this function is
+        fed as the input to Project._parse.
```
maniarathi, line 38: Weird line break here as well... seems like more words can go on the previous line.

maniarathi: There's a few other cases of weird line breaks but I'll stop commenting on them. You can probably do a sweep

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,146 @@
+import typing
```
maniarathi, line 1: Nit for this file, can we avoid using single character names; usually is good practice to fully write out the name for readability (k = key, v = value, etc.).

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+
+        self.assertEqual(project.name, "Single-cell gene expression profiling of SARS-CoV-2 infected human cell lines")
+        self.assertEqual(project.contributors[0]["name"], "Wyler Emanuel")
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
+    def test_get__dne(self):
+        key = ("bad", self.status)
+
+        with self.assertRaises(CorporaException):
+            Project.get(key)
```
maniarathi, line 31: Can you verify the error message here please.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
+    def test_get__invalid_status(self):
+        key = (self.uuid, "bad")
+
+        with self.assertRaises(CorporaException):
+            Project.get(key)
```
maniarathi, line 38: Please verify error message.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_get__ok(self):
+        key = (self.uuid, self.status)
+
+        project = Project.get(key)
+
+        self.assertEqual(project.name, "Single-cell gene expression profiling of SARS-CoV-2 infected human cell lines")
+        self.assertEqual(project.contributors[0]["name"], "Wyler Emanuel")
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
+    def test_get__dne(self):
```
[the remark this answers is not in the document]
maniarathi: Ohhh does not exist.. morning brain. Still worth spelling out though lol

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+        key = ("bad", self.status)
+
+        with self.assertRaises(CorporaException):
+            Project.get(key)
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
+    def test_get__invalid_status(self):
+        key = (self.uuid, "bad")
```
maniarathi, line 35: Sometimes I find explicitly stating what's wrong helpful to understanding why the intentional setup generates the error you are expected.
```suggestion
        key_with_invalid_status = (self.uuid, "nonexistent_project_status")
```

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+        key = (self.uuid, self.status)
+
+        project = Project.get(key)
+
+        self.assertEqual(project.name, "Single-cell gene expression profiling of SARS-CoV-2 infected human cell lines")
+        self.assertEqual(project.contributors[0]["name"], "Wyler Emanuel")
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
+    def test_get__dne(self):
+        key = ("bad", self.status)
```
maniarathi, line 28: Same kind of suggestion as below:
```suggestion
        nonexistent_key = ("invalid_id", self.status)
```

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+sys.path.insert(0, pkg_root)  # noqa
+
+from backend.corpora.common.entities.project import Project
+from backend.corpora.common.utils.exceptions import CorporaException
+
+
+class TestProject(unittest.TestCase):
+    def setUp(self):
+        self.uuid = "f753497f-f8a5-48b5-b344-de9ad5a4354d"
+        self.status = "LIVE"
```
maniarathi, line 15: This is a constant that probably should be in CorporaConstants? Or somewhere else? Best to not hardcode these constants because one day when we change the value, it'll be a nightmare to try and refactor it everywhere.

tests/unit/__init__.py
```diff
@@ -1,4 +1,4 @@
 import os
 
-os.environ["DEPLOYMENT_STAGE"] = "test"
+os.environ["DEPLOYMENT_STAGE"] = "dev"
```
maniarathi, line 3: This actually causes some of the authorizer tests not to run right? If you switch it away from "test"?

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,146 @@
...
+    DbDatasetContributor,
+    DbContributor,
+    ProjectStatus,
+)
+from .entity import Entity
+from ..utils.exceptions import CorporaException
+
+
+class Project(Entity):
+    def __init__(self, **kwargs):
```
Bento007, line 17: There is a lot of redundant packings and unpackings going on. From `_parse` to `__init__` the project variables are packed twice and unpacked twice. It's not efficient. I suggest one of the following:
1.  Naming each argument expected in `__init__` to avoid having to unpack `**kwargs` Then use `**param` on line 125. 
1. Alternatively, pass in `param` from line 125 and unpack it in `__init__`. 
1. Another suggestion is to eliminate `_load` all together and that will reduce the amount of shuffling in half. The output from `_parse` can go straight into `__init__`.

My preference is for 3.

tests/unit/__init__.py
Bento007: Actually it will make them run every time, which we don't want until we fix the issues with running out of generated auth0 access tokens for machine to machine.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
+import os
+import sys
+import unittest
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..", "..", "..", ".."))  # noqa
```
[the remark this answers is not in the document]
Bento007: if we want to run the tests alone it's needed. 

But there is a better way. By adding setting an environment variable `CORPORA_HOME` and pointing to the top level of the project this will be more maintainable. This would become
`pkg_root = os.path.abspath(os.environ["CORPORA_HOME"])

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+sys.path.insert(0, pkg_root)  # noqa
+
+from backend.corpora.common.entities.project import Project
+from backend.corpora.common.utils.exceptions import CorporaException
+
+
```
Bento007, line 11:
```suggestion
@unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
```

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+class TestProject(unittest.TestCase):
+    def setUp(self):
+        self.uuid = "f753497f-f8a5-48b5-b344-de9ad5a4354d"
+        self.status = "LIVE"
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
```
Bento007, line 17: You can move this decorator to `class TestProject` if it's used on all of the tests.

backend/corpora/common/utils/db_utils.py
```diff
@@ -17,7 +17,7 @@ def _get(self, table: Base, entity_id: str) -> typing.Union[Base, None]:
         """
         return self.session.query(table).get(entity_id)
 
-    def _query(self, table_args: typing.List[Base], filter_args: typing.List[bool] = None) -> typing.List[Base]:
+    def query(self, table_args: typing.List[Base], filter_args: typing.List[bool] = None) -> typing.List[Base]:
```
Bento007, line 20: Good idea

tests/unit/backend/corpora/common/entities/test_project.py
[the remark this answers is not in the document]
calvinnhieu: @maniarathi Yup as @Bento007 says (thanks for providing the explanation), it's a convention we've used to run standalone tests (i.e. execution dir becomes test file dir as opposed to top level repo dir). I've found it useful in my development workflow so I continue to inject the suspicious code 🙃.

@Bento007 that's a great idea. This pattern is used in multiple files - can we create a ticket to perform this refactor outside of this PR? Granted @maniarathi is on board with keeping the pattern.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -0,0 +1,38 @@
...
+    def test_get__ok(self):
+        key = (self.uuid, self.status)
+
+        project = Project.get(key)
+
+        self.assertEqual(project.name, "Single-cell gene expression profiling of SARS-CoV-2 infected human cell lines")
+        self.assertEqual(project.contributors[0]["name"], "Wyler Emanuel")
+
+    @unittest.skipIf(True, "Only runnable on local dev env via bastion tunnel. Comment this line to run.")
```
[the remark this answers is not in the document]
calvinnhieu: Nice! While it doesn't guarantee the tunnel is open, it's a start. I'm going to leave this however, as I have a parallel PR that will update these tests to run against a local DB making all of these tests obsolete.

backend/corpora/common/entities/project.py
calvinnhieu: Gotcha - yea I was prematurely line breaking before the char limit, but I'll just go by our configured line char limit.

tests/unit/backend/corpora/common/entities/test_project.py
calvinnhieu: This test file is now deprecated. Several constants defined in the ORM are defined as Enums in `corpora_orm.py`. I noticed you use `namedtuple`s as alternative to this, and they seem to be more lightweight/simpler usage. I would be open to moving the ORM Enum constants to `namedtuple`s `CorporaConstants` in separate refactor PR. Thoughts?

tests/unit/__init__.py
calvinnhieu: Uhh yea this undesired. Deprecating this change - keeping as `test` in this PR. will need to reconcile with Trent's current work.

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,146 @@
...
+
+            contributors[result.DbContributor.id] = {
+                "id": result.DbContributor.id,
+                "name": result.DbContributor.name,
+                "institution": result.DbContributor.institution,
+                "email": result.DbContributor.email,
+            }
+
+        # build dict of Project parameters
+        project = {}
```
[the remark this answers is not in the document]
calvinnhieu: Is this necessary?

backend/corpora/common/entities/project.py
calvinnhieu: Fair point on packing and unpacking. A combination of 1 and 3 seems reasonable! The way `Project` is currently being instantiated felt bloated to add to `_parse` which can get unwieldy. With one, it looks like that can be done in 1 line which I'm ok with.

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: I think enum is the more pythonic way of tracking differences in state. Just import the Enum from corpora_orm.py.

backend/corpora/common/entities/project.py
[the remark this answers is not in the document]
Bento007: It avoid a loop which unpacks all of the information in this dictionary into another dictionary. You could also do `project = query_results[0].DbProject.__dict__.copy()` which does what the loop bellow does.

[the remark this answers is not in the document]
calvinnhieu: Got it! Done

tests/unit/backend/corpora/common/entities/test_project.py
calvinnhieu: Deprecated

backend/corpora/common/entities/project.py
calvinnhieu: Done

tests/unit/backend/corpora/common/entities/test_project.py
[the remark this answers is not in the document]
maniarathi: I'm on board; thanks for the discussion!

tests/unit/backend/corpora/common/entities/test_project.py
maniarathi: Yep, I think @Bento007 's suggestion makes sense. The enums are specific to the database so it makes sense to me to import it directly from the ORM.

calvinnhieu: Done

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,136 @@
...
+        project_params = query_results[0].DbProject.__dict__.copy()
+        project_params.pop("_sa_instance_state")
+        project_params["dataset_ids"] = list(dataset_ids)
+        project_params["links"] = list(links.values())
+        project_params["contributors"] = list(contributors.values())
+
+        # instantiate Project
+        return Project(**project_params)
+
+    @classmethod
```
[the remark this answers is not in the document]
Bento007: list and save

tests/unit/backend/corpora/common/entities/test_project.py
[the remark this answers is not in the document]
calvinnhieu: [Ticket](https://app.zenhub.com/workspaces/corpora-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/406). Resolving

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,129 @@
...
+
+            contributors[result.DbContributor.id] = {
+                "id": result.DbContributor.id,
+                "name": result.DbContributor.name,
+                "institution": result.DbContributor.institution,
+                "email": result.DbContributor.email,
+            }
+
+        # build dict of Project parameters
+        project_params = query_results[0].DbProject.__dict__.copy()
```
maniarathi, line 122: You're probably going to want to deepcopy here instead of shallow copy.

backend/corpora/common/entities/project.py
```diff
@@ -0,0 +1,129 @@
...
+        project_params.pop("_sa_instance_state")
```
maniarathi, line 123: What's this doing?

backend/corpora/common/entities/project.py
calvinnhieu: Done

backend/corpora/common/entities/project.py
calvinnhieu: Removes SQLAlchemy state field to simplify logic for Project instantiation. Added comment.

Issue #352, opened 2020-06-09
[FE] Add TypeScript


tihuan:
Currently our Gatsby FE is using plain JavaScript, but it seems like everyone is interested in using TypeScript for it. So I'll convert it by following https://blog.logrocket.com/set-up-a-typescript-gatsby-app/ and other docs as needed 😄 Thank you!

PR #407, opened 2020-06-29
Add typescript

tihuan:
This PR closes: #352 

1. Add TypeScript
1. Add ESLint + TS plugin
1. Configure TS
1. Add StyleLint
1. Add `Concurrently` to run linter checks in parallel
1. Add [`eslint-plugin-sonarjs`](https://github.com/SonarSource/eslint-plugin-sonarjs) for code smell detection (**triggers a new GPL3 warning by Snyk. This is fine, since data-portal is OSS anyway**)
1. Add [`prettier`](https://github.com/prettier/prettier) for consistent code styling
1. Convert `.js` files to `.ts` and `.tsx`
1. Add a few type definitions needed to provide basic type check
1. Address some code climate refactoring suggestions
1. Add absolute import for modules. E.g., from `import Foo from '../../../../../foo/bar'` to `import Foo from 'src/foo/bar'`
1. Add `prettier-plugin-organize-imports` to auto sort imports


tihuan:
Thanks so much for the review, @Bento007 😄 !

Do you know what `backend` `frontend` checks are?

Screen Shot 2020-07-02 at 1 24 43 PM

It almost feels like deployment, according to: `.github/workflows/deploy.yml`, but given that the PR is `add-typescript` not sure it should trigger the deployment flows that should only happen on `main`?

Not urgent, just wondering lol

Thank you!

PR #409, opened 2020-06-29
Create dataset application class

### LINKED ISSUE #393: Implement Dataset application-level object

Bento007:
- Adding dataset application class
- Adding TransformBase to simplify transforming DBobjects
- Adding custom JSON encoder to handle dateTime object return from DB .

closes #393

### DIFF
--- a/backend/corpora/common/corpora_orm.py
+++ b/backend/corpora/common/corpora_orm.py
@@ -232,7 +232,7 @@ class DbDataset(Base):
 
     artifacts = relationship("DbDatasetArtifact", back_populates="dataset")
     deployment_directories = relationship("DbDeploymentDirectory", back_populates="dataset")
-    contributors = relationship("DbDatasetContributor")
+    contributors = relationship("DbContributor", secondary=lambda: DbDatasetContributor().__table__)
 
 
 class DbDatasetArtifact(Base):
@@ -303,6 +303,3 @@ class DbDatasetContributor(Base):
     dataset_id = Column(ForeignKey("dataset.id"), nullable=False)
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
-
-    contributor = relationship("DbContributor")
-    dataset = relationship("DbDataset", back_populates="contributors")
--- a/backend/corpora/common/corpora_orm.py
+++ b/backend/corpora/common/corpora_orm.py
@@ -230,6 +230,10 @@ class DbDataset(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
+    artifacts = relationship("DbDatasetArtifact", back_populates="dataset")
+    deployment_directories = relationship("DbDeploymentDirectory", back_populates="dataset")
+    contributors = relationship("DbDatasetContributor")
+
 
 class DbDatasetArtifact(Base):
     """
@@ -249,7 +253,7 @@ class DbDatasetArtifact(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
-    dataset = relationship("DbDataset")
+    dataset = relationship("DbDataset", back_populates="artifacts")
 
 
 class DbDeploymentDirectory(Base):
@@ -267,7 +271,7 @@ class DbDeploymentDirectory(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
-    dataset = relationship("DbDataset")
+    dataset = relationship("DbDataset", back_populates="deployment_directories")
 
 
 class DbContributor(Base):
@@ -299,3 +303,6 @@ class DbDatasetContributor(Base):
     dataset_id = Column(ForeignKey("dataset.id"), nullable=False)
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
+
+    contributor = relationship("DbContributor")
+    dataset = relationship("DbDataset", back_populates="contributors")
--- a/backend/corpora/common/corpora_orm.py
+++ b/backend/corpora/common/corpora_orm.py
@@ -22,7 +22,6 @@ sys.path.insert(0, pkg_root)  # noqa
 
 from common.corpora_config import CorporaDbConfig
 
-
 Base = declarative_base()
 deployment_stage = os.environ["DEPLOYMENT_STAGE"]
 
@@ -232,7 +231,9 @@ class DbDataset(Base):
 
     artifacts = relationship("DbDatasetArtifact", back_populates="dataset")
     deployment_directories = relationship("DbDeploymentDirectory", back_populates="dataset")
-    contributors = relationship("DbContributor", secondary=lambda: DbDatasetContributor().__table__)
+    contributors = relationship(
+        "DbContributor", secondary=lambda: DbDatasetContributor().__table__, back_populates="datasets"
+    )
 
 
 class DbDatasetArtifact(Base):
@@ -287,6 +288,9 @@ class DbContributor(Base):
     email = Column(String)
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
+    datasets = relationship(
+        "DbDataset", secondary=lambda: DbDatasetContributor().__table__, back_populates="contributors"
+    )
 
 
 class DbDatasetContributor(Base):
--- a/backend/corpora/common/entities/dataset.py
+++ b/backend/corpora/common/entities/dataset.py
@@ -15,7 +15,7 @@ class Dataset(Entity):
     @classmethod
     def _query(cls, key: str) -> typing.List["Dataset"]:
         """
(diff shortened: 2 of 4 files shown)


backend/corpora/common/corpora_orm.py
```diff
@@ -230,6 +270,10 @@ class DbDataset(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
+    artifacts = relationship("DbDatasetArtifact", back_populates="dataset")
```
maniarathi, line 273: We probably also want to add the `cascade` argument to inform that children in a relationship should be affected when the parent changes (i.e. gets deleted). Ditto in all the child relationships in this file.

backend/corpora/common/corpora_orm.py
```diff
@@ -14,18 +15,57 @@
...
 from common.corpora_config import CorporaDbConfig
+from common.utils.json import CustomJSONEncoder
 
 
-Base = declarative_base()
-deployment_stage = os.environ["DEPLOYMENT_STAGE"]
+class TransformingBase(object):
+    """
+    Add functionality to transform a Base object, and recursively transform its linked entities.
+
```
maniarathi, line 32:
```suggestion
```

backend/corpora/common/corpora_orm.py
```diff
@@ -14,18 +15,57 @@
...
+
```
maniarathi, line 33:
```suggestion
```

backend/corpora/common/entities/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+from .entity import Entity
+from ..corpora_orm import DbDataset
+
+
+class Dataset(Entity):
+    tables = [
+        DbDataset,
+    ]
+
+    def __init__(self, db_obj: DbDataset):
```
maniarathi, line 12: I'm not sure that the Dataset entity class should be initialized with a DbDataset. The primary use case of this class is to be able to take in the input from a user who is inputting a dataset and parse it into a Dataset object that can be transformed into a DbDataset object to be placed into the database. For this, we actually do need all the attributes available to set similar to the function signature in project.py.

backend/corpora/common/corpora_orm.py
```diff
@@ -14,18 +15,57 @@
...
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
 from common.corpora_config import CorporaDbConfig
+from common.utils.json import CustomJSONEncoder
 
 
-Base = declarative_base()
-deployment_stage = os.environ["DEPLOYMENT_STAGE"]
+class TransformingBase(object):
```
maniarathi, line 29: I'm curious about this class as a whole. I don't see `to_dict` or `to_json` being used anywhere -- what's the use case you're thinking of here and why specifically convert the database object to dictionary and JSON instead of the application layer objects?

Also what's the benefit of having a class in the ORM layer versus a utility function that could perform the transformation?

Bento007: The primary use case is when a user make an api request they will want a json response. 

The benefit of having it in the ORM layer is that we only need one conversion function for all of the DB objects. It also allows us to follow the relationships defined by the ORM and convert those object as well in a single function.

backend/corpora/common/corpora_orm.py
Bento007: Good idea, I'll add to all of the children in this class.

backend/corpora/common/entities/dataset.py
Bento007: In that case I think we should create a function call `new` or `create` that takes in all of those parameters, and  generates a DBDataset to pass into the `__init__`. It would be similar to get, except no querying.

backend/corpora/common/corpora_orm.py
Bento007: On second thought I think this should wait until we are handling deletion, since this feature will not be tested until then.

backend/corpora/common/entities/dataset.py
Bento007: These details can be fleshed out when we add save functionality.

backend/corpora/common/corpora_orm.py
```diff
@@ -267,7 +271,7 @@ class DbDeploymentDirectory(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
-    dataset = relationship("DbDataset")
+    dataset = relationship("DbDataset", back_populates="deployment_directories")
 
 
 class DbContributor(Base):
```
maniarathi, line 277: Missing a relationship here to DbDatasetContributor

backend/corpora/common/corpora_orm.py
```diff
@@ -299,3 +303,6 @@ class DbDatasetContributor(Base):
     dataset_id = Column(ForeignKey("dataset.id"), nullable=False)
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
+
+    contributor = relationship("DbContributor")
```
maniarathi, line 307: You'll have to back_populate here once the relationship is added to DbContributors.

backend/corpora/common/entities/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+        self.db_obj = db_obj
+
+    @classmethod
+    def _query(cls, key: str) -> typing.List["Dataset"]:
```
maniarathi, line 16: Add docstring? I find it good practice to have it even for "self-explanatory" functions because it forces that the implementation matches what we meant to do written in plain English.

backend/corpora/common/entities/dataset.py
```diff
@@ -0,0 +1,28 @@
...
+        result = cls.db.query(table_args=[DbDataset], filter_args=[key == DbDataset.id])
+        return result
+
+    @classmethod
+    def _load(cls, db_result: typing.List["Dataset"]) -> "Dataset":
```
maniarathi, line 21: Docstring please?

tests/unit/backend/corpora/fixtures/database/__init__.py
```diff
@@ -98,11 +100,17 @@ def _create_test_datasets(self):
...
+            dataset_id=test_dataset_id,
         )
         self.db.session.add(project_dataset)
         self.db.session.commit()
 
+        deployment_directory = DbDeploymentDirectory(
+            id="test_deployment_directory_id", dataset_id=test_dataset_id, environment="test", url="test_url"
+        )
+        self.db.session.add(deployment_directory)
+        self.db.session.commit()
```
maniarathi, line 112: Since you've added a `commit` here, you could delete line 106 so that a single transaction occurs within this single function.

backend/corpora/common/corpora_orm.py
```diff
@@ -230,6 +230,10 @@ class DbDataset(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
```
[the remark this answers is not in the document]
Bento007: added

backend/corpora/common/corpora_orm.py
Bento007: is that relationship needed here

tests/unit/backend/corpora/fixtures/database/__init__.py
Bento007: fixing

backend/corpora/common/entities/dataset.py
Bento007: fixed

backend/corpora/common/entities/dataset.py
Bento007: fixed

Bento007:
adding issue to apply cascade to relationships https://github.com/chanzuckerberg/corpora-data-portal/issues/414

backend/corpora/common/corpora_orm.py
```diff
@@ -230,6 +230,10 @@ class DbDataset(Base):
...
 
+    artifacts = relationship("DbDatasetArtifact", back_populates="dataset")
+    deployment_directories = relationship("DbDeploymentDirectory", back_populates="dataset")
+    contributors = relationship("DbDatasetContributor")
```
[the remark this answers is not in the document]
Bento007: fix i think

Bento007:
For the issues with project, I will save that for a future PR.

Bento007:
In the sqlAlchemy is suggest using back_populates. If you would like to implement using backref, I wouldn't be offended, but I don't see a compelling reason to make the change at this moment.

backend/corpora/common/entities/dataset.py
```diff
@@ -0,0 +1,38 @@
...
+        DbDataset,
+    ]
+
+    def __init__(self, db_obj: DbDataset):
+        self.db_obj = db_obj
+
+    @classmethod
+    def _query(cls, key: str) -> typing.List["Dataset"]:
+        """
+        Queries the database for required entity data
```
maniarathi, line 18:
```suggestion
        Given a key representing a dataset UUID, queries the database to return the associated dataset data.
```

backend/corpora/common/entities/dataset.py
```diff
@@ -0,0 +1,38 @@
...
+        :param key: dataset.id is the primary key.
+        :return: list of query result rows
+        """
+        result = cls.db.query(table_args=[DbDataset], filter_args=[key == DbDataset.id])
+        return result
+
+    @classmethod
+    def _load(cls, db_result: typing.List["Dataset"]) -> "Dataset":
+        """
+        Parses a database query result into an Entity instance
```
maniarathi, line 28:
```suggestion
        Parses a database query result into a Dataset Entity instance.
```

PR #410, opened 2020-06-30
Setup Swagger UI for Corpora API

maniarathi:


Makefile
```diff
@@ -12,7 +12,11 @@ lint:
 unit-test:
 	-docker run -d -p 5432:5432 --name test_db -e POSTGRES_PASSWORD=test_pw postgres
 	PYTHONWARNINGS=ignore:ResourceWarning python3 -m coverage run \
-		-m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose
+		-m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose || make clean_test_db
```
mckinsel, line 15: I'm not sure I follow here. You only want to clean up the test container if python fails?

maniarathi:
Gonna try and address all the comments in one go:

@Bento007 --

RestyResolver: I actually made a conscious decision not to use RestyResolver because I find the directories and function names to reduce readability versus a function name like `get_project_details` which I find much more readable and clearly states the expected results (versus perhaps a `get` function where you now need to check out the path to figure out what it is getting). I also don't see much benefit of RestyResolver besides the routing. Perhaps you know of other benefits?

Add logging: Done; good idea!

X-AWS-REQUEST-ID: Done! Good call!


@Bento007 and @mckinsel --

I fixed up the Makefile - I was trying to force the cleanup even after the tests failed which wasn't occurring initially but then figured out you just add a dash at the beginning to force the commands to continue. I now realize that the initial PR didn't clean up even if the tests pass. It's fixed now.


@mckinsel --

Yes, it's a bunch of boilerplate code mostly stolen from previous projects haha.

Makefile
```diff
@@ -11,8 +11,14 @@ lint:
 .PHONY: unit-test
 unit-test:
 	-docker run -d -p 5432:5432 --name test_db -e POSTGRES_PASSWORD=test_pw postgres
-	PYTHONWARNINGS=ignore:ResourceWarning python3 -m coverage run \
+	-PYTHONWARNINGS=ignore:ResourceWarning python3 -m coverage run \
 		-m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose
+	-make clean_test_db
```
mckinsel, line 16: I think you're supposed to use `$(MAKE)` for reasons, see https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html

And `clean_test_db` is also `PHONY` I think.

But I think the bigger issue is that `make unit-test` will have a zero exit code even if the test fails, which will make the tests green in a lot of frameworks.

mckinsel: Maybe you can do like a
```
python3 -m ... ; test_result=$$?; $(MAKE) clean_test_db; exit $$test_result
```
or something like that.

maniarathi: TIL. Thanks for the tips - I've switched over to the commands you suggested.

Makefile
```diff
@@ -11,8 +11,16 @@ lint:
 .PHONY: unit-test
 unit-test:
 	-docker run -d -p 5432:5432 --name test_db -e POSTGRES_PASSWORD=test_pw postgres
-	PYTHONWARNINGS=ignore:ResourceWarning python3 -m coverage run \
+	-PYTHONWARNINGS=ignore:ResourceWarning python3 -m coverage run \
 		-m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose
+	test_result=$$?
+	$(MAKE) clean_test_db
+	exit $$test_result
```
mckinsel, line 18: I believe every line of a makefile is executed in a separate shell, so this all needs to be on one line, and we want the exit code from python, so no `-`:

```
PYTHONWARNINGS=ignore:ResourceWarning python3 -m coverage run \
	-m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose; \
test_result=$$?; \
$(MAKE) clean_test_db; \
exit $$test_result
```

maniarathi: Thanks for walking me through it! Seems to work after I purposefully failed a test:

```
Ran 50 tests in 15.007s

FAILED (failures=1)
docker stop test_db
test_db
docker rm test_db
test_db
make: *** [unit-test] Error 1
```

PR #415, opened 2020-07-01
Refactor project to match dataset

Bento007:
By utilizing SQLAlchemy Object Relationship Mapping, CRUD operations will be easier to manage across the various tables.

- Adding more shared code to Entity
- Using SQLAlchemy relationships to query for values in other tables for DBProject
- Factoring out common code from Dataset

### DIFF
--- a/backend/corpora/common/corpora_orm.py
+++ b/backend/corpora/common/corpora_orm.py
@@ -136,6 +136,9 @@ class DbUser(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
+    # Relationships
+    projects = relationship("DbProject", back_populates="user")
+
 
 class DbProject(Base):
     """
@@ -158,8 +161,9 @@ class DbProject(Base):
     created_at = Column(DateTime(True), nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime(True), nullable=False, server_default=DEFAULT_DATETIME)
 
-    user = relationship("DbUser", uselist=False)
-    links = relationship("DbProjectLink")
+    # Relationships
+    user = relationship("DbUser", uselist=False, back_populates="projects")
+    links = relationship("DbProjectLink", back_populates="projects")
     datasets = relationship("DbDataset", secondary=lambda: DbProjectDataset().__table__, back_populates="project")
 
 
@@ -198,6 +202,9 @@ class DbProjectLink(Base):
     created_at = Column(DateTime(True), nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime(True), nullable=False, server_default=DEFAULT_DATETIME)
 
+    # Relationships
+    projects = relationship("DbProject", back_populates="links")
+
     # Composite FK
     __table_args__ = (ForeignKeyConstraint([project_id, project_status], [DbProject.id, DbProject.status]), {})
 
@@ -231,6 +238,7 @@ class DbDataset(Base):
     created_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime, nullable=False, server_default=DEFAULT_DATETIME)
 
+    # Relationships
     project = relationship("DbProject", secondary=lambda: DbProjectDataset().__table__, back_populates="datasets")
     artifacts = relationship("DbDatasetArtifact", back_populates="dataset")
     deployment_directories = relationship("DbDeploymentDirectory", back_populates="dataset")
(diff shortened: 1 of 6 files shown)


brianraymor:
@Bento007 - what issue should this PR be connected to?

backend/corpora/common/entities/entity.py
```diff
@@ -9,12 +9,19 @@ class Entity:
     An abstract base class providing an interface to parse application-level objects to and from their
     database counterparts.
 
+    This class uses a has-a relationship with SQLAlchemy Table object and simplify the CRUD operations performed on the
+    database through these objects. Columns and relationship of the database object can be access as attributes of the
```
maniarathi, line 13:
```suggestion
    database through these objects. Columns and relationships of the database object can be access as attributes of the
```

backend/corpora/common/entities/entity.py
```diff
@@ -34,13 +41,23 @@ def _query(cls, key: typing.Union[str, typing.Tuple[str, str]]) -> typing.List[B
         raise NotImplementedError()
 
     @classmethod
-    def _load(cls, db_result: typing.List[Base]) -> "Entity":
+    def _load(cls, db_result: typing.List["Entity"]) -> typing.Union["Entity", None]:
         """
-        Parses a database query result into an Entity instance
-        :param db_result: list of query result rows
+        Parses query result rows produced by Entity._query into an instantiated Entity object.
```
maniarathi, line 46:
```suggestion

        Parses query result rows produced by Entity._query into an instantiated Entity object.
```
Just adding a line break for readability.

backend/corpora/common/entities/entity.py
```diff
@@ -34,13 +41,23 @@ def _query(cls, key: typing.Union[str, typing.Tuple[str, str]]) -> typing.List[B
...
+        row = query_results[0]
+        project_id = row.DbProject.id
+        :param query_results: list of query result rows
         :return: Entity
         """
-        raise NotImplementedError()
+        try:
+            return cls(db_result[0])
+        except IndexError:
+            return None
```
maniarathi, line 60: Could `log.info` here so that in our logs we know that we weren't able to find a particular result.

backend/corpora/common/entities/entity.py
```diff
@@ -34,13 +41,23 @@ def _query(cls, key: typing.Union[str, typing.Tuple[str, str]]) -> typing.List[B
...
+
+        Example query results access:
+        row = query_results[0]
+        project_id = row.DbProject.id
+        :param query_results: list of query result rows
         :return: Entity
         """
-        raise NotImplementedError()
+        try:
+            return cls(db_result[0])
```
maniarathi, line 58: Might be worth logging when there are multiple results because that would indicate a data anomaly that we should probably be aware of.

backend/corpora/common/entities/entity.py
```diff
@@ -34,13 +41,23 @@ def _query(cls, key: typing.Union[str, typing.Tuple[str, str]]) -> typing.List[B
...
+        If no results are found, None is returned.
+
+        Example query results access:
+        row = query_results[0]
+        project_id = row.DbProject.id
+        :param query_results: list of query result rows
         :return: Entity
         """
-        raise NotImplementedError()
+        try:
```
maniarathi, line 57:
```suggestion

        try:
```
Another line break for readability :)

backend/corpora/common/corpora_orm.py
```diff
@@ -158,7 +158,9 @@ class DbProject(Base):
     created_at = Column(DateTime(True), nullable=False, server_default=DEFAULT_DATETIME)
     updated_at = Column(DateTime(True), nullable=False, server_default=DEFAULT_DATETIME)
 
-    user = relationship("DbUser")
+    user = relationship("DbUser", uselist=False)
```
maniarathi, line 161: Are we still committing to back_populating everywhere for consistency?

maniarathi: Ditto for links below

Bento007: yeah let me add that in.

backend/corpora/common/entities/entity.py
Bento007: When search via the primary key this should be impossible. I'll modify the code to reflect this.

backend/corpora/common/entities/entity.py
Bento007: Will add

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -23,10 +35,4 @@ def test__get__does_not_exist(self):
 
     def test__get__invalid_status(self):
         invalid_status_key = (self.uuid, "invalid_status")
-
```
maniarathi, line 26: I think it's worth retaining the check to make sure the logging was successful - I do this in the DatasetValidator tests to make sure that the right error is logged.

Bento007: Will do

Bento007: fixed

Issue #324, opened 2020-05-29
Implement Corpora API to write project and dataset metadata to persistent store


maniarathi:
Implements part of the `POST /v1/submission/{project_uuid}/save` API.

This ticket is specifically implementing one of two code paths for the above API. This is the code path where given a dataset, we will extract the project-level metadata and dataset-level metadata from the matrix file (as specified [here](https://docs.google.com/document/d/1_I2dnYuCTvO2yABR0FnLRnI8RnMhiASC5Ny4ZwDjsCI/edit) and populate the RDS. Even more specifically, this ticket requires the extraction of metadata from H5AD files ( @mckinsel will be able to provide information about where in the H5AD file you will be able to find each piece of metadata). 

Here is a sample H5AD to prototype with: s3://cellxgene-data-wrangling-prod/Single_cell_gene_expression_profiling_of_SARS_CoV_2_infected_human_cell_lines_Calu_3-28/Single_cell_gene_expression_profiling_of_SARS_CoV_2_infected_human_cell_lines_Calu_3-28-remixed.h5ad

This ticket does not require the implementation of the second path which is to input the project details via the body of the API. This will be implemented at a later date via a different ticket.

maniarathi:
@Bento007 to breakdown further and estimate child tickets.

PR #421, opened 2020-07-03
Create Project and Dataset

Bento007:
Adding creation functions for new Dataset and Project entities. A unique Id is generated when using create for projects and datasets. Existing IDs can be used during joins for example with contributors and datasets.

- For generating unique ids a retry limit of 3 times was arbitrarily set. This can be change by modifying `DbUtils.retry_limit`.


backend/corpora/common/corpora_orm.py
```diff
@@ -163,7 +163,7 @@ class DbProject(Base):
 
     # Relationships
     user = relationship("DbUser", uselist=False, back_populates="projects")
-    links = relationship("DbProjectLink", back_populates="projects")
+    links = relationship("DbProjectLink", back_populates="project")
```
calvinnhieu, line 166: `project` or `projects`?

calvinnhieu: Actually, the correction would be on line 165?

backend/corpora/common/entities/project.py
```diff
@@ -1,9 +1,65 @@
...
+        s3_bucket: str = "",
+        tc_uri: str = "",
+        needs_attestation: bool = False,
+        processing_state: str = "",
+        validation_state: str = "",
+        links: list = None,
+    ) -> "Project":
+        """
+        Need to check if one exists before creating
+        :param id:
```
maniarathi, line 27: Delete lines 27-38 as they don't provide additional context to the docstring.

backend/corpora/common/utils/db_utils.py
```diff
@@ -29,3 +31,25 @@ def query(self, table_args: typing.List[Base], filter_args: typing.List[bool] =
...
+    def generate_id(self, table: Base, *args):
+        """
+        Generates a UUID to enter into the specified table.
+
+        :param table: The table to generate the uuid  for
+        :param arg: Additional primary keys if needed
+        :return:
+        """
+        # Generate the ID
+        retry_attempts = 0
```
maniarathi, line 46: Is the retry here actually necessary? The chance of collision here based on what I've read about uuid4 in the past is extraordinarily small so it seems like unnecessary code. If there does happen to be a collision, it'll get caught at the `commit` layer at which point we bubble up the error to ask the user to re-submit.

backend/corpora/common/corpora_orm.py
mckinsel: The code as proposed seems right to me. Project:ProjectLink and User:Project are both 1:N

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -28,3 +35,91 @@ def test__get__ok(self):
...
+                Dataset.db.session.expire_all()
+
+                dataset = Dataset.get(dataset_id)
+                self.assertIsNotNone(dataset)
+                self.assertEqual(dataset_id, dataset.id)
+                self.assertEqual(artifact_ids, get_ids(dataset.artifacts))
+                self.assertEqual(deployment_directory_ids, get_ids(dataset.deployment_directories))
+                self.assertEqual(contributor_ids, get_ids(dataset.contributors))
+
+        with self.subTest("With existing rows"):
```
maniarathi, line 98: This warrants its own test. It's good practice to keep tests as separate as possible even if there is overlap.

backend/corpora/common/entities/project.py
```diff
@@ -1,9 +1,65 @@
...
+            s3_bucket=s3_bucket,
+            tc_uri=tc_uri,
+            needs_attestation=needs_attestation,
+            processing_state=processing_state,
+            validation_state=validation_state,
+            links=cls._create_sub_objects(links, DbProjectLink, project_id=uuid, project_status=status),
+        )
+
+        cls.db.session.add(new_db_object)
+        cls.db.session.commit()
```
maniarathi, line 64: You'll probably want to catch exceptions here and rollback if the commit was not successful.

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
+        contributors = cls._create_sub_objects(contributors, DbContributor)
+        contributor_dataset_ids = [dict(contributor_id=contributor.id, dataset_id=uuid) for contributor in contributors]
+        dataset_contributor = cls._create_sub_objects(contributor_dataset_ids, DbDatasetContributor)
+
+        cls.db.session.add(new_db_object)
+        cls.db.session.add_all(contributors)
+        cls.db.session.commit()
+
+        cls.db.session.add_all(dataset_contributor)
+        cls.db.session.commit()
```
maniarathi, line 104: You might want to catch exceptions here if the commit fails and log the errors as well as rollback the transaction.

tests/unit/backend/corpora/common/entities/__init__.py
```diff
@@ -0,0 +1,2 @@
+def get_ids(iterable):
```
maniarathi, line 1: Would prefer this to be in a utility class rather than hidden in the `__init__.py` file.

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -28,3 +35,91 @@ def test__get__ok(self):
     def test__get__does_not_exist(self):
         non_existent_key = "non_existent_id"
         self.assertEqual(Dataset.get(non_existent_key), None)
+
+    def test__create__ok(self):
+        """
+        Create a dataset with a variable number of artifacts, contributors, and deployment_directories
+
```
maniarathi, line 42:
```suggestion
```

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -28,3 +35,91 @@ def test__get__ok(self):
...
+            disease_ontology="123",
+            sex="F",
+            ethnicity="ethnicity",
+            ethnicity_ontology="123",
+            source_data_location="location",
+            preprint_doi="preprint",
+            publication_doi="publication",
+        )
+
+        for i in range(3):
```
maniarathi, line 74: Curious, why 3? What are the different test cases that are tested by doing it three times?

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -40,3 +53,53 @@ def test__get__invalid_status(self):
...
+            description="test",
+            owner="test_user_id",
+            s3_bucket="s3://fakebucket",
+            tc_uri="https://fakeurl",
+            needs_attestation=False,
+            processing_state=ProcessingState.IN_VALIDATION.name,
+            validation_state=ValidationState.NOT_VALIDATED.name,
+        )
+
+        for i in range(3):
```
maniarathi, line 76: Ditto here. Why 3?

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -40,3 +53,53 @@ def test__get__invalid_status(self):
...
+
+                # Expire all local object and retireve them from the DB to make sure the transactions went through.
+                Project.db.session.expire_all()
+
+                project = Project.get(project_key)
+                self.assertIsNotNone(project)
+                self.assertEqual(project_key, (project.id, project.status))
+                self.assertEqual(link_ids, get_ids(project.links))
+
+        with self.subTest("With existing rows"):
```
maniarathi, line 91: Ditto here. Would prefer separate test.

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
+        :param deployment_directories:
+        :return:
+        """
+        uuid = cls.db.generate_id(DbDataset)
+
+        """
+        Prevent accidentally linking an existing row to a different Dataset. This maintains the relationship of one
+        to many for artifacts and deployment_directories
+        """
+        artifacts = artifacts if artifacts else []
```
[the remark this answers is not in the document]
Bento007: It's dangerous to set function defaults to mutable types such as a list. Run this a few times and see what happens to `a`:
```python
def foo(bar=[]):
    bar.append("stuff")
    return bar
```

This will lead to unexpected behavior.

backend/corpora/common/utils/db_utils.py
```diff
@@ -29,3 +31,25 @@ def query(self, table_args: typing.List[Base], filter_args: typing.List[bool] =
...
+        while retry_attempts < self.retry_limit:
+            _id = str(uuid.uuid4())
+            key = _id if not args else (_id, *args)
+            if self.get(table, key):
```
mckinsel, line 50: A collision is rather unlikely, no?

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: It's the trinity. Everything is three. Also since python range starts at 0, we are effectively testing 0,1, and 2 foreign keys linked to the dataset. Those are usually the critical test areas.

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: see above

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
+            tissue=tissue,
+            tissue_ontology=tissue_ontology,
+            assay=assay,
+            assay_ontology=assay_ontology,
+            disease=disease,
+            disease_ontology=disease_ontology,
+            sex=sex,
+            ethnicity=ethnicity,
+            ethnicity_ontology=ethnicity_ontology,
+            source_data_location=source_data_location,
```
mckinsel, line 85: This isn't really part of this PR, but what does "source_data_location" represent for a dataset? It's currently only in the schema at the project level.

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
     table = DbDataset
 
     def __init__(self, db_object: DbDataset):
         super().__init__(db_object)
+
+    @classmethod
+    def create(
+        cls,
+        revision: int = 0,
+        name: str = "",
```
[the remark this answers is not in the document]
mckinsel: (Maybe this is just a separate issue to file. You're matching `DbDataset` here)

backend/corpora/common/corpora_orm.py
Bento007: I will keep the naming as is

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
Bento007: That would be a separate issue. We just need to decide what we normalize everything to.

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
+        source_data_location: str = "",
+        preprint_doi: str = "",
+        publication_doi: str = "",
+        artifacts: list = None,
+        contributors: list = None,
+        deployment_directories: list = None,
+    ) -> "Dataset":
+        """
+        Creates a new dataset and related objects
+        :param revision:
```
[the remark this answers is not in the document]
Bento007: will remove

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
+        :param preprint_doi:
+        :param publication_doi:
+        :param artifacts:
+        :param contributors:
+        :param deployment_directories:
+        :return:
+        """
+        uuid = cls.db.generate_id(DbDataset)
+
+        """
```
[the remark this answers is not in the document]
Bento007: done

backend/corpora/common/entities/dataset.py
Bento007: I don't have an answer. @maniarathi any clues?

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,106 @@
...
+        Prevent accidentally linking an existing row to a different Dataset. This maintains the relationship of one
```
calvinnhieu, line 60: Is this comment misleading? Lines 63-65 just set `None`s from the defaults in the func sig to `[]`

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +62,30 @@ def __getattr__(self, name):
         :return:
         """
         return self.db_object.__getattribute__(name)
+
+    @classmethod
+    def _create_sub_objects(
+        cls, rows: typing.List[dict], db_table: Base, keys: typing.List[str] = None, **kwargs
```
calvinnhieu, line 68: It isn't clear to me the usage of each of these parameters from reading the function and current documentation, namely, `rows`, `keys` and `kwargs`. I would recommend adding more examples/adding clarification to the docstring. Suggestions below

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +62,30 @@ def __getattr__(self, name):
...
+    ) -> typing.List[Base]:
+        """
+        Create a list of Table Rows to be added to an Entity Object during an object creation. If id is provided, then
```
calvinnhieu, line 71: "Create or modify N rows in `db_table` to be added to..."

calvinnhieu, line 71:
```suggestion
        Create or modify N `rows` in `db_table` associated with Entity Object during object creation. If id is provided, then
```

backend/corpora/common/entities/project.py
Bento007: done

backend/corpora/common/entities/project.py
Bento007: added

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +62,30 @@ def __getattr__(self, name):
...
+        the row is retrieved and updated.
+
+        :param rows: A list of dictionaries containing columns.
```
calvinnhieu, line 74:
```suggestion
        :param rows: A list of dictionaries each specifying a row to insert or modify
```

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +62,30 @@ def __getattr__(self, name):
...
+        :param db_table: The Table to add or modify rows
+        :param keys: additional primary keys
```
calvinnhieu, line 76: Is this where keys are supplied to invoke "modify" over "insert" functionality? If so, is there an assumed order that must be maintained between `rows` and `keys`? Either way, looking for more clarification in documentation

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
mckinsel: I think your approach is actually idiomatic, but you could also make the default `tuple()`

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +62,30 @@ def __getattr__(self, name):
...
+        :param kwargs: Additional columns attributes to add.
```
calvinnhieu, line 77: Not sure what this is used for. As a general comment on this function, an example usage could be helpful in clarifying if documentation is insufficient. The way I see it is, clean code can be (in this order of importance):

- explicitly named, highly modular and thus self-explanatory
- well-documented
- example usage/tutorial provided

I think the need for each as you go down the list increases with code complexity. This, for example, is a nice, compact, but complex piece of code. My point here is, it's great, but needs to adhere to some standard of readability for the sake of maintainability and developer efficiency.

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
calvinnhieu: @mckinsel oh nice tip

tests/unit/backend/corpora/common/entities/__init__.py
calvinnhieu: +1

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -28,3 +35,90 @@ def test__get__ok(self):
...
+                contributors=[{"id": "test_contributor_id"}],
+                deployment_directories=[{"id": "test_deployment_directory_id"}],
+            )
+
+            dataset_id = dataset.id
+            artifact_ids = get_ids(dataset.artifacts)
+            deployment_directory_ids = get_ids(dataset.deployment_directories)
+            contributor_ids = get_ids(dataset.contributors)
+
+            # Expire all local object and retireve them from the DB to make sure the transactions went through.
```
calvinnhieu, line 110:
```suggestion
            # Expire all local object and retrieve them from the DB to make sure the transactions went through.
```

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -28,3 +35,90 @@ def test__get__ok(self):
...
+                    contributors=[contributor_params] * i,
+                    deployment_directories=[deployment_directory_params] * i,
+                )
+
+                dataset_id = dataset.id
+                artifact_ids = get_ids(dataset.artifacts)
+                deployment_directory_ids = get_ids(dataset.deployment_directories)
+                contributor_ids = get_ids(dataset.contributors)
+
+                # Expire all local object and retireve them from the DB to make sure the transactions went through.
```
calvinnhieu, line 87:
```suggestion
                # Expire all local objects and retrieve them from the DB to make sure the transactions went through.
```

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: moved to its own test

tests/unit/backend/corpora/common/entities/__init__.py
Bento007: moved to utils.py

backend/corpora/common/utils/db_utils.py
Bento007: The chance is low, but the cost of check is also low. The best case is it never happens. The worst case is we silently overwrite someones published project. I'm trying to mitigate the worst case.

backend/corpora/common/utils/db_utils.py
Bento007: I'm not sure if this is how SQLAlchemy works, but wouldn't it update the existing row if it existed? Thus overwriting whatever was there before.

backend/corpora/common/entities/dataset.py
Bento007: added

backend/corpora/common/entities/dataset.py
Bento007: you're right i will move the comment to line 66.

backend/corpora/common/utils/db_utils.py
mckinsel: >The chance is low

"Low" kind of understates it, see https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions

backend/corpora/common/entities/dataset.py
maniarathi: @mckinsel @Bento007 This is to keep track of the original data link. We'll likely supporting transferring files from a drive link or dropbox link that the users give us, so this would keep track of that.

backend/corpora/common/entities/entity.py
Bento007: adding

backend/corpora/common/entities/entity.py
Bento007: adding

tests/unit/backend/corpora/common/entities/test_dataset.py
maniarathi: Wouldn't you want to separate out a test that tests 0 foreign keys to use the None keyword then? And then a separate test for testing 1?

I don't quite understand the benefit of testing having 1 versus 2 foreign keys. If you have documentation on this testing practice I'd be happy to learn too!

backend/corpora/common/corpora_orm.py
calvinnhieu: oh didn't realize that both `project` and `projects` are valid - thought it was a typo/inconsistent. Resolving

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
Bento007: I'm going to stick with how. While a tuple is immutable, it doesn't not have the same functionality as a list. That could cause problems further down the development line if we are assuming we are working with a list.

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: It's not so much a test that the DB can manage foreign keys. It's that our code will not fail when parsing 0, 1 or multiple values from the database into python. It's a general QA best practice to test around the edges. Here is an article that encompases the spirit of the tests https://www.freecodecamp.org/news/a-beginners-guide-to-testing-implement-these-quick-checks-to-test-your-code-d50027ad5eed/.

Bento007: Are you not a fan of using subtest?

maniarathi: Interesting, thank you for the article. You're right, I'm not a fan of subtests. The reason is because when one of the subtests fail, the failure is somewhat hidden by the fact that the tests are lumped into a single unit test. For example, if I had a test `test__create_dataset_with_no_foreign_keys__ok` that failed versus a `test__create_dataset_with_multiple_foreign_keys__ok` that failed, it immediately tells me where to look rather than having to dive more deeply into the code to figure out which `i` value caused the failure and what that `i` value means and what the `i` value means in terms of the created input to the test.

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -28,3 +60,78 @@ def test__get__ok(self):
...
+                dataset = Dataset.get(dataset_id)
+                self.assertIsNotNone(dataset)
+                self.assertEqual(dataset_id, dataset.id)
+                self.assertEqual(artifact_ids, get_ids(dataset.artifacts))
+                self.assertEqual(deployment_directory_ids, get_ids(dataset.deployment_directories))
+                self.assertEqual(contributor_ids, get_ids(dataset.contributors))
+
+    def test__create_ids__ok(self):
+        """
+        Creating a dataet with ids in connect attributes. A new id is generated even if id is provided.
```
maniarathi, line 108:
```suggestion
        Creating a dataset with ids in connect attributes. A new id is generated even if id is provided.
```

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -1,9 +1,37 @@
...
+    DbDataset,
+    DbUser,
+)
 from backend.corpora.common.entities.entity import logger as entity_logger
 from backend.corpora.common.entities.project import Project
+from tests.unit.backend.corpora.common.entities.utils import get_ids
+
+
+class ProjectParams:
+    count = 0
```
maniarathi, line 19: What is the purpose of `count`? There isn't really a need for the project name to be unique...

backend/corpora/common/entities/project.py
```diff
@@ -1,9 +1,58 @@
...
+
+        """
+        uuid = generate_id()
+
+        # Setting Defaults
+        links = links if links else []
+
+        #  Prevent accidentally linking an existing row to a different Project. This maintains the relationship of one
+        #  to many for links
+        [link.pop("id", None) for link in links]  # sanitize of ids
```
maniarathi, line 38: Would remove the inline comment "# sanitize of ids". I actually have no idea what that means and the comment on the previous line makes it clear what this line is doing.

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -1,7 +1,39 @@
...
+    DbDeploymentDirectory,
+    DatasetArtifactType,
+    DatasetArtifactFileType,
+)
 from backend.corpora.common.entities.dataset import Dataset
+from tests.unit.backend.corpora.common.entities.utils import get_ids
+
+
+class DatasetParams:
+    count = 0
```
maniarathi, line 15: Not sure what the purpose of this is? There isn't really a need to have the dataset names be unique.

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +63,45 @@ def __getattr__(self, name):
         :return:
         """
         return self.db_object.__getattribute__(name)
+
+    @classmethod
+    def _create_sub_objects(
+        cls, rows: typing.List[dict], db_table: Base, add_columns: dict = None, primary_keys: typing.List[str] = None
+    ) -> typing.List[Base]:
+        """
+        Create or modify N `rows` in `db_table` associated with Entity Object during object creation. If id is provided,
```
maniarathi, line 72: Is "N" defined somewhere? I'd avoid using mathematical notation unless it is explicitly an argument or the function to derive N is also specified. You could just say "Create or modify rows in the db_table.."

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +63,45 @@ def __getattr__(self, name):
...
+        then the row with that id is retrieved and updated, else a new UUID is generated and a new row is created.
+
+        :param rows: A list of dictionaries each specifying a row to insert or modify
+        :param db_table: The Table to add or modify rows
+        :param primary_keys: Additional columns required to build the primary key. This is used when the primary consist
+        of
```
maniarathi, line 78: Weird line break here and line 79. I think they can be combined.

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +63,45 @@ def __getattr__(self, name):
...
+        multiple columns.
+        :param add_columns: Additional columns attributes or modifications to add to the row.
+
+        This is can be used when there are shared column value that need to be the added across all of the new rows.
```
maniarathi, line 82:
```suggestion
        This can be used when there are shared column values that need to be added across all the new rows.
```

backend/corpora/common/entities/entity.py
```diff
@@ -62,3 +63,45 @@ def __getattr__(self, name):
         :return:
```
maniarathi, line 63: I just noticed this empty return here. Can delete.
```suggestion
```

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,89 @@
...
+        )
+
+        #  Linking many contributors to many datasets
+        contributors = cls._create_sub_objects(contributors, DbContributor)
+        contributor_dataset_ids = [dict(contributor_id=contributor.id, dataset_id=uuid) for contributor in contributors]
+        dataset_contributor = cls._create_sub_objects(contributor_dataset_ids, DbDatasetContributor)
+
+        cls.db.session.add(new_db_object)
+        cls.db.session.add_all(contributors)
+        cls.db.commit()
```
maniarathi, line 84: We wouldn't want to do multiple commits here. We'd like for all sub entities of the dataset to be added at the same time as the entity itself (i.e. a single transaction). That way if the sub entities fail, we don't put the database into a state where some partial information about the dataset still exists.

backend/corpora/common/entities/project.py
```diff
@@ -1,9 +1,58 @@
...
+            needs_attestation=needs_attestation,
+            processing_state=processing_state,
+            validation_state=validation_state,
+            links=cls._create_sub_objects(
+                links, DbProjectLink, add_columns=dict(project_id=uuid, project_status=status)
+            ),
+        )
+
+        cls.db.session.add(new_db_object)
+        cls.db.session.commit()
```
maniarathi, line 57:
```suggestion
        cls.db.commit()
```
To make use of the commit function that catches exceptions, you'll want to use your helper function, not the session commit function.

backend/corpora/common/utils/db_utils.py
Bento007: So we need to explicitly call update in SQLAlchemy otherwise any conflicts will be rejected.

brianraymor:
@Bento007 - would you connect this PR with its issue?

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: If you look at the output of unittest it tells you exactly which iteration failed. There should be no code diving required. It spits out an error message for each failed iteration as if it was a stand alone test. We can always add a more verbose subtest name. I don't see this as a compelling reason to not use subtests.

backend/corpora/common/entities/project.py
```diff
@@ -1,9 +1,58 @@
...
+        tc_uri: str = "",
+        needs_attestation: bool = False,
+        processing_state: str = "",
+        validation_state: str = "",
+        links: list = None,
+    ) -> "Project":
+        """
+        Create a new Project and related objects and store in the database. UUIDs are generated for all new table
+        entries.
+
```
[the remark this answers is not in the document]
Bento007: Why isn't black fixing these?

backend/corpora/common/entities/project.py
Bento007:
```suggestion
        [link.pop("id", None) for link in links]
```

backend/corpora/common/utils/db_utils.py
```diff
@@ -29,3 +31,14 @@ def query(self, table_args: typing.List[Base], filter_args: typing.List[bool] =
...
+    def commit(self):
+        """
+        Commit changes to the database and roll back if error.
+        :return:
+        """
+        try:
+            self.session.commit()
+        except SQLAlchemyError:
+            self.session.rollback()
+            raise
```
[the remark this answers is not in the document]
Bento007: I'll add a log and raise a corpora exception. Though I think this will be better handled once we have some logging standards in place. I'm a fan of using JSON formatted logs myself. They are easier to search and parse.

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: it will be easier to debug to know which one failed.

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: it will be easier to identify during debug. It can also be extended to generate more complicated test inputs in the future.

backend/corpora/common/utils/uuid.py
```diff
@@ -0,0 +1,8 @@
+import uuid
+
+
+def generate_id():
```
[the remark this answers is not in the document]
Bento007: I wanted to avoid code repetition. I was also separating out the concern of generating a UUID and converting it into a usable format from the code of creating a new DB row. I think laying it out like this give us room to grow even if we don't use it.

backend/corpora/common/entities/dataset.py
Bento007: fixed

backend/corpora/common/entities/entity.py
Bento007: removed

backend/corpora/common/entities/entity.py
Bento007: fixed

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: the uuids take longer to compare when look at them at a glance.

Bento007:
Connects to #324

backend/corpora/common/entities/entity.py
```diff
@@ -59,6 +60,46 @@ def __getattr__(self, name):
...
+        >>>>    )
+
+        Another use would be to overwrite column specified in the rows.
+
+       :return: a list of database objects to create or modify.
+        """
+        add_columns = add_columns if add_columns else {}
+        db_objs = []
+        for columns in rows:
+            _columns = dict(**columns, **add_columns)  # add_columns changes takes precedence over columns
```
maniarathi, line 96: What does it mean for add_columns changes to take "precedence" over columns?

tests/unit/backend/corpora/common/entities/test_project.py
maniarathi: Sorry, perhaps I still don't understand. Doesn't each test have it's own error message? So you never need to know what project UUID is associated with each test. 

What are the more complicated test inputs that you envision? We can always bring this back when the complicated test cases you envision actually materialize instead of adding it now.

backend/corpora/common/entities/dataset.py
```diff
@@ -1,9 +1,87 @@
...
+        )
+
+        #  Linking many contributors to many datasets
+        contributors = cls._create_sub_objects(contributors, DbContributor)
+        contributor_dataset_ids = [dict(contributor_id=contributor.id, dataset_id=uuid) for contributor in contributors]
+        dataset_contributor = cls._create_sub_objects(contributor_dataset_ids, DbDatasetContributor)
+
+        cls.db.session.add(new_db_object)
+        cls.db.session.add_all(contributors)
+        cls.db.session.flush()
```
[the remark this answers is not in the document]
Bento007: The tests failed if I didn't flush here. Can you confirm that autoflush is on?

backend/corpora/common/entities/entity.py
Bento007: if there are matching keys in columns and add_columns, the key value in add_columns will be used.

backend/corpora/common/utils/db_utils.py
```diff
@@ -29,3 +34,14 @@ def query(self, table_args: typing.List[Base], filter_args: typing.List[bool] =
...
+
+    def commit(self):
+        """
+        Commit changes to the database and roll back if error.
+        """
+        try:
+            self.session.commit()
+        except SQLAlchemyError:
+            self.session.rollback()
+            logger.exception()
```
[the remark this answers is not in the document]
Bento007: it uses the SQLAlchemy error that was raised.

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: We don't clear the DB after each test and some test generate multiple datasets in order to run.

Bento007: we could clear the DB after every test, but it will be significantly slower.

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: I have already found it useful in debugging the tests. We can remove it later if it stops being useful.

Bento007: It helped me when testing within a time range that only projects within that time range were retrieved.

backend/corpora/common/utils/uuid.py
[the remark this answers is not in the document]
Bento007: this seems more of a stylics change than functional one, but if it bothers you that much, i'll inline it.

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: Also I don't want to spend time generating unique names to help my self identify projects. I think as  we start to add more input validation have a function or class to generate the inputs will be very helpful.

backend/corpora/common/entities/entity.py
maniarathi: Can you update the comment with that please? Your comment above makes it much more clear :)

backend/corpora/common/utils/db_utils.py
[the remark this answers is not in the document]
maniarathi: The error messages tend to be a little convoluted. Adding a hand-crafted message the denotes what happened is usually quite helpful.

backend/corpora/common/utils/uuid.py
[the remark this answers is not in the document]
maniarathi: Stylistic changes are important too, because they have a huge impact on readability, maintainability, debuggability and others. Code reviews are not meant to be only about whether the function performs what it says it performs. If you're interested in learning a bit more on some of the things I look for in code reviews, here's roughly the training I got: https://google.github.io/eng-practices/review/. Functionality is only one of bullet points.

Yes, I'd prefer for it to be inlined. :)

tests/unit/backend/corpora/common/entities/test_dataset.py
maniarathi: Apologies, I still don't understand. What does clearing the database have to do with unique dataset names? The UUIDs are naturally going to be different because they are generated using the uuid4() function. This variable is only used for the dataset name itself and the value isn't even checked anywhere. You check the dataset.id in places but this isn't related to that.

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
maniarathi: https://docs.sqlalchemy.org/en/13/orm/session_api.html#session-and-sessionmaker ?

What does it fail with?

backend/corpora/common/entities/entity.py
Bento007, line 96:
```suggestion
            _columns = dict(**columns, **add_columns)  # if there are matching keys in columns and add_columns, 
                                                                                         # the key value in add_columns will be used.
```

backend/corpora/common/utils/db_utils.py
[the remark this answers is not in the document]
Bento007: I don't know why it would fail to commit.

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
Bento007:
```SubTest error: Traceback (most recent call last):
  File "/Users/trentsmith/workspace/envs/dcp-prototype/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1264, in _execute_context
    cursor, statement, parameters, context
  File "/Users/trentsmith/workspace/envs/dcp-prototype/lib/python3.7/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 873, in do_executemany
    cursor.executemany(statement, parameters)
psycopg2.IntegrityError: insert or update on table "dataset_contributor" violates foreign key constraint "dataset_contributor_contributor_id_fkey"
DETAIL:  Key (contributor_id)=(6e1599fd-4c68-4bc2-9d13-75179a8d4136) is not present in table "contributor".```

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: I use it when I'm using the python debugger to differentiate the datasets.

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
Bento007:
```sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) insert or update on table "dataset_contributor" violates foreign key constraint "dataset_contributor_contributor_id_fkey"
DETAIL:  Key (contributor_id)=(530030b2-3736-4c23-aa9e-bbc0614219fb) is not present in table "contributor".

[SQL: INSERT INTO dataset_contributor (id, contributor_id, dataset_id) VALUES (%(id)s, %(contributor_id)s, %(dataset_id)s)]
[parameters: {'id': '36a13079-5fcd-4fda-abd4-dceb9efca99b', 'contributor_id': '530030b2-3736-4c23-aa9e-bbc0614219fb', 'dataset_id': '569760ea-6816-46ec-90b7-3219cfcdc5a1'}]
(Background on this error at: http://sqlalche.me/e/gkpj)```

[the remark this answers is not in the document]
maniarathi: Chatted offline can resolve.

PR #422, opened 2020-07-03
list dataset and projects

Bento007:


tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -117,3 +117,14 @@ def test__create_ids__ok(self):
 
         self.assertEqual(link_ids, get_ids(project.links))
         self.assertNotEqual(["test_project_link_id"], get_ids(project.links))
+
+
+    def test__list__ok(self):
+        generate = 5
+
+        for i in range(generate):
```
maniarathi, line 125:
```suggestion
        for _ in range(generate):
```
Do not need i here as it is not used.

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -135,3 +135,13 @@ def test__create_ids__ok(self):
 
         self.assertEqual(contributor_ids, get_ids(dataset.contributors))
         self.assertEqual(["test_contributor_id"], get_ids(dataset.contributors))
+
+    def test__list__ok(self):
+        generate = 5
+
+        for i in range(generate):
```
maniarathi, line 142:
```suggestion
        for _ in range(generate):
```
Do not need i here if it not used.

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -135,3 +135,13 @@ def test__create_ids__ok(self):
...
+            Dataset.create(**DatasetParams.get())
+
+        datasets = Dataset.list()
+        self.assertGreaterEqual(len(datasets), generate)
```
maniarathi, line 146: Why should this be "greater than or equal"? Should this test exactly what the length of the datasets should be?

tests/unit/backend/corpora/common/entities/__init__.py
```diff
@@ -0,0 +1,2 @@
+def get_ids(iterable):
```
maniarathi, line 1: I think I've made this comment before, but I'd prefer not to have utility functions like this hidden away in `__init__` files. Utility functions should be in utils.py or util.py or some file that clearly designates the utility status.

Also, I wonder what the value is of having a function that is composed of one line. For testing purposes, code repetition is fine, especially for one line, because it makes the test self-sufficient. I find self-sufficiency in tests to be more valuable that code reuse. I don't want to jumping around files to visit utility functions... how do I know that I haven't introduced a bug in my test via untested utility functions?

Bento007: it's used in 24 different places. It does simply the code significantly.

PR #424, opened 2020-07-08
get/list projects

Bento007:
- Adding CustomJSONEncoder to flask to decode SQLAlchemy objects into json
- Adding TransformingBase to help transform SQLAlchemy objects into dictionaries and eventually json
- Adding Project.list_projects_in_date_range to return a list of projects within a date range, and test cases
- Adding Project.get_projects to return live info.
- Modified Corpora-api.yml Get /v1/project to return an appropriate result. Removed unused parameters
- Implementing GET /v1/project API
- Implementing GET /v1/project/{project_uuid}
- Explicitly converting query_param into a dict in in the chalice app. Flask had an issues handling the query parameters returned from chalice that are type Mulitdict. This is an issue that needs further investigations


backend/chalice/api_server/app.py
```diff
@@ -62,7 +64,7 @@ def dispatch(*args, **kwargs):
         with flask_app.test_request_context(
             path=resource_path,
             base_url="https://{}".format(app.current_request.headers["host"]),
-            query_string=app.current_request.query_params,
+            query_string=dict(app.current_request.query_params) if app.current_request.query_params else None,
```
Bento007, line 67: Explicitly converting query_param, which is a multidict, into a dict in in the chalice app. Flask had an issues handling the query parameters returned from chalice.  This is an issue that needs further investigations

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -118,6 +120,52 @@ def test__create_ids__ok(self):
         self.assertEqual(link_ids, get_ids(project.links))
         self.assertNotEqual(["test_project_link_id"], get_ids(project.links))
 
+    def test__list_in_time_range__ok(self):
```
maniarathi, line 123: As this test is written today, all the projects that are created are returned no matter what the time range is. While you do check to make sure that all the returned projects have a timestamp that is acceptable, you do not check to make sure that certain projects were _excluded_. i.e. you check for true positives but not true negatives in confusion matrix terms.

Here's a way you could potentially frame this test. Create 3 different projects while mocking out the timestamp such that they are created at different times. Control the timestamps yourself. Then conduct the tests such that you have your permutations of the parametrization of the function, but again, control the from and to dates such that certain projects should be excluded each time (or not for the case where you do not parametrize the function). You can check the timestamps of the project but you can also check which projects are included and which ones are excluded explicitly because you controlled the timestamp when you created the projects.

Alternatively, you'd have one test per parametrization of the function where you only need to generate two projects at a time. This would be my preferred route because it tests one functionality at a time, but I know you like subtests so suggesting the former ;)

backend/corpora/common/corpora_orm.py
```diff
@@ -14,17 +14,46 @@
...
 
-Base = declarative_base()
-deployment_stage = os.environ["DEPLOYMENT_STAGE"]
 
+class TransformingBase(object):
+    """
+    Add functionality to transform a Base object, and recursively transform its linked entities.
+    """
+
+    RELATIONSHIPS_TO_DICT = True  # Set to True if entities with a relationship to this should be converted to a dict.
```
maniarathi, line 32: Technically this isn't a global variable so should not be in all-caps.

maniarathi: Also, when is this set to false? As it stands today, it never gets set to anything but true? If so, is this class variable actually needed?

backend/corpora/common/corpora_orm.py
```diff
@@ -14,17 +14,46 @@
...
+
+    def __iter__(self):
+        return self.to_dict().iteritems()
+
+    def to_dict(self, relation=None, backref=None):
```
maniarathi, line 37: Can we use slightly more descriptive argument names here please? The idea behind descriptive names is that in one read, I can tell what the argument is supposed to do. I was really confused for a bit because I thought `relation` represented the relationships themselves and got confused by the boolean nature of it. I'd have preferred something more descriptive, even if it's longer.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -118,6 +120,52 @@ def test__create_ids__ok(self):
...
+        generate = 5
+        sleep = 0.5
+        from_ids = []
+        to_ids = []
+
+        from_date = datetime.now().timestamp()
+        time.sleep(sleep)
+        for i in range(generate):
```
maniarathi, line 131:
```suggestion
        for _ in range(generate):
```
`i` is unused.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -118,6 +120,52 @@ def test__create_ids__ok(self):
...
+
+        with self.subTest("Test from_date"):
+            # Projects from_date are returned.
+            projects = Project.list_in_time_range(from_date=from_date)
+            for project in projects:
+                self.assertGreaterEqual(project["created_at"].timestamp(), from_date)
+
+        to_date = datetime.now().timestamp()
+        time.sleep(sleep)
+        for i in range(generate):
```
maniarathi, line 142:
```suggestion
        for _ in range(generate):
```
`i` is unused.

tests/unit/backend/corpora/common/entities/test_project.py
```diff
@@ -118,6 +120,52 @@ def test__create_ids__ok(self):
...
+            for project in projects:
+                self.assertLessEqual(project["created_at"].timestamp(), to_date)
+
+        with self.subTest("Test no date"):
+            projects = Project.list_in_time_range()
+            test_ids = set([*to_ids, *from_ids])
+            project_ids = set([p["id"] for p in projects])
+            self.assertTrue(test_ids.issubset(project_ids))
+
+        with self.subTest("Verify columns"):
```
maniarathi, line 165: What exactly is this checking? At the surface-level it seems like you're checking the typing of the returned object but I didn't know if there was a more deeper reasoning for this? It's slightly odd to check typing in a test in Python...

backend/config/corpora-api.yml
```diff
@@ -405,10 +405,17 @@ paths:
...
+                  projects:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        project_uuid:
+                          $ref: '#/components/schemas/project_uuid'
+                        creation_date:
+                          type: integer
+                        name:
```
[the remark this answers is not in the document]
Bento007: the name of the project

backend/config/corpora-api.yml
```diff
@@ -462,7 +469,6 @@ paths:
         '404':
           $ref: '#/components/responses/404'
```
Bento007, line 470:
```suggestion
          $ref: '#/components/responses/404'
          
```

backend/corpora/common/corpora_orm.py
Bento007: it was added for future flexibility for Bases that won't be converted to dictionaries. For example, any auth stuff in the DB will not be converted.

backend/corpora/common/corpora_orm.py
Bento007: That's fair, the variable was recycled and changed types.

Bento007: fixed

backend/corpora/common/entities/project.py
```diff
@@ -56,3 +62,50 @@ def create(
...
+    @classmethod
+    def list_in_time_range(
+        cls, to_date: float = None, from_date: float = None, filters: list = None
+    ) -> typing.List[typing.Dict]:
+        """
+
+        :param to_date: Filter dates earlier than this. Unix timestamp since the epoch in UTC timezone.
+        :param from_date: Filter dates later than this. Unix timestamp since the epoch in UTC timezone.
+        :param list_entities: The columns to retrieve from the table.
+        :param filters: additional filters to apply to the query.
```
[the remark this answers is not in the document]
Bento007: they are used by list_projects_in_time_range and will be used by submission in the future to list submissions by a user.

backend/corpora/common/entities/project.py
```diff
@@ -56,3 +62,50 @@ def create(
...
+    @classmethod
+    def list_projects_in_time_range(cls, *args, **kwargs):
+        return cls.list_in_time_range(*args, **kwargs, filters=[DbProject.status == ProjectStatus.LIVE.name])
+
+    @classmethod
+    def list_in_time_range(
+        cls, to_date: float = None, from_date: float = None, filters: list = None
+    ) -> typing.List[typing.Dict]:
+        """
+
```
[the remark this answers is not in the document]
Bento007: added

tests/unit/backend/corpora/common/entities/test_project.py
Bento007: Since I don't know the exact value being returned, but I'm certain of the type that should be returned.

tests/unit/backend/chalice/api_server/test_v1_project.py
```diff
@@ -0,0 +1,38 @@
...
+    def test__GET_project__OK(self):
+        with self.subTest("No Parameters"):
+            test_url = furl(path="/v1/project")
+            response = self.app.get(test_url.url, headers=dict(host="localhost"))
+            response.raise_for_status()
+            body = json.loads(response.body)
+            self.assertEqual(body["projects"][0]["id"], "test_project_id")
+
+        now = int(datetime.now().timestamp()) + 10
+        with self.subTest("With to_date"):
```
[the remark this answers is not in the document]
Bento007: how would you mock the timestamp?

PR #427, opened 2020-07-08
[EASY] Remove body from app logs

Bento007:
- Do not log the body of the request or the response in the logs. It will take up a lot of storage and can contain sensitive information.

### DIFF
--- a/backend/chalice/api_server/app.py
+++ b/backend/chalice/api_server/app.py
@@ -48,7 +48,9 @@ def get_chalice_app(flask_app):
     app.log.setLevel(logging.DEBUG)
 
     def dispatch(*args, **kwargs):
-        app.log.info(f"Request: {app.current_request.to_dict()}")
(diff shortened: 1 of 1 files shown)


backend/chalice/api_server/app.py
```diff
@@ -48,7 +48,9 @@ def get_chalice_app(flask_app):
     app.log.setLevel(logging.DEBUG)
 
     def dispatch(*args, **kwargs):
-        app.log.info(f"Request: {app.current_request.to_dict()}")
+        log_request = app.current_request.to_dict()
```
maniarathi, line 51: Very minor nit: just to avoid duplicate code, you could create a tiny helper function like `clean_entry_for_logging` that does both the dictionary conversion and the removal of the body and use it in both locations.

Bento007: Done

PR #432, opened 2020-07-11
Add initial alembic migration framework.

maniarathi:
Couple notes: much of this code is generated by running `alembic init alembic` and I left most of it as is except for a few formatting things that black decided to change. The main changes here are the `run_migrations_online` function in `env.py` and the scripts for migration which were pulled from the original dcp 2.0 code (which were pulled from upload-service). They provide some nice sugar around being able to execute migrations cleanly.

One thing I'd love to get feedback on is the location of the files. Currently I renamed the `alembic` directory that was created to `database` to more clearly elucidate what the contents are related to and I put the `database.ini` file in the config directory (it was floating in the top level `backend` directory when originally generated). The Makefile with all the database-related commands are in the top-level `backend` directory since they make use of `database/` and `config/` directories. I am open to other arrangements though if the above doesn't make sense.

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -1,0 +1,39 @@
+db/migrate:
+	alembic -x db=${DEPLOYMENT_STAGE} -c=./config/database.ini upgrade head
+
+db/rollback:
+	alembic -x db=${DEPLOYMENT_STAGE}  -c=./config/database.ini downgrade -1
+
+db/new_migration:
+	# Usage: make db/new_migration MESSAGE="purpose_of_migration"
+	alembic -c=./config/database.ini revision --message $(MESSAGE)
+
+db/connect:
+	$(eval DATABASE_URI = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/${DEPLOYMENT_STAGE}/database --region us-east-1 | jq -r '.SecretString | fromjson.database_uri'))
+	psql --dbname $(DATABASE_URI)
+
+db/download:
+	# Usage: make db/download FROM=dev    - downloads DB to corpora_dev-<date>.sqlc
+	$(eval DATABASE_URI = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/${FROM}/database --region us-east-1 | jq -r '.SecretString | fromjson.database_uri'))
+	$(eval OUTFILE = $(shell date +corpora_${FROM}-%Y%m%d%H%M.sqlc))
+	pg_dump -Fc --dbname=$(DATABASE_URI) --file=$(OUTFILE)
+
+db/import:
+	# Usage: make db/import FROM=dev    - imports corpora_dev.sqlc into corpora_local
+	pg_restore --clean --no-owner --dbname corpora_local corpora_$(FROM).sqlc
+
+db/import/schema:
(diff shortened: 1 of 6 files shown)


backend/database/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
Bento007, line 1: Add a link to Alembic. Also why is it generic?

backend/database/env.py
```diff
@@ -0,0 +1,87 @@
...
+    connectable = engine_from_config(alembic_config, prefix="sqlalchemy.", poolclass=pool.NullPool)
+
+    with connectable.connect() as connection:
+        context.configure(connection=connection, target_metadata=target_metadata)
+
+        with context.begin_transaction():
+            context.run_migrations()
+
+
+if context.is_offline_mode():
```
calvinnhieu, line 84: Where is this configured? Thought it might be in the `config/database.ini` file, but I didn't see it.

backend/database/env.py
```diff
@@ -0,0 +1,87 @@
...
+
+    alembic_config["sqlalchemy.url"] = CorporaDbConfig().database_uri
+
+    connectable = engine_from_config(alembic_config, prefix="sqlalchemy.", poolclass=pool.NullPool)
+
+    with connectable.connect() as connection:
+        context.configure(connection=connection, target_metadata=target_metadata)
+
+        with context.begin_transaction():
+            context.run_migrations()
```
calvinnhieu, line 81: Would additional error handling/messaging be helpful here? I suppose on failure the transaction would rollback and the connection would close itself, but just double checking. Can be added as we see necessary in practice.

backend/database/README
maniarathi: This is actually auto-generated by Alembic, so wanted to just leave it as is. Are you OK with that?

backend/Makefile
```diff
@@ -0,0 +1,39 @@
...
+db/new_migration:
+	# Usage: make db/new_migration MESSAGE="purpose_of_migration"
+	alembic -c=./config/database.ini revision --message $(MESSAGE)
+
+db/connect:
+	$(eval DATABASE_URI = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/${DEPLOYMENT_STAGE}/database --region us-east-1 | jq -r '.SecretString | fromjson.database_uri'))
+	psql --dbname $(DATABASE_URI)
+
+db/download:
+	# Usage: make db/download FROM=dev    - downloads DB to corpora_dev-<date>.sqlc
```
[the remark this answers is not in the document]
maniarathi: Great suggestion, thanks! I expanded a bit as well :)

backend/database/env.py
maniarathi: This is actually part of the auto-generated code. The context is offline if the `alembic` command is run with the `--sql` argument which indicates that the script should generate the SQL script instead of actually executing the change. 

In practice, because we have the nice sugar with the Makefile commands that we use, we'll likely not need this often, but it would't be great if at some point we needed to run the `alembic` command manually with `--sql` and then the functionality is broken because we remove this bit of code.

backend/database/env.py
maniarathi: Good question! I wasn't sure what additional error handling here would be helpful. You're right that alembic will handle closing and rollback and does the error throwing as well. Do you have any suggestions by any chance?

PR #446, opened 2020-07-14
401-landing-page

### LINKED ISSUE #312: Update data portal frontend to pull project information via Corpora API
Issue title says it all :)

### LINKED ISSUE #401: Implement Data Portal temporary landing page
The initial "launch" of data portal will contain a single page that shows a list of datasets that can be launched in cellxgene. Each dataset has up to two options: a remix dataset and the original dataset. This ticket tracks the frontend implementation of the design of the front page.

Figma links: https://github.com/chanzuckerberg/corpora-data-portal/issues/365#issuecomment-651378333

Swagger: https://670bpw5cqf.execute-api.us-west-2.amazonaws.com/dev/

tihuan:
This PR aims to create a temporary landing page for Data Portal as described in:

https://www.figma.com/proto/Ys3WOryh4RpT1V5ezlOtEL/M1-Find-Data-explorations?node-id=158%3A2761&scaling=min-zoom

https://www.figma.com/file/Ys3WOryh4RpT1V5ezlOtEL/M1-Find-Data-explorations?node-id=158%3A2441

Changes:
1. Add TS definitions for schemas and entities used to create the page. Such as Project, Dataset, DatasetAsset, etc.. Note that there are properties intentionally commented out in the type definitions, since they're not being used in the FE yet. I'm assuming as we build more features, we can uncomment the properties as they start being used in the app. 

NOTE: I followed `backend/config/corpora-api.yml` to build the types

1. Add `Styled Components` for styling for now. Given that branding / design system discussion is still ongoing, I think it's too early to commit to a certain styling library, so I just picked one that I'm most familiar with 😄 

1. Add lint rule `sort-keys`, so files have consistent object sort order

1. Add hard coded `PROJECTS` data to showcase the UI, this will be removed once we integrate FE with BE API (https://github.com/chanzuckerberg/corpora-data-portal/pull/433)

1. Update `frontend/src/pages/index.tsx` to use the new components to render the landing page

1. Add `.codeclimate.yml` to ignore files

QA:

1. To start the app, navigate to `corpora-data-portal/frontend` folder
2. run `npm i && npm run develop`
3. Visit http://localhost:8000/ to see the landing page

NOTE: Clicking on any cellxgene links won't work right now, because it's hard coded data. This should work once API integration is done!

Thank you for your reviews 😄 🙏 !

Screen Shot 2020-07-14 at 9 16 04 AM

Screen Shot 2020-07-14 at 9 16 16 AM

gif

closes #401
closes #312


backend/config/corpora-api.yml
```diff
@@ -12,7 +12,6 @@ servers:
     url: https://api.staging.corpora.cziscience.com
 paths:
   /v1/submission:
```
tihuan, line 14: this is all linter changes

frontend/package.json
```diff
@@ -7,12 +7,14 @@
   "dependencies": {
```
tihuan, line 7: all related to adding styled components

frontend/src/common/entities.ts
```diff
@@ -0,0 +1,109 @@
+export interface Project {
```
tihuan, line 1: type def file

frontend/src/common/svg.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.svg" {
```
tihuan, line 1: type def for svg files

frontend/src/components/ProjectList/components/Dataset/components/MoreInformation/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+const MoreInformation: FC<Props> = ({ links }) => {
+  return (
+    <Wrapper>
+      {links.map(link => {
+        return (
+          <StyledAnchor
+            key={link.url}
+            href={link.url}
+            target="_blank"
+            rel="noopener"
```
tihuan, line 18: no need for adding `noreferrer`, since it's a security fallback for IE11 <, which I don't think we should support. 

The upside for not adding `noreferrer` is that out linked pages will know our website referred the users to their pages!

frontend/src/components/seo.tsx
```diff
@@ -43,36 +43,36 @@ const SEO: FC<Props> = ({ description, lang, meta = [], title }) => {
       meta={[
         ...meta,
```
tihuan, line 44: resorting object keys a-z

frontend/src/components/styledButton.tsx
```diff
@@ -13,17 +13,17 @@ const StyledButton: FC<Props> = ({ label, disabled, handleClick }) => {
     <Button
```
tihuan, line 13: resorting object keys a-z

frontend/src/components/theme.ts
```diff
@@ -1,116 +1,116 @@
 export default {
```
tihuan, line 1: resorting object keys a-z

frontend/src/contexts/auth0Context.tsx
```diff
@@ -12,12 +12,12 @@ import { User } from "../common/entities";
```
tihuan, line 12: resorting object keys a-z

frontend/src/components/projectsList.tsx
```diff
@@ -1,123 +0,0 @@
-import { SystemStyleObject } from "@styled-system/css";
```
tihuan, line 1: these files are preserved in a separate branch already, so we can remove them from production code

tihuan:
Finally all checks passed ✅ PTAL all! Thank you!

frontend/src/common/entities.ts
```diff
@@ -0,0 +1,109 @@
...
+  // processing_state: string;
+  // s3_bucket_key: string;
+  // validation_state: string;
+  // links: { url: string; type: string }[];
+  // attestation: {
+  //   needed: boolean;
+  //   tc_uri: string;
+  // };
+  datasets: Dataset[];
+  links: { name: string; url: string }[];
```
Bento007, line 30: `links` does not contain `name` yet. Will this work when integrated?

frontend/src/common/entities.ts
```diff
@@ -0,0 +1,109 @@
...
+// }
+
+// interface DOI {
+//   id: string;
+//   title: string;
+//   data: string;
+//   journal: string;
+// }
+
+// id: "HCA-AnalysisFile-001a3fdd-c9e6-4871-9e48-a840d52ecdf9"
```
Bento007, line 90: Why is this the only type with an example?

frontend/src/components/ProjectList/components/Dataset/components/MoreInformation/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const MoreInformation: FC<Props> = ({ links }) => {
+  return (
+    <Wrapper>
+      {links.map(link => {
+        return (
+          <StyledAnchor
+            key={link.url}
+            href={link.url}
+            target="_blank"
```
Bento007, line 17: why _blank?

frontend/src/common/entities.ts
tihuan: Thanks for raising this! I just pushed a commit to temporarily comment out `<MoreInformation />` component, which is the only one that uses `link.name`. So the type definition can stay!

frontend/src/common/entities.ts
tihuan: Ahh good point! I added this type definition to get an example component Colin wrote 5 months ago to work with TypeScript. But since that component is preserved in a branch already, I just deleted it and removed this `File` definition! It doesn't seem to match `api.yml` anyway

frontend/src/components/ProjectList/components/Dataset/components/MoreInformation/index.tsx
tihuan: to open the link in a new tab instead of the current tab 😄

PR #447, opened 2020-07-14
Clarify docs for creating an ssh tunnel via bastion

calvinnhieu:


calvinnhieu:
FYI @maniarathi , in the `pgAdmin` browser view, the connection port can be specified in a field at:

Add New Server > "Connection" tab > "Port" field

However, this PR revises the port to `5432`.

PR #451, opened 2020-07-14
Add security architecture diagram

calvinnhieu:
Illustrating Corpora's usage of `single-cell-infra` and the AWS infrastructure provided by the Shared Infrastructure team, namely the set of resources exported by the `cloud-env` module.


docs/security_arch.md
```diff
@@ -0,0 +1,54 @@
+# Security Architecture
+
+Corpora's infrastructure is managed by the [single-cell-infra repo](https://github.com/chanzuckerberg/single-cell-infra)
+which was provided by the Shared Infrastructure team in order to enable streamlined creation of secure cloud
+environments. This document depicts the application of the template environment in Corpora to illustrate its inner
+complexities. Ideally, general documentation on the `cloud-env` environment from Shared Infrastructure should supplement
```
Bento007, line 6: Can you take out Ideally, it make it sounds wishy washy. Does the documentation exist or doesn't it? Who should we probe to get it?

docs/security_arch.md
```diff
@@ -0,0 +1,54 @@
...
+and/or replace this doc.
+
+## Illustration
+
+The following diagram renders the VPC, subnets, security groups, internet gateway, network address translation
+and their relationships provided by Shared Infrastructure's `cloud-env` environment. Additionally, Corpora-specific
+resources such as the API Lambda and RDS cluster are contextualized against the security infrastructure.
```
Bento007, line 13: I love the diagram, thank you!

docs/security_arch.md
Bento007: can you make cloud-env link to cloud-env in single-cell-infra or somewhere applicable.

calvinnhieu:
> links to the terraform in single-cell-infra that makes all of this happen

For the security architecture (VPC, subnets, etc), these details should be provided by documentation from Shared Infrastructure. Our team's responsibility should begin at importing and using `cloud-env` (i.e. I don't know and probably shouldn't need to know). I've updated the doc to include a link to the `cloud-env' source code however.

PR #454, opened 2020-07-15
[1/3] Simplify Project and Dataset Create

### LINKED ISSUE #453: Simplify Entity._create_sub_object
Remove the ability to update database object from Entity._create_sub_object.

Bento007:
* Removing code path in `Entity._create_sub_object` where `id` is present. Updating Project and Dataset class to no longer sanitize for id's. Id's will be overwritten if present during the creation of new Projects or Datasets.
* Removed unused test cases

* closed #453


backend/corpora/common/entities/entity.py
```diff
@@ -88,19 +88,17 @@ def _create_sub_objects(
...
-            # the key value in add_columns will be used.
-            _id = _columns.get("id")
-            if _id:
-                primary_key = (_id, *primary_keys) if primary_keys else _id
-                row = cls.db.get(db_table, primary_key)
-            else:
-                _columns["id"] = str(uuid.uuid4())
-                row = db_table(**_columns)
+            #  if there are matching keys in columns and add_columns,
+            #  the key value in add_columns will be used.
```
maniarathi, line 97: Nit: can be combined with previous line?

Bento007: fixed

PR #460, opened 2020-07-17
Move schema to github and add its AnnData implementation

### LINKED ISSUE #426: Link Name for Project Links
When links are displayed in the frontend, we want to display a pretty name for the link not a raw url. This will require us to add a new column called `name` to the `DbProjectLink` table. This column will contain the name to display for the link in the frontend. This column should be returned whenever the links are returned.

### LINKED ISSUE #464: Define schema implementation in h5ad.
The schema doc describes the content of the schema but not exactly how it's realized in an h5ad file. cellxgene needs that.

mckinsel:
The schema used to be a google doc, but we're moving docs to markdown on github.

Also, this includes a more detailed description of how a schema-conforming dataset can be realized as an AnnData object.

Also adds `link_name` to the schema, thus resolves #426 

Resolves #464


backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+cell\_type|string
+sex|“male”, “female”, “unknown”, or “na”
+ethnicity|string, “na” if non-human
+development\_stage|string
+
+In addition to these free text fields (except sex), each cell must also have ontology values:
+
+**Field name**|**Constraints**
+:--|:--
+tissue\_ontology|UBERON term
```
bkmartinjr, line 128: these `foo_ontology` field names are not ideal, as they do not define the ontology, but rather the term (from an ontology).  It is also ambiguous if they are supposed to contain a term ID, term name, or be more flexible.

Perhaps rename to `foo_term_id` to match the nomenclature of OWL/OBO?   Or if they allow a more general syntax, something like `foo_term`?

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+assay\_ontology|EFO term
+disease\_ontology|MONDO term or [PATO:0000461](http://bioportal.bioontology.org/ontologies/PATO?p=classes&conceptid=PATO%3A0000461)
+cell\_type\_ontology|CL term
+ethnicity\_ontology|HANCESTRO term, “na” if non-human
+development\_stage\_ontology|HsapDv term if human, child of EFO:0000399 otherwise
```
bkmartinjr, line 133: for terms which may come from multiple ontologies, it might be more explicit to allow the encoding of the actual ontology as well as the ontological term ID.   I would make this optional, with reasonable defaults (eg, assume UBERON for tissue, etc).

In addition, given how fluid ontologies are, do you want to consider the ability to specify the ontology version explicitly?

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+Corpora does not impose any additional constraints on the `X` data matrix. So it may be sparse or dense and any
+numeric `numpy.dtype`.
+
+## `uns`
+
+Recall that `uns` is a mapping with `str`s as keys. Corpora requires the following keys and values in `uns`:
+
+**Key**|**Value Type**|**Notes**
+-----|-----|-----
+title|`str`|Title of the dataset
```
bkmartinjr, line 27: Given that this spec specifies the _encoding_ of data compliant with the corpora schema (defined elsewhere), I don't think it should include notes or additional semantic information like "title of the dataset".  Those should be completely defined in the Corpora schema spec.   I would constrain this spec to only that information related to encoding in H5AD, and push the semantic and more general schema info into the parent spec..

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+contributors|`list` of `dict`s. each `dict` must a "name": `str` item and may optionally have "email": `str` and "institution": `str`|This is the person who gave us the dataset, not necessarily the author or actual data generator.
```
bkmartinjr, line 28: typo:  `... must a "name" ...`

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
+# Corpora Schema AnnData Implementation
+
+Corpora datasets are stored in the HDF5-backed AnnData format as written by version 0.7 of the anndata library. One
```
bkmartinjr, line 3: we need some sort of top-level info in the H5AD to allow introspection on:
* what version of the corpora schema was encoded
* what version of this spec was encoded

I'd just a single semver field in uns that has something like `{ corpora_ver: 0.1, corpora_encoding: 0.4 }`

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+project\_description|Longer description of the project.
+project\_protocol\_links|List of links to protocol descriptions, to protocols.io for example
+project\_raw\_data\_links|List of links to repositories where raw data are available
+project\_other\_links|List of links to other sites the submitter deems relevant, author data sites for example.
+
+Each elements in the `links` fields above are dicts `{"link_name": "...", "link_url": "..."}` where `link_name` is meant to be the anchor text.
+
+#### System Fields
+
+Finally, Corpora inserts a **corpora_dataset_version** field for internal bookkeeping. This must not be present in the original submitted matrix, and Corpora
```
bkmartinjr, line 254: what is the version tracking?   Schema/spec version?  Or dataset (actual data) revision, or?

Somewhere in this spec we need to define schema version, and have that flow into the actual data files.  This allows us to change it down the road.

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+publication\_doi|`str`|Suitable for submission to the CrossRef API
+default\_embedding|`str`|Must match a key to an embedding in `obsm`.
+default\_field|`str`|Must match a column name in `obs`
+tags|`list` of `str`s| 
+<obs\_column>\_colors|`list` of color specifications (see anndata/cellxgene documentation)|<obs\_column> must be a column name in `obs`. There may be multiple keys like this.
+
+
+## `obsm`
+
+`obsm` is a mapping from `str`s to matrices of shape (n, m) where `n` is the number of rows in `X` and `m >= 1`.
```
bkmartinjr, line 56: given that this is an H5AD spec, it might be clearer if you substitute `(nObs, nVar)` for `(n, m)`

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+Corpora requires one value in `obms` to be a two-dimensional embedding, meaning it is of shape (n, 2) and is a
```
bkmartinjr, line 57: does corpora require `m==2`, or `m>=2`?  

currently cellxgene requires `m>=2`, but ignores anything other than the first two.

Also, for embeddings, I think you want to further restrict the values -- they should be numeric, but also non-Infinite.   NaN and finite values make sense in an embedding - not sure what +/- Inf means, and currently cellxgene will refuse to display any embedding-like column containing these values.

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+numeric type.
+
+## `obs`
+
+Corpora requires a number of columns be present in the `obs` dataframe:
+
+**Key**|**Value Type**|**Notes**
+-----|-----|-----
+tissue|`str`|
```
bkmartinjr, line 66: do you literally mean `str` or is Pandas `categorical` also acceptable?

Most datasets we see use categoricals frequently for these types of enumerated string values, and in general, I think they are superior for items such as `sex`.

I recommend accepting either.   Obviously they don't make sense for stuff like cell_type, as you don't want to include the entire ontology in the type definition.

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+  <tr>
+   <td>tags
+   </td>
+   <td>A list of strings that the Portal or other tools may use for searching, sorting, or filtering.
+   </td>
+  </tr>
+  <tr>
+   <td>default_field
+   </td>
+   <td>Name of another field that should be the “default” for display. So for example, when cellxgene starts this is the field that the cells should be
```
bkmartinjr, line 225: missing comma:   `...starts, this ...`

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+   </td>
+  </tr>
+</table>
+
+
+Additionally, submitters can include project-level metadata inside their dataset. This information is used to populate project metadata in Corpora when
+the submitted dataset is used to create a new project.
+
+
+**Field name**|**Description**
```
bkmartinjr, line 242: cellxgene currently as a single "about" link, which it displays for users who wish more information about the dataset.  Which of these links is the "summary" or most general link, and would be appropriate as a "one link to rule them all" choice?

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+layer\_descriptions|dict with string keys and values|One key must be "X" which describes the transformations (if any) performed to produce the X matrix cellxgene displays.
+organism|`str`| 
+organism\_ontology|`str`| 
+project\_name|`str`|Name of the project that contains this dataset
+project\_description|`str`| 
+project\_protocol\_links|`list` of `link`s|May be empty list
+project\_raw\_data\_links|`list` of `link`s|May be empty list
+project\_other\_links|`list` of `link`s|May be empty list
+
+The `project\_protocol\_links`, `project\_raw\_data\_links`, and `project\_other\_links` keys are always present in `uns` as
```
bkmartinjr, line 38: formatting error - don't escape underscores when in backtick quotes.

Ie, you want `project_protocol_links`, not `project\_protocol\_links`

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+tags|`list` of `str`s| 
+<obs\_column>\_colors|`list` of color specifications (see anndata/cellxgene documentation)|<obs\_column> must be a column name in `obs`. There may be multiple keys like this.
+
+
+## `obsm`
+
+`obsm` is a mapping from `str`s to matrices of shape (n, m) where `n` is the number of rows in `X` and `m >= 1`.
+Corpora requires one value in `obms` to be a two-dimensional embedding, meaning it is of shape (n, 2) and is a
+numeric type.
+
```
bkmartinjr, line 59: does this comment about obsm requirements belong in the H5AD encoding spec, or in the main corpora spec?  Ie, is it general to all encodings, or only H5AD?

backend/schema/corpora_schema_h5ad_implementation.md
bkmartinjr: In addition, I believe the embedding names must be prefixed with `X_`.

This is covered in the more general Corpora schema doc - you might want to move that information (`Implementations` section) to this doc.

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+### Note on types
+The types below are python3 types. Note that a python3 `str` is a sequence of Unicode code points, which is stored
+null-terminated and UTF-8-encoded by anndata.
+
+For succinctness, there is a `link` type used below that is a `dict` {"link_name": `str`, "link_url": `str`}.
+
+## `X`
+
+Corpora does not impose any additional constraints on the `X` data matrix. So it may be sparse or dense and any
```
bkmartinjr, line 18: are we following AnnData conventions for layers / raw?  The corpora spec is a bit thin on this, which is something we need to nail down as it is commonly used, and the distinction is important for downstream use cases.

backend/schema/corpora_schema_h5ad_implementation.md
bkmartinjr: I would also consider categorical where the underlying primitive is not a string (probably not useful for these fields, but we do see them occasionally for ad hoc fields).

backend/schema/corpora_schema.md
mckinsel: `ontology` matches what the HCA does, though they also then use `ontology_label`. I suppose we could split the difference and do `ontology_term_id`?

backend/schema/corpora_schema.md
mckinsel: For "term" I pretty much always mean "compact uri", so it would be like "HsapDv:00001". And you can reliably get the ontology by splitting on `:`. I'll clarify that.

As far a version, most of the ontologies seem pretty weakly versioned. There are release dates, but those can be pretty frequent, and they don't do much to show whether one term has changed.

backend/schema/corpora_schema.md
mckinsel: It feels like a lot of commas to me, but I will add.

backend/schema/corpora_schema.md
mckinsel: I don't think any of these links is a reliable candidate for that. Do you need a `project_summary_link` then?

backend/schema/corpora_schema.md
mckinsel: "version" might not be the right term here, I believe the intent is more counting dataset revisions. Is that right @maniarathi ?

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Yeah agree, I will add that in.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Okay that seems reasonable.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Thanks, fixed

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: 🤦 thanks

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Yeah, though `n_obs` and `n_vars`?

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: I suppose we wouldn't want to say having PCA only breaks the spec, so `m>=2`. I'll add the stuff about `X_` and being finite.

backend/schema/corpora_schema.md
bkmartinjr: I'm not super picky - just trying to minimize confusion by downstream users.  I think `ontology_term_id` is very clear.

backend/schema/corpora_schema.md
bkmartinjr: got it - I didn't understand that you were using this term format.  Definitely should specify that.  

Also suggest more prescriptive language around what is required, vs suggested, so we can draw a straight line from the spec to a validator (software)

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
+# Corpora Schema
```
maniarathi, line 1: Are we deprecating the term "corpora" at this point?

backend/schema/corpora_schema.md
bkmartinjr: ah, I totally misunderstood.  Perhaps "revision", "update", "variant", or?

backend/schema/corpora_schema_h5ad_implementation.md
bkmartinjr: yes, you are correct - I mentally mapped it into Javascript :-)

backend/schema/corpora_schema.md
bkmartinjr: eats, shoots, and leaves...?

backend/schema/corpora_schema.md
bkmartinjr: yes, I think so.  That should be the one URL that, if not mandatory, is very strongly encouraged...

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+
+To support data integration, each cell must have the following metadata:
+
+**Field name**|**Constraints**
+:--|:--
+tissue|string
+assay|string
+disease|string
+cell\_type|string
+sex|“male”, “female”, “unknown”, or “na”
```
maniarathi, line 120: You know thinking about it more, does it make sense to have an "other" category which has a different meaning than unknown or not applicable? To be more inclusive?

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+cell\_type\_ontology|CL term
+ethnicity\_ontology|HANCESTRO term, “na” if non-human
+development\_stage\_ontology|HsapDv term if human, child of EFO:0000399 otherwise
+
+
+The ontology fields are all “best effort” fields. We expect that there will be times that no ontology term exists, especially for cell_type. In those cases,
+the ontology field may be left empty. Also, we don’t require submitters to include these fields in their submissions. If the submitter does not include
+ontology fields, then as part of the curation process, Corpora will insert appropriate ontology values based on the text from the submitters. If the
+submitter does include ontology values, then Corpora just uses those. Finally, tissue and tissue_ontology can be appended with “(cell culture)” or
+“(organoid)” if appropriate.
```
maniarathi, line 140: I'd move the sentence about appended with the parenthesis right underneath the tables to make it more clear. Right now this important nugget is a bit hidden. Or even mention it directly in the contraints.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: I think wrt cellxgene, all we need to say is that cellxgene is going to display X, right?

Beyond that, I think we need to get more assertive the schema about raw being required.

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+
+If the features of the dataset are human genes, then the feature ids must be [HGNC](https://www.genenames.org/about/guidelines/#!/#tocAnchor-1-7) approved
+gene symbols. If the original dataset does not use HGNC symbols, Corpora will perform a (possibly lossy) conversion from the submitted features to HGNC
```
maniarathi, line 143: Does it make sense to specify how the conversion will be done (i.e. added together)?

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Yeah, categorical works too. I'll update that.

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+
+
+**Field name**|**Description**
+:--|:--
+project\_name|String
+project\_description|Longer description of the project.
+project\_protocol\_links|List of links to protocol descriptions, to protocols.io for example
+project\_raw\_data\_links|List of links to repositories where raw data are available
+project\_other\_links|List of links to other sites the submitter deems relevant, author data sites for example.
+
```
maniarathi, line 249: Totally optional, more of just a thought:

We could simplify this by having a single `links` field and each element contains the dict `{"link_name": "...", "link_url": "...", "link_type": {PROTOCOL, RAW_DATA, OTHER}}`.

backend/schema/corpora_schema.md
maniarathi: `revision` is what we use internally. And yes, it's to track the number of times the same dataset is changed.

backend/schema/corpora_schema.md
maniarathi, line 1: Are we deprecating Corpora here?

backend/schema/corpora_schema.md
maniarathi, line 140: I'd move the sentence about appending with the parenthesis either directly underneath the tables or right in the Constraints column. Other the nugget of information feels a little hidden to me, especially if a submitted reads it.

backend/schema/corpora_schema.md
maniarathi, line 143: Explain in what way it is lossy? (i.e. we add them up)?

backend/schema/corpora_schema.md
maniarathi: So, for example, when, cellxgene starts, this is, the field, that .. <- do that. Much more dramatic.

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,323 @@
...
+the submitted dataset is used to create a new project.
+
+
+**Field name**|**Description**
+:--|:--
+project\_name|String
+project\_description|Longer description of the project.
+project\_protocol\_links|List of links to protocol descriptions, to protocols.io for example
+project\_raw\_data\_links|List of links to repositories where raw data are available
+project\_other\_links|List of links to other sites the submitter deems relevant, author data sites for example.
```
maniarathi, line 248: More of a thought, rather than requesting a change exactly.. this could be simplified I think. What do you think about having a single `links` field where each element is a dict `{"link_name": "...", "link_url": "...", "link_type": {PROTOCOL, RAW_DATA, OTHER}}`?

backend/schema/corpora_schema.md
maniarathi: Yep. I think we use `revision` internally.

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+
+Recall that `uns` is a mapping with `str`s as keys. Corpora requires the following keys and values in `uns`:
+
+**Key**|**Value Type**|**Notes**
+-----|-----|-----
+title|`str`|Title of the dataset
+contributors|`list` of `dict`s. each `dict` must a "name": `str` item and may optionally have "email": `str` and "institution": `str`|This is the person who gave us the dataset, not necessarily the author or actual data generator.
+layer\_descriptions|dict with string keys and values|One key must be "X" which describes the transformations (if any) performed to produce the X matrix cellxgene displays.
+organism|`str`| 
+organism\_ontology|`str`|
```
maniarathi, line 31: Might as well reiterate the preferred ontologies in the Notes column?

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,78 @@
...
+
+## `uns`
+
+Recall that `uns` is a mapping with `str`s as keys. Corpora requires the following keys and values in `uns`:
+
+**Key**|**Value Type**|**Notes**
+-----|-----|-----
+title|`str`|Title of the dataset
+contributors|`list` of `dict`s. each `dict` must a "name": `str` item and may optionally have "email": `str` and "institution": `str`|This is the person who gave us the dataset, not necessarily the author or actual data generator.
+layer\_descriptions|dict with string keys and values|One key must be "X" which describes the transformations (if any) performed to produce the X matrix cellxgene displays.
```
maniarathi, line 29:
```suggestion
layer\_descriptions|`dict` with string keys and values|One key must be "X" which describes the transformations (if any) performed to produce the X matrix cellxgene displays.
```
OCD kicked in.

backend/schema/corpora_schema_h5ad_implementation.md
bkmartinjr: When we will be able to support more than one "layer"?  Do you want to build in the concept now, or just say "there is one X, and it should be raw"?

backend/schema/corpora_schema.md
bkmartinjr: you still need a name that differentiates from the more agnostic viewpoing cellxgene has.   Corpora is a fine short-hand for that, easier than "the more prescriptive schema we want all data to eventually be converted to so it is integrable."

mckinsel: Yeah I'm open to pretty much any word, but I don't think we can s/corpora/cellxgene/ the way we have everywhere else because this definitely isn't a "cellxgene schema".

backend/schema/corpora_schema.md
mckinsel: We could do [what the HCA does](https://data.humancellatlas.org/metadata/dictionary/biomaterial/donor_organism#donor_organism-sex), which is use "male", "female", "mixed", or "unknown". Or there's the [gender data element](https://cdebrowser.nci.nih.gov/cdebrowserClient/cdeBrowser.html#/search?publicId=2200604&version=3.0) from NIH, which allows "male", "female", "intersex", "not reported", "unknown", and "unspecified".

backend/schema/corpora_schema.md
mckinsel: Yeah I'll add that in.

mckinsel: I think the procedure we settled on is in fact not lossy, so I'll take that out.

backend/schema/corpora_schema.md
mckinsel: Yeah good point, I'll make that more clear.

backend/schema/corpora_schema.md
mckinsel: So 👏 for 👏 example 👏 when 👏 cellxgene 👏 starts 👏

backend/schema/corpora_schema.md
mckinsel: Yeah that'd work for me.

backend/schema/corpora_schema.md
maniarathi: SG!

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: I'd rather not be too prescriptive about it at the moment, since they're on the cusp of redoing the API for layers.

backend/schema/corpora_schema.md
mckinsel: Yeah, added link_type

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Thanks, fixed it.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: I clarified a little in the main schema. I think there can be a note here as well.

backend/schema/corpora_schema.md
mckinsel: Let's punt on this for the time being. We'll need to ask @ambrosejcarr what the new word is for the data integration work part of the project now known as cellxgene.

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,90 @@
...
+Corpora does require that raw counts be present in the h5ad. If the `X` data matrix is not raw counts, then the h5ad
+must have a `raw` attribute where the raw counts are stored in `raw.X`
+
+## `uns`
+
+Recall that `uns` is a mapping with `str`s as keys. Corpora requires the following keys and values in `uns`:
+
+**Key**|**Value Type**|**Notes**
+-----|-----|-----
+version|`dict` with keys "corpora_schema_version" and "corpora_encoding_version" indicating the version of the schema and schema encoding used to create the h5ad.
```
bkmartinjr, line 37: what is the value type for the version keys?   Semver string, eg, "X.Y.Z"?

bkmartinjr: also, should each field in this "encoding" spec be defined in the corpora schema spec?  Most are, but not this one.  That may be the right location to define the version type/encoding?

backend/schema/corpora_schema.md
```diff
@@ -0,0 +1,331 @@
...
+
+To support data integration, each cell must have the following metadata:
+
+**Field name**|**Constraints**
+:--|:--
+tissue|string (see note below regarding cell culture and organoids)
+assay|string
+disease|string
+cell\_type|string
+sex|"male", "female", "mixed", "unknown"
```
bkmartinjr, line 122: is this sufficient for non-mammalian data?

mckinsel: No, though we don't have any of that. I can add back "other".

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,90 @@
...
+tissue\_ontology\_term\_id|`str` or categorical with `str` categories|Not necessarily human-readable
+assay\_ontology\_term\_id|`str` or categorical with `str` categories|Not necessarily human-readable
+disease\_ontology\_term\_id|`str` or categorical with `str` categories|Not necessarily human-readable
+cell\_type\_ontology\_term\_id|`str` or categorical with `str` categories|Not necessarily human-readable
+ethnicity\_ontology\_term\_id|`str` or categorical with `str` categories|Not necessarily human-readable
+development\_stage\_ontology\_term\_id|`str` or categorical with `str` categories|Not necessarily human-readable
+
+## `var`
+
+The Corpora schema requires unique feature identifiers, so the index of `var` must not contain any duplicate values.
```
bkmartinjr, line 90: given that we also need human readable / friendly names, it might be good to make the language stronger around that assumption/requirement.

backend/schema/corpora_schema_h5ad_implementation.md
```diff
@@ -0,0 +1,90 @@
+# Corpora Schema AnnData Implementation
```
bkmartinjr, line 1: it would be good to have a section in this doc on how a tool can unambiguously determine if a dataset _is_ a Corpora dataset.  Just looking for `uns.version` seems likely to trip over other uses of that generic name.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: 👍 Added the SemVer requirement and version to the schema.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Yep, will update.

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Added a note

backend/schema/corpora_schema_h5ad_implementation.md
mckinsel: Per another comment, I think I'm going to leave as much semantic information as possible in the main document rather than the implementation doc.

PR #484, opened 2020-07-27
Implementing DbUtils as a singleton

### LINKED ISSUE #449: Close sessions after interacting.
Currently when a lambda is invoked a DB session is opened. That session will persist until that lambda is torn down. This leave DB sessions for an indefinite amount of time if the lambda is reused a lot. This can cause problems when trying to manage the database or test the database. [SQLAlchemy recommends](https://docs.sqlalchemy.org/en/13/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it) keeping the session open for a "unit of work" before closing, commit, or rolling back. [Flask-SQLAlchemy](https://flask-sqlalchemy.palletsprojects.com/en/2.x/) provide resource to help manage sessions, we can also build out own solution.

Bento007:
This automates the opening and closing of sessions to the project database. The goal is to limit the persistence of a sessions to a single api call or database transaction. The consequence of not properly closing session is an inconsistent database state within the aws lambda. We are following the guidelines of SQLAlchemy to limit the context of a session to a "unit of work".

fixes #449


brianraymor:
@Bento007 - can you connect this pull request to its appropriate issue?

Bento007:
@brianraymor it's not related to an issue it's just a general code improvement, and also breaking it out from other code for simplifying the review process.

mckinsel:
Are you sure we _want_ a single global session? It seems like another piece of state to keep track of.

Bento007:
Since this will run in a lambda I think it's important to keep tight control over the number of database sessions spawned. In general there should only ever be one running in the lambda, and this bit of code helps enforce that. This work is related to https://github.com/chanzuckerberg/corpora-data-portal/issues/449. In the future i'll create a context manager around the api request that will open and close the session automatically.

mckinsel:
Hey can you include some detail in the PR description about the problematic situation you're addressing with this? It says this helps with "creating multiple database sessions", but afaict this change leads to _more_ sessions being opened (and hopefully closed).

PR #494, opened 2020-07-30
Dunitz/cellxgene db  model import

MDunitz:
## Need
- To add cellxgene tables to corpora DB

## Approach
- Import from cellxgene package and create additional base

## Testing
- Drop/Create tables
- Add rows to cellxgene tables
- Check local db to ensure rows created correctly

## Note
- Do not merge until updated cellxgene package is released

### DIFF
--- a/backend/database/versions/f6b189a1d9c1_create_cellxgene_tables.py
+++ b/backend/database/versions/f6b189a1d9c1_create_cellxgene_tables.py
@@ -1,0 +1,56 @@
+"""create_cellxgene_tables
+
+Revision ID: f6b189a1d9c1
+Revises: 899656d37baa
+Create Date: 2020-07-30 12:03:18.275767
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+from sqlalchemy import Column, String, TIMESTAMP, func
+
+revision = 'f6b189a1d9c1'
+down_revision = '899656d37baa'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "cxguser",
+        Column("id", String, nullable=False, primary_key=True),
+        Column("created_at", TIMESTAMP, nullable=False, server_default=func.now()),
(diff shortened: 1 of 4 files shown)


backend/scripts/create_db.py
```diff
@@ -10,16 +10,29 @@
...
+from .corpora_config import CorporaDbConfig
+from .corpora_orm import Base
+
+
+from sqlalchemy.schema import DropTable
+from sqlalchemy.ext.compiler import compiles
+from server.db.cellxgene_orm import Base as CellxGeneBase
+
+
+@compiles(DropTable, "postgresql")
```
Bento007, line 22: So what all is happening in this function?

backend/scripts/create_db.py
```diff
@@ -10,16 +10,29 @@
 pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
 sys.path.insert(0, pkg_root)  # noqa
 
-from corpora.common.corpora_config import CorporaDbConfig
-from corpora.common.corpora_orm import Base
+from .corpora_config import CorporaDbConfig
+from .corpora_orm import Base
```
Bento007, line 14: I recommend renaming to `CorporaBase` for clarity

backend/database/versions/f6b189a1d9c1_create_cellxgene_tables.py
```diff
@@ -0,0 +1,56 @@
...
+from sqlalchemy import Column, String, TIMESTAMP, func
+
+revision = 'f6b189a1d9c1'
+down_revision = '899656d37baa'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
```
Bento007, line 22: wouldn't a cellxgene user be the same thing as a corpora user according to the DB?

backend/scripts/create_db.py
MDunitz: It appends cascade to the drop table command to handle foreign keys/dependencies

PR #510, opened 2020-08-05
psycopg2 fix

Bento007:
Using `psycopg2_binary` for api-server lambda. It will be more update than our previous solution of using [jkehler/awslambda-psycopg2](https://github.com/jkehler/awslambda-psycopg2) and will require less update maintenance on our part. However the psycopg2 developers do not recommend using `psycopg2_binary` for production environments. They recommend building from source which is easier said than done. `psycopg2_binary` has been used in [HumanCellAtlas/query-service](https://github.com/HumanCellAtlas/query-service/blob/dd2c2844b1f6836cb0f151e092f31e9021724f31/requirements.txt.in#L1) which give me confidence that this solution should work for us for the time being.


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - tsmith/psycopg_fix
```
MDunitz, line 7: is this just for testing?

Bento007: just for testing. I'll remove it before submitting

PR #511, opened 2020-08-06
332 cookie banner

### LINKED ISSUE #332: Implement cookie banner on frontend
[UPDATE]
PR is out for review! Probably not worth estimating. Updated points from 3 -> 1

===============

The corpora data portal frontend must get users approval to use tracking cookies in the users browser. These cookies are used for google analytics.

This will require displaying a banner to inform users that tracking cookies are used.

The frontend implementation will match the implementation currently in [cellxgene](https://github.com/chanzuckerberg/cellxgene/blob/master/client/src/components/termsPrompt/index.js). @pdugan20 will update the design for it to gel better with the Corpora temp landing page.

tihuan:
This PR adds the cookie banner! More notes are added as comments 😄 

The pending decision here is whether to use cookie or localStorage to save the user preference, waiting on infosec's advice. The PR has commits for both, so we can switch to the recommended one once we get an answer!

Thank you 🎉 !

closes #332 

Mock:

Screen Shot 2020-08-06 at 9 35 50 AM

Real:

Screen Shot 2020-08-06 at 9 35 36 AM


frontend/.env.development
```diff
@@ -6,3 +6,4 @@ AUTH0_CALLBACK=http://localhost:8000
 # Dev env API_URL: https://api.dev.corpora.cziscience.com
 API_URL=http://localhost:3000
 CXG_URL=http://cellxgene-dev.us-west-2.elasticbeanstalk.com
+ENV=dev
```
tihuan, line 9: Add `ENV` global variable to help tell dev build from prod build

frontend/src/common/constants/localStorageKeys.ts
```diff
@@ -0,0 +1,4 @@
+export enum LOCAL_STORAGE_KEYS {
+  // (thuang): This is shared with `cellxgene` repo
+  COOKIE_DECISION = "cxg.cookieDecision",
```
tihuan, line 3: this is the same key as cellxgene's

frontend/src/common/localStorage/get.ts
```diff
@@ -0,0 +1,7 @@
+export function get(key: string) {
+  try {
```
tihuan, line 2: in case the user's browser doesn't support localStorage, we'll catch it

tihuan:
Turning this into official PR now that infosec is okay with us using LocalStorage 🤩 Thanks, @Bento007 🙏 !

tihuan:
@Bento007 Ah I see unit test is failing due to a missing credential. Not sure what's going on with that, so I'm rerunning the tests now, in case it's just Github being flaky!

Screen Shot 2020-08-06 at 2 23 01 PM

tihuan:
OK it does look like Github was just flaky for the `unit-test` 👍

frontend/src/common/localStorage/set.ts
```diff
@@ -0,0 +1,8 @@
+export enum BOOLEAN {
+  TRUE = "true",
+  FALSE = "false",
+}
+
+export function set(key: string, value: BOOLEAN) {
+  window?.localStorage?.setItem(key, value);
```
Bento007, line 7: do we need another catch, if the browser doesn't support localStorage?

tihuan: Thanks for catching this!

Since we use optional chaining, if the browser doesn't support localStorage, it will not call `setItem()` on `window.localStorage`, which could be `undefined`, so it won't throw a runtime error and thus nothing to catch!

But thanks for catching this, because it made me realize my previous implementation of `get()` had a bug, where `catch()` block will never execute, so the function could return `undefined`, if localStorage is not supported

BEFORE
```ts
export function get(key: string) {
  try {
    return window?.localStorage?.getItem(key); // string, null, or undefined
  } catch (e) {
    return null;
  }
}
```

AFTER:
```ts
export function get(key: string): string | null {
  return window?.localStorage?.getItem(key) || null; // string or null
}

```

PR #513, opened 2020-08-08
First pass at authentication in data portal

bmccandless:
Implements the login, logout, and userinfo endpoints.
It saves the id token in a cookie, and verifies the cookie when it is read.

 Issue:  chanzuckerberg/corpora-data-portal#481


backend/chalice/api_server/app.py
```diff
@@ -49,6 +53,17 @@ def get_chalice_app(flask_app):
     app.debug = flask_app.debug
     app.log.setLevel(logging.DEBUG)
 
+    # set the flask secret key, needed for session cookies
+    flask_secret_key = "OpenSesame"
+    deployment = os.environ["DEPLOYMENT_STAGE"]
+    if deployment != "test":
+        secret_name = f"corpora/backend/{os.environ['DEPLOYMENT_STAGE']}/auth0-secret"
+        auth_secret = json.loads(AwsSecret(secret_name).value)
```
Bento007, line 61: We need a strategy to avoid calling Aws Secrets for every api request. It add an additional remote call to every authenticated endpoint. A timed cache sounds like the appropriate solution. Maybe grab the secret every 5 minutes. This would be a change made to AwsSecret.

backend/config/corpora-api.yml
```diff
@@ -492,11 +492,85 @@ paths:
...
+        "401":
+          $ref: "#/components/responses/401"
+        "403":
+          $ref: "#/components/responses/403"
+
+  /v1/oauth2/callback:
+    get:
+      tags:
+        - authentication
+      summary: callback function from auth server after login
```
Bento007, line 546: can this also be used for logout?

backend/config/corpora-api.yml
```diff
@@ -492,11 +492,85 @@ paths:
...
     user_uuid:
       description: A unique identifier of a logged in User of Corpora.
       type: string
+    userinfo:
+      type: object
+      properties:
+        is_authentifacted:
+          type: bool
+        id:
+          $ref: "#/components/schemas/project_uuid"
```
Bento007, line 567:
```suggestion
          $ref: "#/components/schemas/user_uuid"
```

Bento007: is this the user id from auth0 or from our database? We might want to use the same one for both, or add the auth0 id as a column in the user table.

backend/chalice/api_server/app.py
Bento007: This doesn't need to be done in this PR.

backend/config/corpora-api.yml
bmccandless: No, this is just for login.   The logout is handled in a simpler way.  It logs out of auth0, removes the cookie, then redirects back to "/".

backend/chalice/api_server/app.py
bmccandless: Agreed.   I'll do that in the next pass.

backend/config/corpora-api.yml
bmccandless: It's the id from auth0.  I assumed we would use that as the id for the the database too.   I assumed when we populate the database with users then we would use this as the key.

maniarathi:
Hey -- just noticed that there's a pretty big loss in unit test coverage with this PR. Could you please take a look and add tests for the new functionality? Thanks in advance!

backend/chalice/api_server/app.py
bmccandless: @Bento007 was there a reason you thought it should be a timed cache, rather than just a cache.   I'm asking because a regular cache I can implement with 1 line on the "get_oauth_config" function.   I guess a reason for it to be timed is because we could then change it on the fly without redeploying - but that would lead to other issues for the up to 5 minutes when it would be out of sync.   

Also, I see there is a base class called SecretConfig which is used for CorporaConfig and CorporaDbConfig.  Those also cache permanently.  I didn't see that when I first implemented this.  It looks like I could follow along with that pattern.  Any thoughts on that?

bmccandless:
> Hey -- just noticed that there's a pretty big loss in unit test coverage with this PR. Could you please take a look and add tests for the new functionality? Thanks in advance!

Yes, some of this is hard to test since it requires interaction with the oauth server to really test the flow. I plan to add in more testing before I'm all done.

backend/chalice/api_server/app.py
Bento007: > but that would lead to other issues for the up to 5 minutes when it would be out of sync.

A time cache was the first solution to come to mind but we could also, redeploy the lambdas every time the secret rotates. For now we can wait and see if it become a problem.

>  I didn't see that when I first implemented this. It looks like I could follow along with that pattern. Any thoughts on that?

:+1:

backend/corpora/common/authorizer.py
```diff
@@ -1,54 +1,54 @@
...
+    audience = auth_config.audience
+    public_keys = get_public_keys(auth0_domain)
     public_key = public_keys.get(unverified_header["kid"])
     if public_key:
+        algorithms = ["RS256"]
         try:
-            payload = jwt.decode(
-                token, public_key, algorithms=ALGORITHMS, audience=API_AUDIENCE, issuer=f"https://{AUTH0_DOMAIN}/"
-            )
+            if not auth0_domain.endswith("/"):
```
Bento007, line 26: don't we produce this value? why would we be missing the `/`?

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -0,0 +1,82 @@
...
+from urllib.parse import urlencode
+from functools import lru_cache
+import base64
+from ....common.authorizer import get_userinfo, assert_authorized_token
+from ....common.corpora_config import CorporaAuthConfig
+
+
+@lru_cache(maxsize=1)
+def create_oauth_client(config):
+    oauth = OAuth(current_app)
```
Bento007, line 13: could this be instantiated in a more central location, to prevent creating multiple? Maybe it can be created in corpora_config and imported there.

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def logout():
+    """api call,  logout of the system"""
+    config = CorporaAuthConfig()
+    client = create_oauth_client(config)
+    params = {"returnTo": config.callback_base_url, "client_id": config.client_id}
+    response = redirect(client.api_base_url + "/v2/logout?" + urlencode(params))
+    # remove the cookie
+    response.set_cookie(config.cookie_name, "", expires=0)
```
Bento007, line 43: Ah, so that's how you unset a cookie.

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -0,0 +1,82 @@
...
+    client = create_oauth_client(config)
+    callbackurl = f"{config.callback_base_url}/v1/oauth2/callback"
+    response = client.authorize_redirect(redirect_uri=callbackurl)
+    return response
+
+
+def logout():
+    """api call,  logout of the system"""
+    config = CorporaAuthConfig()
+    client = create_oauth_client(config)
```
Bento007, line 39: Can this oauth_client be reused, or do we need to create a new one for each api call?

backend/corpora/common/authorizer.py
bmccandless: Just to be less picky about the auth0_domain.   Otherwise when setting the value in the secret you'd have to know that the code that used it expected an extra /.  So, this is just to allow either.   True, we control the value.

backend/corpora/lambdas/api/v1/authentication.py
bmccandless: It's only created in this one file.   The lru_cache should make it so that only one is ever created.   Although I could have also implemented as a global variable in this module, which might have been clearer (there can be only one config).   Putting it in corpora_config would be fine, except then I'd have to import a few things into corpora_config (like flask.current_app, and OAuth), which doesn't seem right.  I wanted this file to be the only one that knew about authlib.

backend/corpora/lambdas/api/v1/authentication.py
bmccandless: There should only be one.  create_oauth_client returns one that has already been created.

PR #541, opened 2020-08-20
ci: fix staging auto deploy

Bento007:
The current make file assumes that there will be 3 subnets_ids, which is the case for both dev and prod. However for staging, there are only 3. This change extracts the relevant subnet_ids and adds them to the chalice config.json


brianraymor:
@Bento007 - can this be connected to its related issue?

PR #542, opened 2020-08-20
Change dataset ontology types to JSON

mckinsel:
Datasets can have more than one assay, tissue, etc. This changes the
represention of those attributes in the database from a string to JSON.


brianraymor:
@mckinsel - can this be connected to the related issue?

backend/corpora/common/corpora_orm.py
```diff
@@ -30,7 +31,7 @@ class TransformingBase(object):
     """
 
     def __iter__(self):
-        return self.to_dict().iteritems()
+        return iter(self.to_dict().items())
```
Bento007, line 34: what is the difference?

mckinsel: I believe `iteritems` isn't a thing python3

Bento007: ahhh

backend/database/versions/7968f161b60c_change_ontology_to_json.py
```diff
@@ -0,0 +1,96 @@
...
+            "json_build_array(json_build_object("
+            "'ontology_term_id', development_stage_ontology, 'label', development_stage))"
+        ),
+    )
+    op.drop_column(table_name="dataset", column_name="development_stage_ontology")
+
+    op.alter_column(table_name="dataset", column_name="sex", type_=JSONB(), postgresql_using="json_build_array(sex)")
+
+
+def downgrade():
```
Bento007, line 77: Wouldn't we need to multiplex the json into multiple columns? This would get especially interesting if we have any array of multiple json objects to parse into two columns.

backend/database/versions/7968f161b60c_change_ontology_to_json.py
```diff
@@ -0,0 +1,96 @@
...
+    op.alter_column(table_name="dataset", column_name="organism", type_=String())
```
Bento007, line 78: Doesn't it make sense to parse the column whiles it's json before converting it back into a string.

PR #545, opened 2020-08-24
Add a couple missing fields from the schema

mckinsel:


backend/corpora/common/corpora_orm.py
```diff
@@ -122,11 +122,13 @@ class ProjectLinkType(enum.Enum):
 
     PROTOCOL - A link to a sequencing protocol.
     RAW_DATA - A link to a raw data repository.
+    SUMMARY - The "summary" link cellxgene should display.
     OTHER - Other.
     """
 
     PROTOCOL = "Protocol"
     RAW_DATA = "Raw data"
+    SUMMARY = "Summary"
```
Bento007, line 131: Enums are handles at in the python code, not in the DB schema, so migration is not required.

mckinsel: So after poking around a bit, I think the Enums are also enforced by postgres as a column type. I added a migration script.

mckinsel: Uhh those these string do not match...so now I'm not sure what's going on.

mckinsel: Oh nvm, we pass `name` not `value`.

backend/database/versions/fa0950f6df19_add_summary_link.py
```diff
@@ -0,0 +1,57 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "fa0950f6df19"
+down_revision = "7968f161b60c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    project_link_enum = ENUM(name="projectlink")
```
Bento007, line 20: where does name="projectlink" come from? I don't see it set in corpora_orm.py. Should we be setting the name in corpora_orm.py as well?

mckinsel: Afaict, it's not related to the orm code except that there are identical python enums in there. Searching around a little, it does seem like there are tools or patterns [[1](https://pypi.org/project/SqlAlchemy-Enum-Tables/#description), [2](https://www.thetopsites.net/article/53556214.shtml)] to tighten the link between them. That said, the migration script is just run once.

PR #547, opened 2020-08-24
Pin s3fs dependency

mckinsel:
This mirrors a change in cellxgene. The latest s3fs creates a bunch of errors in aiobotocore.


Bento007:
LGTM

PR #566, opened 2020-08-29
Haromize dev, staging, prod db schemas

mckinsel:
This is an ugly commit, but the schemas of dev, staging, and prod had
drifted apart, weren't properly under the control of alembic, and
differed from the ORM code. This uses some autogenerated revisions
applied to each brach to bring them in line with the ORM code. The
alembic version 9023d0bf61ab was stamped on every environment's db, and
the histories were merged (kind of) via alembic merging.

There were a couple little raw SQL tweaks needed (time with vs without
time zone, names but not content of foreign key constraints, etc) in
addition, but now all three environments have schemas that are identical
per apgdiff and are on the same alembic version num.

One caution is that if you want to downgrade gracefully, you're going to
have a really bad time. But that was true before this commit too.

Note that the history now looks a little complicated. Dev and staging were migrated
and then prod:
```
dev ------
          \_______
          /      /
staging --      /
               /
prod ----------
```


backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
```diff
@@ -0,0 +1,86 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table("dataset_deployment")
+    op.alter_column("dataset", "project_id", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("dataset", "project_status", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("dataset", "revision", existing_type=sa.INTEGER(), nullable=True)
+    op.alter_column("deployment_directory", "created_at", existing_type=postgresql.TIMESTAMP(), nullable=False)
```
maniarathi, line 25: Any `created_at` and `updated_at` fields should have a `server_default=func.now()` I believe so that it gets populated by the database and not us.

(from sqlalchemy import func)

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
```diff
@@ -0,0 +1,86 @@
...
+    op.drop_constraint(None, "deployment_directory", type_="foreignkey")
+    op.alter_column("deployment_directory", "updated_at", existing_type=postgresql.TIMESTAMP(), nullable=True)
+    op.alter_column("deployment_directory", "dataset_id", existing_type=sa.VARCHAR(), nullable=True)
+    op.alter_column("deployment_directory", "created_at", existing_type=postgresql.TIMESTAMP(), nullable=True)
+    op.alter_column("dataset", "revision", existing_type=sa.INTEGER(), nullable=False)
+    op.alter_column("dataset", "project_status", existing_type=sa.VARCHAR(), nullable=True)
+    op.alter_column("dataset", "project_id", existing_type=sa.VARCHAR(), nullable=True)
+    op.create_table(
+        "dataset_deployment",
+        sa.Column("id", sa.VARCHAR(), autoincrement=False, nullable=False),
```
maniarathi, line 73: I think setting autoincrement isn't necessary here (and other locations throughout this PR). It's only does something for INTEGER type Columns.

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
```diff
@@ -0,0 +1,86 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table("dataset_deployment")
+    op.alter_column("dataset", "project_id", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("dataset", "project_status", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("dataset", "revision", existing_type=sa.INTEGER(), nullable=True)
```
maniarathi, line 24: Why is revision nullable?

backend/database/versions/b2b6ffe65512_harmonize_schema_state_staging.py
```diff
@@ -0,0 +1,90 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table("dataset_deployment")
+    op.alter_column("dataset", "project_id", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("dataset", "project_status", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("dataset", "revision", existing_type=sa.INTEGER(), nullable=True)
+    op.alter_column("deployment_directory", "created_at", existing_type=postgresql.TIMESTAMP(), nullable=False)
+    op.alter_column("deployment_directory", "dataset_id", existing_type=sa.VARCHAR(), nullable=False)
+    op.alter_column("deployment_directory", "updated_at", existing_type=postgresql.TIMESTAMP(), nullable=False)
+    op.create_foreign_key(None, "deployment_directory", "dataset", ["dataset_id"], ["id"])
+    op.alter_column("project", "needs_attestation", existing_type=sa.BOOLEAN(), nullable=True)
```
maniarathi, line 29: Why is this nullable? Should just be a true or false and can default to false?

backend/database/versions/d9fb29897244_.py
```diff
@@ -0,0 +1,22 @@
...
+Revision ID: d9fb29897244
+Revises: b2b6ffe65512, 4ddb19dcc4e8
+Create Date: 2020-08-28 13:13:20.903219
+
+"""
+
+
+# revision identifiers, used by Alembic.
+revision = "d9fb29897244"
+down_revision = ("b2b6ffe65512", "4ddb19dcc4e8")
```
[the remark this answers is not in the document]
mckinsel: The history looks like this:
```
d9fb29897244 -> 9023d0bf61ab (head), harmonize_schema_state_prod
b2b6ffe65512, 4ddb19dcc4e8 -> d9fb29897244 (mergepoint), empty message
fa0950f6df19 -> b2b6ffe65512, harmonize_schema_state_staging
fa0950f6df19 -> 4ddb19dcc4e8, harmonize_schema_state_dev
7968f161b60c -> fa0950f6df19 (branchpoint), add_summary_link
899656d37baa -> 7968f161b60c, change_ontology_to_json
a17d25ca2013 -> 899656d37baa, updating_project_dataset_relationship
<base> -> a17d25ca2013, create_initial_tables
```

So it's kind of

```
dev ------
          \_______
          /      /
staging --      /
               /
prod ----------

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
mckinsel: Because [this](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/corpora/common/corpora_orm.py#L241) doesn't have a `nullable = False` kwarg. I agree that it should be non null though. Maybe that's a good candidate for a first post-harmonization migration?

backend/database/versions/b2b6ffe65512_harmonize_schema_state_staging.py
mckinsel: Similar to above, it's just reading the ORM code here, which specifies this as `needs_attestation = Column(Boolean)` which means nullable and no default. But yes, your suggestion sounds sensible.

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
mckinsel: Yeah, none of our code should ever populate it, and in fact postgres gives a DEFAULT=now() of some sort for updated_at. But it's still part of the schema, no?

backend/database/versions/d9fb29897244_.py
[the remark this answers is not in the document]
maniarathi: Oops - I forgot to delete that comment. Though your diagram is fantastic -- you should add it to the top level comment of the PR :D

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
maniarathi: Oh it's part of the schema yes, but I was wondering whether you need to specify the `server_default` argument in the `alter_column` function as well. It sounds like not though.

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
mckinsel: Huh yeah I don't know why that's there. I guess if it were True it might try to make it a `SERIAL`? I feel like that just wouldn't work though.

backend/database/versions/9023d0bf61ab_harmonize_schema_state_prod.py
mckinsel: Oh sorry, somehow I missed...like 80% of your comment. Let me double check.

mckinsel: Well, all the timestamps currently have a `DEFAULT now()`. So I'm guessing this was mostly just trying to set NOT NULL since the default was already set.

Issue #476, opened 2020-07-20
POST /v1/dataset/{dataset_asset_id}


Bento007:
Per the [API Design](https://docs.google.com/document/d/1d8tv2Ub5b3E7Il85adOAUcG8P05N6UBZJ3XbhJSRrFs/edit#heading=h.ybgspejrlzp5)

**Description**: Request to download a file which on success will generate a pre-signed URL to download the dataset.

**Parameters**:
1. Dataset UUID [REQUIRED]: the UUID of the dataset file to be downloaded.

**Responses**:
1. 200 OK: `{dataset_uuid, dataset_presigned_url, dataset_file_size}`
2. 401 Unauthorized user
3. 404 File not found

Tasks:
- [x] Determine what formatting of the API makes it conformant and "kosher".
- [x] Update High Level Arch doc with updated API.
- [x] Implement the API. Based on conversation with @Bento007 and @mckinsel , the s3 API can provide the file size of the file which will be returned in the response (note, this is different than the first draft definition of this API).
- [x] Test the API.

maniarathi:
Thinking about it further, the definition of this POST is probably wrong. We want `POST /v1/dataset/{dataset_asset_id}`. The dataset asset ids will already be available when the frontend call the `GET /v1/project/{project_uuid}`. The user will make a selection using the frontend modal but we only want to return a single pre-signed URL for a single dataset asset since are more than one available per dataset.

maniarathi:
Resource for generating the presigned URLs: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.generate_presigned_url

Bento007:
see #561 for file size download references

maniarathi:
FYI @Bento007 , you should be able to do something as simple as 
```
s3_client = boto3.client('s3')
response = s3_client.head_object(Bucket='bucketname', Key='keyname')
file_size = response['ContentLength']
```

Bento007:
Add Content-Disposition to the header of the response https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

Bento007:
Successfully tested the API on dev.

PR #568, opened 2020-08-31
feat(download): add API to generate presigned-URLs for files in S3

Bento007:
- Change /v1/dataset/{dataset_uuid} to   /v1/dataset/{dataset_uuid}/asset/{asset_uuid}
- Update the response of  /v1/dataset/{dataset_uuid}/asset/{asset_uuid} to include file size, and file name.
- add internal server error exception when a presigned-URL fails to generated.
- add s3_utility function to retrieve S3 file metadata.
- Change the test s3_uri to match the actual format of an S3 URI.
- Add error response details to API details
- Merge s3_utils.py into dataset_asset.py
- increase pre-signed url expiration from 3600 to 604800
- refactor the name of the flask problem error handler to handle_corpora_error
- create Asset entity class to manage dataset assets

Relates to #476


backend/config/corpora-api.yml
```diff
@@ -486,10 +491,16 @@ paths:
               schema:
                 type: object
                 properties:
-                  dataset_uuid:
+                  dataset_id:
                     type: string
```
maniarathi, line 495: Should this just be `$ref: "#/components/schemas/dataset_uuid"`

tests/unit/backend/chalice/api_server/test_v1_dataset.py
```diff
@@ -0,0 +1,60 @@
...
+import unittest
+
+import boto3
+from furl import furl
+from moto import mock_s3
+
+from tests.unit.backend.chalice.api_server import BaseAPITest
+
+
+class TestDataset(BaseAPITest, unittest.TestCase):
```
maniarathi, line 11: Inherit `CorporaTestCaseUsingMockAWS` which already mocks out S3 and more and even has a built in function to create S3 objects.

tests/unit/backend/corpora/common/utils/test_s3_utils.py
```diff
@@ -1,8 +1,9 @@
 import unittest
 
+import boto3
 from moto import mock_s3
 
-from backend.corpora.common.utils.s3_utils import generate_file_url
+from backend.corpora.common.utils.s3_utils import generate_file_url, head_file
 
 
 class TestS3Utils(unittest.TestCase):
```
maniarathi, line 9: Use `CorporaTestCaseUsingMockAWS` which mocks out S3 and other AWS resources. It even has a function to generate S3 objects. That will allow you to delete a ton of this code in this file and DRYs up the codebase.

backend/config/corpora-api.yml
```diff
@@ -771,15 +782,26 @@ components:
...
           type: string
         type:
           type: string
           enum: [REMIX, ORIGINAL]
         s3_uri:
           type: string
+    asset_format:
+      type: string
+      enum: [H5AD, RDS, LOOM, CXG]
+    problem:
```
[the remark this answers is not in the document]
Bento007: This define the shape of the error response from the 40X response. This is base on [Problem Details for HTTP APIs](https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00) which [connexion](https://connexion.readthedocs.io/en/latest/exceptions.html#:~:text=Flask%20by%20default%20contains%20an,problem%20.) uses. I like the standardized error response shape.

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -0,0 +1,48 @@
...
+from flask import make_response, jsonify
+
+from ....common.entities import Dataset
+from ....common.utils.db_utils import db_session
+from ....common.utils.exceptions import NotFoundHTTPException, ServerErrorHTTPException
+from ....common.utils.s3_utils import generate_file_url, head_file
+
+
+@db_session
+def post_dataset_artifact(dataset_uuid: str, asset_uuid: str):
```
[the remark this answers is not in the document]
Bento007: I guess until you get down to the DB this value should be asset. I'll refactor to asset.

backend/config/corpora-api.yml
Bento007: Will fix

backend/config/corpora-api.yml
```diff
@@ -486,10 +491,16 @@ paths:
               schema:
                 type: object
                 properties:
-                  dataset_uuid:
+                  dataset_id:
```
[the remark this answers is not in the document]
Bento007: I've been seeing uuid and id used interchangeably in the spec. In some areas we are using `id`, but when referring to a `id` with other `id`s in the context we've been using `{name}_uuid`. Since there are no other `id`s in the context of this response I justified to a myself a hybrid of `{name}_id` which follows no convention to so far. I think we should decide whether to use `uuid` or `id` in the spec, not both. I think `id` is sufficient, but would require more refectoring since `uuid` is more widely used in the code already. Changing to either one would cause some thrash with the frontend and backend since the API is already being used. For now I'll change it back to `dataset_uuid`

Bento007:
- refactor the name of the flask problem error handler to handle_corpora_error
- create Asset entity class to manage dataset assets
- Asset.delete stub added to prevent miss use until it is implemented

Bento007:
Keeping the s3_utils separate makes it easier to unit test the S3 functionality, but i did make efforts to make the code more object oriented by creating the asset class.

backend/config/corpora-api.yml
```diff
@@ -780,6 +791,18 @@ components:
           enum: [REMIX, ORIGINAL]
         s3_uri:
           type: string
+    problem:
+      type: object
+      description: >-
+        The error response shape based on
```
maniarathi, line 797: Will this actually show up as markdown in the spec? Could just simplify to "Error message container for HTTP APIs.

backend/config/corpora-api.yml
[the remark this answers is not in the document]
maniarathi: Thanks for the references! Makes sense.

backend/corpora/common/entities/asset.py
```diff
@@ -0,0 +1,28 @@
+import typing
+
+from backend.corpora.common.utils.s3_utils import head_file, generate_file_url
+from .entity import Entity
+from ..corpora_orm import DbDatasetArtifact
+import boto3
+
+
+class Asset(Entity):
```
maniarathi, line 9: Might call it a DatasetAsset to be slightly more descriptive.

backend/corpora/common/entities/asset.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def generate_file_url(self, expiration: int = 3600) -> typing.Union[str, None]:
+        return generate_file_url(self.bucket_name, self.file_prefix, expiration, self.s3)
+
+    def get_s3_metadata(self) -> typing.Union[dict, None]:
+        return head_file(self.bucket_name, self.file_prefix, s3=self.s3)
+
+    def delete(self):
+        """
+        TODO: Must also delete s3 objects related to the asset.
```
maniarathi, line 25: Is this placeholder needed right now? I actually imagine that deletion will happen at the dataset level instead of the asset level because fine-grained deletion (i.e. only delete the Loom file) won't be implemented for a long time.

backend/corpora/common/entities/asset.py
```diff
@@ -0,0 +1,28 @@
...
+class Asset(Entity):
+    table = DbDatasetArtifact
+    s3 = boto3.client("s3")
+
+    def __init__(self, db_object: DbDatasetArtifact):
+        super().__init__(db_object)
+        self.bucket_name, self.file_prefix = self.s3_uri[5:].split("/", 1)
+
+    def generate_file_url(self, expiration: int = 3600) -> typing.Union[str, None]:
+        return generate_file_url(self.bucket_name, self.file_prefix, expiration, self.s3)
```
maniarathi, line 18: Why not just add the code here? `generate_file_url` is not used anywhere else in the code and is very DatasetAsset specific for now. If other places need to use that code later, then we can refactor it. Otherwise, I'd prefer to keep it simple.

Ditto for `get_s3_metadata`.

backend/corpora/common/entities/asset.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def get_s3_metadata(self) -> typing.Union[dict, None]:
```
maniarathi, line 20: Is there a reason this function can't just be a very simple `get_dataset_asset_file_size` and returns the file size? We don't make use of any of the metadata and functions should be as narrow as possible in functionality.

backend/corpora/common/entities/dataset.py
```diff
@@ -79,10 +81,25 @@ def create(
...
         return cls(new_db_object)
 
+    def get_asset(self, asset_uuid) -> typing.Union[Asset, None]:
+        """
+        Retrieve the asset if it exists in the dataset.
+        :param asset_uuid: uuid of the asset to find
+        :return: If the asset is found it is returned, else None is returned.
+        """
+        asset = [asset for asset in self.artifacts if asset.id == asset_uuid]
+        if not asset:
```
maniarathi, line 91: Can simplify to `return Asset(asset[0]) if asset else None`.

maniarathi: Though we might want to handle the case that there is more than 1 asset with the same uuid? That should cause an error that we'd want to investigate.

backend/corpora/common/entities/dataset.py
```diff
@@ -79,10 +81,25 @@ def create(
...
+            return
+        else:
+            return Asset(asset[0])
+
     def delete(self):
         """
         Delete the Dataset and all child objects. Contributors connect to a dataset are deleted if they are not longer
         connected to any datasets.
+
+        TODO: Must also delete s3 objects related to the asset.
```
maniarathi, line 101: I'd hold off on adding TODOs related to deletion until we have sorted out our deletion user stories.

backend/corpora/common/entities/dataset.py
```diff
@@ -79,10 +81,25 @@ def create(
...
+        :return:
```
maniarathi, line 102: Delete or fill out?

tests/unit/backend/corpora/common/entities/test_asset.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class TestDataset(unittest.TestCase):
+    def setUp(self):
+        self.uuid = "test_dataset_artifact_id"
+
+    def test__get__ok(self):
+        asset = Asset.get(self.uuid)
+        self.assertEqual(self.uuid, asset.id)
+        self.assertEqual(CorporaTestCaseUsingMockAWS.CORPORA_TEST_CONFIG["bucket_name"], asset.bucket_name)
```
maniarathi, line 14: Ah, ok so what you want to do is have the TestAsset class inherit `CorporaTestCaseUsingMockAWS` ->`class TestAsset(CorporaTestCaseUsingMockAWS):`

Look at `TestDatasetValidator` for an example.

tests/unit/backend/corpora/common/entities/test_asset.py
```diff
@@ -0,0 +1,15 @@
+import unittest
+
+from backend.corpora.common.entities.asset import Asset
+from tests.unit.backend.corpora import CorporaTestCaseUsingMockAWS
+
+
+class TestDataset(unittest.TestCase):
```
maniarathi, line 7:
```suggestion
class TestAsset(unittest.TestCase):
```

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -141,6 +141,17 @@ def test__cascade_delete_dataset__ok(self):
             self.assertRowsDeleted(expected_deleted)
             self.assertRowsExist(expected_exists)
 
+    def test_get_asset__ok(self):
```
maniarathi, line 144:
```suggestion
    def test__get_asset__ok(self):
```

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -141,6 +141,17 @@ def test__cascade_delete_dataset__ok(self):
...
+        dataset = Dataset.get(self.uuid)
+        expected_asset_id = "test_dataset_artifact_id"
+        asset = dataset.get_asset("test_dataset_artifact_id")
+        self.assertEqual(expected_asset_id, asset.id)
+
+    def test_get_asset__not_found(self):
```
maniarathi, line 150:
```suggestion
    def test__get_asset__not_found(self):
```

backend/config/corpora-api.yml
Bento007: yes it shows up as markdown, but i can simplify to that.

backend/corpora/common/entities/asset.py
Bento007: It's just a reminder that this row entry is linked to assets in s3. So if the row is removed, the asset should also be removed.

backend/corpora/common/entities/asset.py
Bento007: parsed out content length

backend/corpora/common/entities/dataset.py
Bento007: removed

backend/corpora/common/entities/dataset.py
Bento007: removed

backend/corpora/common/entities/dataset.py
Bento007: > Though we might want to handle the case that there is more than 1 asset with the same uuid? That should cause an error that we'd want to investigate.

Maintaining unique primary keys is a pretty fundamental database operation. All of the assets are in the same table, therefore I think we can assume there will be no collisions. Do you think this is something we need to test for?

Also I wont be able to write test to verify this error condition, since the DB will prevent me from have duplicate primary keys.

tests/unit/backend/corpora/common/entities/test_asset.py
Bento007: fixed

backend/corpora/common/entities/asset.py
Bento007: I gave it a try and the refactors added more complexity to the tests for dataset_asset class, compared to having them split apart. I chose to favor test simplicity. Mainly I would have to add code to generate assets for testing, which doesn't exist. Adding this additional code wouldn't improve testing or be reused anywhere in the code. So I decided it would be better to keep it as is.

 I also like keeping the error handling for failed s3 request separate from the rest of the server code.

Bento007: Like you said we may use it elsewhere later... maybe. While this isn't the case presently, having it separated makes unit testing easier. So why not reap the benefits of having it split now. It saves us time, code complexity, and it's already written.

Bento007: If you really think this will make things better I can do it, but I don't agree. If it's passing around the s3 client that is the problem, we can add an s3 client to the config and import it from there.

backend/config/corpora-api.yml
maniarathi: Ah good to know! Thanks!

backend/corpora/common/entities/asset.py
maniarathi: Can you tell me more about the complexity of adding code to generate assets for testing? It sounds like something we would actually _want_ to have.

The problem with "refactoring" prematurely is one of obfuscation. The act of generating a presigned URL and getting file size is DatasetAsset specific, not S3 specific. Because of this I state that the functions should belong in DatasetAsset because then the test would specifically test generating a presigned URL for a specific DatasetAsset. That's what we want to test is working.. not whether the boto3 is fine. 

Now on the other hand, if you were telling me that you wanted to write a function that acts upon many different kinds of objects (sanitization functions are a good example) then the functions are static and object-agnostic. This is what warrants a utility function and then DRYs code. 

All of that to say, I'd recommend combining for the following reasons:
1. No other object-agnostic use cases for the functions in the s3_utils except that it specifically acts upon DatasetAsset objects.
2. Adds yet another file to dig into to understand why a presigned URL might not be generating correcting for a DatasetAsset if we need to debug.
3. We actually want to test the functionality of generating a presigned URL for a DatasetAsset object specifically. Not arbitrarily test a presigned url generation because that provides no value.
4. Instantiating the S3 client in the s3_utils is expensive. It almost guarantees roughly a 6 second latency every time a user wants to download a dataset since the frontend call the API 3 times and each API call instantiates it twice. (roughly 1 second instantiation time last time I checked).

backend/corpora/common/utils/s3_utils.py
```diff
@@ -1,28 +1,46 @@
...
+import typing
 
 import boto3
 from botocore.exceptions import ClientError
 
+logger = logging.getLogger(__name__)
 
-def generate_file_url(file_prefix: str, expiration: int = 3600, s3=None) -> str:
+
+def generate_file_url(bucket_name: str, file_prefix: str, expiration: int = 3600) -> typing.Union[str, None]:
```
mckinsel, line 10: Can we make the default expiration 604800? That should the be AWS max.

mckinsel: Guess it would really need to go in dataset_asset.py

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -0,0 +1,71 @@
...
+    DatasetArtifactType,
+    DatasetArtifactFileType,
+)
+from backend.corpora.common.entities.dataset import Dataset
+from backend.corpora.common.entities.dataset_asset import DatasetAsset
+from tests.unit.backend.corpora import CorporaTestCaseUsingMockAWS
+from tests.unit.backend.utils import BogusDatasetParams
+
+
+class TestDatasetAsset(CorporaTestCaseUsingMockAWS):
```
maniarathi, line 11: I think there's a few negative tests missing here for when `generate_presigned_url` or `head_object` fails. You can use `patch` and `side_effect` to force an exception and verify that the DatasetAsset class catches it and returns None.

Reference: https://docs.python.org/3/library/unittest.mock.html

maniarathi (changes requested):
Requesting a few more tests primarily.

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,50 @@
...
+        :param expiration: Presigned URL expiration in seconds
+        :return: Presigned URL to download the requested file
+        """
+        try:
+            response = self.s3.generate_presigned_url(
+                "get_object", Params={"Bucket": self.bucket_name, "Key": self.file_prefix}, ExpiresIn=expiration
+            )
+        except ClientError:
+            logger.exception(f"Failed to generate presigned URL for '{self.file_prefix}'.")
+            return
```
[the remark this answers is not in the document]
Bento007:
```suggestion
            return None
```

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Retrieves the asset content length from the S3 object.
+        :return: The content length in bytes.
+        """
+
+        try:
+            response = self.s3.head_object(Bucket=self.bucket_name, Key=self.file_prefix)
+        except ClientError:
+            logger.exception(f"Failed to retrieve meta data for '{self.file_prefix}'.")
+            return
```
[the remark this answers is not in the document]
Bento007:
```suggestion
            return None
```

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
Bento007: we already test this for get_file_size. I add a mock test for generate_presigned_url

backend/corpora/common/utils/s3_utils.py
Bento007: Done

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -0,0 +1,84 @@
...
+        url = asset.generate_file_url()
+        self.assertIn(file_name, url)
+        self.assertIn("Expires=", url)
+
+    @mock.patch("backend.corpora.common.entities.dataset_asset.DatasetAsset.s3.generate_presigned_url")
+    def test__generate_file_url__ERROR(self, mock_generate_presigned_url):
+        def side_effect(*args, **kwargs):
+            raise ClientError({}, "mock ClientError")
+
+        mock_generate_presigned_url.side_effect = side_effect
```
maniarathi, line 42: I think you can just do: `mock_generate_presigned_url.side_effect = ClientError()`?

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Create the Dataset Asset
```
maniarathi, line 44: I think you want to verify the logs as well because that tell you whether you hit the expected error besides just returning None. So I'd do:

```
asset = self.create_dataset_asset("test__generate_file_url__ERROR.h5ad")

with self.assertLogs(level="ERROR") as logger:
    url = asset.generate_file_url()
self.assertIn("Failed to generate presigned URL", logger.output[0])
self.assertIsNone(url)
```

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -0,0 +1,84 @@
...
+
+        file_size = asset.get_file_size()
+        self.assertEqual(len(content), file_size)
+
+    def test__get_file_size__not_found(self):
+        file_name = "fake_file.txt"
+
+        # Create the Dataset Asset
+        asset = self.create_dataset_asset(file_name)
+        self.assertIs(None, asset.get_file_size())
```
maniarathi, line 68:
```suggestion
        self.assertIsNone(asset.get_file_size())
```

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -0,0 +1,84 @@
...
+        # Create the Dataset Asset
+        asset = self.create_dataset_asset(file_name)
+
+        file_size = asset.get_file_size()
+        self.assertEqual(len(content), file_size)
+
+    def test__get_file_size__not_found(self):
+        file_name = "fake_file.txt"
+
+        # Create the Dataset Asset
```
maniarathi, line 66: Would verify logs here as well:
```
asset = self.create_dataset_asset(file_name)

with self.assertLogs(level="ERROR") as logger:
    file_size = asset.get_file_size()
self.assertIn("Failed to retrieve meta data", logger.output[0])
self.assertIsNone(file_size)

File: tests/unit/backend/corpora/common/entities/test_dataset.py [hunk 27]
```diff
@@ -141,6 +141,17 @@ def test__cascade_delete_dataset__ok(self):
...
+    def test__get_asset__ok(self):
+        dataset = Dataset.get(self.uuid)
+        expected_asset_id = "test_dataset_artifact_id"
+        asset = dataset.get_asset("test_dataset_artifact_id")
+        self.assertEqual(expected_asset_id, asset.id)
+
+    def test__get_asset__not_found(self):
+        dataset = Dataset.get(self.uuid)
+        asset = dataset.get_asset("fake_asset")
+        self.assertEqual(None, asset)
```
maniarathi, line 153:
```suggestion
        self.assertIsNone(asset)
```

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,50 @@
...
+        :return: The content length in bytes.
+        """
+
+        try:
+            response = self.s3.head_object(Bucket=self.bucket_name, Key=self.file_prefix)
+        except ClientError:
+            logger.exception(f"Failed to retrieve meta data for '{self.file_prefix}'.")
+            return None
+        else:
+            return response["ContentLength"]
```
maniarathi, line 50: I think you can delete this and just add `response["ContentLength"] right in the `try` block underneath where you fetch the response.

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # retrieve the artifact
+    asset = dataset.get_asset(asset_uuid)
+    if not asset:
+        raise NotFoundHTTPException(f"'dataset/{dataset_uuid}/asset/{asset_uuid}' not found.")
+
+    # Retrieve S3 metadata
+    file_size = asset.get_file_size()
+    if not file_size:
+        raise NotFoundHTTPException(f"'dataset/{dataset_uuid}/asset/{asset_uuid}' not found.")
```
maniarathi, line 24: We want a different error message here right? Something about fetching the file metadata not working?

maniarathi (changes requested):
Mostly small things this time around :)

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class DatasetAsset(Entity):
+    table = DbDatasetArtifact
+    s3 = boto3.client("s3")
+
+    def __init__(self, db_object: DbDatasetArtifact):
+        super().__init__(db_object)
+
+        self.bucket_name, self.file_prefix = self.s3_uri[5:].split("/", 1)
```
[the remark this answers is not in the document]
mckinsel: Also this it a bit of a nit, but this talks about `file_prefix` a lot, but I think it's really a `key name` or `object key`. The distinction is kind of important because we have things like `s3://corpora/project_id/dataset_id/dataset.loom` which will work with these methods because it is a key. But we also have `s3://corpora/project_id/dataset_id/dataset.cxg/` which will not work because it is just a prefix.

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def generate_file_url(self, expiration: int = 604800) -> typing.Union[str, None]:
+        """
+        Generate a presigned URL for a file for user download.
+        :param expiration: Presigned URL expiration in seconds
```
mckinsel, line 25: Maybe note that it defaults to one week.

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,50 @@
...
+        :return: Presigned URL to download the requested file
+        """
+        try:
+            response = self.s3.generate_presigned_url(
+                "get_object", Params={"Bucket": self.bucket_name, "Key": self.file_prefix}, ExpiresIn=expiration
+            )
+        except ClientError:
+            logger.exception(f"Failed to generate presigned URL for '{self.file_prefix}'.")
```
mckinsel, line 33: Do we want to log the exception's message as well? This kind of swallows all the information we could use to debug.

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -0,0 +1,39 @@
...
+from ....common.utils.exceptions import NotFoundHTTPException, ServerErrorHTTPException
+
+
+@db_session
+def post_dataset_asset(dataset_uuid: str, asset_uuid: str):
+
+    # retrieve the dataset
+    dataset = Dataset.get(dataset_uuid)
+    if not dataset:
+        raise NotFoundHTTPException(f"'dataset/{dataset_uuid}' not found.")
```
mckinsel, line 14: Does these exceptions work with the `problem` schema in the API?

backend/config/corpora-api.yml
```diff
@@ -468,16 +468,21 @@ paths:
         "404":
           $ref: "#/components/responses/404"
 
-  /v1/dataset/{dataset_uuid}:
+  /v1/dataset/{dataset_uuid}/asset/{asset_uuid}:
```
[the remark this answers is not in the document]
mckinsel: Oh nvm, I see it in the lambda. But then shouldn't it be a required path parameter?

backend/corpora/lambdas/api/v1/dataset.py
mckinsel, line 24: Is this really a 404? It seems like more of an internal error if we have the asset in the db but we can't HEAD it.

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -0,0 +1,39 @@
...
+    if not file_size:
+        raise NotFoundHTTPException(f"'dataset/{dataset_uuid}/asset/{asset_uuid}' not found.")
+
+    # Generate presigned URL
+    presigned_url = asset.generate_file_url()
+    if not presigned_url:
+        raise ServerErrorHTTPException()
+
+    return make_response(
+        jsonify(
```
mckinsel, line 32: presigned urls are full of weird characters that need careful escaping. I think it should all work out, but can you add test with a bunch of `% / ' "` etc?

backend/corpora/common/entities/dataset.py
```diff
@@ -1,5 +1,7 @@
+import typing
 import uuid
 
+from backend.corpora.common.entities.dataset_asset import DatasetAsset
```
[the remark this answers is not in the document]
Bento007: it would be a circular dependency.

backend/corpora/common/entities/dataset_asset.py
Bento007: `logging.exception` includes the exception information and stack trace. `logging.error` does not. See https://docs.python.org/3/library/logging.html#logging.Logger.exception

backend/corpora/lambdas/api/v1/dataset.py
Bento007: Yes, connexion defines a problemException that follows that format, and all of the HTTP error inherit it.

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
mckinsel: Ah gotcha

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
Bento007: that works.

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
Bento007: Done

tests/unit/backend/corpora/common/entities/test_dataset.py
Bento007: Done

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
Bento007: done

backend/corpora/common/entities/dataset_asset.py
Bento007: I prefer this way because it keeps the return statement out of the try block, which if it does fail would be a different error.

backend/config/corpora-api.yml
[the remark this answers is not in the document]
Bento007: If it's in the path, isn't it assumed to be required? I can make it explicit.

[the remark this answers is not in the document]
Bento007: if you follow the ref for `dataset_uuid` you'll see that it is required.

backend/corpora/common/entities/dataset_asset.py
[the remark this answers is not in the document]
Bento007: I like it, I'll make the change.

backend/config/corpora-api.yml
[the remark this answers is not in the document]
mckinsel: Oh sorry, totally missed the $ref

backend/corpora/common/entities/dataset_asset.py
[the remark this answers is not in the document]
Bento007: > You could maybe use urlparse so you can assert that the scheme is s3

Shouldn't this be caught during ingestion? If not what kind of error would we return, server error?

backend/corpora/lambdas/api/v1/dataset.py
Bento007: So if this happens, it likely means the the dataset_asset.s3_uri does not reference a valid s3 location in our bucket. This could happen if the underlying object was deleted after the asset was created, or the s3_uri is wrong during ingestion. We should add a check to make sure the resource exists before ingesting into the database which mean the only other possible reason is the object was deleted somehow.

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,53 @@
...
+class DatasetAsset(Entity):
+    table = DbDatasetArtifact
+    s3 = boto3.client("s3")
+
+    def __init__(self, db_object: DbDatasetArtifact):
+        super().__init__(db_object)
+
+        url = urlparse(self.s3_uri)
+        self.bucket_name = url.netloc
+        self.key_name = url.path[1:]
```
mckinsel, line 23: Consider `url.path.lstrip("/")` just in case it's empty.

mckinsel: Eh probably never mind, if it's empty it's just going to break immediately elsewhere.

mckinsel (approved):
LGTM!

tests/unit/backend/corpora/common/utils/test_json.py
```diff
@@ -58,6 +58,40 @@ class Unsupported:
...
             json.dumps(test_unsupported_type, cls=CustomJSONEncoder)
 
+    def test_illegal_characters(self):
+        tests = [
+            ("%", '"%"'),
+            ("/", '"/"'),
+            ('"', '"\\""'),
+            ("'", '"\'"'),
+            (
+                "https://s3.us-west-2.amazonaws.com/bogus_bucket/test_generate_file_url.h5ad?AWSAccessKeyId"
```
maniarathi, line 68: Nit: double check line breaks here.. looks a bit wonky to me.

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -0,0 +1,53 @@
...
+    def get_file_size(self) -> typing.Union[int, None]:
+        """
+        Retrieves the asset content length from the S3 object.
+        :return: The content length in bytes.
+        """
+
+        try:
+            response = self.s3.head_object(Bucket=self.bucket_name, Key=self.key_name)
+        except ClientError:
+            logger.exception(f"Failed to retrieve meta data for '{self.key_name}'.")
```
maniarathi, line 50: Nit: I'd maybe consider logging a more helpful error message since it'll be read by us. Something like `f"Could not locate S3 object s3://{self.bucket_name}/{self.key_name} to fetch metadata."`. Or something like that.

maniarathi (approved):
Couple nits. Nice work!

backend/corpora/common/entities/dataset_asset.py
Bento007: I add the original URL to the response.

PR #569, opened 2020-08-31
thuang-499-smoke-test-part-1

tihuan:
This PR moves away from using `.env.[ENV]` to define non-sensitive app configurations to just copy the variables from `src/configs/[ENV].js` to `src/configs/configs.js`.

This way we decouple build env with config env, in order to allow Gatsby prod build to still hit local API endpoint (will be part 2 smoke test PR)

Changes:
1. Update README to reflect the new setup
2. Update deployment yml accordingly
3. Remove `.env.development`
4. Update `.gitignore`
5. Update the app to use the new config file to get `API_URL`

QA steps:
1. Copy `src/configs/dev.js` to `src/configs/configs.js` (please create the file if not available)
2. Run `npm run develop`
3. Browse `http://localhost:8000/` and you should see a list of datasets that match the `dev` env DB

### DIFF
--- a/frontend/src/configs/dev.js
+++ b/frontend/src/configs/dev.js
@@ -1,0 +1,5 @@
+const configs = {
+  API_URL: "https://api.dev.corpora.cziscience.com",
+};
+
+if (typeof module !== "undefined") module.exports = configs;
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -31,6 +31,7 @@ jobs:
         run: |
           pip install flake8-black
           cd frontend
+          cp src/configs/local.js src/configs/configs.js
           npm install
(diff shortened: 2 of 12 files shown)


frontend/src/configs/dev.js
```diff
@@ -0,0 +1,5 @@
+const configs = {
```
Bento007, line 1: Do we still need API_URL defined in the env.[ENV] file as well?

tihuan: Ah sorry just saw this comment! No, I think configs.js replaces .eng.[ENV] files, so I deleted .env.development file in this PR 😄

Thank you!!

tihuan:
Thanks for the thorough testing, @seve ! I'll fix the linter issue then merge it!

PR #591, opened 2020-09-08
Fix local library imports to be relative

Bento007:
Importing the CorpaAuthConfig used by the chalice app to ensure the same class is used in the test. Otherwise there are two CorpaAuthConfig class. One is defined in backend/chalice/api_server/chalicelib/corpora, the other is defined in backend/corpora. We are interested in testing the former.

### DIFF
--- a/backend/corpora/lambdas/api/v1/authentication.py
+++ b/backend/corpora/lambdas/api/v1/authentication.py
@@ -1,17 +1,18 @@
-from flask import make_response, jsonify, current_app, request, redirect, after_this_request, g, Response
+import base64
 import json
 import os
+from typing import Optional
+from urllib.parse import urlencode
+
 import requests
 from authlib.integrations.flask_client import OAuth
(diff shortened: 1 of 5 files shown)


bmccandless (dismissed):
LGTM

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -25,7 +26,7 @@ def get_oauth_client(config: CorporaAuthConfig) -> FlaskRemoteApp:
     """
 
     global oauth_client
-    if oauth_client and os.environ["DEPLOYMENT_STAGE"] != "test":
+    if oauth_client and os.environ["DEPLOYMENT_STAGE"] == "test":
```
Bento007, line 29:
```suggestion
    if oauth_client and os.environ["DEPLOYMENT_STAGE"] != "test":
```

bmccandless (approved):
LGTM

PR #597, opened 2020-09-09
Add additional footer information and fix box shadow and remove last line from dataset list

maniarathi:
Screen Shot 2020-09-09 at 12 21 42 PM

### DIFF
--- a/frontend/src/components/Footer/index.tsx
+++ b/frontend/src/components/Footer/index.tsx
@@ -15,16 +15,24 @@ const Footer: FC = () => {
           Contact
         </StyledAnchor>
         <StyledAnchor
-          key={"https://cellxgene.cziscience.com/static/cellxgene/deploy/tos.html"}
-          href={"https://cellxgene.cziscience.com/static/cellxgene/deploy/tos.html"}
+          key={
+            "https://cellxgene.cziscience.com/static/cellxgene/deploy/tos.html"
+          }
+          href={
+            "https://cellxgene.cziscience.com/static/cellxgene/deploy/tos.html"
+          }
           target="_blank"
           rel="noopener"
         >
           Terms of Service
         </StyledAnchor>
         <StyledAnchor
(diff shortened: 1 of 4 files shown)


frontend/src/components/Footer/index.tsx
```diff
@@ -1,10 +1,44 @@
...
+import { MainWrapper, StyledAnchor, Wrapper } from "./style";
 
 const Footer: FC = () => {
   return (
     <Wrapper>
-      <MainWrapper>© 2020 cellxgene</MainWrapper>
+      <MainWrapper>
+        © 2020 cellxgene
+        <StyledAnchor
+          key={"mailto:cellxgene@chanzuckerberg.com"}
```
tihuan, line 10: nit: oh we only need `key` when we're generating components in a `map()` call 😄

frontend/src/components/Footer/index.tsx
```diff
@@ -1,10 +1,44 @@
 import React, { FC } from "react";
-import { MainWrapper, Wrapper } from "./style";
+import { MainWrapper, StyledAnchor, Wrapper } from "./style";
 
 const Footer: FC = () => {
   return (
     <Wrapper>
-      <MainWrapper>© 2020 cellxgene</MainWrapper>
+      <MainWrapper>
+        © 2020 cellxgene
```
tihuan, line 8: nit: For future proof, we can do: 

```ts
<MainWrapper>© {new Date().getFullYear()} cellxgene</MainWrapper>
```

tihuan (approved):
Looks super good! Thanks for updating this quickly, @maniarathi 🙏 ✨ !! Small nits otherwise LGTM!

frontend/src/components/Footer/index.tsx
maniarathi: Ahh oops! Deleted!

frontend/src/components/Footer/index.tsx
maniarathi: Ooh fancy! Thanks!

PR #599, opened 2020-09-10
fix header logo anchor

tihuan:
Adding `display: inline-block` changes the browser's default display type from `inline`, which takes up the whole space to `inline-block`, which only takes up the space of its children!

gif

PTAL! Thank you 😄 🙏 !


tihuan:
This was fixed in Pat's [PR](https://github.com/chanzuckerberg/corpora-data-portal/pull/601)!

PR #608, opened 2020-09-14
feat(download): Add download column to data portal

Bento007:
- Add download header for dataset list
- Add download button styling
- Add tests for UI elements
- Add Dataset Title with page description.

Screen Shot 2020-09-14 at 1 33 44 PM


frontend/src/components/ProjectList/index.tsx
```diff
@@ -12,6 +13,10 @@ const ProjectsList: FC<Props> = ({ projects }) => {
 
   return (
     <>
+      <h1>Datasets</h1>
+      <p>
+        Explore public dataset or add your own. <Link to="">Learn more</Link>
```
tihuan, line 18: Ah sorry Trent, can we use the following instead please 😄 🙏 ?

>The cellxgene data portal is a repository of public, explorable single-cell datasets. If you have a public dataset which you would like hosted for visualization on this site, please let us know at cellxgene@chanzuckerberg.com.

I think that's what Ambrose wants in the ticket. Thank you!

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/607

Bento007: Fixed Screen Shot 2020-09-16 at 10 10 02 AM

tihuan (approved):
Thanks again for the fantastic work, @Bento007 !! I added auto merge 👍 🚀 Thank you!

PR #610, opened 2020-09-16
thuang-fix-FE-makefile

tihuan:
https://stackoverflow.com/questions/20389064/using-gnu-make-functions-to-check-if-variables-are-defined explains why `define ifndef` doesn't work with `$(call)`, and it's easier to just do `check-var-defined = $(if $(1),,$(error $(1) is not defined))`

PTAL thank you!


Bento007 (approved):
ooo shell magic

PR #613, opened 2020-09-16
thuang-download-feedback

tihuan:
Tiny updates:

1. Add a space between "please let us know at" and "cellxgene@chanzuckerberg.com"
2. Add max-width to name column, otherwise it could take up more space than intended (I missed that 😆 )
3. Change "More Information" column to stack links vertically 

For #3, I don't know if that's the best solution, so would love input from @pdugan20 🙏  One thing I was debating with myself is whether we could sort the link by text length, so at least shorter ones are on top and can take a tiny pyramid 🔺 But also unsure if the info links are already sorted by importance? CC: @maniarathi @mckinsel 

Thanks all!

BEFORE:
Screen Shot 2020-09-16 at 10 38 02 AM

AFTER:
Screen Shot 2020-09-16 at 10 34 12 AM

### DIFF
--- a/frontend/src/components/ProjectList/components/Dataset/components/MoreInformation/index.tsx
+++ b/frontend/src/components/ProjectList/components/Dataset/components/MoreInformation/index.tsx
@@ -1,7 +1,7 @@
 import React, { FC } from "react";
 import { Project } from "src/common/entities";
 import { SmallColumn } from "../../common/style";
-import { StyledAnchor } from "./style";
+import { StyledAnchor, Wrapper } from "./style";
 interface AnchorProps {
   url: string;
 }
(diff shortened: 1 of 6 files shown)


maniarathi:
I don't think the links are sorted in any particular order nor do we have that sorting in our database, so doesn't matter much there. At some point, we could order by type but I don't think it matters much (imo).

pdugan20 (approved):
LGTM! I think the ideal scenario for the list of links would be to all have them left justified and starting from the same position, and to comma separate them. But honestly the visuals in the attached screenshot look good to me so not worth going back to make changes!

tihuan:
Thanks @pdugan20 ! Yeah I was playing with that idea a little, but wasn't super successful lol But at least I added the comma back in the meantime!

Screen Shot 2020-09-16 at 12 40 05 PM


Also currently when resizing the screen, it doesn't look as good, because we haven't defined responsive design for different screen sizes. Do you see a need to prioritize that?

Thank you!

gif

tihuan:
Sync'd with Pat offline! Since this is a temp page, the work above is good enough and we'll have more robust handling for the permanent pages! Mergingg

PR #614, opened 2020-09-16
Dunitz/565 auto scale test

MDunitz:
## Need
- automated scale test to ensure latency remains low for dp api

## Approach
- create scheduled (weekly) github action that runs a 5 minute scale test and sends an aggregation of the results to slack

## Todo
- [ ] change SLACK_WEBHOOK in github secrets once single-cell-ops webhook is approved (currently sent to a channel I use for testing slack webhooks)
- [ ] add test for staging/prod once the cloudfront code is promoted 

Screen Shot 2020-09-16 at 16 27 14

### DIFF
--- a/scripts/scale_test.py
+++ b/scripts/scale_test.py
@@ -1,6 +1,6 @@
 import random
 import time
-from locust import HttpUser, between, task, events
+from locust import HttpUser, between, task
 
 random.seed(time.time())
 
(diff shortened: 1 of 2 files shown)


scripts/scale_test.py
```diff
@@ -0,0 +1,29 @@
+import random
+import time
+from locust import HttpUser, between, task
+
+random.seed(time.time())
+
+
+class WebsiteUser(HttpUser):
+    wait_time = between(1, 2)
```
Bento007, line 9: If we want the test to run faster we modify or remove this value.

Bento007 (approved):
LGTM

scripts/scale_test.py
MDunitz: I'm just running the test for 5 minutes

PR #620, opened 2020-09-24
Point to cloudfront distribution of apis

MDunitz:
## Need
- as a user I want data portal to load in under 1 second

## Approach
- we set up a cloudfront distribution to cache the api content this pr will point the data portal frontend at the cached version of the api

## Testing
- deployed to dev and saw a significant speed up in load time


Bento007 (approved):
LGTM

PR #628, opened 2020-10-01
thuang-621-dataset-column-fix

tihuan:
Looks like we just needed to add `min-width` and `max-width` as lower and upper bounds for `flex` CSS property!

The reason we can't just use `width: 620px;` is because `width` is ignored when `flex-basis` is set, in this case, `flex: 2` will automatically sets `flex-basis` to `auto`.

Screen Shot 2020-09-30 at 5 29 07 PM

Resources:
1. https://mastery.games/post/the-difference-between-width-and-flex-basis/
2. https://css-tricks.com/snippets/css/a-guide-to-flexbox/

AFTER in Chrome:
Screen Shot 2020-09-30 at 5 19 25 PM

BEFORE and AFTER in Safari:
gif

### DIFF
--- a/frontend/src/components/ProjectList/components/Dataset/style.ts
+++ b/frontend/src/components/ProjectList/components/Dataset/style.ts
@@ -20,6 +20,8 @@ export const Name = styled.div`
   align-items: center;
   line-height: 1.3;
   padding: 12px 0 12px 10px;
+  /* (thuang): width is ignored when flex is used */
+  min-width: 620px;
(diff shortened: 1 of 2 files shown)


pdugan20 (approved):
Sweet!

tihuan:
Thanks for the quick review, @pdugan20 ! I'm seeing BE unit test is failing, which is not relevant to the PR. But I'll just run it again to see if it's just flaky lol

PR #673, opened 2020-10-29
First batch of changes to support local dev

jgadling:
This PR makes a few changes to the corpora codebase to support local development environments:

## OAuth path customization
The local dev environment uses different authorization and token paths than Auth0, so we need a way to customize this. This PR adds support for setting default values for SecretsManager secrets if a key isn't present in the secret's JSON document. This makes the change safe to deploy without having to migrate existing secrets, and allows the local dev environment to set those keys and override the defaults.

## BOTO_ENDPOINT_URL
The dev environment runs a local version of the AWS API's. This URL needs to be passed into AWS clients to direct them to use the local version of AWS for development instead of the actual AWS cloud. If this environment variable isn't set, the application should use default AWS endpoints.

## run_local_server.py
This is an update to `run_local_server.py` that supports automatically reloading the app when its python source files change.


tests/unit/backend/corpora/common/utils/test_json.py
```diff
@@ -65,7 +65,7 @@ def test_illegal_characters(self):
             ('"', '"\\""'),
             ("'", '"\'"'),
             (
-                "https://s3.us-west-2.amazonaws.com/bogus_bucket/test_generate_file_url.h5ad?AWSAccessKeyId"
+                "https://s3.us-west-2.amazonaws.com/bogus-bucket/test_generate_file_url.h5ad?AWSAccessKeyId"
```
jgadling, line 68: underscores are illegal in s3 bucket names

tests/unit/backend/corpora/__init__.py
```diff
@@ -7,25 +8,29 @@
 
 
 class CorporaTestCaseUsingMockAWS(unittest.TestCase):
-    CORPORA_TEST_CONFIG = {"bucket_name": "bogus_bucket"}
+    CORPORA_TEST_CONFIG = {"bucket_name": "bogus-bucket"}
```
jgadling, line 11: underscores are illegal in s3 bucket names

backend/chalice/api_server/run_dev_server.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
```
Bento007, line 1: how is this different from https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/chalice/api_server/run_local_server.py If this is a replacement then please remove or update the previous.

backend/chalice/api_server/.chalice/config.json.dev
```diff
@@ -0,0 +1,25 @@
+{
```
Bento007, line 1: isn't this file generated using https://github.com/chanzuckerberg/corpora-data-portal/blob/7e9afe90d9589b850d0dabcf225e239af30f92e7/backend/chalice/api_server/Makefile#L49. How is this different?

backend/corpora/common/utils/aws_secret.py
```diff
@@ -23,7 +24,7 @@ class AwsSecret:
     def __init__(self, name):
         self._debug("AwsSecret.__init__({})".format(name))
         self.name = name
-        self.secrets_mgr = boto3.client(service_name="secretsmanager")
+        self.secrets_mgr = boto3.client(service_name="secretsmanager", endpoint_url=os.getenv("BOTO_ENDPOINT_URL"))
```
Bento007, line 27: what does the endpoint_url do? I have an idea, but please indulge me.

backend/corpora/common/utils/secret_config.py
```diff
@@ -47,13 +48,24 @@ def config(self):
 
     def __getattr__(self, name):
         if self.config_is_loaded():
-            return self.value_from_config(name) or self.value_from_env(name) or self.raise_error(name)
+            return (
+                self.value_from_config(name)
+                or self.value_from_env(name)
+                or self.value_from_defaults(name)
```
Bento007, line 54: I feel like this could have some unintended consequences if we don't set the vars correctly for a deployment. Like setting a dummy auth and allowing unrestricted access. Are these concerns valid or do we have additional fail-safes or documentation to prevent this?

backend/corpora/common/utils/secret_config.py
```diff
@@ -47,13 +48,24 @@ def config(self):
...
+                or self.raise_error(name)
+            )
         else:
-            return self.value_from_env(name) or (self.load() and self.value_from_config(name)) or self.raise_error(name)
+            return (
+                self.value_from_env(name)
+                or (self.load() and self.value_from_config(name))
+                or self.value_from_defaults(name)
```
Bento007, line 61: See last comment

Bento007 (commented):
I have questions

backend/corpora/common/utils/aws_secret.py
jgadling: This is explained in the PR description -- basically it allows us to override default AWS service url's with a local "fake-aws" service

backend/corpora/common/utils/secret_config.py
jgadling: This just adds an additional call to self.value_from_defaults(), which is a no-op for most subclasses of SecretConfig, but CorporaAuthConfig has support to [create oauth default URL's](https://github.com/chanzuckerberg/corpora-data-portal/blob/c351a695ed17c289a3bc54ef131a3a721a32a8b6/backend/corpora/common/corpora_config.py#L41) that follow the previous formatting *if api_authorize_url and api_token_url aren't present in the secret json document* and I've updated the [backend/config/auth0-secret-template.json](https://github.com/chanzuckerberg/corpora-data-portal/pull/673/files#diff-2ae991b600dc1ef0749ef18449e03bc8f9f6f2edcb525d92daca245427584698R3-R5) docs to explain how the defaults are set.

backend/chalice/api_server/.chalice/config.json.dev
jgadling: The build-chalice-config target has a dependency on being able to talk to AWS:
https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/chalice/api_server/Makefile#L13

The dockerized local dev environment is designed to have no (or as few as possible) external dependencies, so we can't depend on making AWS calls to populate it. This template just gives us a simple way to sidestep that dependency, until we get more comfortable with the docker dev env.

backend/chalice/api_server/run_dev_server.py
jgadling: fair enough. Updated!

backend/corpora/common/utils/secret_config.py
Bento007: i'm sorry (>_<), yeah, I should have seen that.

backend/corpora/common/utils/aws_secret.py
Bento007: I'm sorry (>_<), yeah, I should have seen that.

Bento007 (approved):
LGTM

PR #676, opened 2020-11-02
Allow obfuscated_uuid to be none

Bento007:


backend/corpora/common/entities/project.py
```diff
@@ -133,6 +133,7 @@ def reshape_for_api(self) -> dict:
         # Reshape the data to match.
         result.pop("user", None)
         result.pop("owner", None)
+        result.pop("obfuscated_uuid", None)
```
mckinsel, line 136: Can you make this `result["obfuscated_id"] = result.get("obfuscated_id") or ""`? I don't think we necessarily want to drop it from the response, just not return a `None`.

mckinsel (approved):
LGTM

PR #677, opened 2020-11-02
fix FE collection runtime error

tihuan:
Doing a better response check to avoid runtime error! 

I forgot BE would return error response like the following, so I need to update how I check for empty datasets:

```ts
{
  detail: "None is not of type 'string'↵↵Failed validating 'type' in schema['properties']['obfuscated_uuid']:↵    {'type': 'string'}↵↵On instance['obfuscated_uuid']:↵    None",
  status: 500,
  title: "Response body does not conform to specification",
  type: "about:blank"
}
```

Thank you!

Screen Shot 2020-11-02 at 10 23 43 AM


tihuan:
Just merged this one since it's a low risk fix! Please still help review in case I missed something lol Thank you!

Issue #644, opened 2020-10-19
Refactor `project` APIs to read `collections`


maniarathi:
## Need
We are moving away from the notion of "projects" and instead using the term "collections" to represents groups of datasets. In order to reduce future confusion, we'll refactor current use of `project` in the API to instead read `collections`.

## Definition of Done
- [x] Rename all current `/project/` APIs to instead be `/collections/`.
- [x] Update the frontend calls to make use of the renamed APIs.
- [x] Update the High level architecture RFC to reflect the latest API definitions.

MDunitz:
Do we need/want to update the API version when we do this?

maniarathi:
@MDunitz Good question.. I'm inclined to say keep it at the same version since we really don't have any external users.

bmccandless:
@maniarathi Does this refactor only refer to the API?
The Database will still have Projects?

bmccandless:
@maniarathi The images in the high level architecture are in png format.   I won't be able to change those, but I can change the text.

PR #682, opened 2020-11-03
Revert recent  changes to the database from "collection" back to "pro…

bmccandless:
…ject".

 #644


backend/corpora/common/corpora_orm.py
```diff
@@ -73,7 +73,7 @@ def session(self, **kwargs):
         return self.session_maker(**kwargs)
 
 
-class CollectionVisibility(enum.Enum):
+class ProjectVisibility(enum.Enum):
```
Bento007, line 76: I made this change in the database so this change is not needed.

Bento007 (changes requested):
One change to undo

Bento007 (approved):
LGTM

PR #683, opened 2020-11-03
allow empty fields

Bento007:
due to migration, there are empty fields that are returned. Changing the openapi spec to accommodate these empty fields.


MDunitz (approved):
lgtm - do you want to add a unit test for this?

Bento007:
yes we should https://github.com/chanzuckerberg/corpora-data-portal/issues/684

PR #691, opened 2020-11-05
Thuang 651 create collection component

tihuan:
1. Add `gatsby-plugin-sass` to allow Blueprint.js scss overwrite
1. Add `@blueprintjs/icons`
1. Add `lodash-es` for ES6 module version of lodash, which is tree shaking friendly and, in theory, reduces js bundle size
1. Introduce @hthomas-czi 's design system by adding standardized colors in `frontend/src/components/common/theme.ts` and Blueprint style overwrite in `frontend/src/global.scss`
1. Add `CreateCollectionModal`, which is hidden behind URL param feature flag: `?cc=true` 
1. Build common `Form` components in `frontend/src/components/common/Form`, which allows us to do sync validation and show error when requirement is not met. This includes `Input` and `TextArea` components
1. Add policy attestation where we hard code policy version in the FE
1. Add `.snyk` file to ignore non-fixable warnings

QA steps:
1. Please pull branch `thuang-651-part2`
2. Run `npm i && npm start`
3. Visit: `http://localhost:8000/?cc=true`
4. You should see "Create Collection" button on the upper right corner
5. Play with the form and see if anything is weird for you
6. Currently we hard code POST response to redirect to an existing collection, since BE endpoint is WIP

Form:
Screen Shot 2020-11-04 at 2 42 23 PM

Error state:
Screen Shot 2020-11-04 at 2 41 31 PM

Demo:
gif

Thank you!!


tihuan:
Previous discussion here:

https://github.com/chanzuckerberg/corpora-data-portal/pull/690

Thank you!

frontend/src/common/queries/collections.ts
```diff
@@ -39,3 +39,28 @@ async function fetchCollection(_: unknown, id: string): Promise<Collection> {
...
+    setTimeout(() => {
+      resolve();
+    }, 2 * 1000);
+  });
+
+  return "673637cf-dcb7-45e1-bb88-72a27c50c8ca";
+}
+
+export const formDataToObject = function (formData: FormData) {
+  const payload: { [key: string]: unknown } = {};
```
seve, line 57: Is the main difference between the `unknown` and `any` the heightened restrictions on unknown?

tihuan:
@hthomas-czi I transferred your comment from https://github.com/chanzuckerberg/corpora-data-portal/pull/690 here! Thank you!

> Does it show the error state if the form field receives focus and then loses focuses while empty? When does the logic first run?

It currently only runs after you type something in and make it empty again! But I can add the behavior you describe if needed! Should be a quick change 🙆‍♂️

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  index: number;
+};
+
+export enum TYPES {
+  DOI = "DOI",
+  RAW_DATA = "RAW_DATA",
+  PROTOCOL = "PROTOCOL",
+  LAB_WEBSITE = "LAB_WEBSITE",
+  OTHER = "OTHER",
+}
```
seve, line 22: This enum is a duplicate of the one on entities

hthomas-czi:
@tihuan Gotcha. I'm just concerned about users not knowing which fields are required (all of them) and getting stuck without any guidance. Could we make the primary button enabled and show the error states if users click it without completing the required fields?

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+export const OPTIONS = {
+  [TYPES.DOI]: { text: "DOI", value: TYPES.DOI },
+  [TYPES.RAW_DATA]: { text: "Raw Data", value: TYPES.RAW_DATA },
+  [TYPES.PROTOCOL]: { text: "Protocol", value: TYPES.PROTOCOL },
+  [TYPES.LAB_WEBSITE]: { text: "Lab Website", value: TYPES.LAB_WEBSITE },
```
seve, line 28: Is this assigning the value of the enum keys to the keys of `OPTIONS`? Wild!

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    isValid: boolean;
+    value: string;
+  }) {
+    handleChange({ id, index, isValid: isValidFromInput, value });
+  }
+};
+
+function isDOILink(type: TYPES): (value: string) => true | string {
+  return (value: string) => {
+    if (type !== TYPES.DOI) return true;
```
seve, line 88: We're returning true on non-doi links because we're checking validity not presence right? Maybe the function name can be a bit clearer

frontend/src/components/CreateCollectionModal/components/Content/components/Policy/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const Policy: FC<Props> = ({ handleChange }) => {
+  const [isChecked, setIsChecked] = useState(false);
+  const [isOpen, setIsOpen] = useState(false);
+
+  return (
+    <Wrapper>
+      <Checkbox checked={isChecked} onChange={handleChange_}>
+        I agree to cellxgene&#39;s data submission policies.{" "}
```
seve, line 36: Is the trailing space on purpose?

frontend/src/components/common/Form/TextArea/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isValid?: boolean;
+  syncValidation?: Array<(value: string) => boolean>;
+  handleChange: ({ isValid, value, name }: Value) => void;
+}
+
+const TextArea: FC<Props> = (props) => {
+  const [isValid, setIsValid] = useState(true);
+  const inputRef = useRef<HTMLTextAreaElement>(null);
+
+  const { handleChange = noop, ...restProps } = props;
```
seve, line 26: This noop is temporary, correct? Should we mark it as so?

frontend/src/components/common/Form/Input/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  handleChange: ({ isValid, value, name }: Value) => void;
+  syncValidation?: Array<(value: string) => true | string>;
+  noNameAttr?: boolean;
+  className?: string;
+  placeholder?: string;
+}
+
+const DangerIcon = () => {
+  return <StyledIcon icon={IconNames.ISSUE} intent={Intent.DANGER} />;
+};
```
seve, line 23: should this be a directory level constant?

seve (changes requested):
Just a few questions

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
tihuan: Yes! It's fixed in this [PR](https://github.com/chanzuckerberg/corpora-data-portal/pull/689) :D Thanks for the callout!

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
tihuan: Yup! Yeah pretty convenient lol

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
tihuan: Ooh great question! So Checking for `isValidHttpUrl()` is a separate sync validation function that we pass to the `<Input />` component below, so I was thinking `if (type !== TYPES.DOI) return true;` means skipping validation if the type is not `DOI`. Maybe I should add a comment there? E.g.,

```ts
// Skip validation if type is not DOI
if (type !== TYPES.DOI) return true;
```

```tsx
<Input
        name={value}
        text={text}
        handleChange={debounce(handleChange_, DEBOUNCE_TIME_MS)}
        syncValidation={[isValidHttpUrl, isDOILink(value)]} <-- HERE
        noNameAttr
        placeholder={value === TYPES.DOI ? DOI_PLACEHOLDER : LINK_PLACEHOLDER}
      />
```

seve: I think the comment would work 👍

frontend/src/common/queries/collections.ts
tihuan: Great question! So `any` is looser than `unknown` in the sense that when an `any` value is assigned to another typed value like `string`, it'll pass type checking, but if it was an `unknown`, then passing that value to a `string` type will not pass type checking!

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
tihuan: Somehow Github is not reflecting the fix from your PR, but I just verified that the TYPE is now consolidated!

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
tihuan: Awesome thank you! Done!

frontend/src/components/CreateCollectionModal/components/Content/components/Policy/index.tsx
tihuan: Thank you for the eagle eye!! It's not needed 😆 Fixed!

frontend/src/components/common/Form/Input/index.tsx
tihuan: Ah I see that it's already defined at the top level of this file, so we're not creating a new function per re-render. What do you mean by directory level constant 😄 ? Thank you!

frontend/src/components/common/Form/TextArea/index.tsx
tihuan: Ah I forgot to make `handleChange` prop optional!

So since `props.handleChange` is optional, I want the default callback function to be `noop`, so I don't have to check if `props.handleChange` is defined before calling it in the block:

```ts
handleChange({ isValid: result, name, value });
```

tihuan:
@seve I think I've replied to your comments! PTAL again 🙏 Thank you!

I slacked Harley to discuss the form behavior 👌

frontend/src/components/common/Form/Input/index.tsx
seve: I saw the same snippet in a few places,might be worth pulling out into its own file

hthomas-czi (approved):

frontend/src/components/common/Form/Input/index.tsx
tihuan: Oh I see! Yeah there are two places that use this, but their `StyledIcon` actually has different styles!

seve (approved):

tihuan:
Thanks all! Merging now 🤩🎉🙌

PR #695, opened 2020-11-06
Dunitz/466 post collections

MDunitz:
closes all of https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/466 other than the functional tests

### DIFF
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -587,7 +587,6 @@ components:
     visibility:
       type: string
       enum: [PUBLIC, PRIVATE]
-      default: PRIVATE
     collection:
       type: object
       required:
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -441,6 +441,27 @@ paths:
         "404":
           $ref: "#/components/responses/404"
 
+    post:
+      tags:
+        - collections
+      summary: Create a collection
+      description: >-
+        Creates a new collection in corpora. Once a collection is created it can be linked to datasets.
+      operationId: corpora.lambdas.api.v1.collection.create_collection
+      parameters:
+        - requestBody:
+            content:
+              application/json:
+                schema:
+                  description: >-
+                    A schema that contains a field dictating the visibility of the deployments.
+                  type: object
+                  properties:
+                    visibility:
+                      description: >-
+                        Determines the visibility of the collection, that will either make the collection visible on the public
+                        sites or only viewable through obfuscated URLs.
+                      $ref: "#/components/schemas/collection"
     delete:
       tags:
         - collections
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -602,11 +602,7 @@ components:
       properties:
         access_type:
           type: string
-<<<<<<< HEAD
           enum: [ READ, WRITE ]
-=======
-          enum: [ read, write ]
->>>>>>> WIP
         created_at:
           type: number
         updated_at:
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -413,6 +413,67 @@ paths:
         "400":
           $ref: "#/components/responses/400"
 
+    post:
+      tags:
+        - collections
+      summary: Create a collection
+      security:
+        - cxguserCookie: []
+      description: >-
+        Creates a new collection in corpora. Once a collection is created it can be linked to datasets.
+      operationId: corpora.lambdas.api.v1.collection.create_collection
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - name
+                - description
+                - contact_name
+                - contact_email
+              properties:
+                name:
+                  type: string
+                  description: name of the collecton
+                description:
+                  type: string
+                  description: description of the collection
+                contact_name:
+                  type: string
+                  description: name of the primary person of contact for the collection
+                contact_email:
+                  type: string
+                  description: email of contact person for the collection
+                data_submission_policy_version:
+                  type: string
+                  description: Which data submission policy was the collection submitted under
+                links:
+                  type: array
+                  items:
+                    type: object
+                    properties:
+                      name:
+                        type: string
+                      url:
+                        type: string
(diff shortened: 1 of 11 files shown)


tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -190,3 +190,27 @@ def test__get_collection_uuid__403_not_found(self):
...
+            "contact_name": "person human",
+            "contact_email": "person@human.com",
+            "data_submission_policy_version": 1,
+            "links": [
+                {"name": "DOI Link", "url": "http://doi.org/10.1016", "type": "DOI"},
+                {"name": "DOI Link 2", "url": "http://doi.org/10.1017", "type": "DOI"}]
+        })
+        response = self.app.post(
+            test_url.url,
+            headers={"host": "localhost"},
```
Bento007, line 208: you need to add the `Content-Type` header. For example: https://github.com/HumanCellAtlas/fusillade/blob/b5ec9ed8fa3a45fc0f449ed01b828c2bbafc80a5/tests/test_user_api.py#L130

That way the backend knows how to convert it.

```suggestion
            headers={"host": "localhost", 'Content-Type': "application/json"},
```

backend/config/corpora-api.yml
```diff
@@ -413,6 +413,65 @@ paths:
         "400":
           $ref: "#/components/responses/400"
 
+    post:
+      tags:
+        - collections
+      summary: Create a collection
```
Bento007, line 419: add the security scheme somewhere around here. Once you do and pass in a token, add the user field to `corpora.lambdas.api.v1.collection.create_collection`

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -25,6 +28,29 @@ def get_collection_details(collection_uuid: str):
         raise ForbiddenHTTPException()
 
 
+@db_session
+def create_collection(body: object):
```
Bento007, line 32: - `body` is automatically converted to a `dict` by connexions. So you don't need line 35.
- Once you add security in the openapi spec, the `user` param will be magically populated with the owner id.

```suggestion
def create_collection(body: dict, user: str):
```

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -25,6 +28,29 @@ def get_collection_details(collection_uuid: str):
...
+            description=collection_data.description,
+            owner=collection_data.contact_name,
+            links=collection_data.links,
+            data_submission_policy_version=collection_data.data_submission_policy_version,
+        )
+
+        return make_response(jsonify({"collection_uuid": collection.id}), 201)
+    # except NotAuthorized:
+    #     return make_response(401)
+    except Exception as e:
```
Bento007, line 48: you don't need to catch exceptions here, the errors are returned in the request body.

tests/unit/backend/chalice/api_server/test_v1_collection.py
MDunitz: I keep get a 500 when I did that but now it magically works. Thank you!!

backend/corpora/lambdas/api/v1/collection.py
MDunitz: oh awesome, Thanks!

backend/config/corpora-api.yml
```diff
@@ -412,6 +412,66 @@ paths:
...
+                contact_name:
+                  type: string
+                  description: name of the primary person of contact for the collection
+                contact_email:
+                  type: string
+                  description: email of contact person for the collection
+                data_submission_policy_version:
+                  type: string
+                  description: Which data submission policy was the collection submitted under
+                links:
```
Bento007, line 450: I think there is link datatype you can reference, rather than duplicating it here.

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -32,6 +33,20 @@ def get_collection_details(collection_uuid: str, visibility: str, user: str):
         raise ForbiddenHTTPException()
 
 
+@db_session
+def create_collection(body: object, user: str = ""):
+    collection = Collection.create(
```
Bento007, line 38: remove the default value for user. If a user value is not passed in then, there must be something wrong with the token and we shouldn't proceed,

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -278,3 +279,32 @@ def test__get_collection_uuid__403_not_found(self):
...
+            "data_submission_policy_version": "0.0.1",
+            "links": [
+                {"link_name": "DOI Link", "link_url": "http://doi.org/10.1016", "link_type": "DOI"},
+                {"link_name": "DOI Link 2", "link_url": "http://doi.org/10.1017", "link_type": "DOI"}]
+        })
+        response = self.app.post(
+            test_url.url,
+            headers={"host": "localhost", 'Content-Type': "application/json", "Cookie": get_auth_token(self.app)},
+            data=data)
+        self.assertEqual(201, response.status_code)
```
Bento007, line 299: see if you can retrieve your newly created collection and verify the owner was set.

Bento007 (changes requested):
You should expand on the testing to verify the collection was created.

backend/corpora/common/entities/collection.py
```diff
@@ -66,7 +72,7 @@ def list_collections_in_time_range(cls, *args, **kwargs):
 
     @classmethod
     def list_attributes_in_time_range(
-        cls, to_date: int = None, from_date: int = None, filters: list = None, list_attributes: list = None
+            cls, to_date: int = None, from_date: int = None, filters: list = None, list_attributes: list = None
```
Bento007, line 75: ???

Bento007 (approved):
LGTM

backend/corpora/common/entities/collection.py
MDunitz: I think this is a pycharm formatting change?

PR #696, opened 2020-11-06
feat(collections): extend GET collection/{collection_id} with access_type and visibility

### LINKED ISSUE #645: Extend `GET /v1/collections/{collection_uuid}`
## Description
Returns all available metadata information about a project submission, including URIs of datasets that are attached to the project.

## Parameters:
- collection_uuid [REQUIRED]: the UUID of the project on which information is being requested.
- visibility [REQUIRED]: whether the collection to query is a private collection or a public collection.
- user_id [OPTIONAL]: the OAuth id of the user trying to access a the private collection.

Responses:
200 OK: Response payload will be a schema with all fields that are associated with a submission. Also return the `access_type` which is one of `{READ, WRITE}`.
401 Unauthorized user
404 Submission not found

Implementation details:
- Given the collection_id and the user_id, find the appropriate collection data (and children data like datasets, etc.) from the database.
- If the collection_uuid matches a collection_uuid in the database and the `owner` of the collection does not match the inputted `user_id` (or none is given) return a 401.
- Otherwise if collection_uuid matches a collection_uuid in the database and the `owner` of the collection matches the `used_id`, return the collection data with `access_type=WRITE`.

Bento007:
- Add security schema  `publicAuth` to the Openapi spec to allow unauthenticed users to use the GET collections/{collection_id} endpoint
 - Add `visibility` as query parameter to the Openapi spec to GET collections/{collection_id}
 - Add `access_type` as a property of the `collection` schema to the Openapi spec.
- Update collection endpoint to return read or write access depending on if the user owns the collection or not.
- 403 is returned, if a project does not exists or the user does not have access.
- Visibility query parameter default is public.

fixes #645


backend/chalice/api_server/app.py
```diff
@@ -80,12 +80,15 @@ def dispatch(*args, **kwargs):
         # incompatible with the werkzeug.MultiDict expected by Flask.
         query_string = list(app.current_request.query_params.items()) if app.current_request.query_params else None
 
+        # set dummy auth token value for optional security endpoints
+        headers = [*app.current_request.headers.items(), ('cxgdummy', 'dummy')]
```
tihuan, line 84: since we need to document our cookie usage in cookie policy page, maybe we need to pick a more PC name lol? Like `cxgpublic = public` or something better 😆

backend/config/corpora-api.yml
```diff
@@ -418,11 +418,15 @@ paths:
...
       summary: Get a collection's full details
+      security:
+        - cxguserCookie: []
+        - dummyAuth: []
       description: >-
         This will return all datasets and associated attributes.
       operationId: corpora.lambdas.api.v1.collection.get_collection_details
       parameters:
         - $ref: "#/components/parameters/path_collection_uuid"
+        - $ref: "#/components/parameters/query_visibility"
```
tihuan, line 429: For clarification: Is `query_visibility` expecting FE to pass a param to tell BE if the collection is private or not? If so, given that FE doesn't know if a collection is private or public, BE might need to look this up in the database?

backend/chalice/api_server/app.py
Bento007: sounds good

backend/config/corpora-api.yml
Bento007: we may need to update list collections to include the visibility and retrieve the users collections if they are logged in.

tihuan: In that case, would FE still need to pass `visibility` for the collection?

I'm thinking about the case that a user bookmarks a private collection URL `/collections/private-collection-id` and then hit that URL directly, then the FE won't have a chance to get the collection's visibility. So it doesn't seem like FE is a reliable source of truth for knowing whether a collection is public or private

Another question is, if a malicious user passed `visibility = Public` for a private collection, would they gain access to the private collection?

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -179,6 +179,19 @@ def apikey_info_func(tokenstr: str, required_scopes: list) -> dict:
     return payload
 
 
+def apikey_dummy_info_func(tokenstr: str, required_scopes: list) -> dict:
+    """Function used by connexion in the securitySchemes.
+    This acts a a NOP when the user is not logged in.
```
bmccandless, line 184: nit:  "acts as a NOP"

bmccandless (approved):
LGTM

backend/config/corpora-api.yml
Bento007: > Another question is, if a malicious user passed visibility = Public for a private collection, would they gain access to the private collection?

N, they would need to be the owner or use the obfuscated id to be able read.

Bento007: > I'm thinking about the case that a user bookmarks a private collection URL /collections/private-collection-id and then hit that URL directly, then the FE won't have a chance to get the collection's visibility. So it doesn't seem like FE is a reliable source of truth for knowing whether a collection is public or private

You mean they save the API request. Since the API is not public, we don't guarantee anything if they use the API directly.
Does the frontend URL need to mirror the backend API request format? Couldn't the front end try to get the public and if that fails. try to get the private, or something to that effect.

backend/corpora/lambdas/api/v1/authentication.py
Bento007:
```suggestion
    This acts as a NOOP when the user is not logged in.
```

PR #697, opened 2020-11-06
Extracting auth test code so it can be reused.

Bento007:
- Also change the name of the fake user to `test_user_id` so it can be used in more tests.


MDunitz (approved):
lgtm - just curious why is the token expiration 2 seconds?

Bento007:
that's what was set for the auth test. We can modify the code to make it adjustable.

PR #698, opened 2020-11-07
Pin nodejs to version 14

jgadling:
NPM in the nodejs:15 docker image has a bug that causes it to OOM. This should fix problems with launching the local dev environment using `make dev-init`


tihuan (approved):
LGTM!! Thanks so much for figuring this out, @jgadling 🤩 🙏 !

Issue #702, opened 2020-11-09
authentication path requires better error handling


bmccandless:
Check that any exception that may be thrown when retrieving and validating the token is caught, and appropriate actions are taken.   If the token is malformed or otherwise invalid, it should be removed from the cookie.

In the current implementation the exception is not caught and actually displays on the browser as a backtrace.

PR #707, opened 2020-11-10
authentication path requires better error handling

bmccandless:
#702

### DIFF
--- a/backend/corpora/lambdas/api/v1/authentication.py
+++ b/backend/corpora/lambdas/api/v1/authentication.py
@@ -83,6 +83,7 @@ def oauth2_callback() -> Response:
         current_app.logger.warning(f"Unable to authorize access token: {str(e)}: {str(token)}")
         # remove the token
         remove_token(config.cookie_name)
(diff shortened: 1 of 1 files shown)


backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -75,9 +75,15 @@ def oauth2_callback() -> Response:
...
     client = get_oauth_client(config)
-    token = client.authorize_access_token()
-    # write the cookie
-    save_token(config.cookie_name, token)
+    try:
+        token = client.authorize_access_token()
+        # write the cookie
+        save_token(config.cookie_name, token)
+    except Exception as e:
+        current_app.logger.warning(f"Unable to authorize access token: {str(e)}: {str(token)}")
```
Bento007, line 83: do we want to continue raising the expectation so we return an error response.

bmccandless: We can raise an UnauthorizedError here.  I agree that is better.

Bento007 (approved):
LGTM

PR #709, opened 2020-11-12
thuang-update-dev-API-url

tihuan:
Seems like we need to remove the `/` at the end of the API URL, otherwise we get `//` in the API request! E.g., 

Screen Shot 2020-11-12 at 11 06 34 AM

This somehow prevents Auth cookies from being sent to BE on Dev now

PTAL thank you!


Bento007 (approved):

tihuan:
Thanks so much, Trent!! Merging!

Issue #637, opened 2020-10-07
Extend `GET /v1/collections` endpoint to return public collections and those owned by the user


maniarathi:
GET /v1/collections doesn’t need to accept optional visibility param, BUT, Back end needs to return all private collections owned by the user if logged in, so on the homepage, the user gets to see all public collections (whether they own or not) + all private collections they own

Definition of Done:
- [ ] Backend `GET /v1/collections` endpoint returns all public collections and all private collections owned by the user

## API Description
Lists all collections by their UUIDs that currently exist in Corpora. 

## Responses
200 OK: {request_id, collection_uuid, name, description, user_id}
400 Invalid parameter

tihuan:
The following ticket will rename `/project` with `/collections`!

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/644

mckinsel:
One possible additional filter: there may be datasets with a status that is "incomplete" in some way, either not done uploading or validating or converting or something failed.

bmccandless:
Can a user see their own private projects from this API?

maniarathi:
@mckinsel  They those projects wouldn't be able to get published publicly right? This endpoint would only be for public projects.

@bmccandless No. They would use `GET /v1/submissions`

mckinsel:
@maniarathi , then I think the filter needs to go somewhere else. Suppose:

1. Submitter uploads a picture of a dog.
2. Upload starts, a row with a new dataset_id and appropriate collection_id gets inserted into dataset table.
3. Validation fails (hopefully), and error message gets written into dataset row.

Submitter then publishes the collection. What does this endpoint return?

maniarathi:
Chatted offline so adding details here -- 

We haven't really talked about what "publishing" entails, but likely, it will involve
a) validating that the collection is valid (including the datasets)
b) rescinding the capability URL
c) database finagling if there are issues with changing the visibility of the collection since it's part of the primary key.

Regardless, it won't be possible to return a project that is **public** with errors.

HOWEVER, this API has been updated to just be the same regardless of whether we're working with private or public collections. You'll need to pass in the visibility to filter. So @bmccandless, I take back my previous statement and yes, you would be able to see private collections if you filter by private.

PR #711, opened 2020-11-12
Extend `GET /v1/collections` endpoint to return public collections an…

bmccandless:
…d those owned by the user

 #637

### DIFF
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -1,18 +1,34 @@
 from flask import make_response, jsonify
+from typing import Optional
 
-from ....common.corpora_orm import CollectionVisibility
+from ....common.corpora_orm import DbCollection, CollectionVisibility
 from ....common.utils.db_utils import db_session
 from ....common.entities import Collection
 from ....common.utils.exceptions import ForbiddenHTTPException
 
 
 @db_session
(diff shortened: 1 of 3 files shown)


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -1,18 +1,34 @@
...
+    all_collections = Collection.list_attributes_in_time_range(
+        from_date=from_date,
+        to_date=to_date,
+        list_attributes=[DbCollection.id, DbCollection.visibility, DbCollection.owner, DbCollection.created_at],
+    )
+
+    collections = []
+    for coll_dict in all_collections:
+        visibility = coll_dict["visibility"]
+        owner = coll_dict["owner"]
```
Bento007, line 22: this might be a premature optimization, but wouldn't it be better to write a query that filters by owner or public rather than returning all of the collections within a time range. I think @mckinsel could help write that query.

Bento007 (approved):
LGTM

backend/corpora/lambdas/api/v1/collection.py
bmccandless: I thought about do that, but then I figured that when the get to the point where there are a large number of collections, then ideally the number of public collections should by much greater than the number of private, so the optimization wouldn't save too much.

PR #714, opened 2020-11-13
fix: Null link_name is string

Bento007:


backend/corpora/common/entities/collection.py
```diff
@@ -139,10 +139,9 @@ def reshape_for_api(self) -> dict:
         # Reshape the data to match.
         result.pop("user", None)
         result.pop("owner", None)
-        result["links"] = [
-            dict(link_url=link["link_url"], link_name=link["link_name"], link_type=link["link_type"])
```
mckinsel, line 143: how about just doing `link_name=link["link_name"] or ''`?

Bento007: look at you, with the minimal solution 👍

mckinsel (approved):

Issue #719, opened 2020-11-14
Improve auth tests


bmccandless:
They currently do not reliably pass.

PR #721, opened 2020-11-14
Improve the mock oauth server tests

bmccandless:
It can now be used simultaneously by multiple tests, and the
port used by the mock server can be immediately reused.
This should avoid the flaky tests.

 #719

### DIFF
--- a/tests/unit/backend/chalice/api_server/mock_auth.py
+++ b/tests/unit/backend/chalice/api_server/mock_auth.py
@@ -1,75 +1,97 @@
-import random
 import urllib
-
 import jose.jwt
 import time
-
+import random
+import sys
+import requests
 from flask import Flask, request, redirect, make_response, jsonify
-
-# Create a mocked out oauth server, which servers all the endpoints needed by the oauth type.
-mock_oauth_app = Flask("mock_oauth_app")
-
-# The port that the mock oauth server will listen on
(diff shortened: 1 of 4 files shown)


tests/unit/backend/chalice/api_server/mock_auth.py
```diff
@@ -1,75 +1,97 @@
...
+
+class MockOauthServer:
+    def __init__(self):
+        self.process = None
+        self.port = None
+        self.server_okay = False
+
+    def start(self):
+        self.port = random.randint(10000, 20000)
+        self.process = subprocess.Popen([sys.executable, __file__, str(self.port)])
```
mckinsel, line 78: Does `multiprocessing.Process` not work here?

bmccandless: It should, but I ran into an error when running the full test suite.  I spent a little bit of time 
trying to figure out the issue, but I couldn't spot the problem so I switched to subprocess.

Bento007 (approved):
LGTM

tests/unit/backend/chalice/api_server/mock_auth.py
maniarathi: Just curious -- do you remember what the error was?

PR #729, opened 2020-11-18
thuang-fix-fetch-api-cookies PART 2

tihuan:
Part 2 adds `include` to collections related API calls too! Thank you!

### DIFF
--- a/backend/chalice/api_server/app.py
+++ b/backend/chalice/api_server/app.py
@@ -22,7 +22,7 @@ from corpora.common.utils.json import CustomJSONEncoder
 from corpora.common.utils.aws_secret import AwsSecret
 from corpora.common.corpora_config import CorporaAuthConfig
 
-cors_config = CORSConfig(allow_origin="http://localhost", max_age=600, allow_credentials=True)
+cors_config = CORSConfig(allow_origin="*", max_age=600, allow_credentials=True)
 
 
 def requires_auth():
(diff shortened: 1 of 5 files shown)


tihuan:
I have a feeling the smoke test is failing right now, because of the `*` for `Access-Control-Allow-Origin` from BE (not verified 😆 ) But @bmccandless is helping us to fix that in this PR!

Thank you!

tihuan:
@bmccandless OH! Sorry I revert it back to `*` 🙏 That was an unintended change 😆 

Also since we actually set `CORS` in `get_chalice_app()` call anyway, I think `CORSConfig()`'s  `allow_origin="*"` will always get overridden by `CORS`?

Bento007:
I don't think we need two CORS configurations. We are already using Chalice CORS, can you use that one or replace it.

tihuan:
@bmccandless @Bento007 PTAL again!

I removed Chalice CORS and used `# pragma: no cover` to ignore the code block 🎉 

Looks like CodeCov uses [`coverage.py`](https://github.com/codecov/example-python) under the hood, and `coverage.py` has a way to exclude code! https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html

Hopefully it's good to go now! Thank you!

bmccandless:
> @bmccandless @Bento007 PTAL again!
> 
> I removed Chalice CORS and used `# pragma: no cover` to ignore the code block 🎉
> 
> Looks like CodeCov uses [`coverage.py`](https://github.com/codecov/example-python) under the hood, and `coverage.py` has a way to exclude code! https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html
> 
> Hopefully it's good to go now! Thank you!

Thanks, good to know about "no cover"

bmccandless (approved):
LGTM

tihuan:
Yay! Thanks for the quick review, @bmccandless 🙏 Added auto merge 🚀 Have a wonderful evening!

Bento007 (approved):
LGTM

PR #730, opened 2020-11-18
thuang-update-create-collection-policy

tihuan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/22

Updated the copy per [ticket](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/22) "final policy" items:

UPDATE (11/18):

Screen Shot 2020-11-18 at 8 54 51 PM

Screen Shot 2020-11-18 at 8 55 39 PM

Screen Shot 2020-11-18 at 8 55 43 PM

demo

### DIFF
--- a/frontend/src/common/queries/common.ts
+++ b/frontend/src/common/queries/common.ts
@@ -1,3 +1,3 @@
 export const DEFAULT_FETCH_OPTIONS = {
-  credentials: "include" as RequestCredentials,
+  credentials: "same-origin" as RequestCredentials,
 };
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
   rules: {
     // Disable prop-types as we use TypeScript for type checking
     "@typescript-eslint/explicit-function-return-type": "off",
+    "react/jsx-no-target-blank": 0,
     "react/prop-types": "off",
     "sort-keys": [
       "error",
--- a/frontend/src/components/CreateCollectionModal/components/Content/components/AddLink/index.tsx
+++ b/frontend/src/components/CreateCollectionModal/components/Content/components/AddLink/index.tsx
@@ -42,7 +42,7 @@ const AddLink: FC<Props> = ({ handleClick }) => {
       content={<LinkTypes handleClick={handleClick} />}
       position={Position.BOTTOM_LEFT}
     >
-      <Button minimal intent={Intent.PRIMARY}>
+      <Button outlined intent={Intent.PRIMARY}>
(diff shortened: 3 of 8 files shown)


hthomas-czi (commented):
@tihuan Looks good! I have a couple comments.
1. The `Hide Details` button should still have the `minimal` style, it looks like it's switching to the `outlined` style. 
2. Can we put a border (with 8px padding) around the scroll view to help indicate that it's scrollable?

tihuan:
@hthomas-czi 

Done 😄 ! 

I had to take out the arrow in the popover, because Blueprint doesn't seem to be able to place the arrow at the anchor -.- And I also tried to modify the Popover to at least not cover its anchor, but it doesn't seem to be working 🤦 

Hope this still looks OK! Thank you!

Screen Shot 2020-11-18 at 8 54 51 PM
Screen Shot 2020-11-18 at 8 55 39 PM
Screen Shot 2020-11-18 at 8 55 43 PM


demo

frontend/src/common/queries/common.ts
```diff
@@ -1,3 +1,3 @@
 export const DEFAULT_FETCH_OPTIONS = {
-  credentials: "include" as RequestCredentials,
+  credentials: "same-origin" as RequestCredentials,
```
seve, line 2: Is this going to prevent us from working on hosted dev from local dev?

seve (approved):
LGTM, just one question

frontend/src/common/queries/common.ts
tihuan: Thanks for the call out! I think given that [AuthN BE PR](https://github.com/chanzuckerberg/corpora-data-portal/pull/729) is still in flux and this PR will likely get merged first, temporarily having `same-origin` would allow FE to use Requestly to send cookies, instead of getting the CORS error from BE!

hthomas-czi:
Looks good to me @tihuan! Could we override the `minimal` button styles so it doesn't have the padding and give it a different hover state? I didn't realize that was the behavior of the component when I was designing.

tihuan:
@hthomas-czi Hurray! Oh I could just not use AnchorButton and style it with the same blue and pointer on hover?

tihuan:
Button style updated! PTAL again 😄 🙏 Thank you! @hthomas-czi 

demo

tihuan:
@hthomas-czi LGTM'd in Slack! Thanks all 🎉 😄 👏 !!

PR #731, opened 2020-11-18
Invoking upload step function from backend

Bento007:
partial fix for https://github.com/chanzuckerberg/corpora-data-portal/issues/654

- CorporaConfig uses `corpora/backend/${DEPLOYMENT_STAGE}/config` secret to retrieve the `upload_sfn_arn` populated by single-cell-infra.
- The backend lambda has been given permission to invoke the step function and to describe the execution. This is used in a test.
- The upload step function is invoked with the collection_uuid, dataset_uuid, and URL.
- Adding step function support to localstack container
- Generating secret `corpora/backend/test/config` for testing


tests/functional/backend/corpora/test_upload_sfn.py
```diff
@@ -0,0 +1,54 @@
...
+    client_ssm = boto3.client("secretsmanager", endpoint_url=os.getenv("BOTO_ENDPOINT_URL"))
+    secret_id = "corpora/backend/test/config"
+    old_secret = client_ssm.get_secret_value(SecretId=secret_id)["SecretString"]
+
+
+    def setUpModule():
+        response = client_sfn.create_state_machine(
+            name="fake_step",
+            definition=json.dumps(step_definition),
+            roleArn="arn:aws:iam::123456789012:role/corpora-dataset-uploader-test-sfn-service",
```
MDunitz, line 33: Does it work with this Arn? How?

Bento007: When the environment is test, it works with the test framework created by central infra using localstack. The docker containers must be running in the background for it to work. The fake role isn't used on any of the deployed environments. 

This test will not work in github since the containers aren't used. Central infra is working on that. The tests will work on all other deployments if the step function exists.

MDunitz (approved):
lgtm

PR #738, opened 2020-11-20
thuang-update-BE-server-CORS

tihuan:
Ah, just realized that we need to add `localhost` to the allow list, so localhost can hit dev/staging/prod BE as well!

PTAL thank you 😄 !

### DIFF
--- a/backend/chalice/api_server/app.py
+++ b/backend/chalice/api_server/app.py
@@ -53,6 +53,7 @@ def get_chalice_app(flask_app):
     flask_app.debug = True
     app.debug = flask_app.debug
     app.log.setLevel(logging.DEBUG)
+    localhost_origin = r"^http://localhost:\d+"
 
     # set the flask secret key, needed for session cookies
     flask_secret_key = "OpenSesame"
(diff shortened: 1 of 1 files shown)


backend/chalice/api_server/app.py
```diff
@@ -22,6 +22,16 @@
 from corpora.common.utils.aws_secret import AwsSecret
 from corpora.common.corpora_config import CorporaAuthConfig
 
+ALLOWED_ORIGINS = [
+    "http://localhost:8000",
+    "http://localhost:9000",
```
seve, line 27: In the future, could we check the env of the server and only append localhost origins if it is dev or staging? Just to prevent other clients from accessing prod.

tihuan: Yes! @Bento007 is helping to investigate a complete solution 👌 This change is a temp measure to ease development!

Bento007:
Please look at https://github.com/chanzuckerberg/corpora-data-portal/pull/740

Bento007:
This can be closed now that https://github.com/chanzuckerberg/corpora-data-portal/pull/740 merged

tihuan:
Thanks so much, @Bento007 🔥 🙏 !! Closing!

PR #741, opened 2020-11-20
Add update to dataset

Bento007:
This is needed to update the dataset row after is has been processed by the upload service.


MDunitz (approved):
lgtm -- can you add a test or create a ticket to test the _create_sub_object(s) and the update function (separate from updating a dataset)?

tihuan:
@Bento007 is this PR related to a ticket? If so, can we link this PR to the ticket please? Thank you!

PR #742, opened 2020-11-20
POST /dp/v1/collections/{collection_uuid}/upload/link

Bento007:
- add endpoint POST /dp/v1/collections/{collection_uuid}/upload/link
- Check owner
- Check file size
- Check file name
- Create Dataset and processing_status row
- Verify Dropbox url


backend/config/corpora-api.yml
```diff
@@ -515,6 +453,54 @@ paths:
         "404":
           $ref: "#/components/responses/404"
 
+  /dp/v1/collections/{collection_uuid}/upload/link:
```
tihuan, line 456: discussion: I think I didn't think of this earlier 😆 But does it make sense to change the URL to `/dp/v1/collections/{collection_uuid}/links`, since that feels closer to API best practice of using plurals?

Another thing I'm thinking is, using plurals and perhaps accepting an array of links would be future proof, since we want to support selecting multiple dropbox links in one go or something

But this is definitely not a blocker!

backend/config/corpora-api.yml
```diff
@@ -515,6 +453,54 @@ paths:
...
+    post:
+      summary: Start an dataset upload
+      tags:
+        - collections
+      description: >-
+        An authenticated user can upload a file from a shared link to a dataset in their collection.
+
+        If the upload is in an error state, this endpoint can be used with the dataset_id to restart the collection.
+        If a new link is provided, the new link will be used. If the collection is not in an error state, this endpoint
```
tihuan, line 465: nit:

> If the collection is not in an error state, this endpoint will return an error.

I'm trying to unpack what this means 😆 

So if the collection is **not** in an error state, this endpoint will return an error (400, 401, 409, 413)
But if the collection is in an error state, what will the endpoint return?

backend/corpora/common/corpora_config.py
```diff
@@ -4,11 +4,18 @@
 
 
 class CorporaConfig(SecretConfig):
+    environ_source = "CORPORA_CONFIG"
+
     def __init__(self, *args, **kwargs):
         super().__init__("backend", secret_name="config", **kwargs)
 
+    def get_defaults_template(self):
+        return {"upload_file_formats": ["h5ad"], "upload_max_file_size": 30}
```
tihuan, line 13: nit: do we need to specify the unit is GB?

e.g., `upload_max_file_size_gb`

backend/corpora/common/utils/dropbox.py
```diff
@@ -0,0 +1,31 @@
...
+    params = {"dl": "1"}
+    query.update(params)
+
+    url_parts = list(parsed_url)
+    url_parts[4] = urlencode(query)
+    result = urlunparse(url_parts)
+    return result
+
+
+dropbox_link_rx = regex.compile(r"https://www.dropbox.com/s/[\w\d]+/.*\?dl=[01]")
```
tihuan, line 19: nice!!

MDunitz: for those of us who are terrible at regex, would you mind including a comment about what this is doing?

tihuan: Really great point! +1 😄 

My stab is: 

Example Dropbox share link: https://www.dropbox.com/s/jnm2svlcr45blee/Get%20Started%20with%20Dropbox.pdf?dl=0

`[\w\d]+` matches 1 or more alphanumeric character -- `/jnm2svlcr45blee`
`.*` matches 0 or more any char, which takes care of encoded file name -- `/Get%20Started%20with%20Dropbox.pdf`
`\?dl=[01]` matches URL param `?dl=0` or `?dl=1` --  `?dl=0`

backend/config/corpora-api.yml
Bento007: How about `/dp/v1/collections/{collection_uuid}/upload-links`?

If we accepted and array, we would need to return a list of dataset_uuids and how it maps to each link. Things we can consider in the future. I recommend writing up a ticket for this future work.

backend/corpora/common/corpora_config.py
Bento007: I'll change it to gb

backend/config/corpora-api.yml
Bento007: The upload must be in an error state. If it's not in an error state, 409 is the correct response to return, but for now you can't retry an upload. All requests to the upload/link endpoint result in creating a new dataset.

I added a todo since the above text is not yet implemented.

backend/corpora/common/utils/dropbox.py
Bento007: added some test cases for the regex

backend/config/corpora-api.yml
tihuan: `/dp/v1/collections/{collection_uuid}/upload-links` looks great! Thank you!

Yeah multiple uploads will be for the future us to worry about 😆

backend/config/corpora-api.yml
tihuan: ahh got it! That makes sense now 👌 Thanks for the clarification!

backend/corpora/common/utils/dropbox.py
```diff
@@ -0,0 +1,31 @@
...
+
+def verify(url):
+    return True if dropbox_link_rx.fullmatch(url) else False
+
+
+def get_file_info(url):
+    resp = requests.head(url, allow_redirects=True)
+    return {
+        "size": int(resp.headers["content-length"]),
+        "name": resp.headers["content-disposition"].split(";")[1].split("=", 1)[1][1:-1],
```
bmccandless, line 30: How are failures handled here?

We may want to do some error checking.
Say the url was an invalid, and the resp status code is 404.  
Then the resp.headers will not have context-disposition (for example), and the code will throw an exception
Probably should check the status code, and also check that the headers contain the keys you expect.

backend/corpora/lambdas/api/v1/collection_uuid/upload.py
```diff
@@ -0,0 +1,37 @@
...
+    url = body["url"]
+    if not dropbox.verify(url):
+        raise InvalidParametersHTTPException("The dropbox shared link is invalid.")
+
+    # Get file info
+    url = dropbox.get_download_url_from_shared_link(url)
+    resp = dropbox.get_file_info(url)
+    if resp["size"] > CorporaConfig().upload_max_file_size_gb * GB:
+        raise TooLargeHTTPException()
+    if resp["name"].rsplit(".")[-1].lower() not in CorporaConfig().upload_file_formats:
```
bmccandless, line 26: May need to check the name.  What if name doesn't have a "." character.  This will thrown an exception.

tests/unit/backend/corpora/common/utils/test_dropbox.py
```diff
@@ -0,0 +1,28 @@
...
+        ]
+
+        for test in positive_tests:
+            with self.subTest(test):
+                self.assertTrue(verify(test))
+
+    def test_regex_negative(self):
+        negative_tests = [
+            "https://www.notdropbox.com/s/abcd1234/abcd1234?dl=0",
+            "https://www.dropbox.com/b/abcd1234/abcd1234?dl=0" "https://www.dropbox.com/s/!bcd1234/abcd1234?dl=0",
```
bmccandless, line 21: I think you are missing a comma between the two strings.

bmccandless (approved):
Looks good.
I think there is some additional error checking that can be done,
mentioned in the comments.  But other than that LGTM.

backend/corpora/lambdas/api/v1/collection_uuid/upload.py
bmccandless: I take that back, a missing "." in name won't cause a problem.

backend/corpora/common/utils/dropbox.py
Bento007: good catch, I'll add an error handler and forward the errror

bmccandless (approved):
LGTM.

backend/corpora/common/utils/dropbox.py
mckinsel: @Bento007 would you mind testing this quick with a dropbox file with non-ascii characters in its name (if such a thing is possible)?

PR #743, opened 2020-11-22
Fix scale testing. Was referring to non-existent collection id.

maniarathi:


MDunitz (approved):
lgtm -- at some point we should grab the collection ids from /collections list instead of hard coding them

PR #744, opened 2020-11-23
thuang-517-AuthN-bugfix

### LINKED ISSUE #517: [AuthN] Publishers want to sign up or sign in on the portal to publish their datasets
What:

1. User needs to be able to log in / log out from Data Portal. Log in button will hit `GET API_DOMAIN/v1/login` and Log out button will hit `GET API_DOMAIN/v1/logout`

1. On app initialization, makes an API call to `GET API_DOMAIN/v1/userinfo` to know whether the user is logged in or not

1. When a user logs in from cellxgene and visits Data Portal, they need to be treated as logged in at Data Portal too

1. On the other hand, if a user logs in from Data Portal and visits cellxgene, they need to be treated as logged in at cellxgene

1. We need to implement Call to Action for user to sign in / sign up when clicking on "Create Collection" button

1. If a user is logged in, follow mock design to indicate the user is logged in. If name is available, show **FULL** name, otherwise **show emoji** like `cellxgene` does

   Screen Shot 2020-10-20 at 12 07 53 PM

1. If a user is not logged in, follow mock design and implement log in and create account buttons
1. Check `backend/config/corpora-api.yml` for available endpoints! This [PR](https://github.com/chanzuckerberg/corpora-data-portal/pull/513/files) is the original BE AuthN implementation. 
    - Should be just `/v1/login`, `/v1/logout`, and `/v1/userinfo`. 
    - Prod example: `https://api.cellxgene.cziscience.com/dp/v1/userinfo`

1. Create a `react-query` to get `userinfo` from API

Design:
https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Collections?node-id=0%3A1
https://www.figma.com/file/6wPUe08qSM8V9G2G0UfYOP/Consumer-Collection?node-id=36%3A585
Screen Shot 2020-10-20 at 11 52 44 AM

========== Background Info ==========

**Story**

_Publishers want to create an account on the portal to publish their datasets, increasing their impact through review, evaluation, and exploration by scientists with broad expertise._

**Scheduling Guidance**
We may want to schedule this _just-in-time_ after the minimum self-publishing flow is implemented. This must not be released to publishers if implemented earlier. This is the remaining _frontend_ registration issue to complete this story. 

**UX and BE Designs:**

- [Authentication](https://docs.google.com/document/d/19cC_tH7DOHIYqojdo-6T9eTzPmoMkwrAj83Ozvqth8s/edit#heading=h.t849jh3lwzxu)
- [Flow](https://github.com/chanzuckerberg/corpora-data-portal/issues/356)

tihuan:
This PR fixes the edge case where a user logs out from another tab, the current tab will still be logged in

The issue was that React Query caches the successful response, so even if subsequent calls fail (since the user has logged out from another tab) React Query will still return cached successful response.

The solution is to detect when there's both `userInfo` and `error`, and force refresh the page to log the user out!

closes #517

### DIFF
--- a/frontend/src/components/CreateCollectionModal/index.tsx
+++ b/frontend/src/components/CreateCollectionModal/index.tsx
@@ -34,7 +34,6 @@ const CreateCollection: FC = () => {
         canEscapeKeyClose: false,
         canOutsideClickClose: false,
         content: AsyncContent,
-        isCloseButtonShown: false,
       }
     : {
         canEscapeKeyClose: true,
--- a/frontend/src/components/Header/components/AuthButtons/index.tsx
+++ b/frontend/src/components/Header/components/AuthButtons/index.tsx
@@ -19,7 +19,12 @@ import { ButtonWrapper, Initial } from "./style";
 const AuthButtons = () => {
   const hasAuth = get(FEATURES.AUTH) === BOOLEAN.TRUE;
 
-  const { data: userInfo, isLoading } = useUserInfo(hasAuth);
+  const { data: userInfo, isLoading, error } = useUserInfo(hasAuth);
+
+  if (userInfo && error) {
+    // (thuang): Force refresh page to log user out
+    window.location.reload();
+  }
 
   if (!hasAuth || isLoading) return null;
 


Bento007 (approved):
LGTM

tihuan:
Thanks so much for the quick review, Trent!

PR #748, opened 2020-12-01
Turn entity tests back on and fix them

mckinsel:
Also run the docker process tests for the first time.

### DIFF
--- a/backend/corpora/common/entities/dataset.py
+++ b/backend/corpora/common/entities/dataset.py
@@ -85,7 +85,6 @@ class Dataset(Entity):
         if artifacts or deployment_directories or processing_status:
             if artifacts:
                 for af in self.artifacts:
-                    print(af)
                     self.db.delete(af)
                 new_db_objects = self._create_sub_objects(
                     artifacts, DbDatasetArtifact, add_columns=dict(dataset_id=self.id)
@@ -94,7 +93,6 @@ class Dataset(Entity):
                 kwargs["artifacts"] = new_db_objects
             if deployment_directories:
                 for dd in self.deployment_directories:
-                    print(dd)
                     self.db.delete(dd)
                 new_db_objects = self._create_sub_objects(
                     deployment_directories,
--- a/backend/corpora/common/entities/dataset.py
+++ b/backend/corpora/common/entities/dataset.py
@@ -70,7 +70,7 @@ class Dataset(Entity):
         return cls(new_db_object)
 
     def update(
-        self, artifacts: list = None, deployment_directories: list = None, processing_status: list = None, **kwargs
+        self, artifacts: list = None, deployment_directories: list = None, processing_status: dict = None, **kwargs
     ) -> None:
         """
         Update an existing dataset to match provided the parameters. The specified column are replaced.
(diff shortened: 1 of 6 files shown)


tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -93,16 +94,51 @@ def test__update__ok(self):
...
+            conversion_loom_status=ConversionStatus.NA,
+            conversion_rds_status=ConversionStatus.NA,
+            conversion_cxg_status=ConversionStatus.NA,
+            conversion_anndata_status=ConversionStatus.NA,
+        )
         dataset_params = BogusDatasetParams.get()
         dataset = Dataset.create(
-            **dataset_params, artifacts=[artifact_params] * 1, deployment_directories=[deployment_directory_params] * 1
+            **dataset_params,
+            artifacts=[artifact_params] * 1,
```
bmccandless, line 109: What does the `* 1` do here?

bmccandless (approved):
LGTM

backend/corpora/common/entities/dataset.py
```diff
@@ -85,6 +85,7 @@ def update(
         if artifacts or deployment_directories or processing_status:
             if artifacts:
                 for af in self.artifacts:
+                    print(af)
```
Bento007, line 88: Change these to logging statements.

mckinsel: Oh I actually meant to delete these

tests/unit/backend/corpora/common/entities/test_dataset.py
mckinsel: Hmm nothing I don't think. Looks like it's maybe copied from a pattern where different counts of entities are added.

tests/unit/backend/corpora/common/entities/test_dataset.py
```diff
@@ -93,16 +94,51 @@ def test__update__ok(self):
             s3_uri="some_uri",
         )
         deployment_directory_params = dict(url="test_url")
+        processing_status = dict(
```
Bento007, line 97: to avoid duplication can you make a BogusProcessingStatus Class [here]( https://github.com/chanzuckerberg/corpora-data-portal/blob/ceb8955a9bcecfca6d47d3181741c7541eec4d31/tests/unit/backend/utils.py#L6) and use it here.

Bento007 (changes requested):
A few code cleanup changes

tests/unit/backend/corpora/common/entities/test_collection.py
```diff
@@ -119,9 +119,9 @@ def test__list_in_time_range__ok(self):
 
     def test__list_collections_in_time_range___ok(self):
         """Public collections are returned"""
-        from_date = 10
-        created_at = datetime.fromtimestamp(20)
-        to_date = 30
+        from_date = 10010
```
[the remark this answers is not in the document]
mckinsel: There are two tests that create collections. The new collections persist over tests and you get unexpected values.

Bento007 (approved):
LGTM

PR #749, opened 2020-12-01
Connect upload batch container to db

mckinsel:


backend/corpora/dataset_processing/process.py
```diff
@@ -9,12 +9,30 @@
...
+
+except ImportError:
+    # We're in the container
+    sys.path.append("/code")
+    from common.entities.dataset import Dataset
+    from ..common.corpora_orm import DatasetArtifactFileType, DatasetArtifactType
+
+# This is unfortunate, but this information doesn't appear to live anywhere
+# accessible to the uploader
+DEPLOYMENT_STAGE_TO_URL = {
```
Bento007, line 24: Can we put this in the corpora/${stage}/config

Bento007: or it could be passed to the container by the step function.

Bento007: but this works for now.

backend/corpora/dataset_processing/process.py
```diff
@@ -87,14 +122,21 @@ def extract_metadata(filename):
         numerator += numpy.count_nonzero(chunk)
         denominator += chunk.shape[0]
 
+    def get_term_pairs(base_term):
```
Bento007, line 125:
```suggestion
    def _get_term_pairs(base_term):
```

Bento007: and apply to all places it's used.

backend/corpora/dataset_processing/process.py
```diff
@@ -46,6 +64,23 @@ def fix_dropbox_url(url):
     return pr.geturl()
 
 
+def update_db(metadata=None, processing_status=None):
```
Bento007, line 67: you'll need to wrap this in @db_session to make sure the changes are committed or rolled back.

Bento007: https://github.com/chanzuckerberg/corpora-data-portal/blob/7b3255bfe5c075de4da6c8ea65defa5111ae67b0/backend/corpora/common/utils/db_utils.py#L71

backend/corpora/dataset_processing/process.py
```diff
@@ -46,6 +64,23 @@ def fix_dropbox_url(url):
...
+
+    dataset = Dataset.get(os.environ["DATASET_ID"])
+
+    if metadata:
+
+        # TODO: Delete this line one mean_genes_per_cell is in the db
```
Bento007, line 73: What does this mean?

Bento007:
```suggestion
        # TODO: Delete this line once mean_genes_per_cell is in the db
```

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -28,38 +39,127 @@ def test_extract_metadata(self, mock_read_h5ad):
...
+            extracted_metadata["development_stage"],
+            [{lab: "adult", ont: "HsapDv:0"}, {lab: "baby", ont: "HsapDv:1"}, {lab: "tween", ont: "HsapDv:2"}],
+            self.assertDictEqual,
+        )
         self.assertEqual(extracted_metadata["cell_count"], 50001)
         self.assertAlmostEqual(extracted_metadata["mean_genes_per_cell"], numpy.count_nonzero(df) / 50001)
 
+    def test_update_db(self):
+
+        collection = Collection.create(visibility=CollectionVisibility.PUBLIC)
```
Bento007, line 132: We should only update Private collections, but this check is handled earlier in the upload process. We should change to PRIVATE for consistency.

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -28,38 +39,127 @@ def test_extract_metadata(self, mock_read_h5ad):
...
+        dataset = Dataset.create(collection_id=collection.id, collection_visibility=CollectionVisibility.PUBLIC)
+        dataset_id = dataset.id
+
+        fake_env = patch.dict(os.environ, {"DATASET_ID": dataset_id, "DEPLOYMENT_STAGE": "test"})
+        Dataset.db.session.expire_all()
+        fake_env.start()
+
+        process.update_db(metadata={"sex": ["male", "female"]})
+        Dataset.db.session.expire_all()
```
Bento007, line 141: After you wrap `update` in `db_session` you shouldn't need to expire the sessions anymore

Bento007 (changes requested):

Bento007:
Since both of our PRs' are touching the dropbox URL's we similar code in processing the dropbox URL. We should consolidate.

backend/corpora/dataset_processing/process.py
mckinsel: Currently there's no column for this in the DB. There will be one, and then we should start populating it.

backend/corpora/dataset_processing/process.py
mckinsel: Yeah I think I would be better to to live in some deployment config or terraform and get passed to the container.

Bento007 (approved):
LGTM

PR #751, opened 2020-12-02
fix-fe-dataset

tihuan:
Actually this runtime error is breaking the app completely lol PTAL! Thank you!

Screen Shot 2020-12-01 at 5 32 52 PM


Bento007 (approved):
Great Fix. 

Nit: I'm not sure about displaying N/A, maybe None, or - would be better.

tihuan:
Thanks @Bento007 ! Ahh yeah great feedback!

@hthomas-czi would love your input here 🙏 Thank you!

Discussion context in Slack here: https://chanzuckerbergteam.slack.com/archives/C01CPJURT0S/p1606871738049700

Issue #745, opened 2020-11-23
The data portal must support the intention to create a collection when the publisher is not signed in


brianraymor:
There was an **implicit** assumption if a publisher clicked on _Create Collection_ and was **not** signed in, then the portal would support their intention to create a collection by:

1. Prompting them to sign in or sign up
2. Then upon success, redirecting them to the _Create Collections_ dialog to [achieve their original goal](https://chanzuckerbergteam.slack.com/archives/C01CPJURT0S/p1605571496375600?thread_ts=1605564688.372100&cid=C01CPJURT0S)

Instead, the current implementation prompts the publisher to sign in or sign up and then afterwards returns the publisher to the portal home page where they must click  _Create Collection_ again.

Harley has now added an [**explicit** flow](https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Publisher-Collection?node-id=906%3A192) demonstrating the desired behavior. 

Implementation details:
1. We will need BE update to support RedirectURL  -- probably don't need a separate ticket, since @bmccandless mentioned it was a < 1 point change
1. FE will need to pass redirect URL to include new URL param `?ccModal=true`, so when Auth0 redirects the user back to DP, our app can parse the param and trigger the modal programmatically

tihuan:
@brianraymor I added tech details and pointed a 3! Thank you!

maniarathi:
Reassigning to Seve for the frontend portion.

PR #755, opened 2020-12-02
Add a redirect option to the login endpoint

bmccandless:
usage:  /dp/v1/login?redirect=\<redirect\>
 example: /dp/v1/login?redirect=?showCC=1

Once the login is successful, the user will be redirected to the
frontend base url + the redirect value.

 #745


tihuan (approved):
YAY LGTM!! This will definitely work 🙏 Thanks so much, @bmccandless !!

PR #758, opened 2020-12-04
Merge Dropbox code

Bento007:


backend/corpora/common/utils/dropbox.py
```diff
@@ -1,26 +1,32 @@
-from urllib.parse import urlparse, parse_qs, urlencode, urlunparse
-
+from urllib.parse import urlparse
 import regex
 import requests
 
+dropbox_link_rx = regex.compile(r"/s/[\w\d]+/.*")
```
mckinsel, line 5: I think `\d` is a subset of `\w`, so you'd get the same effect with `\w+`. But, it doesn't include `-` or `.`, which would be valid, common characters in the url. Generally regex matching on urls is a little perilous.

More generally though, does dropbox make any promises about the structure of their shared links? Like if they all started with `/t/...` tomorrow, should that break the system?

backend/corpora/dataset_processing/process.py
```diff
@@ -3,12 +3,13 @@
 import os
 import subprocess
 import sys
-import urllib
 
 import boto3
 import numpy
 import scanpy
 
+from ..common.utils.dropbox import get_download_url_from_shared_link
```
mckinsel, line 11: This is a very tiny nit, but could you just import the module here? Eventually we might have a few ways of parsing these submitted urls, and keeping the context of `dropbox.<fn_name>` when we call it is kind of nice.

mckinsel (changes requested):

backend/corpora/common/utils/dropbox.py
Bento007: > Generally regex matching on urls is a little perilous.

I agree, and dropbox does not make any guarantee. Should we drop the regex all together. My goal here was to [prevent serverside request forgery](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html#available-protections) by matching against a regex as recommended by OWASP. Maybe the best thing to do it %encode the URL once we have it before sending to dropbox

> I think \d is a subset of \w, so you'd get the same effect with \w+. 

you're right \w includes everything.

> But, it doesn't include - or ., which would be valid, common characters in the url. 

No in thus particular part of the URL.

Bento007: After doing more investigating into SSRF prevention, strictly matching the netloc and schema seems sufficient. However I did find an interesting wrapper for requests that claims to prevent SSRF https://github.com/JordanMilne/Advocate.

mckinsel (approved):

mckinsel:
Looks good, note the lint issues though.

PR #760, opened 2020-12-04
thuang-fix-my-collections-only-if-auth

tihuan:
Changes:
1. Only show "My Collections" button when the user is logged in

Current:

Screen Shot 2020-12-04 at 10 11 30 AM

Fixed:
Screen Shot 2020-12-04 at 10 10 24 AM
Screen Shot 2020-12-04 at 10 10 34 AM


seve:
@tihuan Do we want to keep the existing check and append the auth check?

tihuan:
@seve ooh! Great idea 🎉 Thanks for that! DONE 😄 ✨ !

seve (approved):
LGTM!

PR #761, opened 2020-12-04
fix import

Bento007:


MDunitz (approved):
lgtm -- do you need to update the reqs file?

Bento007:
no, `re` is python standard. Our test environment was picking up regex from another package.

Bento007:
the happy path work will help prevent this in the future.

Bento007:
**Cause:** `regex` exists in the requirements needed by the development environment, but does not exist in the requirements for the python lambda environment.

This bug made it through both local testing and the github test environment. Why? Because both of those environments included additional python packages that are not installed on the deployed lambda. These are the packages installed by [requirements.txt](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/requirements.txt) which are needed for development.

In the future, the lambda will be containerized and only contain the minimum required packages. This would have caught the fact that the package that caused the error does not exist in the current python environment.

PR #762, opened 2020-12-05
auth to collection creation flow

seve:
Passed query param to auth backend which is sent with redirect.

Front end reads query param and resumes the user's state in the collection creation  flow.

---

QA:

Navigate to `http://localhost:8000/?showCCModule&auth=true&cc=true`

 _also_

On unauthenticated session, authenticating after attempting to create a collection should redirect to:
`https://cellxgene.dev.single-cell.czi.technology/?showCCModule` (redirect is only available on dev as of now)

### DIFF
--- a/frontend/src/components/CreateCollectionModal/components/CTA/index.tsx
+++ b/frontend/src/components/CreateCollectionModal/components/CTA/index.tsx
@@ -1,6 +1,7 @@
 import { AnchorButton, Button, Classes, Intent } from "@blueprintjs/core";
 import React, { FC } from "react";
 import { API } from "src/common/API";
+import { QUERY_PARAMETERS } from "src/common/constants/queryParameters";
 import { API_URL } from "src/configs/configs";
 import { ContentWrapper } from "./style";
 
@@ -27,7 +28,7 @@ const CTA: FC<Props> = ({ onClose }) => {
           </Button>
           <AnchorButton
             intent={Intent.PRIMARY}
-            href={`${API_URL}${API.LOG_IN}`}
+            href={`${API_URL}${API.LOG_IN}?redirect=?${QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT}`}
           >
             Continue
           </AnchorButton>
--- a/frontend/src/common/constants/queryParameters.ts
+++ b/frontend/src/common/constants/queryParameters.ts
@@ -1,0 +1,3 @@
+export enum QUERY_PARAMETERS {
+  LOGIN_MODULE_REDIRECT = "showCCModule",
+}
--- a/frontend/src/components/CreateCollectionModal/index.tsx
+++ b/frontend/src/components/CreateCollectionModal/index.tsx
@@ -25,7 +25,7 @@ const CreateCollection: FC<{ className?: string }> = ({ className }) => {
   const urlParams = new URLSearchParams(window.location.search);
   const param = urlParams.get(QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT);
 
-  const shouldModuleOpen = param?.toLowerCase() === "true";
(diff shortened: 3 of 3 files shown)


frontend/src/components/CreateCollectionModal/components/CTA/index.tsx
```diff
@@ -27,7 +28,7 @@ const CTA: FC<Props> = ({ onClose }) => {
           </Button>
           <AnchorButton
             intent={Intent.PRIMARY}
-            href={`${API_URL}${API.LOG_IN}`}
+            href={`${API_URL}${API.LOG_IN}?redirect=?${QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT}`}
```
tihuan, line 31: I haven't checked out the code yet 😆 But I see a question mark after `redirect=`, is that intended? 

Feels like that'd interpolate to `?redirect=?showCCModule` instead of `?redirect=showCCModule`

Thank you!!

seve: I'm 85% sure that the backend requires the query `?` as it would also accept something along the lines of `route?param`

tihuan: And I think we might need `?redirect=showCCModule=true`, so the redirect URL will be `?showCCModule=true`, which is a valid URL param?

tihuan: Ahh sorry my bad, you're right! We do need two question marks 😆

And then likely `=true` too?

seve: I'm just doing a check to check that the query param is present and not set to false, lmk if you prefer it more verbose

tihuan: Yeah if you don't mind, let's make it consistent with other URL param implementations, so we follow the "convention over customization" idea that Rails framework popularized!

Also, checking the query param exists and not set to false sounds like the deny list approach over the preferred allow list approach, which potentially means a typo in param chaining `showCCModule=foo=true` could trigger the modal too!

What do you think 😄 ? Thank you!

seve: Consistency is key.  Sounds good to me!

Interesting thoughts about deny/allow list approaches and that edge case. I took a look and you're right about failing open.

`http://localhost:8000/?showCCModule=auth=true`
results in
image

tihuan (approved):
Looks really awesome 🎉 !! Thanks for testing it out ✨ !

One last nit, can we use `BOOLEAN` enum from `frontend/src/common/localStorage/set.ts` instead of hardcoded `"true"` too 🙏 ?

The enum location is probably not the best place now, so please feel free to move that to somewhere better if you feel like it!

```
export enum BOOLEAN {
  TRUE = "true",
  FALSE = "false",
}
```

Thank you!

seve:
@brianraymor ready for QA! You can follow QA steps lined out above in the main body.

frontend/src/components/CreateCollectionModal/components/CTA/index.tsx
tihuan: @seve Ah I think we forgot to add `=true` here 😆 

`href={`${API_URL}${API.LOG_IN}?redirect=?${QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT}`=true}`

PTAL! Thank you!

brianraymor:
Using` https://cellxgene.dev.single-cell.czi.technology/?auth=true&cc=true `, the flow appears correct.

tihuan:
Hurray! Thanks so much for the confirmation, @brianraymor 🎉 🚀

PR #764, opened 2020-12-08
add true value to redirect query param

seve:
Query param wasn't assigned a value to reflect when we changed the param digest to operate only on true values


tihuan (approved):
LGTM!!! Thanks Seve :D

PR #767, opened 2020-12-09
Download Progress

### LINKED ISSUE #661: Download Progress in Upload service
# Need

As a publisher of data on the data portal, I would like to know the progress of my upload, so I can know when my data will be available on the data portal.

# Approach
- In the container, track the file size and the total number of bytes uploaded as the file is uploaded from the cloud storage provider.
- Update the `upload_progress` field in the `dataset_processing_status` table, with the percentage of the upload that is complete when the `upload_status` is uploading.
- `upload_progress` must be updated twice as frequently as the frontend is querying for it. The initial frontend query frequency is 10 seconds, therefore the `upload_status` should be updated every 5 seconds.
- Do not update the database too frequently to the point we DOS the database.
- If an upload fails, or needs to be retried, the progress is reset back to 0.

# DoD
- The progress of an upload can be observed when using the [GET datasets/{dataset_id}/status](https://github.com/chanzuckerberg/single-cell/blob/main/rfcs/0002-upload-from-dropbox/text.md#get-datasetsdataset_idstatus) endpoint during an upload.

Bento007:
### Reviewers:
Functional: @mckinsel 
Readability: @MDunitz 

---

## Changes

- fixes #661 
- Add `upload` to processing code. The file upload happens in two threads. The `uploader` downloads the file from the URL while `update_progress` updates the database on a regular interval.
- Add `db_session_manager` to make it easier to keep database transactions contained.
- Use a local file server for testing.


backend/corpora/lambdas/api/v1/collection_uuid/upload.py
```diff
@@ -24,6 +24,8 @@ def link(collection_uuid: str, body: dict, user: str):
         resp = dropbox.get_file_info(url)
     except requests.HTTPError:
         raise InvalidParametersHTTPException("The URL provided causes an error with Dropbox.")
+    except KeyError:
```
MDunitz, line 27: Can you add a test case for this?

backend/corpora/common/utils/db_utils.py
```diff
@@ -84,3 +85,17 @@ def wrapper_decorator(*args, **kwargs):
...
     return wrapper_decorator
+
+
+@contextmanager
+def db_session_manager():
+    try:
+        db = DbUtils()
+        yield db
+    except SQLAlchemyError:
+        db.session.rollback()
```
MDunitz, line 96: Can you add a test case for errors/rollback?

Bento007: yes I can

backend/corpora/dataset_processing/process.py
```diff
@@ -122,19 +124,18 @@ def update_db(metadata=None, processing_status=None):
         dataset.update(processing_status=status)
 
 
-def fetch_dropbox_url(dropbox_url, local_path):
+def fetch_dropbox_url(dataset_uuid: str, dropbox_url: str, local_path: str) -> str:
```
MDunitz, line 127: can you update the name/signature of this function to make it clear that it kicks off the upload process? And maybe add a test to make it clear that the function always returns the local path you pass in as an arg? Also why does it need to return the local path at all?

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+from ..common.utils.math_utils import MB
+
+logger = logging.getLogger(__name__)
+
+
+class ProgressTracker:
+    def __init__(self, file_size: int):
+        self.file_size: int = file_size
+        self._progress: int = 0
+        self.progress_lock: threading.Lock = threading.Lock()  # prevent concurrent access of _progress
```
MDunitz, line 18: These comments are super helpful! Thanks

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+    :param tracker: Tracks information about the progress of the upload.
+    :return:
+    """
+    try:
+        with requests.get(url, stream=True) as resp:
+            resp.raise_for_status()
+            with open(local_path, "wb") as fp:
+                for chunk in resp.iter_content(chunk_size=chunk_size):
+                    if tracker.stop_uploader.isSet():
+                        logger.debug("Upload ended early!")
```
MDunitz, line 47: Can you test that this is logged properly? https://docs.python.org/3.4/library/unittest.html#unittest.TestCase.assertLogs

tihuan (commented):
Do we need to update `corpora-api.yml` as well or is it not relevant? Thank you!

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+    progress_thread = threading.Thread(
+        target=updater,
+        kwargs=dict(upload_uuid=upload_uuid, tracker=progress_tracker, frequency=update_frequency),
+    )
+    progress_thread.start()
+    upload_thread = threading.Thread(
+        target=uploader, kwargs=dict(url=url, local_path=local_path, tracker=progress_tracker, chunk_size=chunk_size)
+    )
+    upload_thread.start()
+    upload_thread.join()
```
MDunitz, line 128: can you add a comment about why join is used here?

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+
+def upload(dataset_uuid: str, url: str, local_path: str, file_size: int, chunk_size: int = 10 * MB, update_frequency=3):
+    """
+    Upload a file from a url and update the processing_status upload fields in the database
+
+    :param dataset_uuid: The uuid of the dataset the upload will be associated with.
+    :param url: The URL of the file to be uploaded.
+    :param local_path: The local name of the file be uploaded
+    :param file_size: The size of the file in bytes.
+    :param chunk_size: The size of downloaded data to copy to memory before saving to disk.
```
MDunitz, line 110: can you add update_frequency to the param descriptions

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+        manager.session.commit()
+
+
+def updater(upload_uuid: str, tracker: ProgressTracker, frequency: float = 3):
+    """
+    Update the progress of an upload to the database using the tracker.
+
+    :param upload_uuid: The uuid of the upload_progress row.
+    :param tracker: Tracks information about the progress of the upload.
+    :param frequency: The frequency in which the database is updated
```
MDunitz, line 72: is this seconds or ms?

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+                DbDatasetProcessingStatus.upload_progress: progress,
+                DbDatasetProcessingStatus.upload_message: "The file size, does not match the size of the upload.",
+                DbDatasetProcessingStatus.upload_status: UploadStatus.FAILED,
+            }
+        elif progress == 1 and tracker.stop_updater.isSet():
+            processing_status = {
+                DbDatasetProcessingStatus.upload_progress: progress,
+                DbDatasetProcessingStatus.upload_status: UploadStatus.UPLOADED,
+            }
+        else:
```
mckinsel, line 90: I think we need to handle the case where the upload completes, but the local file it too small. I think currently is just stays UPLOADING forever.

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ProgressTracker:
+    def __init__(self, file_size: int):
+        self.file_size: int = file_size
+        self._progress: int = 0
+        self.progress_lock: threading.Lock = threading.Lock()  # prevent concurrent access of _progress
+        self.stop_updater: threading.Event = threading.Event()  # Stops the update_progress thread
+        self.stop_uploader: threading.Event = threading.Event()  # Stops the uploader threads
+        self.error: queue.Queue = queue.Queue(maxsize=1)  # Track errors
```
mckinsel, line 21: Why is this a queue?

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+                for chunk in resp.iter_content(chunk_size=chunk_size):
+                    if tracker.stop_uploader.isSet():
+                        logger.debug("Upload ended early!")
+                        return
+                    elif chunk:
+                        fp.write(chunk)
+                        chunk_size = len(chunk)
+                        tracker.update(chunk_size)
+                        logger.debug(f"chunk size: {chunk_size}")
+    except requests.HTTPError as ex:
```
mckinsel, line 54: Are there other errors to catch here? For example what if we run out of disk?

backend/corpora/lambdas/api/v1/collection_uuid/upload.py
mckinsel, line 27: Could you make this a more explicit exception, like maybe a `dropbox.MissingHeaderException`?

tests/unit/backend/corpora/dataset_processing/test_upload.py
```diff
@@ -0,0 +1,78 @@
...
+        self.addCleanup(self.cleanup_local_file, local_file)
+        url = f"http://localhost:{self.port}/fake.h5ad"
+        upload.upload("test_dataset_id", url, local_file, 100, chunk_size=1024, update_frequency=1)
+        processing_status = Dataset.get("test_dataset_id").processing_status
+        self.assertEqual(UploadStatus.FAILED, processing_status.upload_status)
+
+    def test__dataset_does_not_exist__error(self):
+        local_file = "local.h5ad"
+        self.addCleanup(self.cleanup_local_file, local_file)
+        url = f"http://localhost:{self.port}/upload_test_file.h5ad"
```
mckinsel, line 75: Would you mind giving this file a `.txt` extension? I'm picturing myself trying to open it with scanpy over and over in the future.

mckinsel (changes requested):
Left a few comments.

One general thing: from the perspective of all these upload functions, aren't we _downloading_? To me it's odd for a function named `upload` to copy a file from remote to local.

backend/corpora/common/utils/db_utils.py
```diff
@@ -84,3 +85,17 @@ def wrapper_decorator(*args, **kwargs):
             db.close()
 
     return wrapper_decorator
+
+
+@contextmanager
+def db_session_manager():
```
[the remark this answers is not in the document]
Bento007: done

backend/corpora/dataset_processing/upload.py
Bento007: I thought it was easier than creating another lock to prevent concurrent access.

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+    :param local_path: The local name of the file be uploaded
+    :param tracker: Tracks information about the progress of the upload.
+    :return:
+    """
+    try:
+        with requests.get(url, stream=True) as resp:
+            resp.raise_for_status()
+            with open(local_path, "wb") as fp:
+                for chunk in resp.iter_content(chunk_size=chunk_size):
+                    if tracker.stop_uploader.isSet():
```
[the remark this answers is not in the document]
Bento007: changed to is_set

backend/corpora/dataset_processing/upload.py
Bento007: done

backend/corpora/dataset_processing/upload.py
Bento007: with a 30Gb limit on file size, is this a possibility? Any other errors I should catch? I could put a blanket `except Exception`.

backend/corpora/dataset_processing/upload.py
Bento007: added

backend/corpora/dataset_processing/upload.py
Bento007: added

backend/corpora/dataset_processing/upload.py
```diff
@@ -0,0 +1,139 @@
...
+    :param url: The URL of the file to be uploaded.
+    :param local_path: The local name of the file be uploaded
+    :param file_size: The size of the file in bytes.
+    :param chunk_size: The size of downloaded data to copy to memory before saving to disk.
+    """
+    with db_session_manager() as mananger:
+        processing_status = Dataset.get(dataset_uuid).processing_status
+        processing_status.upload_status = UploadStatus.UPLOADING
+        processing_status.upload_progress = 0
+        mananger.commit()
```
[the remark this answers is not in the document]
Bento007: If there isn't anything to [commit](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.commit) then an [InvalidRequestError](https://docs.sqlalchemy.org/en/13/core/exceptions.html#sqlalchemy.exc.InvalidRequestError) will be thrown. We will need to add a  catch for this error in the context manager and pass it through. I worry that this will lead to us throwing a bunch of false positives and missing a real error.

backend/corpora/dataset_processing/upload.py
Bento007: added

backend/corpora/dataset_processing/upload.py
Bento007: also if multiple errors occur we want to catch them.

backend/corpora/dataset_processing/upload.py
Bento007: fixed

Bento007:
> One general thing: from the perspective of all these upload functions, aren't we downloading? To me it's odd for a function named upload to copy a file from remote to local.

I 100% agree, but when I was doing the upload endpoint someone said it confusing to read in the upload in the download endpoint. I will gladly change it!

tests/unit/backend/corpora/dataset_processing/test_upload.py
Bento007: fixed

backend/corpora/dataset_processing/upload.py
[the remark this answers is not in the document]
Bento007: I add a commit flag that can be set if you want to commit at the end of the transaction

backend/corpora/lambdas/api/v1/collection_uuid/upload.py
Bento007: I can, but I'm not really sure How I would test this.

Bento007: I can try mocking the requests library to return a response without headers.

Bento007: This was changed, and a test was added to verify the new error condition

backend/corpora/dataset_processing/process.py
Bento007: I update the name

mckinsel (commented):
Just a couple remaining things.

backend/corpora/common/utils/db_utils.py
```diff
@@ -68,19 +69,36 @@ def __getattr__(self, name):
...
+    except SQLAlchemyError:
+        db.session.rollback()
+        msg = "Failed to commit."
+        logger.exception(msg)
+        raise CorporaException(msg)
+    finally:
+        db.close()
+
+
+def db_session(func, commit=False):
```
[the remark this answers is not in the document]
Bento007: yeah, I was doing it kinda wrong. I fixed it. Someday I'll get a decorator right on the first shot. When using the decorator use `@db_session()` for no commit and `@db_session(commit=True)` or `@db_session(True)` for commit

backend/corpora/dataset_processing/download.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class ProgressTracker:
+    def __init__(self, file_size: int):
+        self.file_size: int = file_size
+        self._progress: int = 0
+        self.progress_lock: threading.Lock = threading.Lock()  # prevent concurrent access of ProgressTracker._progress
+        self.stop_updater: threading.Event = threading.Event()  # Stops the update_progress thread
+        self.stop_downloader: threading.Event = threading.Event()  # Stops the downloader threads
+        self.error: queue.Queue = queue.Queue()  # Track errors
```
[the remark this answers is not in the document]
Bento007: > but isn't this written once by a single thread and read once by another?

This was the impression I was under but since two threads are using the error queue so it's possible they could both fail.

backend/corpora/dataset_processing/download.py
```diff
@@ -0,0 +1,157 @@
...
+                for chunk in resp.iter_content(chunk_size=chunk_size):
+                    if tracker.stop_downloader.is_set():
+                        logger.info("Download ended early!")
+                        return
+                    elif chunk:
+                        fp.write(chunk)
+                        chunk_size = len(chunk)
+                        tracker.update(chunk_size)
+                        logger.debug(f"chunk size: {chunk_size}")
+    except requests.HTTPError as ex:
```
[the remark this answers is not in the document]
Bento007: added

mckinsel (approved):

Issue #724, opened 2020-11-16
Integration: Publisher can choose a dataset from dropbox and start an upload


brianraymor:
Placeholder for any required FE-BE integration. 

***TEAM COMMENTS***

1. feature flags are: `cc=true` and `auth=true`
2. I think it's currently working in development env, and we're waiting to deploy to staging on Wednesday for QA!

QA steps once in staging:

1. Use this feature flag enabled staging link: https://cellxgene.staging.single-cell.czi.technology/?cc=true&auth=true
1. Log into your account
1. If you already have collections, click on My Collections link to a access a collection page
1. Once you're on a collection page, click on "Add dataset from Dropbox" to start the upload flow

![Screen Shot 2021-01-04 at 9.59.25 AM.png](https://images.zenhubusercontent.com/5ebed8780d2d150fc281ad55/efd0698f-0ae0-4bd2-95ec-7b236395b683)

Thank you!

tihuan:
I'll do this ticket along with [#642](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/642undefined)! Thank you!

tihuan:
Deployed to staging and can confirm that it's working in staging. But @brianraymor please feel free verify too as product owner (QA steps in ticket description) 🙏 Thank you!

tihuan:
Confirmed working in prod! Closing!

PR #769, opened 2020-12-10
Thuang 642 dropbox integration

### LINKED ISSUE #642: Upload Dataset Part 1 [Dropbox Integration]
Design: https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Publisher-Collection?node-id=39%3A793

What:

**After finishing this PART 1 ticket, the user will be able to upload a dataset via Dropbox Chooser and FE will make an API call to the BE to kick start the upload process**

1. Please implement Dropbox Chooser following the [design](https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Publisher-Collection?node-id=39%3A894) - which includes the "Add Dataset" button. Right now we only allow Dropbox upload, so I don't think we need dropdown - @hthomas-czi please help confirm 🙏 Thank you!

1. Restrict to select ONE file for now / Direct download link

1. Call the relevant API to kick off the cloud to cloud transfer -- API TBD, since some discussion is still ongoing in Slack `#tmp-sc-collection-flow-ux`

1. API [here](https://github.com/chanzuckerberg/single-cell/blob/main/rfcs/0002-upload-from-dropbox/text.md#data-portal-apis), please also double check with Marcus and Trent to make sure that's the latest/correct API doc

Research notes:
1. Dropbox Chooser library: https://www.dropbox.com/developers/chooser
2. Need a Dropbox app in App Console
3. Need to create a CZI Dropbox account
4. Need to specify deployment web app links (dev, staging, prod)
5. Embed their script through Dropbox CDN
6. The Chooser option config section allows direct download link, multiselect, file types, size limit, etc.. Super convenient!
7. Direct download link expires in 4 hours
8. Looks like we need to call `Dropbox.isBrowserSupported()` to see if the browser supports it. (I've confirmed that the latest Chrome, Firefox, and Safari all work)
9. Contact IT to create Dropbox account for us!

tihuan:
fixes #642 #724 

Changes:

1. Adds `dropins.js` in `frontend/src/components/DropboxChooser/components/dropins.js` instead of using a script tag to load it, because it works better that way with React + code splitting

1. Extracts `EmptyDatasets` to its own Component file for organization

1. Creates `DropboxChooser` Component that accepts lazy loads the Chooser library when its children (button) are hovered over. It's also responsible for interacting with the Chooser library and edits/passes the dropbox link to its ancestor component via prop `onSelectUploadLink(dropboxLink)`

1. Creates POST upload-links mutation query

1. `views/Collection` Component now has `useEffect` that listens to new `dropboxLink` to make a POST call to upload the dataset

demo

Screen Shot 2020-12-09 at 4 46 20 PM

QA Steps:

1. This is actually not easy to test locally, because Dropbox authentication doesn't like Requestly for some reason. So we have to deactivate Requestly, physically comment out some code to show the upload button on a Collection page even if the collection is not available (since all WRITE collections are private now, when we're logged out, the BE won't return the collection)

1. Open your dev console, so it will record the following network activities

1. Now you should be able to see the upload button and clicking on it will trigger the Dropbox window

1. Log in to your Dropbox account

1. Select an `.h5ad` file and click "Choose" button

1. Check network tab, and you should see a request to `***/upload-links` with a dropbox share link as `url`

OR: We can just merge this in `main` and test it on Dev lol

Thank you!

@Bento007 for read
@seve  for functionality


frontend/src/components/DropboxChooser/components/Content.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { FC, useEffect } from "react";
+import dropins from "./dropins";
+
+// (thuang): My personal Dropbox app key. Will be replaced with
+// Single Cell app key once the contract goes through
+const TEMP_APP_KEY = "66mv8tdrtxi2s0m";
```
Bento007, line 6: Are there any security risks to having this exposed?

frontend/src/components/DropboxChooser/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    )
+);
+
+type DropboxFile = {
+  link: string;
+};
+
+const DROPBOX_OPTIONS = {
+  extensions: [".h5ad"],
+  sizeLimit: 30 * 1024 * 1024 * 1024, // 30GB
```
Bento007, line 26: 2^30 == 1 GB if that's less confusing

frontend/src/components/DropboxChooser/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+const DropboxChooser: FC<Props> = ({ children, onSelectUploadLink }) => {
+  const [isContentShown, setIsContentShown] = useState(false);
+
+  const handleMouseOver = () => {
+    setIsContentShown(true);
+  };
+
+  const handleClick = () => {
+    if (!window.Dropbox.isBrowserSupported()) {
+      alert("Sorry, your browser does not support Dropbox upload :(");
```
Bento007, line 42: Is there an alternative option to submit if their browser is not supported?

frontend/src/components/DropboxChooser/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    const options = {
+      ...DROPBOX_OPTIONS,
+      success(files: DropboxFile[] = []) {
+        // (thuang): We need to set `dl=1` to be direct download
+        const link = files[0].link.replace("dl=0", "dl=1");
```
Bento007, line 49: This is handled on the backend as well if you want to avoid repetition.

Bento007 (commented):
Does frontend/src/components/DropboxChooser/components/dropins.js need to be reviewed? Is this from dropbox?

frontend/src/components/DropboxChooser/components/Content.tsx
tihuan: Great question! I'm thinking since the app-key is not a secret (publicly available in the rendered HTML), I think it should be fine. On Dropbox's tutorial, the key is hard coded in the `<script />` tag as well :D The dangerous part of configs is done on Dropbox Developer website!

frontend/src/components/DropboxChooser/index.tsx
tihuan: Oh super great feedback 😆 Thanks a lot! Done!

frontend/src/components/DropboxChooser/index.tsx
tihuan: Great question! We don't haves alternatives right now, since Dropbox Chooser is the only UI we have. And given that our web app only supports modern browsers, they should all work for Dropbox Chooser 😄

frontend/src/components/DropboxChooser/index.tsx
tihuan: ooh perfect thank you! I removed the edit here!

tihuan:
Thanks so much for the detailed review, @Bento007 !

`frontend/src/components/DropboxChooser/components/dropins.js ` is straight from Dropbox, so we don't need to review it 😄 🙏

seve (approved):
LGTM!

PR #777, opened 2020-12-15
Getting 500 errors for GET Collection

Bento007:
#### Reviewers
**Functional:** @MDunitz   

**Readability:** @mckinsel 

---

## Changes
- add test to return a collection with the minimum number of field allowed
- add tests to verify the fix.
- remove required fields from dataset schema in openAPI
- This is part of the dropbox integration work

When reading a dataset already in the database, not all fields expected according to the openAPI spec were present. Since the only way to add a dataset to a collection is through the upload of an H5AD file, which does its own validation, we can relax the fields required in the openAPI spec. The spec has been modified to allow more field in the dataset to be null. This presently only affect responses we return since users cannot population these field through the API.

**This PR currently breaks the `My Collections` page. I'll need to work with Frontend to resolve this before merging.**


MDunitz (approved):
lgtm 
out of curiosity, on line 105 of test_v1_collection.py why are you using .addCleanup instead of just running clean up after each test in the teardown function?

Bento007:
@MDunitz it's a super handy feature of unittests that let's you add a cleanup handler to a specific test. Since these tests create a resources that are only relevant to that single test we want to clean up that test data afterward so they don't pollute the test environment for the rest of the tests.

tihuan (approved):
LGTM!! Thanks so much @Bento007 and @seve for fixing this 🤩  🙏 🎉 !

PR #787, opened 2020-12-22
Remove null fields from JSON Response

Bento007:
#### Reviewers
**Functional:** @mckinsel 

**Readability:** @bmccandless 

---

Fields that would normally be returned as null in the JSON response will be removed by the backed before returning the response. Prior to this change, we were returning null values for some APIs and remove the null fields in other APIs. The goal is to handle null fields in the response uniformly. 

This is part of an effort to follow a better [coding style](https://google.github.io/styleguide/jsoncstyleguide.xml#Empty/Null_Property_Values). 
## Changes
- Remove nullable from openapi spec
- Remove None values from JSON responses
- Add option to remove None values when converting a SQLAlchemy Base class to a dictionary


bmccandless (approved):
LGTM

PR #796, opened 2021-01-05
update validation and conversion status during upload

Bento007:
#### Reviewers
**Functional:** @mckinsel 

**Readability:** @MDunitz 

---

## Changes
- Add `DatasetAsset.create` to add assets to a dataset.
- Fix code that converts an ORM model to a dictionary with None values. If `remove_none` is True, continue, else If `remove_none` is False, add it to the returned dictionary.
- Add `generate_uuid` function to generate uuids to reduce code duplication.
- dataset_processing/process.py
  - Add updates to the database when conversion starts and finishes for cxg, seurat, loom, and h5ad.
  - Add updates to the database when validation starts and finishes for h5ad. 
  - Add `logger` to dataset_processing/process to log exceptions.
  - Modify `update_db` to use `processing_status_updater`
  - Moved s3_client to global scope
  - Add `create_artifact` functions to upload an artifact to s3 and update the dataset.
  - Localized variables to reduce the use of environment variables, and simplify testing.
  - Modify the return value of `convert_file_ignore_exceptions` to `file_name` and `ConversionStatus`
  - Add `get_bucket_prefix` function to reduce code duplication
  - Add `process_cxg` function to handle processing the CXG for code readability.
- Moved *generate_data_mixing* so it can be used in *test_process*
- Fix`GenerateDataMixin* to work even if there is no active session.


backend/corpora/common/utils/uuid.py
```diff
@@ -0,0 +1,5 @@
+import uuid
+
+
+def generate_uuid():
```
MDunitz, line 4: nit -- does this need to be in its own file? can it go in db_utils? or as a function on the base of the orm?

MDunitz (approved):
lgtm

backend/corpora/common/utils/uuid.py
Bento007: I could put it in ORM, but it doesn't need to be apart of base.

mckinsel (approved):
LGTM. One tiny nit, and we otherwise discussed in slack.

PR #799, opened 2021-01-05
Dunitz/660 delete endpoint

MDunitz:
#### Reviewers
**Functional:** 
@Bento007 

**Readability:** 
@bmccandless 

---

## Changes
- add an endpoint to cancel a dataset upload while it is in progress

### DIFF
--- a/backend/chalice/api_server/Makefile
+++ b/backend/chalice/api_server/Makefile
@@ -40,7 +40,6 @@ package: build-chalice-config
 	mkdir -p chalicelib/corpora
 	cp -R ../../corpora/common chalicelib/corpora
 	cp -R ../../corpora/lambdas chalicelib/corpora
-	cp -R ../../corpora/dataset_processing chalicelib/corpora
 	mkdir -p chalicelib/config
 	cp -a ../../config/$(APP_NAME).yml chalicelib/config/$(APP_NAME).yml
 	rm -rf vendor dist/deployment
--- a/backend/chalice/api_server/Makefile
+++ b/backend/chalice/api_server/Makefile
@@ -40,6 +40,7 @@ package: build-chalice-config
 	mkdir -p chalicelib/corpora
 	cp -R ../../corpora/common chalicelib/corpora
 	cp -R ../../corpora/lambdas chalicelib/corpora
+	cp -R ../../corpora/dataset_processing chalicelib/corpora
 	mkdir -p chalicelib/config
 	cp -a ../../config/$(APP_NAME).yml chalicelib/config/$(APP_NAME).yml
 	rm -rf vendor dist/deployment
--- a/backend/corpora/common/utils/aws.py
+++ b/backend/corpora/common/utils/aws.py
@@ -96,7 +96,8 @@ def delete_many_from_s3(bucket_name: str, dataset_uuid: str) -> None:
     This deletes everything with a specific prefix from the given bucket
 
     """
-    assert dataset_uuid
+    if not dataset_uuid:
+        raise ValueError
     s3 = boto3.resource("s3")
     bucket = s3.Bucket(bucket_name)
     bucket.objects.filter(Prefix=f"{dataset_uuid}/").delete()
--- a/backend/corpora/common/utils/aws.py
+++ b/backend/corpora/common/utils/aws.py
@@ -89,3 +89,13 @@ class AwsSecret:
     def _debug(self, message):
         if self.debug_logging:
             print(message)
+
+
+def delete_many_from_s3(bucket_name, dataset_uuid) -> None:
+    """
+    This deletes everything with a specific prefix from the given bucket
+
+    """
+    s3 = boto3.resource("s3")
+    bucket = s3.Bucket(bucket_name)
+    bucket.objects.filter(Prefix=f"{dataset_uuid}/").delete()
--- a/backend/corpora/common/utils/aws.py
+++ b/backend/corpora/common/utils/aws.py
@@ -91,11 +91,12 @@ class AwsSecret:
             print(message)
 
 
-def delete_many_from_s3(bucket_name, dataset_uuid) -> None:
+def delete_many_from_s3(bucket_name: str, dataset_uuid: str) -> None:
     """
     This deletes everything with a specific prefix from the given bucket
 
     """
+    assert dataset_uuid
     s3 = boto3.resource("s3")
     bucket = s3.Bucket(bucket_name)
     bucket.objects.filter(Prefix=f"{dataset_uuid}/").delete()
--- a/backend/corpora/dataset_processing/download.py
+++ b/backend/corpora/dataset_processing/download.py
@@ -4,17 +4,19 @@ import requests
 
 from backend.corpora.common.corpora_orm import DbDatasetProcessingStatus, UploadStatus
 from backend.corpora.common.entities import Dataset
-from backend.corpora.common.utils.db_utils import db_session_manager
+from backend.corpora.common.utils.aws import delete_many_from_s3
+from backend.corpora.common.utils.db_utils import db_session_manager, processing_status_updater
 from backend.corpora.common.utils.math_utils import MB
-
 
 logger = logging.getLogger(__name__)
 
 
 class ProgressTracker:
-    def __init__(self, file_size: int):
+    def __init__(self, file_size: int, artifact_bucket: str, cellxgene_bucket: str):
         self.file_size: int = file_size
         self._progress: int = 0
+        self.artifact_bucket = artifact_bucket
+        self.cellxgene_bucket = cellxgene_bucket
         self.progress_lock: threading.Lock = threading.Lock()  # prevent concurrent access of ProgressTracker._progress
         self.stop_updater: threading.Event = threading.Event()  # Stops the update_progress thread
         self.stop_downloader: threading.Event = threading.Event()  # Stops the downloader threads
@@ -31,24 +33,6 @@ class ProgressTracker:
     def cancel(self):
         self.stop_downloader.set()
         self.stop_updater.set()
-
-
-def cancel_upload(dataset_uuid: str):
-    print(f"cancelling the upload for {dataset_uuid}")
-
-    # set db to cancelled
-    status = {
-        DbDatasetProcessingStatus.upload_progress: 0,
-        DbDatasetProcessingStatus.upload_status: UploadStatus.CANCELED,
-        DbDatasetProcessingStatus.upload_message: "Canceled by user",
-    }
-
-    dataset = Dataset.get(dataset_uuid)
-    processing_status_updater(dataset.processing_status.id, status)
-    # delete from s3
-    delete_many_from_s3(os.environ["ARTIFACT_BUCKET"], dataset_uuid)
-    delete_many_from_s3(os.environ["CELLXGENE_BUCKET"], dataset_uuid)
-
 
 
 def downloader(url: str, local_path: str, tracker: ProgressTracker, chunk_size: int):
@@ -81,11 +65,6 @@ def downloader(url: str, local_path: str, tracker: ProgressTracker, chunk_size: 
         tracker.stop_updater.set()
 
 
-def processing_status_updater(uuid: str, updates: dict):
-    with db_session_manager(commit=True) as manager:
-        manager.session.query(DbDatasetProcessingStatus).filter(DbDatasetProcessingStatus.id == uuid).update(updates)
-
-
 def updater(processing_status_uuid: str, tracker: ProgressTracker, frequency: float):
     """
     Update the progress of an upload to the database using the tracker.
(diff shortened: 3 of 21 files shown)


backend/corpora/dataset_processing/download.py
```diff
@@ -38,6 +39,31 @@ def update(self, progress):
         with self.progress_lock:
             self._progress += progress
 
+    def cancel(self):
+        self.stop_downloader.set()
+        self.stop_updater.set()
+
+
+def cancel_upload():
+    dataset_uuid = os.environ["DATASET_ID"]
```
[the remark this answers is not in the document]
MDunitz: The tracker only has access to to the ProcessStatus but I can use that to access the dataset_uuid

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -48,3 +55,24 @@ def get_status(dataset_uuid: str, user: str):
...
+@db_session()
+def cancel_dataset_download(dataset_uuid: str, user: str):
+    dataset = Dataset.get(dataset_uuid)
+    if not dataset:
+        raise NotFoundHTTPException(f"'dataset/{dataset_uuid}' not found.")
+    if not Collection.if_owner(dataset.collection.id, dataset.collection.visibility, user):
+        raise ForbiddenHTTPException()
+    curr_status = dataset.processing_status
+    if curr_status.upload_status is UploadStatus.UPLOADED:
+        raise MethodNotAllowedException(f"'dataset/{dataset_uuid}' upload is complete and can not be cancelled.")
```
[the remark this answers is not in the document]
MDunitz: Thats a good point, I set it up this way because the ticket specified `an upload can be canceled if it has not reached the uploaded, upload_status`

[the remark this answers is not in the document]
Bento007: At the time of writing the ticket, deleting datasets was not something we were trying to accomplish. I believe the need to download a dataset is coming soon.

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -1,8 +1,15 @@
...
 from ....common.entities import Dataset, Collection
 from ....common.utils.db_utils import db_session
-from ....common.utils.exceptions import NotFoundHTTPException, ServerErrorHTTPException, ForbiddenHTTPException
+from ....common.utils.exceptions import (
+    NotFoundHTTPException,
+    ServerErrorHTTPException,
+    ForbiddenHTTPException,
+    MethodNotAllowedException,
+)
+from ....dataset_processing.download import processing_status_updater
```
Bento007, line 12: I'd avoid importing from data_processing, we might start running into weird circular dependencies in the future if we do. You can access and update the `upload_status` through dataset.

`processing_status_updater` was written as an optimized way of updating the upload_status since we were modifying it frequently. It wasn't supposed to really be used outside of the downloader code.

To modify the upload_status you can do something like this 
```python
from corpora.common.entities.dataset import Dataset
from corpora.common.corpora_orm import UploadStatus
from corpora.common.utils.db_utils import db_session_manager

with db_session_manager(commit=True):
  data_uuid = "test_dataset_id"
  dataset = Dataset.get(dataset_uuid)
  dataset.processing_status.upload_status = UploadStatus.CANCEL_PENDING
```

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -48,3 +55,24 @@ def get_status(dataset_uuid: str, user: str):
     for remove in ["dataset", "created_at", "updated_at"]:
         status.pop(remove)
     return make_response(jsonify(status), 200)
+
+
+@db_session()
+def cancel_dataset_download(dataset_uuid: str, user: str):
```
Bento007, line 61: Follow the naming convention of the other APs
```suggestion
def delete_dataset(dataset_uuid: str, user: str):
    """Cancels an inprogress upload."""
```

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -1,8 +1,15 @@
...
 
+from ....common.corpora_orm import DbDatasetProcessingStatus, UploadStatus
 from ....common.entities import Dataset, Collection
 from ....common.utils.db_utils import db_session
-from ....common.utils.exceptions import NotFoundHTTPException, ServerErrorHTTPException, ForbiddenHTTPException
+from ....common.utils.exceptions import (
+    NotFoundHTTPException,
+    ServerErrorHTTPException,
+    ForbiddenHTTPException,
+    MethodNotAllowedException,
```
Bento007, line 10: Why did you choose method not allowed?

backend/corpora/dataset_processing/download.py
```diff
@@ -38,6 +39,31 @@ def update(self, progress):
         with self.progress_lock:
             self._progress += progress
 
+    def cancel(self):
+        self.stop_downloader.set()
+        self.stop_updater.set()
+
+
+def cancel_upload():
```
Bento007, line 47: All of this code and live in `_update`, it doesn't need to be its own function

backend/corpora/dataset_processing/download.py
```diff
@@ -85,6 +111,11 @@ def updater(processing_status_uuid: str, tracker: ProgressTracker, frequency: fl
     """
 
     def _update():
+        with db_session_manager(commit=True) as db:
+            curr_status = db.get(DbDatasetProcessingStatus, processing_status_uuid)
+            if curr_status.upload_status is UploadStatus.CANCEL_PENDING:
```
Bento007, line 116: After removing `sys.exit(1)` from `cancel_upload` you will need to add 
```python
elif curr_status.upload_status is UploadStatus.Canceled:
  return
```

backend/corpora/dataset_processing/download.py
```diff
@@ -9,6 +9,7 @@
     from ..common.entities import Dataset
     from ..common.utils.db_utils import db_session_manager
     from ..common.utils.math_utils import MB
+    from ..lambdas.upload_failures.upload import delete_many_from_s3
```
Bento007, line 12: if `delete_many_from_s3` will be used by multiple deployed resources then it should live in `common`. I think in `common.utils` would be a good start. This helps us avoid circular dependencies and keeps the code's hierarchy in tact.

Bento007 (changes requested):
A few things you may have overlooked

backend/corpora/lambdas/api/v1/dataset.py
MDunitz: discussed over slack
-> I am going to move the processing_status_updater into db_utils

backend/corpora/common/utils/aws.py
```diff
@@ -89,3 +89,13 @@ def _restore(self):
...
+
+
+def delete_many_from_s3(bucket_name, dataset_uuid) -> None:
+    """
+    This deletes everything with a specific prefix from the given bucket
+
+    """
+    s3 = boto3.resource("s3")
+    bucket = s3.Bucket(bucket_name)
+    bucket.objects.filter(Prefix=f"{dataset_uuid}/").delete()
```
bmccandless, line 101: We should check the inputs here, and verify that dataset_uuid is not an empty string.

https://developers.slashdot.org/story/16/04/14/1542246/man-deletes-his-entire-company-with-one-line-of-bad-code

backend/corpora/dataset_processing/download.py
```diff
@@ -74,6 +76,28 @@ def updater(processing_status_uuid: str, tracker: ProgressTracker, frequency: fl
...
+
+                    dataset = Dataset.get(curr_status.dataset.id)
+                    processing_status_updater(dataset.processing_status.id, status)
+                    # delete from s3
+                    delete_many_from_s3(tracker.artifact_bucket, dataset.id)
+                    delete_many_from_s3(tracker.cellxgene_bucket, dataset.id)
+                    tracker.cancel()
+                elif curr_status is UploadStatus.Canceled:
+                    return
+            except AttributeError:
```
bmccandless, line 99: Maybe add a comment.  What in the above can raise an attribute error and why is that not a problem.

bmccandless (changes requested):
Looks good.  I just had two small things to address.

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -48,3 +54,27 @@ def get_status(dataset_uuid: str, user: str):
...
+
+
+@db_session()
+def delete_dataset(dataset_uuid: str, user: str):
+    """
+    Cancels an inprogress upload.
+    """
+    dataset = Dataset.get(dataset_uuid)
+    if not dataset:
+        raise NotFoundHTTPException(f"'dataset/{dataset_uuid}' not found.")
```
Bento007, line 66: For systems that require auth for some resource, you should return ForbiddenHTTPException both when an item is to found and when the user is not authorized. This prevents users from probing the system for uuids.

```suggestion
         raise ForbiddenHTTPException()
```

Bento007 (changes requested):

bmccandless (approved):
LGTM

backend/chalice/api_server/Makefile
```diff
@@ -40,6 +40,7 @@ package: build-chalice-config
 	mkdir -p chalicelib/corpora
 	cp -R ../../corpora/common chalicelib/corpora
 	cp -R ../../corpora/lambdas chalicelib/corpora
+	cp -R ../../corpora/dataset_processing chalicelib/corpora
```
Bento007, line 43: do you still need to include data_processing?

MDunitz: oh good catch-- removed

backend/corpora/common/utils/aws.py
```diff
@@ -89,3 +89,14 @@ def _restore(self):
...
         if self.debug_logging:
             print(message)
+
+
+def delete_many_from_s3(bucket_name: str, dataset_uuid: str) -> None:
+    """
+    This deletes everything with a specific prefix from the given bucket
+
+    """
+    assert dataset_uuid
```
Bento007, line 99: Don't use assert in production code. Use and `if not dataset_uuid: raise ValueError` in this case.  Using asserts in production code can make debug difficult since python removes asserts when converting to when it is compiled. See https://docs.python.org/3/using/cmdline.html#cmdoption-o. We don't want to accidentally delete the contents of a  bucket because someone used the `-o` flag before deploying.

backend/corpora/dataset_processing/download.py
```diff
@@ -4,16 +4,18 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 class ProgressTracker:
-    def __init__(self, file_size: int):
+    def __init__(self, file_size: int, artifact_bucket: str, cellxgene_bucket: str):
         self.file_size: int = file_size
         self._progress: int = 0
+        self.artifact_bucket = artifact_bucket
```
Bento007, line 17: you also don't need the buckets anymore since you aren't removing anything from s3.

backend/corpora/dataset_processing/download.py
```diff
@@ -74,6 +75,27 @@ def updater(processing_status_uuid: str, tracker: ProgressTracker, frequency: fl
...
+                if curr_status is UploadStatus.CANCEL_PENDING:
+                    logger.info(f"cancelling the upload for {curr_status.dataset.id}")
+                    # set db to cancelled
+                    status = {
+                        DbDatasetProcessingStatus.upload_progress: 0,
+                        DbDatasetProcessingStatus.upload_status: UploadStatus.CANCELED,
+                        DbDatasetProcessingStatus.upload_message: "Canceled by user",
+                    }
+
+                    dataset = Dataset.get(curr_status.dataset.id)
```
Bento007, line 90: remove line 90-91. The database will be updated at the bottom of this function on line 122.

Bento007 (changes requested):
A few more things to fix

backend/corpora/dataset_processing/download.py
MDunitz: wont status get overwritten by one of the following if statements?

backend/corpora/dataset_processing/process.py
```diff
@@ -119,7 +119,9 @@ def update_db(metadata=None, processing_status=None):
         dataset.update(processing_status=status)
 
 
-def download_from_dropbox_url(dataset_uuid: str, dropbox_url: str, local_path: str) -> str:
+def download_from_dropbox_url(
```
Bento007, line 122: are the artifact bucket names used at all?

backend/corpora/dataset_processing/download.py
Bento007: you make a good point. You can shorten the scope of the db_session_manager context to only retrieve the curr_status. Then everything will be at the same level as the main if statement. Then add `if curr_status is UploadStatus.CANCEL_PENDING:` as the first if statement. I think this will also resolve your need for the `try except` block.

Bento007 (changes requested):
a few more things

Bento007 (approved):
LGTM

PR #805, opened 2021-01-11
lambda upload failures support for happy path

srm78:
Implement lambda upload failures support. Verified lambda is invoked and s3 objects getting cleaned up.


backend/corpora/dataset_processing/process.py
```diff
@@ -226,8 +226,6 @@ def main():
         os.environ["DATASET_ID"],
         os.environ["DROPBOX_URL"],
         "local.h5ad",
-        os.environ["ARTIFACT_BUCKET"],
-        os.environ["CELLXGENE_BUCKET"],
```
srm78, line 230: These lines are added by #799. the called function `download_from_dropbox_url` is only taking 3 arguments resulting in a failure and also not using these parameters. for now I have removed these.

jgadling (approved):

backend/corpora/common/utils/aws.py
```diff
@@ -98,6 +98,6 @@ def delete_many_from_s3(bucket_name: str, dataset_uuid: str) -> None:
     """
     if not dataset_uuid:
         raise ValueError
-    s3 = boto3.resource("s3")
+    s3 = boto3.resource("s3", endpoint_url=os.getenv("BOTO_ENDPOINT_URL"))
```
MDunitz, line 101: will this cause errors if BOTO_ENDPOINT_URL isnt set?

MDunitz (approved):
one small question, otherwise lgtm

backend/corpora/common/utils/aws.py
jgadling: Nope, os.getenv() will return None if that env var isn't set.

PR #809, opened 2021-01-12
Thuang feedback 675 frontend dataset view

tihuan:
#### Reviewers
**Functional:** 
@seve 

**Readability:** 
@MDunitz 

**UX:**
@hthomas-czi 

---

## Changes
- add
- remove
- modify

The GIF shows that the app will poll for collection metadata until it's populated!

demo

Also sorting datasets from old to new:

Screen Shot 2021-01-12 at 12 37 09 PM

### DIFF
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -375,7 +375,7 @@ paths:
               properties:
                 name:
                   type: string
-                  description: name of the collecton
+                  description: name of the collection
                 description:
                   type: string
                   description: description of the collection
@@ -465,7 +465,7 @@ paths:
         If a new link is provided, the new link will be used. If the collection is not in an error state, this endpoint
         will return an error.
       security:
-        - cxguserCookie: [ ]
+        - cxguserCookie: []
       operationId: corpora.lambdas.api.v1.collection_uuid.upload.link
       parameters:
         - $ref: "#/components/parameters/path_collection_uuid"
@@ -546,7 +546,7 @@ paths:
       summary: Get the status of a dataset upload.
       operationId: corpora.lambdas.api.v1.dataset.get_status
       security:
-        - cxguserCookie: [ ]
+        - cxguserCookie: []
       parameters:
         - $ref: "#/components/parameters/path_dataset_uuid"
       responses:
@@ -561,7 +561,16 @@ paths:
                     $ref: "#/components/schemas/dataset_uuid"
                   upload_status:
                     type: string
-                    enum: [WAITING, UPLOADING, UPLOADED, FAILED, CANCEL PENDING, CANCELED, NA]
+                    enum:
+                      [
+                        WAITING,
+                        UPLOADING,
+                        UPLOADED,
+                        FAILED,
+                        CANCEL PENDING,
+                        CANCELED,
+                        NA,
+                      ]
                   upload_message:
                     type: string
                   upload_progress:
@@ -582,7 +591,7 @@ paths:
                     enum: [CONVERTING, CONVERTED, FAILED, NA]
                   conversion_rds_status:
                     type: string
-                    enum: [ CONVERTING, CONVERTED, FAILED, NA ]
+                    enum: [CONVERTING, CONVERTED, FAILED, NA]
         "400":
           $ref: "#/components/responses/400"
         "401":
--- a/frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
+++ b/frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
@@ -1,197 +1,40 @@
-import {
-  Button,
-  Checkbox,
-  Classes,
-  Colors,
-  Icon,
-  Intent,
-  PopoverInteractionKind,
-  Spinner,
-  Tooltip,
-} from "@blueprintjs/core";
+import { Button, Checkbox, Intent } from "@blueprintjs/core";
 import { IconNames } from "@blueprintjs/icons";
 import loadable from "@loadable/component";
 import React, { FC, useEffect, useState } from "react";
-import { useQueryCache } from "react-query";
-import {
-  CONVERSION_STATUS,
-  Dataset,
-  DatasetUploadStatus,
-  UPLOAD_STATUS,
-  VALIDATION_STATUS,
-} from "src/common/entities";
+import { QueryCache, useQueryCache } from "react-query";
+import { Dataset, VALIDATION_STATUS } from "src/common/entities";
 import {
   useDatasetStatus,
   USE_DATASET_STATUS,
 } from "src/common/queries/datasets";
+import { aggregateDatasetsMetadata } from "src/components/Collections/components/Grid/common/utils";
 import {
   DetailsCell,
-  LeftAlignedDetailsCell,
   RightAlignedDetailsCell,
   StyledRow,
-} from "src/components/Collections/components/Grid/common/style";
-import { aggregateDatasetsMetadata } from "src/components/Collections/components/Grid/common/utils";
+} from "src/components/Collections/components/Grid/components/Row/common/style.ts";
 import { UploadingFile } from "src/components/DropboxChooser";
 import DatasetUploadToast from "src/views/Collection/components/DatasetUploadToast";
-import { DatasetStatusTag, StyledAnchor, TitleContainer } from "./style";
+import CellCount from "./components/CellCount";
+import Popover from "./components/Popover";
+import UploadStatus from "./components/UploadStatus";
+import { TitleContainer } from "./style";
+import { checkIfFailed, checkIfLoading, FailReturn } from "./utils";
 
-const AsyncPopover = loadable(
+const FETCH_COLLECTION_INTERVAL_MS = 5 * 1000;
+
+const AsyncTooltip = loadable(
   () =>
-    /*webpackChunkName: 'CollectionRow/components/Popover' */ import(
-      "src/components/Collections/components/Grid/components/Popover"
+    /*webpackChunkName: 'Grid/Row/DatasetRow/Tooltip' */ import(
+      "./components/Tooltip"
     )
 );
 
-type FailReturn =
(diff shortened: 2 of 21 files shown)


backend/config/corpora-api.yml
```diff
@@ -375,7 +375,7 @@ paths:
               properties:
```
tihuan, line 375: prettier and spellchecker fixes!

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
 
-  const datasetStatus = queryResult.data ?? dataset.processing_status;
-  const isLoading = checkIfLoading(datasetStatus);
-  const hasFailed = checkIfFailed(datasetStatus);
+  const { upload_progress } = datasetStatus;
 
-  // `nameIsPopulated` is a temp-workaround for no "complete status"
-  // const isComplete = checkIfComplete(datasetStatus);
-  const nameIsPopulated = dataset.name !== undefined && dataset.name !== "";
+  const [uploadProgress, setUploadProgress] = useState(upload_progress);
```
tihuan, line 64: Set default `uploadProgress` to be `upload_progress` instead of `-1`, I think that works and saves us one `setState` cycle?

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
 
-  const name = nameIsPopulated ? dataset.name : file?.name ?? dataset.id;
+  if (queryResult.isError) console.error(queryResult.error);
 
-  useEffect(() => {
```
tihuan, line 238: Split 1 `useEffect` into 3 custom hooks, with focused responsibility

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
+  const isNamePopulated = Boolean(dataset.name);
 
-    if (hasFailed.isFailed || nameIsPopulated) {
-      queryCache.cancelQueries([USE_DATASET_STATUS, dataset.id]);
-    }
-  }, [
-    dataset.id,
-    datasetStatus.upload_progress,
-    hasFailed.isFailed,
+  const name = dataset.name || file?.name || dataset.id;
```
tihuan, line 70: I think this is equivalent to `const name = nameIsPopulated ? dataset.name : file?.name ?? dataset.id;` but easier to read?

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
+        {isLoading && (
+          <UploadStatus
+            isValidating={
+              datasetStatus.validation_status === VALIDATION_STATUS.VALIDATING
+            }
+            progress={datasetStatus.upload_progress}
+          />
+        )}
       </DetailsCell>
-      {conditionalPopover(tissue, isLoading, hasFailed.isFailed)}
```
tihuan, line 271: Change helper function to Component for React convention

tihuan (commented):
Still adding more comments!

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
-        {nameIsPopulated && (
+        {isNamePopulated && (
           <Button intent={Intent.PRIMARY} outlined text="Explore" />
         )}
       </RightAlignedDetailsCell>
     </StyledRow>
   );
 };
 
+function useCheckCollectionPopulated({
```
seve, line 137: Is this supposed to be checkCollectionPopulated or checkDatasetPopulated?

Why are we invalidating the collection query on an interval?

seve: Oh I see, this is a workaround not having a signifier that we're done with the upload process and that the metadata is populated.

Once the backend is implemented, the useEffect will be called by an update to the datasetStatus and the interval won't be needed. Correct?

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
+  isNamePopulated: boolean;
+  upload_progress: number;
+}) {
+  useEffect(() => {
+    let intervalId: number | undefined = undefined;
+
+    if (!intervalId && upload_progress === 1 && !isNamePopulated) {
+      intervalId = window?.setInterval(() => {
+        if (upload_progress === 1 && !isNamePopulated) {
+          invalidateCollectionQuery();
```
seve, line 152: Instead of using `invalidateCollectionQuery` we might be better off calling the `refetch` function that `useQuery` returns

https://react-query-v2.tanstack.com/docs/api#usequery
image

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
seve: Yep! The logic I attached to `nameIsPopulated` was unnecessary as we have type checking :)

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
tihuan: I think BE is not planning to implement the complete status we were hoping for, and want us to just rely on checking for `isNamePopulated`!

frontend/src/views/Collection/index.tsx
```diff
@@ -102,6 +103,13 @@ const Collection: FC<Props> = ({ id = "" }) => {
   const datasetPresent =
     collection.datasets?.length > 0 || Object.keys(uploadedFiles).length > 0;
 
+  const invalidateCollectionQuery = memoize(
```
tihuan, line 106: We need to memoize `invalidateCollectionQuery` here, since `<DatasetsGrid />` is shallow diffing (`===`) it to decide whether it needs to re-render

If we don't memoize the function, the following code will always generate a new function and therefore `props.invalidateCollectionQuery === prevProps.invalidateCollectionQuery` will always be `false`

```tsx
invalidateCollectionQuery={() => {
              queryCache.invalidateQueries([USE_COLLECTION, id, visibility]);
            }}
```

frontend/src/components/common/theme.ts
```diff
@@ -50,3 +50,11 @@ export enum BLUE {
   E = "#73B7FF",
   F = "#DFEFFF",
 }
+
+export enum RED {
```
tihuan, line 54: We can't use Blueprint's `Colors` object, since Blueprint only has 5 colors and our `BLUE` has 6 colors!

frontend/src/views/Collection/index.tsx
```diff
@@ -1,6 +1,7 @@
 import { Button, Classes, H3, Intent } from "@blueprintjs/core";
 import { IconNames } from "@blueprintjs/icons";
 import { RouteComponentProps } from "@reach/router";
+import { memoize } from "lodash-es";
```
tihuan, line 4: we use `lodash-es` for better tree shaking

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/utils.ts
```diff
@@ -0,0 +1,50 @@
+import {
```
tihuan, line 1: Split util functions out into its own file for readability

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -207,79 +50,155 @@ const DatasetRow: FC<Props> = ({
...
+        }
+      }, FETCH_COLLECTION_INTERVAL_MS);
+    }
+
+    return () => clearInterval(intervalId);
+  }, [invalidateCollectionQuery, isNamePopulated, upload_progress]);
+}
+
+function useCancelDatasetStatusQuery({
```
tihuan, line 161: Keeping all custom hooks via function declaration at the bottom for keeping critical information about this file "above the fold"

tihuan:
@seve @MDunitz @hthomas-czi just updated PR per your comments and added all the explanations I think are needed. PTAL again! Thank you! 🙏 🔥

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
seve: Wouldn't `cxg_conversion_status` be the substitute?

tihuan: I asked BE this question in [Slack](https://chanzuckerbergteam.slack.com/archives/C01CPJURT0S/p1609791352221300), and Marcus suggested `title` or I think he meant `name`, since it's a mandatory field:

https://chanzuckerbergteam.slack.com/archives/C01CPJURT0S/p1609795978226100?thread_ts=1609791352.221300&cid=C01CPJURT0S

So we can continue using `isNamePopulated` for checking if the dataset's metadata have been populated

For whether to show `Explore` button or not, that will be based on `cxg_conversion_status`!

seve (approved):

tihuan:
Thanks for the quick review, @seve ! I'll merge this into your branch and get final LGTM from @hthomas-czi there 👍

PR #820, opened 2021-01-15
Let SQLAlchemy handle generating the uuid for new rows

Bento007:
#### Reviewers
**Functional:** @MDunitz 

**Readability:** @mckinsel 

---

## Changes
- modify where the primary key is generated for new rows in the database. 
- remove unused code


backend/corpora/common/entities/dataset.py
```diff
@@ -51,23 +42,20 @@ def create(
...
-            ),
-            processing_status=cls._create_sub_object(
-                processing_status, DbDatasetProcessingStatus, add_columns=dict(dataset_id=primary_key)
-            ),
             **kwargs,
         )
-
-        cls.db.session.add(new_db_object)
-        cls.db.session.flush()
+        if artifacts:
```
mckinsel, line 47: Not a big deal, but do you even need the `if artifacts` here?

backend/corpora/common/entities/dataset.py
```diff
@@ -86,31 +74,21 @@ def update(
...
                 for af in self.artifacts:
                     self.db.delete(af)
-                new_db_objects = self._create_sub_objects(
-                    artifacts, DbDatasetArtifact, add_columns=dict(dataset_id=self.id)
-                )
-                self.db.session.add_all(new_db_objects)
-                kwargs["artifacts"] = new_db_objects
+                new_objs = [DbDatasetArtifact(dataset_id=self.id, **art) for art in artifacts]
+                self.db.session.add_all(new_objs)
+                kwargs["artifacts"] = new_objs
```
mckinsel, line 79: Why are you updating `kwargs` here?

mckinsel (approved):
Left a couple questions, but LGTM and congrats on net code deletion

backend/corpora/common/entities/dataset.py
Bento007: No we don't

Bento007: it turns out it is needed otherwise we might try iterating over None.

backend/corpora/common/entities/dataset.py
Bento007: it can be removed. I was trying to avoid making lots of extra changes

MDunitz (approved):
lgtm
discussed moving id into base over slack, decided to do that in separate pr

PR #832, opened 2021-01-19
Seve/706 fe delete cancel upload

seve:
#### Reviewers
**Functional:**  @tihuan 

**Readability:** @Bento007 

---

## Changes
* `frontend/src/common/API.ts` + `frontend/src/common/entities.ts` + `frontend/src/common/queries/common.ts`
  * Add consts and enums for dataset deletion
* `frontend/src/common/queries/datasets.ts`
  * Add dataset deletion query and query function with query cancellation and invalidation
* `frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/components/UploadStatus/index.tsx` + `frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx`
  * Add cancellation button
* `frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/utils.ts`
  * Add util for checking cancellation state
* `frontend/src/views/Collection/index.tsx`
  * Add dataset deletion button with alert
* `frontend/src/components/Collections/components/Grid/components/DatasetsGrid/index.tsx` + `frontend/src/components/Collections/components/Grid/common/style.ts` + `frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/components/UploadStatus/style.ts` + `frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/components/common/Skeleton.tsx` + `frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/style.ts` + `frontend/src/components/Collections/components/Grid/components/Row/common/style.ts` + `frontend/src/views/Collection/style.ts`
  * Style changes/additions for cancellation button

---

image

image

image


Bento007 (approved):
LGTM on readability

frontend/src/views/Collection/index.tsx
```diff
@@ -76,6 +78,10 @@ const Collection: FC<Props> = ({ id = "" }) => {
 
   const [mutate] = useCollectionUploadLinks(id, visibility);
 
+  const [isAlertOpen, openAlert] = useState(false);
+
+  const [deleteDataset] = useDeleteDataset(collection?.id ?? "");
```
tihuan, line 83: is there a reason we need `??` instead of just `||` here 😄 ?

frontend/src/views/Collection/index.tsx
```diff
@@ -144,14 +149,31 @@ const Collection: FC<Props> = ({ id = "" }) => {
...
+      <StyledAlert
+        cancelButtonText={"Cancel"}
+        confirmButtonText={"Delete Dataset"}
+        intent={Intent.DANGER}
+        isOpen={isAlertOpen}
+        onCancel={() => {
+          openAlert(!isAlertOpen);
+        }}
+        onConfirm={() => {
+          deleteDataset(selected ?? "");
```
tihuan, line 170: Also perhaps `deleteDataset()` could loosen its argument to accept an optional string, so we don't need to pass `""` at callsite?

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -65,7 +71,8 @@ const DatasetRow: FC<Props> = ({
 
   const [uploadProgress, setUploadProgress] = useState(upload_progress);
 
-  if (queryResult.isError) console.error(queryResult.error);
+  if (queryResult.isError && !(queryResult.error instanceof CancelledError))
```
tihuan, line 74: For readability, if the condition and its block span over one line, can we add curly braces please?

E.g.,

```ts
if (queryResult.isError && !(queryResult.error instanceof CancelledError)) {
  console.error(queryResult.error);
}
```

tihuan (approved):
Looks super good! Thanks so much for the detailed PR description, @seve 🤩 🥇 🎉 

Just some small nits, otherwise LGTM! 👍

frontend/src/views/Collection/index.tsx
seve: Just to be more specific that we're setting it to "" when collection is nullish.  Would you prefer:
```suggestion
  const [deleteDataset] = useDeleteDataset(collection?.id || "");
```

frontend/src/views/Collection/index.tsx
seve: Technically, `deleteDataset()` should never be called with no collection attribution as the delete action can only be enacted when there is a collection.  The nullish coalescing is just to meet the needs at compile time

tihuan: Ah yeah, I'm thinking it's probably better to handle `""` or `undefined` in `deleteDataset()`, otherwise by passing `""` to the function and not handle the edge case, we would end up sending a malformed request to BE?

e.g., 

```ts
async function deleteDataset(
  dataset_uuid?: string // update type here to allow undefined to come through
): Promise<DatasetUploadStatus> {
  // handle edge case here
  if (!dataset_uuid) return Promise.reject('`dataset_uuid` is not valid')

  // real API call
}
```

frontend/src/views/Collection/index.tsx
tihuan: Hmm yeah personally when I see `??`, it makes me wonder what other possible falsey values `collection?.id` could have that we **don't** want to use the fallback value `""`. Such as `0`, `false`, and `""`.

But in our case, I think the intention is to use the fallback value for all falsey values, thus `||`?

seve:
Gross commit name 😐

PR #836, opened 2021-01-20
Various cleanup of rdev script

mbarrien:
* Use boto3 session instead of setting env var
* Instantiate cloudformation client in a single place
* Add ctx as argument of EnvMeta
* Remove unused variables


jgadling (approved):
🙇

PR #841, opened 2021-01-22
Remove DBUtil

Bento007:
#### Reviewers
**Functional:** @MDunitz 

**Readability:** @tihuan   

---

## Changes
- DBSessionMaker create a single sessionMaker
- An Entity class method must have an open session passed into it.
- Remove DbUtils from Entity
- Use the session that tracks the db_object in place of DbUtils
- Use DbSessionMaker to create a new session for db_session_manager
- Pass through SQLAlchemy session flags to session_maker
- processing_status_updater requires a session to be passed in.
- Use db_session_manager for tighter database session management in the processing container
- generate_collection, and generate_dataset take a session parameter.
- GenerateDataMixin is inherited by DataPortalTestCase
- Remove DBUtils


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -1,16 +1,16 @@
...
 from ....common.corpora_orm import DbCollection, CollectionVisibility
-from ....common.utils.db_utils import db_session
 from ....common.entities import Collection
 from ....common.utils.exceptions import ForbiddenHTTPException
 
 
-@db_session()
 def get_collections_list(from_date: int = None, to_date: int = None, user: Optional[str] = None):
-
+    session = g.db
```
MDunitz, line 11: instead of importing g from flask what if you created a get_db function and added it to the app?
Something like
```
def get_db():
    if 'db' not in g:
        g.db = connect_to_database()

    return g.db
```
https://flask.palletsprojects.com/en/1.1.x/appcontext/

Bento007: I saw that, and did that for a while. I decided to remove it because it added another path for managing a session other than the existing `db_session_manager`. Using the `existing db_session_manager` adds less redundant code to the project. I also couldn't decide where in the project a `get_db` function should live.

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -32,26 +32,27 @@ def get_collections_list(from_date: int = None, to_date: int = None, user: Optio
...
-            access_type = "WRITE"
-        elif visibility != "PUBLIC":
-            raise ForbiddenHTTPException()
-        else:
-            access_type = "READ"
-        result = collection.reshape_for_api()
-        result["access_type"] = access_type
-        return make_response(jsonify(result), 200)
-    else:
+    session = g.db
```
mckinsel, line 36: So `session` means one thing in flask and another thing in sqlalchemy. It could a little confusing pull the sqlalchemy meaning out of a flask variable.

Bento007: Since we are not using a flask session, I thought it was ok to use that variable name. I can rename it to db_session for clarity.

backend/chalice/api_server/app.py
```diff
@@ -108,7 +109,9 @@ def dispatch(*args, **kwargs):
             data=req_body,
             environ_base=app.current_request.stage_vars,
         ):
-            flask_res = flask_app.full_dispatch_request()
+            with db_session_manager() as session:
+                g.db = session
```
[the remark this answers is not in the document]
Bento007: The context isn't assigned to `g.db`. The output of the context manager is assigned, which is a database session. `__exit__` will close the session.

backend/corpora/lambdas/api/v1/collection.py
MDunitz: I think it would be ok to put it in app.py

I think adding it as a function will improve readability. Right now g.db is all over the codebase and I think its not immediate clear what it is doing. Because its in so many places I think it makes more sense to rename it than to add comments about what it is doing

Bento007: it can't live in app, unless we put in where `db_session_manager` is. `flask.g` only exist within the context of `flask_app.test_request_context`.

tihuan (approved):
LGTM for readability! Thanks for adding all the tests too, @Bento007 🤩 🎉 🚀

backend/corpora/lambdas/api/v1/collection.py
MDunitz: discussed further over slack. Summary -- its not ideal but its the best of bad options. In the future we may refactor to use flask-sqlalchemy or fast api

MDunitz (approved):
lgtm

PR #843, opened 2021-01-22
Feature Branch for Delete Collection

Bento007:


frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import { goToPage } from "tests/utils/helpers";
+import { getText } from "tests/utils/selectors";
+
+const describeIfCalledByDevEnv = TEST_ENV === "dev" ? describe : describe.skip;
+
+describeIfCalledByDevEnv("CRUD Collection", async () => {
+  it("creates and deletes a collection", async () => {
+    await goToPage(TEST_URL);
+    // LOG IN
+    await page.click(getText("Log In"));
```
tihuan, line 11: do we need to type email and password?

Also it'd be great to extract the `login()` function as a utility function for reusability like we have in Explorer 🙏 

Thank you!

frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
+import { TEST_ENV, TEST_URL } from "tests/common/constants";
```
tihuan, line 1: maybe the file name can just be `Collection.test.ts`? Adding `crud` makes me wonder what non-CRUD tests we're expecting to write and if they should live in a separate file or just in the same file. Personally it feels all Collection related tests can live in this file, whether it's CRUD or not. What do you think?

Thank you!

frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    // LOG IN
+    await page.click(getText("Log In"));
+    await page.waitForNavigation({ url: TEST_URL, waitUntil: "networkidle" });
+    // Create collection
+    await page.click(getText("Create Collection"));
+    await page.fill("#name", "TEST COLLECTION");
+    await page.fill("#description", "TEST DESCRIPTION");
+    await page.fill("#contact-name", "TEST");
+    await page.fill("#contact-email", "TEST@example.coms");
+    await page.click("input[type=checkbox]");
```
tihuan, line 19: For robustness and readability, maybe we can add a `test-id` attribute to the checkbox and click on that, instead of relying on the form currently having only one checkbox to click? Otherwise this line won't work if the form gets a second checkbox. Also using  `test-id` attribute make this line more readable, because people will then know we're clicking on the ToS element!

frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    const [request, response] = await Promise.all([
```
tihuan, line 20: great use of `Promise.all()`!

frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      page.waitForEvent("request"),
+      page.waitForEvent("response"),
+      await page.click(getText("Create")),
+    ]);
+
+    const cookie = request.headers().cookie;
+
+    const collectionID = (await response.json()) as object;
```
tihuan, line 28: `(await response.json())` is an `object` not a `string`?

frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    await expect(page).toHaveSelector("h3");
```
tihuan, line 30: For robustness and readability, let's also add `test-id` attribute for the element we're checking here too. `h3` doesn't convey the meaning and intention of why checking that element is necessary in this test. Also whatever that element is, if the design decides to change it to `h4` or `h3`, this test will break 🙏

frontend/tests/features/crudCollection.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    // CLICK DELETE COLLECTION BUTTON HERE
+    // TEMP: FETCH ENDPOINT TO DELETE COLLECTION
```
tihuan, line 32: nice comment!

tihuan:
@Bento007 did you mean to assign @seve to this instead of me 😄 ? Just wanted to double check, since I see you and Seve are working on this branch. Thank you!

Issue #850, opened 2021-01-25
Download button is enabled for failed uploads


brianraymor:
I uploaded `NaN` which failed as expected. 

The Download button is enabled in both the collection view and on the landing page, but the Explore button is correctly not available in both cases.

![Screen Shot 2021-01-25 at 11.26.37 AM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/b8d90d8c-ee51-4a9f-8c7a-2908fe02bef6)

![Screen Shot 2021-01-25 at 11.23.24 AM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/d4f42197-014f-42db-a9f2-f958a090870f)

tihuan:
Actually this is a quick fix! I'll work on that now. Thanks Brian 🎉!

PR #852, opened 2021-01-25
thuang-850-fix-download-button

tihuan:
#850 

#### Reviewers
**Functional:** 
@seve 

**Readability:** 
@MDunitz 

---

## Changes
- add
  - Add selector `hasAssets()` to only enable the Download button if `true`
  - Add new folder and file `frontend/src/common/modules/datasets/selectors.ts `

- modify
  - Disable download button if no asset is available for download

Homepage:
Screen Shot 2021-01-25 at 1 26 42 PM

demo

Collection Page:
demo

UPDATE (1/26):
New disabled style:

Screen Shot 2021-01-26 at 1 44 48 PM

### DIFF
--- a/frontend/src/common/modules/datasets/selectors.ts
+++ b/frontend/src/common/modules/datasets/selectors.ts
@@ -1,0 +1,7 @@
+import { Dataset } from "src/common/entities";
+
+export function hasAssets(dataset?: Dataset) {
+  if (!dataset) return false;
+
+  return Boolean(dataset?.dataset_assets?.length);
+}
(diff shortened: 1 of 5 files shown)


tihuan:
PTAL all! Thank you!

hthomas-czi (changes requested):
That looks like the default button style rather than the [disabled button style](https://blueprintjs.com/docs/#core/components/button) to me.

tihuan:
Ooh!! That's a super great point! Thank you for pointing that out. I could sense that it didn't look right, but I couldn't tell why. Now I know lol

I followed how Blueprint does with disabled primary button and came up with this, does that work? Thank you!

Screen Shot 2021-01-26 at 1 44 48 PM

hthomas-czi (approved):
Beautiful! 🎉

tihuan:
Yay! Thanks so much, Harley!!

PR #853, opened 2021-01-25
Exiting if upload fails or is cancelled

Bento007:
#### Reviewers
**Functional:** @MDunitz 

**Readability:** @mckinsel 

---

## Changes
- Add exceptions to catch when the processing container fails or processing is canceled.
  - Return 0 if  upload was canceled this will end the upload.
  - Return 1 if upload failed, this will restart the processing container.


backend/corpora/dataset_processing/exceptions.py
```diff
@@ -0,0 +1,8 @@
+class ProcessingCancelled(Exception):
+    def __init__(self, status, *args, **kwargs) -> None:
```
mckinsel, line 2: Can't you just say `pass` here?

backend/corpora/dataset_processing/process.py
```diff
@@ -373,12 +376,20 @@ def process_cxg(local_filename, dataset_id, cellxgene_bucket):
...
+        local_filename = download_from_dropbox_url(
+            dataset_id,
+            os.environ["DROPBOX_URL"],
+            "local.h5ad",
+        )
+    except ProcessingCancelled as ex:
+        logging.info(ex.status)
+        sys.exit(0)
+    except ProcessingFailed as ex:
+        logging.error(ex.status)
```
mckinsel, line 389: What does this do with the actual error and traceback?

Bento007: the traceback and error are in ex.status.upload_message which are populated by the download code.

MDunitz (approved):
lgtm!

mckinsel (approved):
One tiny comment, but not really a big deal either way.

PR #858, opened 2021-01-26
Refactor remote-dev to use Terraform instead of CloudFormation

jgadling:
This is a semi-rewrite of remote dev environment handling. This is a *breaking change* in that all old remote-dev environments will become unmanaged after merging this, but we feel this is a worthwhile tradeoff:

- We're now using Terraform instead of CloudFormation to manage infrastructure. Terraform is easier to debug and we have a lot of support for it at CZI.
- Terraform code for dev stacks is now committed right alongside the code. Make changes to terraform job definitions or lambdas and just run `./scripts/happy update <myenv>` to apply!
- Terraform creates/updates environments faster than CloudFormation

Also, the name of the remote-dev utility has changes. It's now `./scripts/happy` rather than `./scripts/rdev`

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


Bento007:
Does this mean if we make a terraform change in single-cell-infra we will need to make the change here as well?

jgadling:
> Does this mean if we make a terraform change in single-cell-infra we will need to make the change here as well?

At the moment since there are two different sets of infrastructure - existing lambdas and Happy infra, yes.

In the future we're hoping the Happy infra will be the primary deployment platform, and the majority of infra changes will be handled by this repo. Some long-lived services (RDS, s3 buckets, etc) and access control (IAM roles, etc) will probably continue to be managed by single-cell-infra over the long term, but anything directly related to this service and its deployment will live here.

mbarrien (approved):

PR #859, opened 2021-01-26
thuang-update-dropbox-app-key and favicon

tihuan:
#### Reviewers
**Functional:** 
@seve 

**Readability:** 
@Bento007 
---

## Changes
- add
  - favicon support -- it only works locally now and not in prod mode -.- I've done an hour worth of research and still can't figure it out. So maybe something to work on later!
- remove
- modify
  - Use my work Dropbox account's app key
 
Screen Shot 2021-01-26 at 12 47 56 PM

PTAL thank you!


seve (approved):
LGTM

Issue #827, opened 2021-01-16
Delete a Private Collection


tihuan:
Context:

This ticket will allow a publisher to delete their private collections

[Design](https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Publisher-Collection?node-id=102%3A2384)

What:

1. On a private collection page, the user sees a red delete button
1. Clicking on the red delete button shows a Blueprint `Alert` component that confirms if a user wants to proceed with such deletion
1. Clicking "Delete Collection" button will send an API call to delete, and the UI will optimistically update the FE to pretend the deletion is complete and proceed with navigating a user to their "My Collections" page
1. Clicking "Cancel" button will close the modal
1. Write E2E tests

tihuan:
This works!

PR #860, opened 2021-01-26
Delete Collection Button

seve:
#827

#### Reviewers
**Functional:**  

@tihuan 

**Readability:** 

@Bento007 
---

## Changes
- add
  - New Component `DeleteCollection` to house the delete button and confirmation alert
  - New React Query methods to mutate the collection and send the DELETE request 
  - Adds integration testing for collection deletion
- ~remove~
- modify
  - Adds delete, share, and publish buttons to collection page
  - Moves Styled Alert to a more general directory

---

image

image


frontend/src/components/Collections/components/Collection/common/components/DeleteCollection/index.tsx
```diff
@@ -0,0 +1,57 @@
+import { Button, H6, Intent } from "@blueprintjs/core";
```
tihuan, line 1: ah I'm thinking `/DeleteCollection` can just live at `src/components/Collections/components/DeleteCollection` to prevent too much nesting. What do you think? 

Thank you!

seve: That works with me!

frontend/src/components/Collections/components/DeleteCollection/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface Props {
+  id: Collection["id"];
+}
+
+const DeleteCollection: FC<Props> = ({ id }) => {
+  const [deleteMutation] = useDeleteCollection(id);
+
+  const handleDelete = () => {
+    deleteMutation(id);
```
tihuan, line 18: I'm assuming we need to await here until the DELETE call succeeds before we navigate, otherwise we might still see the collection on My Collections page?

e.g., 

```ts
const handleDelete = async () => {
  await deleteMutation(id);
  ...
}
```

frontend/tests/features/collection.test.ts
```diff
@@ -0,0 +1,57 @@
+import { ROUTES } from "src/common/constants/routes";
+import { apiTemplateToUrl } from "src/common/utils/apiTemplateToUrl";
+import { API_URL } from "src/configs/configs";
+import { TEST_ENV, TEST_URL } from "tests/common/constants";
+import { goToPage } from "tests/utils/helpers";
+import { getText } from "tests/utils/selectors";
+
+const describeIfCalledByDevEnv = TEST_ENV === "dev" ? describe : describe.skip;
+
+/*  eslint-disable sort-keys */
```
tihuan, line 10: can we just sort the keys alphabetically so we have consistency?

frontend/tests/features/collection.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  contactName: "TEST NAME",
+  contactEmail: "TEST@example.com",
+};
+/* eslint-enable sort-keys */
+
+describeIfCalledByDevEnv("Collection", async () => {
+  it("creates and deletes a collection", async () => {
+    await goToPage(TEST_URL);
+    // LOG IN
+    await page.click(getText("Log In"));
```
tihuan, line 23: do we need to type email and password to log in?

tihuan (changes requested):
Overall looks great! Have you tested Trent's mocked API? (I only tested the FE part and didn't run Local Happy Path to spin up local BE)

Thank you!

frontend/tests/features/collection.test.ts
seve: This is assuming we're using Trent's mock auth which will just give us an auth token with no user input needed

seve:
> Have you tested Trent's mocked API? (I only tested the FE part and didn't run Local Happy Path to spin up local BE)

@tihuan: Wasn't working for me yet, have a slack conversation with him rn

frontend/tests/features/collection.test.ts
tihuan: Ahh I see! That's super awesome!

tihuan (approved):
LGTM! Thanks so much, @seve 🎉 ✨ 🚀 !!

frontend/src/components/Collections/components/DeleteCollection/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface Props {
+  id: Collection["id"];
+}
+
+const DeleteCollection: FC<Props> = ({ id }) => {
+  const [deleteMutation] = useDeleteCollection(id);
+
+  const handleDelete = async () => {
+    await deleteMutation(id);
+    navigate(ROUTES.MY_COLLECTIONS);
```
tihuan, line 19: ah actually now thinking about it, `await deleteMutation(id);` doesn't guarantee `deleteMutation()` succeeds, because React Query catches the error already, so `deleteMutation()` won't throw. 

So we might need to use:

```ts
import { useNavigate } from "@reach/router";
...
const [deleteMutation, { isSuccess }] = useDeleteCollection(id);
const navigate = useNavigate()

if (isSuccess) {
  navigate(ROUTES.MY_COLLECTIONS)
}
```

tihuan: @seve just wanted to double check you saw this 😆 Thank you!

seve: Oh! Totally right, I'll implement this now

tihuan (approved):
Awesome thanks for the quick change!

tihuan:
@seve I force pushed your branch after updating both the feature branch and your branch 🙏 Thank you!

Bento007 (approved):
LGTM

PR #862, opened 2021-01-27
fix backend unit tests

srm78:
1. Fix backend unit tests.
2. Updated make target to run individual test(s). Updated local dev readme on examples of how to run all/some/individual test cases.


mbarrien (approved):

jgadling (approved):

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -64,6 +64,22 @@ def setup_s3_bucket(self, bucket_name):
         )
         return s3
 
+    def delete_s3_bucket(self, bucket_name):
+        # Mock S3 service if we don't have a mock api already running
+        if os.getenv("BOTO_ENDPOINT_URL"):
```
Bento007, line 69: Most of this code is duplicated from `setup_s3_bucket`. If this is only used to clean up `setup_s3_bucket`, use [addCleanup](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addCleanup) in `setup_s3_bucket`. Some think like this

```python
def setup_s3_bucket(self, bucket_name):
   ...
   def _delete_s3_bucket(bn):
        bucket = s3.Bucket(bn)
        if bucket.creation_date is not None:
            bucket.objects.all().delete()
            bucket.delete()

   self.addCleanup(_delete_s3_bucket, bucket_name)
   return s3
```

Bento007 (changes requested):
Once change to make

Bento007 (approved):
LGTM

PR #863, opened 2021-01-27
Add a test to capture the case where the raw layer name is provided in the layer_description and not in X or raw.

maniarathi:
#### Reviewers
**Functional:** @mckinsel 

**Readability:** @MDunitz 

---

## Changes
- add
- remove
- modify


maniarathi:
lol I was reading an outdated main. Closing.

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -178,6 +178,74 @@ def list_equal(list1, list2, cmp_func):
...
+            ],
+            index=(str(i) for i in range(11)),
+        )
+        uns = {
+            "title": "my test dataset",
+            "organism": "Homo sapiens",
+            "organism_ontology_term_id": "NCBITaxon:8505",
+            "layer_descriptions": {"my_awesome_wonky_layer": "raw"},
+        }
+        adata = anndata.AnnData(X=X_df, obs=obs, uns=uns, layers={"my_awesome_wonky_layer": awesome_layer_df})
```
mckinsel, line 240: Do we expect `X_df` and `awesome_layer_df` have different nonzero genes per cell? It seems like they're both full, nonzero matrices.

maniarathi: Yep -- lines 184 and 185 initialize the `X_df` and `awesome_layer_df` with 2s and 3s respectively. Which is why I assert below that the `mean_genes_per_cell` is 3 and not 2.

mckinsel: Right, but `mean_genes_per_cell` is the mean count of nonzero genes per cell. Doesn't matter if it's 3, 2, or a million. I think the test is passing because the matrix has three columns.

maniarathi: Ohhhhhhhhhhhhh

maniarathi: Good catch -- will fix :P

mckinsel (approved):
LGTM, thanks very much!

PR #874, opened 2021-02-02
make target to retrieve cxguser cookie

srm78:
created new target `make local-cxguser-cookie`. This will printout `cxguser-cookie` for a user configured in the oidc mock server. we can use this cookie to make further requests.

Testing
1. Retrieved cxguser-cookie by running `make local-cxguser-cookie`
2. Used the above cookie and called successfully called multiple endpoints. Some sample examples called are in the script.


Bento007 (commented):
LGTM

Bento007 (approved):

login.py
```diff
@@ -0,0 +1,104 @@
...
+
+f = open(CLIENTS_CONFIGURATION_PATH)
+data = json.load(f)
+client_data = data[0]
+client_id = client_data['ClientId']
+client_secret = client_data['ClientSecrets'][0]
+
+
+# Calling http://localhost:5000/dp/v1/login
+session = requests.session()
```
jgadling, line 27: We're creating sessions, which *should* allow us to persist some cookies between requests, but then we're creating a new session for every request. Can we either use the session as it's intended, or drop it?

Bento007, line 27: Do you need to create a new session for every request?

login.py
```diff
@@ -0,0 +1,104 @@
...
+location = location.replace("https://localhost:8443", "https://oidc")
+response = session.get(location, verify=False, allow_redirects = False, cookies=login_cookies)
+location = response.headers['Location']
+
+# Calling Account/login
+session = requests.session()
+response = session.get(location, verify=False, allow_redirects = False, cookies = login_cookies)
+authorize_cookies = response.cookies
+
+# Parsining login page
```
Bento007, line 45:
```suggestion
# Parsing login page
```

login.py
```diff
@@ -0,0 +1,104 @@
...
+  'scope': 'openid profile email offline_access',
+  'state': params['state'],
+  'nonce': params['nonce'],
+  'code_challenge': params['code_challenge'],
+  'code_challenge_method': params['code_challenge_method'],
+  'response_type': params['response_type'],
+  'redirect_uri': params['redirect_uri']
+}
+
+combine_cookies = {}
```
jgadling, line 71: Can we use dict.update() here?

Bento007 (changes requested):
With a few changes, it'll be perfect. If you could make this into a python function that we could import, that would make it more usable for our backend unit tests.

login.py
```diff
@@ -0,0 +1,104 @@
...
+from urllib import parse
+
+requests.packages.urllib3.disable_warnings(requests.packages.urllib3.exceptions.InsecureRequestWarning)
+
+OIDC_LOGIN_URL="http://localhost:5000/dp/v1/login"
+USERS_CONFIGURATION_PATH="oauth/users.json"
+CLIENTS_CONFIGURATION_PATH="oauth/clients-config.json"
+
+f = open(USERS_CONFIGURATION_PATH)
+data = json.load(f)
```
jgadling, line 13: This script can't be imported by another module. Can we convert it to an executable module? (https://docs.python.org/3/tutorial/modules.html#executing-modules-as-scripts)

jgadling (changes requested):

login.py
```diff
@@ -0,0 +1,104 @@
...
+f = open(CLIENTS_CONFIGURATION_PATH)
+data = json.load(f)
+client_data = data[0]
+client_id = client_data['ClientId']
+client_secret = client_data['ClientSecrets'][0]
+
+
+# Calling http://localhost:5000/dp/v1/login
+session = requests.session()
+response = session.get(OIDC_LOGIN_URL, verify=False, allow_redirects = False)
```
jgadling, line 28: Can we run python's `black` formatting utility against this script to fix some of the formatting inconsistencies?

login.py
```diff
@@ -0,0 +1,80 @@
...
+        "redirect_uri": params["redirect_uri"],
+    }
+
+    # calling /Account/login with all the login details.
+    response = session.post(location, data=login_payload, verify=False, allow_redirects=False)
+
+    # Calling /connect/authorize/
+    location = response.headers["Location"]
+    response = session.get("https://oidc" + location, verify=False, allow_redirects=False)
+    location = response.headers["Location"]
```
jgadling, line 69:
```suggestion
```

srm78: Why remove??. This is the callback for the app and you need to call this location to get the cookie.

jgadling (approved):

Bento007 (approved):
LGTM

Issue #824, opened 2021-01-16
BE + FE Integration


tihuan:
What:

1. Given that Seve will complete the FE feature in [Upload Datasets Part 4](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/706undefined), we just need this integration ticket as a buffer to make sure the FE implementation works for both cancelling a converting dataset and deleting a converted dataset

2. Please add E2E test(s) to ensure the user flows work! (Please kindly make sure to add a dataset back after you delete one, so the account doesn't run out of datasets to delete!)

tihuan:
@seve I updated this from 3 points to 1 point, since we're just missing porting over the AuthN flow from Explorer and make it work with Happy Path mocked AuthN page right? Thank you!

PR #886, opened 2021-02-08
Enable auth on front end smoke tests

seve:
#### Reviewers
**Functional:** 
@tihuan 

**Readability:** 
@Bento007 

---

## Changes
- add
  - Helper function to click through auth
  - Add setup and config to add feature flags into local storage before running tests
- remove
- modify
   -  Enable auth adjacent test
   - Add rdev options for e2e test
   - Add Auth0 test credentials
   - General test tweaks to correctly follow auth and collection flow

----

## Notes

I'm adding a note section just to capture some info I learned while working on this:

* Playwright explicitly handles a lot of waiting needed to ensure elements are ready to be interacted with
   * Make sure not to copy these little hacks from puppeteer code
*  Sometimes blueprint will interfere with these checks, sometimes the `force` option will work, other times adding a `data-testid` attribute to the semantically correct element will fix it
* Playwright has a really cool system for capturing and loading in persisted browser data (i.e. Cookies & local storage)
   * In the future, it might be advantageous of us to utilize this in lieu of walking through Auth0
   * [Docs](https://playwright.dev/docs/auth#reuse-authentication-state)
*  Our create collection modal has input checking and was failing to notice when playwright had filled out the inputs.  To fix this I used the `page.type()` API instead of the `page.fill()`. make sure to utilize the `delay` option
* I set up a rdev option for the e2e tests, but rdev requires authentication from okta-duo.  To get around this I ran playwright's codegen to grab any necessary cookies and inserted the relevant output into `frontend/jest/featureFlags.js`.  If you do this make sure not to commit the changes
   * [Codegen docs](https://playwright.dev/docs/cli?_highlight=codegen#preserve-authenticated-state)

---

#824


frontend/jest-playwright.config.js
```diff
@@ -4,6 +4,8 @@
  * and `page` objects
  */
 
+const storageState = JSON.parse(process.env.STORAGE);
```
seve, line 7: Load in cookies and localstorage to playwright's browser context

frontend/tests/utils/helpers.ts
```diff
@@ -17,19 +33,22 @@ export async function tryUntil(assert: () => void) {
...
-
+      savedError = error;
       await page.waitForTimeout(WAIT_FOR_MS);
     }
   }
 
   if (retry === MAX_RETRY) {
-    throw Error("tryUntil() assertion failed!");
+    savedError.message += " tryUntil() failed";
+    throw savedError;
```
seve, line 52: I was receiving the,
> tryUntil() failed

message and it wasn't quite useful,  instead threw the last saved error

seve:
Just a reminder that the smoke-test reflected in the PR checks, doesn't reflect the changes made in the PR.  The tests currently fail as collection deletion has not been implemented

frontend/tests/utils/helpers.ts
```diff
@@ -1,14 +1,30 @@
-import { TEST_URL } from "tests/common/constants";
+import { TEST_EMAIL, TEST_URL } from "../common/constants";
 import { getText } from "./selectors";
 
 export async function goToPage(url: string = TEST_URL) {
   await page.goto(url);
 }
 
-export async function login(url: string = TEST_URL) {
-  await goToPage(url);
```
tihuan, line 9: What's the reason to remove goToPage() here? Thank you!

tihuan (approved):
Overall looks awesome! Just some tiny non-blocking questions 😄 Thank you!! 🎉👏👍

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
```diff
@@ -144,6 +144,7 @@ const Content: FC<Props> = (props) => {
             disabled={!isValid}
             onClick={onSubmit}
             loading={isLoading}
+            data-test-id="create button"
```
tihuan, line 147: Ah actually can we make this kebab case for consistency with other `data-test-id` please? Similar to how `id` attribute is kebab case too 👌 Thank you!

frontend/tests/features/collection.test.ts
```diff
@@ -18,25 +19,27 @@ const TEST_COLLECTION = {
...
+      await page.click(getText("Delete"));
+      await page.click(getText("Delete Collection"));
 
       await goToPage(
         apiTemplateToUrl(API_URL + ROUTES.PRIVATE_COLLECTION, {
           id: collectionId,
         })
       );
-      await expect(page).not.toHaveText(TEST_COLLECTION.name);
+      await expect(page).not.toHaveSelector(getText(TEST_COLLECTION.name));
```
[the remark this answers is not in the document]
seve: It wasn't. IIRC, it was comparing the arg to some function string, really bizzare

frontend/tests/utils/helpers.ts
seve: I forget the exact reasoning, but I believe I did it because the login action is available at any page and I don't believe login should own any page navigation as it's unrelated to the action itself and would just be a side effect

tihuan: ahh I see! 

Right now I think logging in will always dump the user back to the homepage regardless of which page the user is previously on before clicking the login button, because FE is not setting `redirect` param for the `/login` API call based on the user's current url. So `login()` needs to be called before we navigate to the destination page

For example, the following code will end up being on the homepage, even if the intention is to stay on the collection page after logging in:

```ts
await goToPage('/collections/foo');
await login();

// the user will still be on homepage at this point
```

Unless we update FE's call to `/login` to include the user's current URL as `redirect`

Otherwise we'd need to do:

```ts
await goToPage();
await login();
await goToPage('/collections/foo');
```

or:

```ts
await goToPage('/collections/foo');
await login();
await goToPage('/collections/foo');
```

which is more redundant than just keeping `await goToPage();` in `login()`, so we only need to:

```ts
await login();
await goToPage('/collections/foo');
```

?

frontend/tests/features/collection.test.ts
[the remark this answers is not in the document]
tihuan: hmm interesting, their doc seems to show that `await expect(page).not.toHaveText('foo')` should work 🤔 

https://github.com/playwright-community/expect-playwright#toHaveText

Let me try!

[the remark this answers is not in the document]
tihuan: Hmm yeah I see that 4 months ago I was using `getText()` too, but now `toHaveText()` seems to work for those tests. Maybe Playwright fixed it?

https://github.com/chanzuckerberg/corpora-data-portal/pull/899

frontend/tests/utils/helpers.ts
```diff
@@ -1,14 +1,30 @@
...
-  await page.click(getText("Log In"));
-  await page.waitForNavigation({ url, waitUntil: "networkidle" });
+export async function login() {
+  try {
+    await expect(page).toHaveSelector(getText("My Collections"));
+  } catch (error) {
+    const password = "Test1111";
```
Bento007, line 12: is this account created in auth0?

frontend/tests/common/constants.ts
```diff
@@ -7,7 +7,25 @@ const TEST_ENV_TO_TEST_URL = {
...
 
+const TEST_ENV_TO_TEST_EMAIL = (env: TEST_ENV) => {
+  switch (env) {
+    case "dev":
+    case "rdev":
+      return "cellxgene-smoke-test+dev@chanzuckerberg.com";
+    case "staging":
+      return "cellxgene-smoke-test+staging@chanzuckerberg.com";
+    case "prod":
+      return "cellxgene-smoke-test+prod@chanzuckerberg.com";
```
Bento007, line 22: If this test will run on prod we should avoid creating test collections. We don't want to end up with a bunch of junk collections if this test fails.

Bento007 (changes requested):
a few comments to address

frontend/tests/utils/helpers.ts
seve: Somewhere along the line I thought we were redirecting, I guess I made that assumption at some point because we only have a single non-authenticated page. I'll implement:
```
await login();
await goToPage('/collections/foo');
```
in the meantime

frontend/tests/utils/helpers.ts
seve: Yep!

Bento007: this value shouldn't be in the public repo, especially if this user also exists on prod.

Bento007: https://github.com/chanzuckerberg/corpora-data-portal/issues/900

Bento007 (approved):
LGTM but we need to resolve #900 ASAP

frontend/tests/common/constants.ts
tihuan: that's a good concern! Maybe we can only create a new collection only if there's less than 2 collections, otherwise some other tests might rely on the account having at least 1 collection to run their tests?

frontend/tests/utils/helpers.ts
tihuan: hmm what's the downside of exposing this prod password?

tihuan:
@seve can you take a look at https://github.com/chanzuckerberg/corpora-data-portal/pull/899 before we merge this PR please :D? It seems like `toHaveText()` is working! Thank you!

frontend/tests/utils/helpers.ts
maniarathi: [Here's](https://www.beyondtrust.com/blog/entry/hardcoded-and-embedded-credentials-are-an-it-security-hazard-heres-what-you-need-to-know) a good article on why exposing hardcoded passwords is a bad idea with some case studies.

frontend/tests/common/constants.ts
seve: For now, I've just disabled the tests on prod

PR #890, opened 2021-02-10
Workaround for pip dependency conflict.

jgadling:
This is a temporary workaround for errors installing the requirements for the happy script:
```
ERROR: Cannot install boto3 and terrasnek because these package versions have conflicting dependencies
```
This should be enough to make the happy script work for the moment, but needs more investigation in the future.

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


Bento007 (approved):
It's certainly a workaround.

PR #895, opened 2021-02-12
Dunitz/830 collection deletion

MDunitz:
#### Reviewers
**Functional:** 

@Bento007 

**Readability:** 

@seve 
---

## Changes
- add endpoint to tombstone collections
### Definition of Done
- The collection will no longer appear in the list of private Collections nor public collections
- The collection cannot be retrieved by the user
- Datasets that were in the collection cannot be retrieved.
- Artifacts cannot be download from direct links
- Get collection return 403 for a deleted collection
- Get dataset/status returns 403 for a deleted dataset
- Return status 202 to confirm the deleted or if the collection has already been deleted.

### DIFF
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -151,12 +151,15 @@ paths:
       summary: Delete a collection
       security:
         - cxguserCookie: [ ]
+        - dummyAuth: [ ]
       description: >-
         Deletes an entire private collection from cellxgene data-portal, including any generated artifacts/assets and
         deployments.
       operationId: corpora.lambdas.api.v1.collection_uuid.delete
       parameters:
         - $ref: "#/components/parameters/path_collection_uuid"
+        - $ref: "#/components/parameters/query_visibility"
+
       responses:
         "202":
           $ref: "#/components/responses/202"
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -155,7 +155,7 @@ paths:
       description: >-
         Deletes an entire private collection from cellxgene data-portal, including any generated artifacts/assets and
         deployments.
-      operationId: corpora.lambdas.api.v1.collection_uuid.delete
+      operationId: corpora.lambdas.api.v1.collection.delete_collection
       parameters:
         - $ref: "#/components/parameters/path_collection_uuid"
         - $ref: "#/components/parameters/query_visibility"
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -151,15 +151,12 @@ paths:
       summary: Delete a collection
       security:
         - cxguserCookie: [ ]
-        - dummyAuth: [ ]
       description: >-
         Deletes an entire private collection from cellxgene data-portal, including any generated artifacts/assets and
         deployments.
       operationId: corpora.lambdas.api.v1.collection.delete_collection
       parameters:
         - $ref: "#/components/parameters/path_collection_uuid"
-        - $ref: "#/components/parameters/query_visibility"
-
       responses:
         "202":
           $ref: "#/components/responses/202"
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -68,3 +68,16 @@ def create_collection(body: object, user: str):
 
 def get_collection_dataset(dataset_uuid: str):
     raise NotImplementedError
+
+
+def delete_collection(collection_uuid: str, visibility: str, user: str):
+    if visibility == CollectionVisibility.PUBLIC.name:
+        return make_response(jsonify("Can not delete a public dataset"), 405)
+    collection = Collection.get_collection(collection_uuid, visibility)
+    if not collection:
+        raise ForbiddenHTTPException()
+    if not Collection.if_owner(collection.uuid, collection.visibility, user):
+        raise ForbiddenHTTPException()
+    if not collection.tombstone:
+        collection.tombstone_collection()
+    return "", 202
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -70,11 +70,11 @@ def get_collection_dataset(dataset_uuid: str):
     raise NotImplementedError
 
 
-def delete_collection(collection_uuid: str, visibility: str, user: str):
-    if visibility == CollectionVisibility.PUBLIC.name:
-        return make_response(jsonify("Can not delete a public dataset"), 405)
+def delete_collection(collection_uuid: str, user: str):
     db_session = g.db_session
-    collection = Collection.get_collection(db_session, collection_uuid, visibility, include_tombstones=True)
+    collection = Collection.get_collection(
+        db_session, collection_uuid, CollectionVisibility.PRIVATE.name, include_tombstones=True
+    )
     if not collection:
         raise ForbiddenHTTPException()
     if not Collection.if_owner(db_session, collection.id, collection.visibility, user):
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -73,10 +73,11 @@ def get_collection_dataset(dataset_uuid: str):
 def delete_collection(collection_uuid: str, visibility: str, user: str):
     if visibility == CollectionVisibility.PUBLIC.name:
         return make_response(jsonify("Can not delete a public dataset"), 405)
-    collection = Collection.get_collection(collection_uuid, visibility)
+    db_session = g.db_session
+    collection = Collection.get_collection(db_session, collection_uuid, visibility, include_tombstones=True)
     if not collection:
         raise ForbiddenHTTPException()
-    if not Collection.if_owner(collection.uuid, collection.visibility, user):
+    if not Collection.if_owner(db_session, collection.id, collection.visibility, user):
         raise ForbiddenHTTPException()
     if not collection.tombstone:
         collection.tombstone_collection()
--- a/tests/unit/backend/chalice/api_server/test_v1_collection.py
+++ b/tests/unit/backend/chalice/api_server/test_v1_collection.py
@@ -1,15 +1,13 @@
 import itertools
 import json
-import uuid
 from datetime import datetime
 
 from furl import furl
 
-from backend.corpora.common.corpora_orm import CollectionVisibility, UploadStatus
+from backend.corpora.common.corpora_orm import CollectionVisibility, UploadStatus, generate_uuid
 from backend.corpora.common.entities import Collection
 from tests.unit.backend.chalice.api_server.mock_auth import get_auth_token
 from tests.unit.backend.chalice.api_server.base_api_test import BaseAuthAPITest
-from tests.unit.backend.fixtures.generate_data_mixin import GenerateDataMixin
 
 
 class TestCollection(BaseAuthAPITest):
@@ -431,7 +429,7 @@ class TestCollection(BaseAuthAPITest):
             self.assertNotIn(private_not_owned, ids)
 
 
-class TestCollectionDeletion(BaseAuthAPITest, GenerateDataMixin):
+class TestCollectionDeletion(BaseAuthAPITest):
     def test_delete_collection__ok(self):
         # Generate test collection
         collection = self.generate_collection(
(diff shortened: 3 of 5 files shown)


backend/config/corpora-api.yml
```diff
@@ -151,12 +151,15 @@ paths:
...
         - cxguserCookie: [ ]
+        - dummyAuth: [ ]
       description: >-
         Deletes an entire private collection from cellxgene data-portal, including any generated artifacts/assets and
         deployments.
-      operationId: corpora.lambdas.api.v1.collection_uuid.delete
+      operationId: corpora.lambdas.api.v1.collection.delete_collection
       parameters:
         - $ref: "#/components/parameters/path_collection_uuid"
+        - $ref: "#/components/parameters/query_visibility"
```
Bento007, line 161: this endpoint should only work for private collections. Why do we need visibility?

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -68,3 +68,17 @@ def create_collection(body: object, user: str):
 
 def get_collection_dataset(dataset_uuid: str):
     raise NotImplementedError
+
+
+def delete_collection(collection_uuid: str, visibility: str, user: str):
+    if visibility == CollectionVisibility.PUBLIC.name:
+        return make_response(jsonify("Can not delete a public dataset"), 405)
```
Bento007, line 75: Create an exception for this, inherit ProblemException, and then raise it here. This will ensure the format of the exception is uniform.

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -1,12 +1,15 @@
 import itertools
 import json
+import uuid
```
Bento007, line 3: use generate_uuid from corpora_orm.py 
```suggestion
```

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -427,8 +430,143 @@ def test__list_collection__check_owner(self):
             self.assertIn(private_owned, ids)
             self.assertNotIn(private_not_owned, ids)
 
-    def test__delete_collection__OK(self):
-        path = furl(path="/dp/v1/collections/test_collection_id")
+
+class TestCollectionDeletion(BaseAuthAPITest, GenerateDataMixin):
```
Bento007, line 434: `BaseAuthAPITest`  includes `GenerateDataMixin` 
```suggestion
class TestCollectionDeletion(BaseAuthAPITest):
```

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -1,12 +1,15 @@
...
 from datetime import datetime
 
 from furl import furl
 
-from backend.corpora.common.corpora_orm import CollectionVisibility
+from backend.corpora.common.corpora_orm import CollectionVisibility, UploadStatus
+from backend.corpora.common.entities import Collection
 from tests.unit.backend.chalice.api_server.mock_auth import get_auth_token
 from tests.unit.backend.chalice.api_server.base_api_test import BaseAuthAPITest
+from tests.unit.backend.fixtures.generate_data_mixin import GenerateDataMixin
```
Bento007, line 12:
```suggestion
```

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -1,12 +1,15 @@
 import itertools
 import json
+import uuid
 from datetime import datetime
 
 from furl import furl
 
-from backend.corpora.common.corpora_orm import CollectionVisibility
+from backend.corpora.common.corpora_orm import CollectionVisibility, UploadStatus
```
Bento007, line 8:
```suggestion
from backend.corpora.common.corpora_orm import CollectionVisibility, UploadStatus, generate_uuid
```

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -427,8 +430,143 @@ def test__list_collection__check_owner(self):
...
+
+        test_url = furl(path=f"/dp/v1/collections/{collection.id}")
+        test_url.add(query_params=dict(visibility="PRIVATE"))
+        response = self.app.get(test_url.url, headers=headers)
+        response.raise_for_status()
+
+        body = json.loads(response.body)
+        dataset_ids = [dataset["id"] for dataset in body["datasets"]]
+        self.assertIn(dataset_1.id, dataset_ids)
+        self.assertIn(dataset_2.id, dataset_ids)
```
Bento007, line 455: Remove. We don't need to test if get collection works.

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -427,8 +430,143 @@ def test__list_collection__check_owner(self):
...
+            self.session, visibility=CollectionVisibility.PRIVATE.name, owner="test_user_id"
+        )
+        processing_status = {"upload_status": UploadStatus.UPLOADED, "upload_progress": 100.0}
+
+        dataset = self.generate_dataset(self.session, collection=collection, processing_status=processing_status)
         headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
-        response = self.app.delete(path.url, headers)
+        dataset_url = furl(path=f"/dp/v1/datasets/{dataset.id}/status")
+        response = self.app.get(dataset_url.url, headers=headers)
+        self.assertEqual(response.status_code, 200)
```
Bento007, line 486: Remove. This is already tested in dataset.

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -427,8 +430,143 @@ def test__list_collection__check_owner(self):
...
+        collection = self.generate_collection(
+            self.session, visibility=CollectionVisibility.PRIVATE.name, owner="test_user_id", tombstone=True
+        )
+        test_url = furl(path=f"/dp/v1/collections/{collection.id}")
+        test_url.add(query_params=dict(visibility="PRIVATE"))
+        headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
+        response = self.app.delete(test_url.url, headers=headers)
+        self.assertEqual(response.status_code, 202)
+
+    def test_delete_collection__public(self):
```
Bento007, line 507:
```suggestion
    def test_delete_collection__public__405(self):
```

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -427,8 +430,143 @@ def test__list_collection__check_owner(self):
...
+        collection = self.generate_collection(
+            self.session, visibility=CollectionVisibility.PUBLIC.name, owner="test_user_id"
+        )
+        test_url = furl(path=f"/dp/v1/collections/{collection.id}")
+        test_url.add(query_params=dict(visibility="PRIVATE"))
+        headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
+        response = self.app.delete(test_url.url, headers=headers)
+        self.assertEqual(response.status_code, 403)
```
Bento007, line 515:
```suggestion
        self.assertEqual(response.status_code, 405)
```

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -427,8 +430,143 @@ def test__list_collection__check_owner(self):
...
+            self.session, visibility=CollectionVisibility.PRIVATE.name, owner="test_user_id"
+        )
+        headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
+        response = self.app.get("/dp/v1/collections/", headers=headers)
+
+        collection_ids = [collection["id"] for collection in json.loads(response.body)["collections"]]
+        self.assertIn(private_collection.id, collection_ids)
+        self.assertIn(public_collection.id, collection_ids)
+        self.assertIn(collection_to_delete.id, collection_ids)
+
```
Bento007, line 552: Remove, already tested in previous tests.

tests/unit/backend/corpora/common/entities/test_collection.py
```diff
@@ -197,3 +197,23 @@ def test__reshape_for_api(self):
...
         self.assertEqual([], response["datasets"])
+
+    def test_tombstone_collection_tombstones_all_datasets_in_collection(self):
+        collection = self.generate_collection(
+            self.session, visibility=CollectionVisibility.PRIVATE.name, owner="test_user_id"
+        )
+        self.generate_dataset(self.session, collection=collection)
+        self.generate_dataset(self.session, collection=collection)
+
+        self.assertEqual(len(collection.datasets), 2)
```
Bento007, line 208: Remove. We don't need to test this again.

Bento007 (approved):
Lost of extra test code that can be removed. Nothing major to fix.

Bento007 (changes requested):
Make changes please

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -68,3 +68,17 @@ def create_collection(body: object, user: str):
...
+    db_session = g.db_session
+    collection = Collection.get_collection(db_session, collection_uuid, visibility, include_tombstones=True)
+    if not collection:
+        raise ForbiddenHTTPException()
```
seve, line 79: Is 403 the most accurate exception we can throw here? I feel like it's a little misleading especially since the default message is saying the user is not authorized.  Maybe a 404 is more accurate?

seve, line 79: Is Forbidden the most accurate exception we can throw here, it's a bit misleading since the message sent with it is "User is not authorized to access this resource."  Perhaps a 404 would be better?

Bento007: This is common practice to throw forbidden rather than 404. It prevents users from probing for used UUIDs.

MDunitz: I think its actually meant to be a little misleading, my understanding is we send the same error message for a variety of errors so a malicious user cant tell from the error message which resources exist and they arent allowed to see and which just dont exist, but @Bento007 may have a better explanation/another opinion

backend/config/corpora-api.yml
MDunitz: To allow for a clear error message/that the user was specific about what they were trying to do and to avoid an unnecessary db call. But I can remove it

backend/corpora/lambdas/api/v1/collection.py
MDunitz: if we arent including visibility in the api  this will just be removed. If the collection is public it wont be retrieved and the error will be sent from the next lines

tests/unit/backend/chalice/api_server/test_v1_collection.py
MDunitz: When testing deletions I like to do an initial test to ensure it was created to make sure the test isnt passing due to the resource never being created

tests/unit/backend/chalice/api_server/test_v1_collection.py
MDunitz: see above comment

tests/unit/backend/corpora/common/entities/test_collection.py
MDunitz: see above comment about testing existence before testing for deletion

tests/unit/backend/chalice/api_server/test_v1_collection.py
MDunitz: see above comment

tests/unit/backend/chalice/api_server/test_v1_collection.py
MDunitz: based on the changes you recommend above (removing visibility as a param) this will return a 403 not a separate error

backend/corpora/lambdas/api/v1/collection.py
Bento007: No that was a good explanation.

tests/unit/backend/chalice/api_server/test_v1_collection.py
Bento007: but you can retrieve the dataset and collection and see that it's tombstoned using `get(tombstone=True)`, that will verify that it was both created and tombstoned. It looks like you're already doing that after the test. I still think this extra verify is superfluous

tests/unit/backend/corpora/common/entities/test_collection.py
Bento007: but you can and are retrieving the rows and can see that they are tombstoned after the delete. This confirms that they were both created and tombstoned. I still think this extra verify is superfluous. What you could add `tombstone=False` to the dataset and collections to make sure you didn't generate a tombstoned dataset or collection from the start. Then you could remove these extra lines from the tests. I guess this is a bit nit. I guess it's fine either way.

Bento007 (approved):
LGTM

seve (approved):

tihuan:
@MDunitz thanks for adding this endpoint 🎉 !! I just changed the PR to merge into the feature branch `feature-delete-collection `, so it gets the code to enable the delete button in FE 😄 Please help merge the feature branch into `main` 🙏! Thank you!

PR #896, opened 2021-02-12
Pin tiledb to version 0.7.4

mckinsel:
#### Reviewers

@Bento007 

---

## Changes
We can't have the tiledb of the processing image creep ahead of what's deployed in hosted cellxgene, or we create cxgs that can't be read. So, pin to the version used in hosted cellxgene, which is 0.7.4


Bento007 (approved):
LGTM

PR #897, opened 2021-02-12
conditionally render buttons based on collection visibility and access_type

Bento007:
#### Reviewers
**Functional:** @tihuan 

**Readability:** @seve 

---

## Changes
- modify collections page to hide delete and dataset error details.
- View of a **private** dataset as an **owner**. Errors and delete are visible.
Screen Shot 2021-02-16 at 10 52 28 AM

- View of a **published** dataset as an **owner**. Delete is not visible, and errors would be visible
Screen Shot 2021-02-16 at 11 35 52 AM

- View of a **published** dataset as a **non-owner**. Delete and error are not visible.
Screen Shot 2021-02-16 at 11 13 28 AM

### DIFF
--- a/frontend/src/common/entities.ts
+++ b/frontend/src/common/entities.ts
@@ -81,7 +81,6 @@ export interface Dataset {
   dataset_assets: DatasetAsset[];
   processing_status: DatasetUploadStatus;
   collection_id: Collection["id"];
-  collection_access_type: Collection["access_type"];
   // contributors: Contributor[];
   // preprint_doi: DOI;
   // publication_doi: DOI;
--- a/frontend/src/common/entities.ts
+++ b/frontend/src/common/entities.ts
@@ -81,6 +81,7 @@ export interface Dataset {
   dataset_assets: DatasetAsset[];
   processing_status: DatasetUploadStatus;
   collection_id: Collection["id"];
+  collection_visibility: Collection["visibility"];
   // contributors: Contributor[];
   // preprint_doi: DOI;
   // publication_doi: DOI;
--- a/frontend/src/common/entities.ts
+++ b/frontend/src/common/entities.ts
@@ -81,7 +81,7 @@ export interface Dataset {
   dataset_assets: DatasetAsset[];
   processing_status: DatasetUploadStatus;
   collection_id: Collection["id"];
-  collection_visibility: Collection["visibility"];
+  collection_access_type: Collection["access_type"];
   // contributors: Contributor[];
   // preprint_doi: DOI;
   // publication_doi: DOI;
--- a/frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
+++ b/frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
@@ -7,6 +7,7 @@ import {
   Dataset,
   UPLOAD_STATUS,
   VALIDATION_STATUS,
+  VISIBILITY_TYPE,
 } from "src/common/entities";
 import { useDatasetStatus } from "src/common/queries/datasets";
 import { aggregateDatasetsMetadata } from "src/components/Collections/components/Grid/common/utils";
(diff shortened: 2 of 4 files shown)


seve (approved):

frontend/src/common/entities.ts
```diff
@@ -81,7 +81,7 @@ export interface Dataset {
   dataset_assets: DatasetAsset[];
   processing_status: DatasetUploadStatus;
   collection_id: Collection["id"];
-  collection_visibility: Collection["visibility"];
+  collection_access_type: Collection["access_type"];
```
seve, line 84: Is collection_access_type a field on dataset? I don't see it on the [corpora-api.yaml](https://github.com/chanzuckerberg/corpora-data-portal/blob/8ad48e003283b328d7bd75ffc44140a6400479b0/backend/config/corpora-api.yml#L533)

Bento007: it's a field on collections. Does it also need to be a field on dataset?

Bento007: I was pulling this value from the collection this dataset is associated with.

tihuan:
Thanks so much for the PR, Trent!

Given that we don't allow a private collection with invalid/failed dataset to be published, we can remove the logic for hiding the error message based on write access, I believe

The following PR shows the conditions that need to be met in order to publish a private collection:

https://github.com/chanzuckerberg/corpora-data-portal/pull/893

Hope this makes sense! Thank you!

frontend/src/common/entities.ts
tihuan: Ah since the property is on the collection, we don't need to add `collection_access_type` on dataset object, since that implies `dataset.collection_access_type` is available, which it's not!

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -154,11 +155,12 @@ const DatasetRow: FC<Props> = ({
             checked={selected === dataset.id}
           />
           <div>{name}</div>
-          {!isLoading && (
-            <ErrorTooltip isFailed={isFailed} error={error} type={type} />
-          )}
+          {!isLoading &&
+            dataset.collection_access_type === ACCESS_TYPE.WRITE && (
```
tihuan, line 159: I think we can remove this change given that conditions needed to publish a private collection

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
```diff
@@ -154,11 +155,12 @@ const DatasetRow: FC<Props> = ({
...
+              <ErrorTooltip isFailed={isFailed} error={error} type={type} />
+            )}
         </TitleContainer>
-        {isLoading && (
+        {isLoading && dataset.collection_access_type === ACCESS_TYPE.WRITE && (
```
tihuan, line 163: This can be removed too

frontend/src/views/Collection/index.tsx
```diff
@@ -99,7 +99,9 @@ const Collection: FC<Props> = ({ id = "" }) => {
       </CollectionInfo>
 
       <CollectionButtons>
-        <DeleteCollection id={id} />
+        {collection.access_type === ACCESS_TYPE.WRITE && isPrivate && (
```
tihuan, line 102: Feels like this is the only change we need for this PR!

tihuan (approved):
LGTM after keeping only the delete button change 🙏 Thanks so much for fixing this!

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/index.tsx
Bento007: in the future when we share a private collection will we want users to see the errors?

frontend/src/common/entities.ts
```diff
@@ -81,6 +81,7 @@ export interface Dataset {
   dataset_assets: DatasetAsset[];
   processing_status: DatasetUploadStatus;
   collection_id: Collection["id"];
+  collection_access_type: Collection["access_type"];
```
Bento007, line 84:
```suggestion
```

PR #901, opened 2021-02-17
Don't do work on module import

jgadling:
create_engine() is being called as soon as this module is imported. Is this intentional?

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/corpora/common/utils/db_session.py
+++ b/backend/corpora/common/utils/db_session.py
@@ -12,10 +12,11 @@ logger = logging.getLogger(__name__)
 
 
 class DBSessionMaker:
-    engine = create_engine(CorporaDbConfig().database_uri, connect_args={"connect_timeout": 5})
+
     _session_make = None
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


backend/corpora/common/utils/db_session.py
```diff
@@ -12,10 +12,11 @@
 
 
 class DBSessionMaker:
-    engine = create_engine(CorporaDbConfig().database_uri, connect_args={"connect_timeout": 5})
+
     _session_make = None
 
     def __init__(self):
+        self.engine = create_engine(CorporaDbConfig().database_uri, connect_args={"connect_timeout": 5})
```
Bento007, line 19: Only one engine needs to be built for the whole process. It was intentional but we can change to initialize once when the first `DBSessionMaker` is created.

Bento007 (changes requested):

Bento007 (approved):
LGTM

PR #902, opened 2021-02-17
Refactor terraform for remote dev

jgadling:
This switches the remote dev environment to a new ECS cluster

#### Reviewers
**Functional:** 
This refactor allows data portal stacks to be launched with different configurations in different environments (dev, stage, prod, etc) and paves the way for staging deploys using the happy cli script. Unfortunately it's a hard cutover, so existing remote-dev stacks have been removed, and new ones will require this PR to be merged before they can be launched.

In addition, it contains some fixes to improve docker layer caching when launching remote dev stacks, and allows some environments to prevent stack deletion.

**Readability:** 

---

## Changes
- add
- remove
- modify


.happy/terraform/modules/batch/main.tf
```diff
@@ -30,7 +30,7 @@ resource aws_batch_job_definition batch_job_def {
     },
     {
       "name": "REMOTE_DEV_PREFIX",
-      "value": "/${var.custom_stack_name}"
+      "value": "${var.remote_dev_prefix}"
```
srm78, line 33: This is used as db schema prefix also right? Not sure if any python code needs changes as a result of this?

jgadling: Nope - we just add the leading '/' earlier on in the workflow here:
https://github.com/chanzuckerberg/corpora-data-portal/pull/902/files#diff-667e21ce425cdbf70b0b62e040edb6a4ec8c1100a20dd94fc36cda064f98e100R13

srm78 (approved):

Issue #855, opened 2021-01-25
Refactor "All Collections" to use collection grid


seve:
Context:

The homepage should implement a version of collection grid instead of a list of all readable datasets

What:

1. Import/utilize collection view from `src/components/Collections/components/Grid/components/DatasetsGrid/index.tsx` 
   1. This may require refactoring and added generalization of the component
   1. Memoize/cache dataset metadata fetching/calculating
1. Grid should display all published and readable collections in descending order by updated date
    1. This will require addition of a visibility prop to the grid
1. Also need to alter the header and description to [meet the design by Harley](https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Publisher-Collection?node-id=906%3A192)
1. E2E testing ensuring that collections are correctly displaying

Design Notes:
- Change collection title to `$pt-text-color` and `font-weight: semibold`
- Change DOI text to `$pt-text-color`
- Make entire collection row click target (hover state should show make background `@light-gray5`
- Rows should be top aligned by text baseline

PR #918, opened 2021-02-17
update homepage to show public collections

seve:
#### Reviewers
**Functional:**  @tihuan 

**Readability:** @MDunitz 

---

## Changes
- add
- remove
- modify
  - Fixes our font import from google fonts
  - Various style changes to collection grid to meet mock 
  - Memoized dataset metadata aggregation
  - Utilized collection grid on collections component
  - Add `displayVisibility` prop to `collectionsGrid`

---

image

---

#855


frontend/src/components/Collections/components/Grid/common/style.ts
```diff
@@ -23,29 +23,32 @@ export const StyledCollectionsGrid = styled(HTMLTable)`
   margin-top: ${PT_GRID_SIZE_PX * 2}px;
 `;
 
-export const CollectionHeaderCell = styled.th`
+const HeaderCell = styled.th`
+  padding: 12px 0 !important;
+  font-weight: 500;
+  font-size: 14px;
+  line-height: 18px;
+`;
```
seve, line 31: Pulled common header styles into own style

frontend/src/components/Collections/components/Grid/common/style.ts
```diff
@@ -23,29 +23,32 @@ export const StyledCollectionsGrid = styled(HTMLTable)`
...
   ${detailsColWidthCSS}
   text-align: left !important;
-  padding: 0 !important;
   padding-left: ${PT_GRID_SIZE_PX * 2}px !important;
 `;
 
-export const RightAlignedHeaderCell = styled.th`
-  ${datasetTitleColWidthCSS}
+export const RightAlignedHeaderCell = styled(HeaderCell)`
+  ${detailsColWidthCSS}
```
seve, line 51: Was using incorrect width previously

frontend/src/components/Collections/components/Grid/common/utils.ts
```diff
@@ -34,6 +35,12 @@ export function aggregateDatasetsMetadata(datasets: Dataset[]) {
     organism: unique(result.organism),
     tissue: unique(result.tissue),
   };
+}, hashFn);
+
+function hashFn(datasets: Dataset[]): string {
+  return datasets.reduce((acc, dataset) => {
+    return acc + dataset.id;
+  }, "");
```
seve, line 43: hash function is just a string of all ids, this implementation is sensitive to ordering, but that shouldn't be changing between api calls

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -11,10 +11,15 @@ import CollectionRow from "../Row/CollectionRow";
 
 interface Props {
   collections: CollectionResponse[];
-  accessType: ACCESS_TYPE;
+  accessType?: ACCESS_TYPE;
+  displayVisibility?: VISIBILITY_TYPE;
```
seve, line 15: Optional display visibility will filter out visibilities that don't match the specified type
Access type is now also optional

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -28,17 +33,29 @@ const CollectionsGrid: FC<Props> = ({ collections, accessType }) => {
...
       <tbody>
-        {collections?.map(({ id, visibility }) => (
-          <CollectionRow
-            id={id}
-            key={id + visibility}
-            visibility={visibility}
-            accessType={accessType}
-          />
-        ))}
+        {sortByUpdatedAtDescending(collections)?.reduce(
```
seve, line 36: Sorting collections by updated date

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -28,17 +33,29 @@ const CollectionsGrid: FC<Props> = ({ collections, accessType }) => {
...
+          (acc, { id, visibility }) => {
+            if (!displayVisibility || visibility === displayVisibility) {
```
seve, line 38: Filter out any non-matching visibilities if applicable

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -12,13 +13,13 @@ import {
...
   StyledCell,
-  StyledRow,
+  StyledCollectionRow,
 } from "../common/style";
 import { CollectionTitleText } from "./style";
 
 interface Props {
   id: string;
-  accessType: ACCESS_TYPE;
+  accessType?: ACCESS_TYPE;
```
seve, line 22: continue optional accessType

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -75,13 +76,14 @@ const CollectionRow: FC<Props> = (props) => {
   } = aggregateDatasetsMetadata(datasets);
 
   return (
-    <StyledRow>
+    <StyledCollectionRow
+      onClick={(e) => {
+        if ((e.target as Element).tagName !== "A")
+          navigate(`/collections/${id}${isPrivate ? "/private" : ""}`);
+      }}
+    >
```
seve, line 84: Entire row will now navigate to the collection page, ignore row clicks if the user is clicking on the DOI link

tihuan:
Ah actually, does the homepage need to show both public and private collections if the user is logged in?

@brianraymor @hthomas-czi @seve 

Thank you!

brianraymor:
I would expect to see my private collections on the landing page if I was logged in. Basically, I should see all collections that I'm authorized to see.

brianraymor:
@hthomas-czi - do you think that the private collections on the landing page should have a "private" label so that the curator doesn't believe that the collection was accidentally made public?

hthomas-czi:
@brianraymor @tihuan @seve 

I explored using [tags on My Collections view](https://www.figma.com/file/g0L0CfVzRc8r6HQa4NH1wJ/Publisher-Collection?node-id=202%3A4040) like you described, Brian. I think we could easily adopt the same concept for the All Collections view.

tihuan:
Thanks both! @seve PTAL 😄 🙏 Thank you!

hthomas-czi:
@brianraymor Following up a little more, @seve and I both thought the landing page would show **all public collections** rather than **all collections available to me** as you described. 

Do you feel strongly about that behavior? @tihuan thoughts?

frontend/tests/features/homepage.test.ts
```diff
@@ -1,64 +1,15 @@
...
-    await expect(page).toHaveSelector(getText("More information"));
-
-    await expect(page).toHaveSelector(getTestTag("dataset-name"));
-    await expect(page).toHaveSelector(getTestTag("view-dataset-link"));
-    await expect(page).toHaveSelector(
-      getTestTag(DATASET_ROW_DOWNLOAD_BUTTON_ID)
-    );
-  });
-
-  describe("renders the download dataset modal", () => {
```
seve, line 25: no longer showing download modal on homepage

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -66,6 +67,11 @@ const CollectionRow: FC<Props> = (props) => {
 
   const isPrivate = visibility === VISIBILITY_TYPE.PRIVATE;
 
+  const handleRowClick = (e: React.MouseEvent) => {
```
tihuan, line 70: It seems like a navigation that's not an anchor link is not a11y friendly, and nested anchor links is an antipattern: https://ux.stackexchange.com/questions/102252/more-info-button-vs-making-the-entire-table-row-a-link

Can you help work with @hthomas-czi to prevent the nesting please?

I'm thinking maybe we can just make the name column an anchor element instead of the whole row or something!

Thank you!

tihuan (changes requested):
Really great PR! I haven't gone through the whole change yet, but some comments I think we can address first! Thanks a lot 🤩 🙏 🚀 !

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
seve: I'll be removing the DOI anchor to avoid nested anchors and will try changing the entire row to an anchor

MDunitz (approved):
lgtm

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
tihuan: Thanks Seve! Sorry I forgot to mention that I think the UX stackexchange link seems to suggest that making the entire row an anchor is an antipattern in UX world too 😆 @hthomas-czi PTAL too thank you!

seve: HTML semantics actually doesn't allow anything but a td or tr as a child of table, so I'll be putting an anchor in each individual cell

tihuan: Got it! I think we need to resolve the UX level issue first before we resolve the eng implementation. The UX issue here is that when an entire row is clickable, it's unclear if clicking in different column navigates to different pages or the same page, thus the UX antipattern and the Stackexchange discussion!

hthomas-czi: @seve Ah, if we can't make the entire row the anchor element, I agree with @tihuan that the expected behavior is unclear. Can we return to the [original design](https://www.figma.com/file/6wPUe08qSM8V9G2G0UfYOP/Consumer-Collection?node-id=478%3A0) where title and DOI and links?

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
tihuan: @hthomas-czi test notif lol No action needed

tihuan: @hthomas-czi again

tihuan: same thread that has mention

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -1,5 +1,6 @@
 import { Intent, Tag } from "@blueprintjs/core";
 import loadable from "@loadable/component";
+import { navigate } from "@reach/router";
```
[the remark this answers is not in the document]
tihuan: Harley shouldn't get this comment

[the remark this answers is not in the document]
tihuan: @hthomas-czi now you should see this!

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -1,5 +1,6 @@
...
 import React, { FC } from "react";
```
tihuan, line 4: Harley shouldn't see this too!! hehe

tihuan: Hi Harley you shouldn't see this

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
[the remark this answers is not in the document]
tihuan: @hthomas-czi should see this!!

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -75,17 +81,20 @@ const CollectionRow: FC<Props> = (props) => {
   } = aggregateDatasetsMetadata(datasets);
 
   return (
-    <StyledRow>
+    <StyledCollectionRow
+      onClick={handleRowClick}
+      data-test-class="collection-row"
```
tihuan, line 86: I haven't found the need to have both `data-test-id` and `data-test-class`, especially since I think `data-test-class` doesn't support css class selector `.class-name` (e.g., `page.$$('.collection-row')` 🤔 So perhaps if we really need it, we could use attribute `className="collection-row"` instead?

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
tihuan: Harley don't see this!

frontend/src/components/Collections/index.tsx
```diff
@@ -15,22 +15,16 @@ const Collections: FC = () => {
...
-          cellxgene@chanzuckerberg.com
-        </a>
-        .
-      </p>
-      <Heading />
-      {collections?.map(({ id, visibility }) => (
-        <Collection id={id} visibility={visibility} key={id} />
-      ))}
+      <CollectionsGrid
+        collections={collections}
```
tihuan, line 25: Hmm personally I prefer to do the filtering at this level, since it's business logic instead of `<CollectionsGrid />`'s own reusable component logic. This way the shared component is free of business logic as much as possible!

So something like:

```ts
<CollectionGrid
  collections={publicOnlyCollections(collections)}
/>
```

frontend/src/components/seo.tsx
```diff
@@ -75,8 +75,9 @@ const SEO: FC<Props> = ({ description, lang, meta = [], title }) => {
         },
       ]}
     >
+      <link rel="preconnect" href="https://fonts.gstatic.com" />
       <link
-        href="https://fonts.googleapis.com/css?family=Montserrat:500,700|Roboto&display=swap"
+        href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
```
tihuan, line 80: Thanks for fixing this! Are we sure we need all of these font weights?

CC: @hthomas-czi @seve 

Thanks!

frontend/tests/features/homepage.test.ts
tihuan: Awesome thank you!

tihuan (commented):
Done reviewing! PTAL @seve and @hthomas-czi 😄 🙏 Thank you!!

frontend/src/components/seo.tsx
hthomas-czi: We don't need all of those. 400, 500, 600, and 700 is sufficient.

hthomas-czi: Just for reference, those correspond to Regular, Medium, Semibold, and Bold, respectively.

frontend/src/components/Collections/index.tsx
seve: I had the `CollectionsGrid` own visibility filtering because the component already passes its `accessType` prop to `CollectionRow`  since it owns the access type filtering as access type is only accessible after fetching the collection details.  I believe we're also planning on adding user filters/sorting to the grids in the future, so I thought leaving this modularity accessible would be beneficial

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
seve: Sounds good! But I don't think `data-test-id` supports css class selector `#id-name` either

seve: https://playwright.dev/docs/selectors/#id-data-testid-data-test-id-data-test-selectors

tihuan: Yeah that's correct! Which is why we have to use `getTestTag()` helper function 👍

frontend/src/components/Collections/index.tsx
tihuan: Ahh got it! That sounds good 👍  I forgot about the broader context of the component 😂 Thanks for the reminder!

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
tihuan: Actually we can use `sortByUpdatedAtDescending(collections)?.filter()` instead of `.reduce()` here for clarity, I believe!

E.g., 

```tsx
{ renderCollections(sortByUpdatedAtDescending(collections)) }

function renderCollections(collections = []) {
  if (!collections.length) return null;

  return collections.filter((collection) => isConditionMet(collections))
}
```

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -75,24 +75,28 @@ const CollectionRow: FC<Props> = (props) => {
...
+            minimal
+            intent={isPrivate ? Intent.PRIMARY : Intent.SUCCESS}
+            data-test-id="visibility-tag"
+          >
             {isPrivate ? "Private" : "Published"}
           </Tag>
         ) : (
           dois?.map((doi) => (
-            <a key={doi.doi} href={doi.link}>
+            <DOILink key={doi.doi} href={doi.link}>
```
tihuan, line 97: maybe the DOI link needs to open in a new tab 😄 ?

tihuan (approved):
LGTM!! Thanks for the quick iterations on this, @seve !

I think one last thing to check with @hthomas-czi is whether we want the collection DOI links to open in a new tab or navigate from the current page 🙆‍♂️ Thank you!

hthomas-czi:
> LGTM!! Thanks for the quick iterations on this, @seve !
> 
> I think one last thing to check with @hthomas-czi is whether we want the collection DOI links to open in a new tab or navigate from the current page 🙆‍♂️ Thank you!

@seve @tihuan New tab would be ideal since it's navigating to external site.

hthomas-czi (approved):

PR #942, opened 2021-02-24
allow non owners to access private collections

MDunitz:
#### Reviewers
**Functional:** 
- @Bento007 

**Readability:** 
- @seve 
---

## Changes
- modify collections endpoint to allow non owners to have read access 

When a non-owner visits a private collection link, they should still see the page, even though they don't have WRITE access. But they shouldn't see WRITE access related actions, such as Publish


Bento007 (approved):
LGTM

seve (approved):
Readability LGTM

Issue #519, opened 2020-08-12
[Infra] Integrate DP Smoke Test with CI/CD


tihuan:
Given that we're deferring DP's CI/CD deployment setup to Release 2+, we can hook smoke test integration into CI/CD once the setup is done too

Requirement:
1. After the deployment stage of our CI/CD pipeline, add a step that runs the smoke test on the new deployment

TEAM COMMENTS:
1. We're missing FE e2e tests against staging and prod! We already have the e2e commands in `package.json`, so we just need to use them in Github Actions!

tihuan:
Ah @brianraymor I see this issue was removed from Release 2, can you remind me the reason please 😄? Thank you!

maniarathi:
Add the frontend tests -- we currently execute the backend tests.

Bento007:
dev: https://github.com/chanzuckerberg/corpora-data-portal/blob/2c6a0e7588737aaa5ac5cb13f134c73f6699eadb/.github/workflows/deploy.yml#L88

PR #946, opened 2021-02-25
run smoke test on staging and prod

Bento007:
#### Reviewers
**Functional:** @tihuan 

**Readability:** @seve 

---

Fix for #519

## Changes
- add smoke test to staging and dev environments


.github/workflows/deploy-prod.yml
```diff
@@ -68,3 +68,27 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: backend
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup node.js
+        uses: actions/setup-node@v1
+      - name: Install Dependencies
+        run: |
+          cd frontend
+          cp src/configs/dev.js src/configs/configs.js
```
tihuan, line 82: We can remove copying the config, because we're not building the app for smoke test! But if we want to keep the config, we just need to change it to use `cp src/configs/prod.js src/configs/configs.js` instead 😄 

Thank you!

.github/workflows/deploy-staging.yml
```diff
@@ -68,3 +68,27 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: backend
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup node.js
+        uses: actions/setup-node@v1
+      - name: Install Dependencies
+        run: |
+          cd frontend
+          cp src/configs/dev.js src/configs/configs.js
```
tihuan, line 82: same here but `cp src/configs/staging.js src/configs/configs.js` if needed!

tihuan (approved):
2 tiny comments otherwise LGTM! Thanks so much, Trent 🎉 !!

.github/workflows/deploy-prod.yml
Bento007:
```suggestion
```

.github/workflows/deploy-staging.yml
Bento007:
```suggestion
```

tihuan (approved):
LGTM to the max!! Thanks so much, @Bento007 🙏 ✨ 🚀 !!

seve (approved):
Cool cool cool

Issue #940, opened 2021-02-24
Integration Tests


Bento007:
- [x] Create a collection
- [x] Delete a collection
- [x] Publish a collection
- [x] upload a dataset
- [x] delete a dataset
- [ ] delete a pending dataset
- [ ] upload a bad dataset
- [ ] delete a bad dataset
- [x] cannot retrieve a tombstoned collection
- [ ] cannot retrieve a tombstoned dataset
- [x] login
- [ ] logout
- [ ] list owners collections
- [ ] only owner can delete and upload datasets to a collection

## Definition of Done
- [ ] Write an integration test that run on a deployed environment.
- [ ] Ensure that cleanup happens after the test.

maniarathi:
We should ensure that these use cases are covered both on the frontend and backend.

Bento007:
I'll do the backend version.

tihuan:
Actually writing the test for uploading a dataset and then cancelling it will take a few more hours, which I think is lower priority than working on the validation script. So I just sent out a PR for the E2E test fixes and will stop working on this ticket for now! Thank you!

https://github.com/chanzuckerberg/corpora-data-portal/pull/950

Bento007:
created backend issue https://github.com/chanzuckerberg/corpora-data-portal/issues/970

tihuan:
FE issue here! https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/971

Will close this ticket now 🤩 Thank you!

PR #950, opened 2021-03-01
thuang-940-moar-e2e-tests

tihuan:
#940 

#### Reviewers
**Functional:** 
@seve 

**Readability:** 
@Bento007 

---

## Changes
- add

- remove

- modify

1. update `expect-playwright` and `playwright` to the latest versions. Turned out that `expect(page).toHaveText('foo', { timeout: TIMEOUT_MS })` didn't work, because `npm upgrade expect-playwright` wouldn't upgrade from `0.2.5` to `0.3.2`, which accept the option. And I had to do `npm remove expect-playwright && npm i -D expect-playwright` 😂 

    However, `toHaveText()` API is not as smart as expected, since the `timeout` option is for waiting for the `selector` to exist, instead of waiting for the `expected text` to appear. So when we omit the selector, `toHaveText()` will use `body` as default, so it will just grab whatever is in `body` tag and assert if the HTML string includes our expected text, which won't work for our use case. So we might as well just continue using `toHaveSelector(getText())`!

1. Update Collection test to navigate to the right Collection page. Previously it's navigating to an invalid URL, which still passes the test, since the invalid page also doesn't have the deleted collection name!

1. Add timeout option of 5 seconds to `toHaveSelector()` to shorten the test to pass, since the default is 30 seconds, which makes the test run longer than needed

1. My mistake: Helper function `createCollection()` was returning `undefined` as collection id, since I was extracting from an invalid property key 🤦 

1. Helper function `login()` also gets the same 5 seconds timeout now, so it logs in faster instead of waiting for 30 seconds!

All E2E tests should pass, but should complete much faster now!

## QA steps (optional)
N/A


frontend/tests/features/collection.test.ts
```diff
@@ -15,7 +17,7 @@ const TEST_COLLECTION = {
   name: "TEST COLLECTION",
 };
 
-describe("Collection", async () => {
+describe("Collection", () => {
```
seve, line 20: thank you!

seve (approved):
What a ridiculous serious of bugs that resulted in tests passing 🤣 .  Thanks for cleaning up our hiccups

image

tihuan:
Ahaha Windows XP!! Oh man lol

I happened to see the bugs because I was running headful mode 😆 I definitely didn't catch this in code reviews too!!

Thanks for the quick review 🚀

PR #962, opened 2021-03-03
thuang-fix-collection-test

tihuan:
#### Reviewers
**Functional:** 
@seve 

**Readability:** 
@Bento007 

---

## Changes
- add
- remove
- modify
1. Update test code to prevent race condition

The problem we had before is that right after clicking on the delete button, we immediately navigate to the private collection page without waiting for the success API callback. This means the collection page sometimes will still be there, since the BE hasn't deleted the collection yet, thus failing the test.

The solution is to wait for navigations!


Bento007 (approved):
LGTM

PR #976, opened 2021-03-09
feat: update a private collection

Bento007:
#### Reviewers
**Functional:** @MDunitz  

**Readability:** @tihuan 

---

## Changes
- Remove if_owner: Rolled the functionality into Collection.get_collection
- Added PUT collection/collection_uuid endpoint

## Definition of Done
- New API PUT /collections/{collection_uuid} is added to openapi.yml specification.
- The user must be the owner of the collection to modify
- The collection must be private to modify.
- A user can update a private collection info
- A user can add, remove, and update links in a private collection


backend/corpora/common/entities/collection.py
```diff
@@ -58,31 +57,28 @@ def create(
...
-        :param user: the uuid of the user.
-        :return: a collection if the user is the owner of the collection else None
+        :param include_tombstones: If true, the collection is returned even if it has been tombstoned.
+        :param owner: A user id use to check if the user is the owner of the collection. If the user id matches the
+        owner then the collection is returned.
+        :return: the collection if it matches the filter.
         """
-        filters = [cls.table.id == collection_uuid, cls.table.owner == user, cls.table.visibility == visibility]
+        filters = [cls.table.id == collection_uuid, cls.table.visibility == visibility]
+        if owner:
```
MDunitz, line 78: Will non owners still have read only access for pre publication link sharing?

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -71,12 +71,21 @@ def get_collection_dataset(dataset_uuid: str):
...
     if not collection:
         raise ForbiddenHTTPException()
-    if not Collection.if_owner(db_session, collection.id, collection.visibility, user):
-        raise ForbiddenHTTPException()
     if not collection.tombstone:
         collection.tombstone_collection()
     return "", 202
+
+
+def update_collection(collection_uuid: str, body: dict, user: str):
```
tihuan, line 83: discussion: also personally prefer `userId` or `user_uuid` to be consistent with `collection_uuid`

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -71,12 +71,21 @@ def get_collection_dataset(dataset_uuid: str):
...
+    db_session = g.db_session
+    collection = Collection.get_collection(db_session, collection_uuid, CollectionVisibility.PRIVATE.name, owner=user)
```
tihuan, line 85: feels like we just need to check if the user has write access to the collection before updating the collection here, instead of restricting the collection access to `WRITE only`?

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -577,3 +577,81 @@ def test_deleted_collection_does_not_appear_in_collection_lists(self):
         self.assertIn(public_collection.id, collection_ids)
         self.assertNotIn(private_collection.id, collection_ids)
         self.assertNotIn(collection_to_delete.id, collection_ids)
+
+
+class TestUpdateCollection(BaseAuthAPITest):
```
tihuan, line 582: awesome tests!!

tihuan (changes requested):
Thanks for adding this, @Bento007 🙏 Just some tiny discussion and readability stuff 😄 Thank you!

backend/corpora/lambdas/api/v1/collection.py
Bento007: `user` is because fo [connexion](https://connexion.readthedocs.io/en/latest/security.html)

backend/corpora/lambdas/api/v1/collection.py
Bento007: At this point, we are not updating the collection. We are only fetching it from the DB. If the user is the owner of the collection then they have write access.

backend/corpora/common/entities/collection.py
Bento007: `owner` is used as a filter if it is provided, otherwise it's not used in the query.

Bento007: Yes, non-owners will still have read-only access for pre-published link sharing.

backend/corpora/lambdas/api/v1/collection.py
tihuan: Oh I didn't realize `owner` argument was optional 😆  

My thought was instead of figuring out when to pass `owner` as an optional argument to `Collection.get_collection` as a way to indirectly check if the user is the owner, would it be more straightforward to just do:

```py
collection = Collection.get_collection(db_session, collection_uuid, CollectionVisibility.PRIVATE.name)

if not (check user is the owner of the collection):

# update collection here
```
or maybe update description of `Collection.get_collection()` to include notes that `owner` argument is optional and is only needed if the callsite needs to return WRITE only collection?

Bento007: By including the owner in the filter we are performing the check for ownership in the query. I'll update the documentation.

backend/corpora/common/entities/collection.py
```diff
@@ -58,31 +57,28 @@ def create(
...
     @classmethod
     def get_collection(
-        cls, session, collection_uuid, visibility=CollectionVisibility.PUBLIC.name, include_tombstones=False
-    ):
+        cls,
+        session,
+        collection_uuid: str,
+        visibility: str = CollectionVisibility.PUBLIC.name,
+        include_tombstones: bool = False,
+        owner: typing.Optional[str] = None,
```
Bento007, line 65: @tihuan owner is an optional parameter.

tihuan: oh!! Thank you 🙏  I didn't scroll up high enough to see this 😆 My bad!

backend/corpora/lambdas/api/v1/collection.py
tihuan: Awesome thank you!

tihuan (approved):
LGTM! Thanks for the quick discussion, @Bento007 👍 💯 🚀 !

MDunitz (approved):
lgtm

PR #980, opened 2021-03-10
Override config from env by default, functional-test reads from chamber

mbarrien:
* FUNCTIONALITY CHANGE: Makes data portal read config from environment by default,
  overriding values in config file. Uses existing functionality with minor code changes.
* Fix code and unit tests that were broken when use_env=True
  * Fix non-deterministic behavior when reading config
  * Unset variables that are overrridden by EnvironmentSetup decorator if not previously set
  * Explicitly set use_env in tests that were relying on implicit behavior.
* Uses chamber to read environment variables for auth0 for client ID and secret
  when running make local-functional-test. This is a stopgap that will unblock
  local testing while we work on a longer term solution. This has been tested
  to pass tests successfully, and is currently hard coded to using the secrets
  from the dev stage. Requires Chamber v2.9.0 or above to work
  (needs secretsmanager backend introduced in that version).
  This can be installed via brew install on Macs.


jgadling (approved):

Makefile
```diff
@@ -149,8 +153,9 @@ local-unit-test: ## Run backend tests in the dev environment
 	fi
 
 .PHONY: local-functional-test
+local-functional-test: export AWS_PROFILE=$(TEST_AWS_PROFILE)
 local-functional-test: ## Run functional tests in the dev environment
-	docker-compose exec -T backend bash -c "cd /corpora-data-portal && export DEPLOYMENT_STAGE=test && make container-functionaltest"
+	chamber -b secretsmanager exec corpora/backend/$${DEPLOYMENT_STAGE}/auth0-secret -- docker-compose exec -T -e CLIENT_ID -e CLIENT_SECRET -e TEST_ACCOUNT_PASSWORD -e DEPLOYMENT_STAGE backend bash -c "cd /corpora-data-portal && make container-functionaltest"
```
Bento007, line 158: what does `--` do?

Bento007 (changes requested):
Since there are additional dependencies required to run the functional tests, please update the documentation with instructions on running the functional tests which include how to install the dependencies.

Makefile
mbarrien: Tells chamber that everything past the `--` is the command to run in the shell it creates, with the environment variables set from the values in the secrets manager

mbarrien:
> Since there are additional dependencies required to run the functional tests, please update the documentation with instructions on running the functional tests which include how to install the dependencies.

Done. Please re-review.

Bento007 (approved):
LGTM

PR #998, opened 2021-03-19
fix: Add secret manager to FE e2e testing

seve:
#### Reviewers
**Functional:** @tihuan  + @jgadling 

---

## Changes
- add
  - jq to dockerfile `apt install` command 
  - env vars for localstack to frontend container
  - users.json volume to backend container
  - e2e-on-local-dev Makefile target
  - secretsmanager fetching to playwright global setup
  - test env for happy localhost
  - users.json parsing to setup dev data
- remove
- modify
 - re-enabled auth testing 
 - Pointed `smoke-test-with-local-dev` target to `npm run e2e-happy`


seve:
This is almost ready for merge.  The tests currently pass on our deployments, but need some work to work on local-dev. 

LocalStack SecretsManager is successfully creating and serving the secrets needed to log into the mock OIDC, but we're having difficulty getting the frontend container to communicate with the OIDC container.  We're able to get communication working between the frontend and backend containers by replacing the config.js with local-docker.js via the `smoke-test-with-local-dev` Makefile target.  This changes the api url from localhost:5000 to backend:5000.  The difficulty with OIDC stems from the fact that the backend provides the URL to connect to the OAuth service and that URL is loaded into the backend through the `setup_dev_data` script when we run `make local-init`.

_Now that I'm looking at the commands behind that target, there might be a chance that we can run `export OIDC_BROWSER_URL=http://oidc` before we run the test and then replace it back with `https://localhost:8443` after the tests have completed_

I've changed/created a few Makefile targets for testing, but haven't put enough thought into the workflow to be certain they'll end up on main. Right now the workflow is calling `make e2e-on-local-dev` from `/frontend` with the containers already running.

I last left off on talking with Infra regarding setting all the containers to `network_mode: host` to remove the need to use the container name when connecting between containers.  This means we won't be replacing instances of `localhost` with `<container-name>`.

After this PR is merged we need to port over the setup and utilization code over to explorer.  This will just be including `    "aws-sdk/client-secrets-manager` as a devDep and loading in the secret from aws (we won't be needing any connection to localstack since local-dev doesn't support explorer currently).  This secret loading in will be in a new file that will be referenced in `e2eJestConfig.json:5` and then we'll grab the username/pass from env in`login()` in `cellxgeneActions.js` 

Once both explorer and data portal PRs are merged, the ticket can be closed 🥳 

P.S. The estimation for the ticket has been updated and the remaining work is pointed at a 2

Dockerfile
```diff
@@ -8,8 +8,8 @@ ENV LC_ALL=C.UTF-8
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y gettext moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm && \
-    rm -rf /var/lib/apt/lists/* 
+    apt-get install -y gettext moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm jq && \
```
seve, line 11: we need to parse json now

docker-compose.yml
```diff
@@ -49,6 +49,11 @@ services:
     command: ["./container_init.sh"]
     environment:
       - DEPLOYMENT_STAGE=dev
+      - AWS_REGION=us-west-2
+      - AWS_DEFAULT_REGION=us-west-2
+      - AWS_ACCESS_KEY_ID=dev_access_key_id
+      - AWS_SECRET_ACCESS_KEY=dev_secret_access_key
+      - BOTO_ENDPOINT_URL=http://frontend.corporanet.local:4566/
```
seve, line 56: container needs access to localstack

docker-compose.yml
```diff
@@ -156,6 +161,8 @@ services:
       - ./backend/chalice/api_server/.chalice-dev:/corpora-data-portal/backend/chalice/api_server/.chalice
       - ./backend/corpora:/corpora-data-portal/backend/chalice/api_server/chalicelib/corpora
       - ./backend/config:/corpora-data-portal/backend/chalice/api_server/chalicelib/config
+      # SecretsManager population relies on oauth json
+      - ./oauth/users.json:/chalice/users.json
```
seve, line 165: copy the users json for consumption

frontend/tests/utils/helpers.ts
```diff
@@ -9,21 +15,30 @@ export async function goToPage(url: string = TEST_URL) {
...
+      ]);
+    } else {
+      await page.click(getText("Log In"));
+      await page.fill('[name="email"]', TEST_EMAIL);
+      await page.fill('[name="password"]', TEST_PASSWORD);
+      await Promise.all([
+        page.waitForNavigation({ waitUntil: "networkidle" }),
+        page.click('[name="submit"]'),
+      ]);
+    }
```
seve, line 41: mock OIDC uses username, Auth0 uses email

scripts/setup_dev_data.sh
```diff
@@ -51,9 +58,12 @@ ${local_aws} secretsmanager update-secret --secret-id corpora/backend/dev/auth0-
     "internal_url": "'"${OIDC_INTERNAL_URL}"'",
     "cookie_name": "cxguser",
     "callback_base_url": "'"${BACKEND_URL}"'",
-    "redirect_to_frontend": "'"${FRONTEND_URL}"'"
+    "redirect_to_frontend": "'"${FRONTEND_URL}"'",
+    "test_account_username": '"${TEST_ACCOUNT_USERNAME}"',
+    "test_account_password": '"${TEST_ACCOUNT_PASSWORD}"'
```
seve, line 63: load into secretsmanager

frontend/Makefile
```diff
@@ -50,9 +50,13 @@ smoke-test-with-local-backend-ci:
 e2e:
 	npm run e2e-$(DEPLOYMENT_STAGE)
 
+.PHONY: e2e-on-local-dev
+e2e-on-local-dev:
```
jgadling, line 54: Just a comment / point for discussion: We currently have make targets that are expected to be run sometimes inside containers and sometimes outside.  I think this has gotten a little confusing / assumes a lot of background for the engineers running `make` commands. I propose that we set up a convention where the Makefile in the root of the repo should expose targets that are meant to be run _outside_ of containers, and the Makefiles in the frontend/backend/other dirs should be expected to run _inside_ their respective containers.

Dockerfile
tihuan: Nice comment!

docker-compose.yml
```diff
@@ -49,6 +49,11 @@ services:
     command: ["./container_init.sh"]
     environment:
       - DEPLOYMENT_STAGE=dev
```
tihuan, line 51: Out of scope question for @jgadling : since we don't have `dev` anymore, given the new deployment flow of `main` automatically deploys to `staging`, should we change `DEPLOYMENT_STAGE` to something else? 

Thank you!

frontend/Makefile
tihuan: Ahh that's a great idea and makes total sense 🔥 Encapsulation FTW 🎉 !

docker-compose.yml
```diff
@@ -49,6 +49,11 @@ services:
...
+      - AWS_REGION=us-west-2
+      - AWS_DEFAULT_REGION=us-west-2
+      - AWS_ACCESS_KEY_ID=dev_access_key_id
+      - AWS_SECRET_ACCESS_KEY=dev_secret_access_key
+      - BOTO_ENDPOINT_URL=http://localstack:4566
```
tihuan, line 56: For readability, would `AWS_SECREMT_MANAGER_URL` work better? I feel `BOTO` is a Python library that implements AWS Secret Manager and in JS world it's just AWS SDK, so using `BOTO_ENDPOINT_URL` in JS world doesn't convey meaning as much as `AWS_SECREMT_MANAGER_URL`? What do you think?

frontend/tests/utils/helpers.ts
tihuan: Hmm OIDC's login button doesn't have `name='submit'`, does `page.click('[name="submit"]'` still work?

If not, yesterday I used:

```ts
await Promise.all([
      page.waitForNavigation({ waitUntil: "networkidle" }),
      page.click(':is(button:has-text("Log In"), button:has-text("Login"))'), // <--- THIS!
    ]);
```

Screen Shot 2021-04-02 at 10 35 45 AM

tihuan (commented):
Looks really good so far! Thanks so much for updating the tests too 😆 🙏

frontend/tests/utils/helpers.ts
```diff
@@ -9,21 +15,30 @@ export async function goToPage(url: string = TEST_URL) {
...
-    await page.click(getText("Log In"));
-
-    await page.fill('[name="email"]', TEST_EMAIL);
-    // await page.fill('[name="password"]', password);
-
-    await Promise.all([
-      page.waitForNavigation({ waitUntil: "networkidle" }),
-      page.click('[name="submit"]'),
-    ]);
+    if (TEST_ENV === "happy") {
```
tihuan, line 25: I think we can simplify this with Playwright's new selector too!

```ts
await page.fill(`[name="email"], [name="Username"]`, TEST_EMAIL);
await page.fill(`[name="password"], [name="Password"]`, TEST_PASSWORD);
```

Found this yesterday 😆 

https://playwright.dev/docs/selectors#selecting-elements-matching-one-of-the-conditions

docker-compose.yml
jgadling: I think we've got a lot of cleanup to do for the config system in general, but we should probably tackle that separately. poking @mbarrien for visibility.

tihuan: That makes sense 👍 ! Thanks @jgadling 😄

frontend/tests/utils/helpers.ts
seve: Wicked, I'll do that now, same with the submit button!

frontend/tests/utils/helpers.ts
seve: For OIDC we use value=login, auth0 uses name-submit

docker-compose.yml
seve: I'm probably going to be leaving it, so we don't accidentally forget it in the case grep'ing for BOTO_ENDPOINT.  We could also rename it so it's named after what its value represents and not where its consumed

tihuan: Ah I see! Yeah sounds like if we change it, it'll have to be across all languages and systems. SGTM!

tihuan (approved):
LGTM!! I see that there's 19K delete on `package-lock.json`, now I wonder if that's because we're using different versions of npm? Mine is `npm --version // 7.76`!

PR #1001, opened 2021-03-22
Unify DNS for local dev

jgadling:
** Experimental! **

This is a draft PR to get some feedback from the team about usability.

We've been having some trouble getting playwright to work properly when run against local dev due to some inconsistencies in the way DNS works inside a docker network, and on our laptops.

When containers connect to each other, they use service names defined in docker-compose.yml (for example, "backend", and "database"), but when we connect to the containers from our laptops, we use the 'localhost' name with different ports for each container. This means that OAuth redirects inside the docker network need to be different than redirects for browsers running on our laptops.

This PR makes the container names consistent inside and outside of the docker network by adding /etc/hosts entries to our systems to create aliases for {service_name}.corporanet.local, and using docker network aliases inside the docker network with the same name. 

After pulling this branch down, engineers will need to re-run `make local-init` to apply all the necessary changes to their environment.

The `corporanet.local` suffix allows us to use a Samesite=Lax policy for the session cookie. Without domain suffixes, browsers don't allow sending cookies to the backend on redirects.

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/DEV_ENV.md
+++ b/DEV_ENV.md
@@ -4,7 +4,7 @@
 1. [install docker](https://docs.docker.com/get-docker/). If brew is installed run `brew install docker`.
 1. [install chamber](https://github.com/segmentio/chamber). If brew is installed run `brew install chamber`. (This is needed for running functional tests.)
 1. From the root of this repository, run `make local-init` to build and run the dev environment. The first build takes awhile, but subsequent runs will use cached artifacts.
-1. Visit [http://localhost:5000](http://localhost:5000) to view the backend, and [http://localhost:8000](http://localhost:8000) for the frontend.
+1. Visit [http://backend.corporanet.local:5000](http://backend.corporanet.local:5000) to view the backend, and [http://frontend.corporanet.local:8000](http://frontend:8000) for the frontend.
 1. `make local-dbconsole` starts a connection with the local postgresql db.
 1. **Open the source code and start editing!**
   - Modify code in the `frontend/src` directory, save your changes and the browser will update in real time.
--- a/Makefile
+++ b/Makefile
@@ -89,8 +89,20 @@ local-ecr-login:
 		aws ecr get-login-password --region us-west-2 --profile single-cell-dev | docker login --username AWS --password-stdin $$(aws sts get-caller-identity --profile single-cell-dev | jq -r .Account).dkr.ecr.us-west-2.amazonaws.com; \
 	fi
 
+.PHONY: local-hostconfig
+local-hostconfig: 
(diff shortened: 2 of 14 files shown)


tihuan:
Hi @jgadling ! Just wondering when this PR will be ready to merge 😄 Thank you so much again for the fix!

jgadling:
> Hi @jgadling ! Just wondering when this PR will be ready to merge 😄 Thank you so much again for the fix!

I just haven't had a chance to sort out the test failures, but it's otherwise good to go. I probably won't have time for that until tomorrow. If someone else has the bandwith, go for it.

tihuan:
Ooh perfect! OK I'll take a look at the failed tests now. Thank you!

tihuan:
@Bento007 not urgent, but if you have time could you help take a quick look at why the unit test on `session` cookie didn't get `Secure` for cookie even though, `True` seems to be passed? https://github.com/chanzuckerberg/corpora-data-portal/pull/1001/files#diff-63893bfe0ed0a90f3c17e64ae8d17d4a41894ecb4a296f9146fa94b24a053d14R61

Thank you!

tihuan:
Hi @jgadling ! Quick update: @Bento007 and I chatted about the failing unit test. 

It seems like the reason it's failing is because unit tests run inside the containers, which always has `DEV_MODE_COOKIES=1`, thus `require_secure_cookies = False`, but the session cookie unit test still expects `Secure`, thus failing.

So we're wondering if it's easy to run unit tests on HTTPS, using the already generated cert? Because if that works, then we can remove the code:

```py
if os.getenv("DEV_MODE_COOKIES"):
            require_secure_cookies = False
```
?

Context: https://chanzuckerbergteam.slack.com/archives/C01CPJURT0S/p1616699680039000

jgadling:
> So we're wondering if it's easy to run unit tests on HTTPS, using the already generated cert? Because if that works, then we can remove the code:

I think there might be a simpler option ... PR in a sec.

tihuan (approved):
LGTM!! Thanks so much for all the fixes, @jgadling 🤩 👏 ✨ !

Love the solution to reset `DEV_MODE_COOKIES` to undefined in test command, so we don't need to alter the build env variable config 💡

jgadling:
> LGTM!! Thanks so much for all the fixes, @jgadling 🤩 👏 ✨ !
> 
> Love the solution to reset `DEV_MODE_COOKIES` to undefined in test command, so we don't need to alter the build env variable config 💡

I think we need a better solution soon... but this should unstick us for the moment :'(

PR #1030, opened 2021-03-31
increase batch memory

srm78:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.happy/terraform/modules/batch/main.tf
+++ b/.happy/terraform/modules/batch/main.tf
@@ -10,7 +10,7 @@ resource aws_batch_job_definition batch_job_def {
 {
   "jobRoleArn": "${var.batch_role_arn}",
   "image": "${var.image}",
-  "memory": 230000,
+  "memory": "${var.batch_container_memory_limit}",
   "environment": [
     {
       "name": "ARTIFACT_BUCKET",
--- a/.happy/terraform/modules/batch/main.tf
+++ b/.happy/terraform/modules/batch/main.tf
@@ -10,7 +10,7 @@ resource aws_batch_job_definition batch_job_def {
 {
   "jobRoleArn": "${var.batch_role_arn}",
   "image": "${var.image}",
-  "memory": 28000,
+  "memory": 230000,
   "environment": [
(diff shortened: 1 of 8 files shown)


.happy/terraform/modules/batch/main.tf
```diff
@@ -10,7 +10,7 @@ resource aws_batch_job_definition batch_job_def {
 {
   "jobRoleArn": "${var.batch_role_arn}",
   "image": "${var.image}",
-  "memory": 28000,
+  "memory": 230000,
```
jgadling, line 13: 1) Can/should we make this configurable per environment?
2) Let's update the vcpu setting to match the old config as well.

srm78: vcpu setting is same as old config. i'll make this configurable per environment.

maniarathi: Random question -- is this field not deprecated? https://docs.aws.amazon.com/batch/latest/APIReference/API_ContainerProperties.html#Batch-Type-ContainerProperties-memory

srm78: @maniarathi for ec2 resources it is still supported. not for fargate though.

.happy/terraform/modules/batch/main.tf
```diff
@@ -10,7 +10,7 @@ resource aws_batch_job_definition batch_job_def {
 {
   "jobRoleArn": "${var.batch_role_arn}",
   "image": "${var.image}",
-  "memory": 28000,
+  "memory": "${var.batch_container_memory_limit}",
```
jgadling, line 13: should this be an int or a string? can we remove the quotes?

jgadling (approved):
Approved once these changes are tested.

.happy/terraform/modules/batch/main.tf
srm78: yes. quotes needs to be removed. will test on rdev and update the PR.

Issue #988, opened 2021-03-17
Using auto-fill for Contact in Create Collection sometimes disables the Create button ?


brianraymor:
Contact was auto-filled on chrome. `Create` is disabled:

![Screen Shot 2021-03-16 at 6.05.01 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/ef80ae5d-c9fb-48e5-841b-6a2e56da24e2)

Manually typed in the same contact information which enables `Create`:

![Screen Shot 2021-03-16 at 6.05.27 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/c0a7a8fd-0d1c-432d-bd3b-9f8c58f2e973)

maniarathi:
@brianraymor -- I tagged @tihuan about this in his PR but this is still an open problem in prod. Chrome does some weird overrides such that even though autocomplete is disabled, Chrome still allows for auto-complete. This will likely require some more digging into either to turn off auto-complete or to allow auto-complete so does it make sense to label this as a spike and move it to the product backlog to be prioritized later?

brianraymor:
Also see [Turning Off Autocomplete in Chrome with React](https://medium.com/paul-jaworski/turning-off-autocomplete-in-chrome-ee3ff8ef0908)

tihuan:
Morning all! I found another 2021 solution Tuesday night and sent out a new PR for it 😄 

https://github.com/chanzuckerberg/corpora-data-portal/pull/1073

https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off/30873633#30873633

PTAL thank you!

maniarathi:
Thank you @tihuan !!

PR #1033, opened 2021-03-31
thuang-disable-create-collection-autocomplete

tihuan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/988

#988 

### Reviewers
**Functional:** 
@seve 

**Readability:** 
@MDunitz 

---

## Changes
- add
- remove
- modify

The form should no longer provide autocomplete suggestion for any field!

Screen Shot 2021-03-31 at 4 57 20 PM


MDunitz (approved):
lgtm

tihuan:
Thanks so much, @MDunitz ! Mergingg

maniarathi:
Hey @tihuan this doesn't actually work in Chrome unfortunately :( .. I did a tiny bit of research into this and it seem to be a persistent problem: https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off. Just an FYI.. I will check in with Brian to see if we want to still tackle this or accept it as a bug.

Quick question though -- do you have a quick resource on how to tackle _accepting_ auto complete info?

tihuan:
Ah thanks for catching this, @maniarathi ! Looks like form level `autocomplete="off"` was not enough 🤦 

I did some more research and found another solution suggested in 2021 by someone on Stackoverflow:

https://github.com/chanzuckerberg/corpora-data-portal/pull/1073

Seems to be working better now! PTAL! Thank you!

PR #1036, opened 2021-04-01
Fix: oidc runs on port 8443 rather than portmapping

jgadling:
### Reviewers
**Functional:** 
This makes the OIDC app run natively on port 8443 instead of mapping the container's internal port 443 to the OSX hosts's port 8443. This makes it possible for other containers to reach the oidc service on port 8443 as well as engineers' web browsers

**Readability:** 

---

## Changes
- add
- remove
- modify


tihuan (approved):
Woop woop! LGTM!! Thanks so much, @jgadling 🎉 👏 !!

seve (approved):

mbarrien (approved):

PR #1046, opened 2021-04-05
Handle workspaces without any latest config version

mbarrien:
The happy path script breaks if a previous run created a stack's workspace
without actually creating any workspace state or workspace config versions,
and you try to delete that workspace.

This PR prevents us from crashing in these cases, instead printing a warning
that the workspace is empty. Since the workspace is empty, this should not
be harmful.

### DIFF
--- a/scripts/happy
+++ b/scripts/happy
@@ -266,6 +266,9 @@ class Stack:
 
     def destroy(self):
         self._ensure_workspace()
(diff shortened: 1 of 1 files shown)


jgadling (approved):

MDunitz (approved):
lgtm

PR #1065, opened 2021-04-06
remove Flush=true from logs

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:** 

---

## Changes
-  Flush=true from logs


MDunitz (approved):
lgtm

PR #1073, opened 2021-04-07
thuang-988-disable-autocomplete

tihuan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/988

#988 

Related to https://github.com/chanzuckerberg/corpora-data-portal/pull/1033

### Reviewers
**Functional:** 
@seve 

**Readability:** 
@Bento007 

---

## Changes
Following https://stackoverflow.com/a/30873633 to add `type="search"` and `autocompete="off"` to each input field. This seems to stop autocomplete for me!

Seems like Chrome (and any browser) has the right to ignore `autocomplete="off"` for user protection, so the latest workaround is the combination above 🤞 

Another backup option is https://flaviocopes.com/react-form-browser-autofill/, which is not ideal, since it touches things that should be controlled by React, but should reliably work. Definitely the last resort and not a scalable solution

## QA steps (optional)
demo

### DIFF
--- a/frontend/src/components/common/Form/Input/style.ts
+++ b/frontend/src/components/common/Form/Input/style.ts
@@ -1,4 +1,4 @@
-import { Icon, Label } from "@blueprintjs/core";
+import { Icon, InputGroup, Label } from "@blueprintjs/core";
 import { DARK_GRAY, PT_GRID_SIZE_PX } from "src/components/common/theme";
 import styled from "styled-components";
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/Form/Input/style.ts
```diff
@@ -25,3 +25,9 @@ export const IconWrapper = styled.div`
 export const StyledIcon = styled(Icon)`
   margin-right: ${PT_GRID_SIZE_PX}px;
 `;
+
+export const StyledInputGroup = styled(InputGroup)`
+  && input {
```
tihuan, line 30: This is needed because blueprint applies border-radius to `type="search"`

Bento007 (approved):

seve (approved):

PR #1078, opened 2021-04-09
thuang-fix-nextjs-multiservers

tihuan:
### Reviewers
**Functional:** 
@seve 
**Readability:** 
@maniarathi 

---

## Changes

Fixes deployed FE containers sometimes not being able to resolve static files such as `/_ssgManifest.js` and `buildManifest.js`, returning 404

Screen Shot 2021-04-09 at 8 47 52 AM

The root cause is that every Nextjs prod build has a unique build id for their static assets, which means if we have multiple FE instances running in a cluster, load balancer will sometimes route a static asset request to a different instance, which won't have the static asset with that build id.

The solution is to have build id based on the last commit hash, so the build id is the same across different FE instances!

Sources:

https://github.com/vercel/next.js/issues/18389
https://nextjs.org/docs/api-reference/next.config.js/configuring-the-build-id

After deploying to staging and prod, we should no longer get 404 for `ssgManifest.js` and `buildManifest.js`

## Known Issues
N/A

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -1,5 +1,9 @@
 const withImages = require("next-images");
 
+const execSync = require("child_process").execSync;
+
+const lastCommitCommand = "git rev-parse HEAD";
+
 const configs = require(__dirname + "/src/configs/configs.js");
 const nodeEnv = require(__dirname + "/src/common/constants/nodeEnv.js");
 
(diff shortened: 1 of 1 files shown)


maniarathi (approved):
Can you also make a note to remove the workaround once the issue you linked is released in a new version of next.js please? Ideally, next.js will deploy the fix soon and we can remove this code!

Otherwise, LGTM! Thanks!

tihuan:
@maniarathi Ah actually it's a feature not a bug, so Next.js won't fix anything! And the "workaround" is the actual right thing to do according to Next.js' documentation https://nextjs.org/docs/api-reference/next.config.js/configuring-the-build-id 👍 !

PR #1091, opened 2021-04-13
Run more backend containers in prod.

jgadling:
This increases the number of backend workers available to service requests. Each backend container can process a max of 40 parallel requests, and prod is regularly exceeding our available worker count. We should look into adding an autoscaling policy for these services, but this PR will reduce errors in the meantime.

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


maniarathi (approved):
LGTM, thank you!

PR #1096, opened 2021-04-14
e2e tests use secret for current deployment stage

mbarrien:
* Move the test_account_username into a secret, even for test stage
* Modify the playwright global setup to dynamically change secret
  name to read based on deployment stage


MDunitz (approved):
lgtm

PR #1097, opened 2021-04-14
Pass git commit at container build time, use a next.js build id

mbarrien:
* Make the current Git branch, commit hash, and tag description be
  added as Docker build-time arguments.
* Modify containers to add those as labels embedded within the image
* Override next.js build ID to use the commit hash, falling back to
  default random ID behavior if commit hash not provided.

Ideally we'd rather build the next build artifact right into the
container image, so a 2nd container using the same image would have
the same build ID for free (and there's no build time at startup).
However, we cannot run `npm run build` on the bare image because
even running this requires a configs/config.js file to be present,
which it isn't at the time we initially build the Dockerfile.

### DIFF
--- a/Dockerfile.upload_failures
+++ b/Dockerfile.upload_failures
@@ -13,7 +13,7 @@ COPY backend/corpora/dataset_processing ./chalicelib/backend/corpora/dataset_pro
 RUN mkdir -p ./.chalice
 
 ARG HAPPY_BRANCH="unknown"
-ARG HAPPY_COMMIT="unknown"
+ARG HAPPY_COMMIT=""
 LABEL branch=${HAPPY_BRANCH}
 LABEL commit=${HAPPY_COMMIT}
 ENV COMMIT_SHA=${HAPPY_COMMIT}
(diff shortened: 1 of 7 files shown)


tihuan (approved):
WOW!! Thanks so much for making this possible, @mbarrien 🥳 🚀 Super exciting! You the best 🙏 🔥 

LGTM to the max!!

Dockerfile.upload_failures
```diff
@@ -12,4 +12,11 @@ COPY backend/corpora/common ./chalicelib/backend/corpora/common
 COPY backend/corpora/dataset_processing ./chalicelib/backend/corpora/dataset_processing
 RUN mkdir -p ./.chalice
 
+ARG HAPPY_BRANCH="unknown"
```
jgadling, line 15: We also have `make` targets for building images and working with docker-compose. Are we going to (or going to want to?) add these labels to containers in local dev?

mbarrien: Current thought was not to do it for local dev. In our specific case, we run with a dev server locally and with a single replica so this build ID wouldn't come into use there. That said, I am going to modify the Dockerfiles to set HAPPY_COMMIT's default to empty string instead of "unknown" to allow next to resort to its normal random ID.

jgadling (approved):

PR #1105, opened 2021-04-16
Processing container file size check

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:**  @MDunitz 

---

## Changes

- Change download errors to warning
- Catch all error during download
- Upload fails if the downloader has stopped and progress is not 100%
- Check if there is enough disk space to download before starting.
- log status if upload failed


MDunitz (approved):
lgtm

PR #1108, opened 2021-04-19
Fix happy-deploy script

mbarrien:
* Use last 50 deployments instead of first 30 deployments
* Do not break prematurely
* Add debug output and more error output


seve (approved):

seve:
@mbarrien Looks like the merge bot is broken on this one

Issue #1106, opened 2021-04-19
bug(fe-tests): reduce flakiness of frontend tests


seve:
Front end tests can be a little flaky around creating and deleting collections

## To Reproduce
Deploy workflow can randomly fail on smoke-tests

## Expected behavior
Test results are consistent

PR #1113, opened 2021-04-20
bug(fe-tests): reduce flakiness of frontend tests

seve:
### Reviewers
**Functional:**  @tihuan 

---

## Changes
- add
 - Unique identifier for test collections
 - timeout to check for collection to give react query time to clear cache

----

#1106

### DIFF
--- a/frontend/tests/features/collection.test.ts
+++ b/frontend/tests/features/collection.test.ts
@@ -4,7 +4,7 @@ import {
   TEST_ENV,
   TEST_URL,
 } from "tests/common/constants";
-import { goToPage, login, TIMEOUT_MS } from "tests/utils/helpers";
+import { goToPage, login } from "tests/utils/helpers";
 import { getTestID, getText } from "tests/utils/selectors";
 
 const describeIfDeployed =
@@ -20,9 +20,10 @@ const TEST_COLLECTION = {
 describe("Collection", () => {
   describeIfDeployed("Logged In Tests", () => {
     it("creates and deletes a collection", async () => {
+      const timestamp = Date.now();
       await login();
 
-      const collectionId = await createCollection();
+      const collectionId = await createCollection(timestamp.toString());
 
       // Try delete
       await page.click(getTestID("collection-more-button"));
@@ -33,9 +34,13 @@ describe("Collection", () => {
       await goToPage(
         TEST_URL + ROUTES.PRIVATE_COLLECTION.replace(":id", collectionId)
       );
-      await expect(page).not.toHaveSelector(getText(TEST_COLLECTION.name), {
-        timeout: TIMEOUT_MS,
-      });
+
+      // wait five seconds for cache to clear
+      await setTimeout(async () => {
+        await expect(page).not.toHaveSelector(
+          getText(TEST_COLLECTION.name + timestamp)
+        );
+      }, 5000);
     });
 
     describe("Publish a collection", () => {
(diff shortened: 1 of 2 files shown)


frontend/tests/features/collection.test.ts
```diff
@@ -20,9 +20,10 @@ const TEST_COLLECTION = {
 describe("Collection", () => {
   describeIfDeployed("Logged In Tests", () => {
     it("creates and deletes a collection", async () => {
+      const timestamp = Date.now();
       await login();
 
-      const collectionId = await createCollection();
+      const collectionId = await createCollection(timestamp.toString());
```
tihuan, line 26: prefer `String(timestamp)` for best practice, since values that don't have `toString()` method will throw runtime error E.g., `undefined.toString() // runtime error`

frontend/tests/features/collection.test.ts
```diff
@@ -33,9 +34,13 @@ describe("Collection", () => {
       await goToPage(
         TEST_URL + ROUTES.PRIVATE_COLLECTION.replace(":id", collectionId)
       );
-      await expect(page).not.toHaveSelector(getText(TEST_COLLECTION.name), {
-        timeout: TIMEOUT_MS,
-      });
+
+      // wait five seconds for cache to clear
+      await setTimeout(async () => {
```
tihuan, line 39: I remember in Jest, functions in setTimeout() don't work, so tests will always pass!

https://stackoverflow.com/questions/46077176/jest-settimeout-not-pausing-test/46077769

Also, using a hard coded wait time is typically a code smell, because it masks the underlying flakiness and at times makes a test run longer than needed

What if we retry the test until the condition is met or timeout?

E.g.,

```ts

// This replaces 5 second wait, and will retry every 200ms until it either passes, or 
// reaches max retries and throws
await retryAssert(async () => { 
  await expect(page).not.toHaveSelector(
    getText(TEST_COLLECTION.name + timestamp)
  )
}, { maxRetry: 25 })

export async function retryAssert(assert, { maxRetry = 10, waitForMs = 200 }) {
  let retry = 0;

  while (retry < maxRetry) {
    try {
      await assert();

      break;
    } catch (error) {
      retry += 1;

      await page.waitFor(waitForMs);
    }
  }

  if (retry === MAX_RETRY) {
    throw Error("retryAssert() assertion failed!");
  }
}
```

frontend/tests/features/collection.test.ts
```diff
@@ -56,9 +61,11 @@ describe("Collection", () => {
   });
 });
 
-async function createCollection(): Promise<string> {
+async function createCollection(unique?: string): Promise<string> {
```
tihuan, line 64: discussion: should we update the API to accept an optional collection object?

E.g.,

```ts
interface Collection {
  name: string
  description: string
  ...
}

async function createCollection(collection?: Collection = TEST_COLLECTION) {
  ...
}

// callsite
const name = TEST_COLLECTION + timestamp

createCollection({ ...TEST_COLLECTION, name })

...

await expect(page).not.toHaveSelector(
  getText(name)
)
```

tihuan (commented):
Thanks for fixing the flakiness, @seve 😄 👍 

Some minor comments and discussion. PTAL! Thank you!

frontend/tests/features/collection.test.ts
seve: How about a Partial<Collection> that overrides the TEST_COLLECTION?

tihuan: That sounds even better! 🔥

tihuan (approved):
Niiice looks amazing! LGTM 🚀 🚀 🚀

Issue #1051, opened 2021-04-05
Update POST /dp/v1/collections/{collection_uuid}/publish for revision


Bento007:

Bento007:
This ticket could have been a 2. The changes were minimal.

PR #1116, opened 2021-04-21
Allow Publishing of a revision

Bento007:
### Reviewers
**Functional:** @seve 

**Readability:** @seve 

---

## Changes
- modify publish to update a public collection if a public collection already exists.

- #1051
- Return 201 when a revision has successfully replaced a published collection, all columns should be updated and new links should be visible from the public collection page.
- Return 403 if trying to publish when no private collection exists
- Return 403 if the user is not the owner.
- Published a regular private collection still works.


seve (approved):

tests/unit/backend/chalice/api_server/test_v1_collection.py
```diff
@@ -763,3 +763,42 @@ def test__revision_not_owner__403(self):
         headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
         response = self.app.post(test_url, headers=headers)
         self.assertEqual(403, response.status_code)
+
+    def test__publish_revision__201(self):
```
MDunitz, line 767: Could you add links to the original collection and check that one of them is copied over and one of them is deleted to ensure `remove_relationships=True` is working as expected?

Issue #1062, opened 2021-04-06
Merge revision feature flag into curator feature flag when feature is released


brianraymor:
From https://github.com/chanzuckerberg/corpora-data-portal/issues/983#issuecomment-800480654

There was a rationale to define separate feature flags for `cc` (Create Collection) and `auth` in the past. Moving forward, I would recommend that we merge released features to one feature flag such as `curator=true`.  (Or perhaps something slightly harder to guess) Features under development would still have an internal feature flag. When the feature is ready for release to curators, then the internal feature flag would be deprecated and replaced by `curator`.

For example:

1. `?cc=true&auth=true` -> `?curator=true`
2. Gene sets in development -> `?curator=true&gs=true`, make sure this can be enabled from the top level of the frontend.
3. Gene sets released to curators -> `?curator=true` which now includes create collections, auth, gene sets
4. Revision in development -> `?curator=true&revision=true`
5. Revision released to curators -> `?curator=true` which now includes create collections, auth, gene sets, revise collections

## Definition of Done

- [ ] Merge `rc` flag into `curator` flag.
- [ ] Remove `rc` flag.

brianraymor:
This issue was incorrectly closed when addressing [Update feature flags to work on ALL routes](https://github.com/chanzuckerberg/corpora-data-portal/issues/983).

Re-opening.

tihuan:
in prod!

Issue #983, opened 2021-03-15
Update feature flags to work on ALL routes


tihuan:
Currently feature flag `gs` only works on `/collections/:id` route, and doesn't work on homepage route `/`, because the collection page is the only page reading `gs` flag

We should update FE to detect feature flags on all routes, which means adding the related code in `frontend/src/components/AppContainer/index.tsx`

DoD:

1. Feature flags should be set on ALL routes, regardless of which page reads it. For example, homepage route `/?gs=true` should persist the feature flag in `localStorage` even though homepage doesn't use `gs` feature flag

1. Consolidate launched features under `curator=true` see https://github.com/chanzuckerberg/corpora-data-portal/issues/983#issuecomment-800480654 for explanation and examples

brianraymor:
There was a rationale to define separate feature flags for `cc` (Create Collection) and `auth` in the past. Moving forward, I would recommend that we merge released features to one feature flag such as `curator=true`.  (Or perhaps something slightly harder to guess) Features under development would still have an internal feature flag. When the feature is ready for release to curators, then the internal feature flag would be deprecated and replaced by `curator`.

For example:

1. `?cc=true&auth=true` -> `?curator=true`
2. Gene sets in development -> `?curator=true&gs=true`
3. Gene sets released to curators -> `?curator=true` which now includes create collections, auth, gene sets
4. Revision in development -> `?curator=true&revision=true`
5. Revision released to curators -> `?curator=true` which now includes create collections, auth, gene sets, revise collections

@maniarathi and @tihuan - thoughts?

maniarathi:
LGTM!

tihuan:
Thanks for the great idea @brianraymor ! Added that as a DoD item in the description 🔥

brianraymor:
We may want to make the change once gene sets are ready for release. That way, we're sharing a new feature when we require the curators to make the change.

PR #1118, opened 2021-04-22
feat: create curator flag, always check for flags

seve:
### Reviewers
**Functional:**  @tihuan 
---

## Changes
- add
  - General remove params util function
  - Curator flag (`rubber-chicken`)
- modify
  - Check for all feature flags whenever `get()` is called 
  - Remove feature flags when read
---
#1062 #983

### DIFF
--- a/frontend/src/common/featureFlags/index.ts
+++ b/frontend/src/common/featureFlags/index.ts
@@ -1,29 +1,37 @@
 import { get as getLocalStorage } from "src/common/localStorage/get";
 import { BOOLEAN, set as setLocalStorage } from "src/common/localStorage/set";
 import { isSSR } from "../utils/isSSR";
-import { removeParams } from "../utils/removeParams";
 import { FEATURES } from "./features";
 
 const FEATURE_FLAG_PREFIX = "cxg-ff-";
 
-// Checks both URL and localStorage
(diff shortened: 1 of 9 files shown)


frontend/src/common/featureFlags/index.ts
```diff
@@ -1,23 +1,29 @@
...
+      if (params.has(feature)) {
+        const URLValueAsBooleanString =
+          params.get(feature) === "true" ? BOOLEAN.TRUE : BOOLEAN.FALSE;
+        setLocalStorage(FEATURE_FLAG_PREFIX + feature, URLValueAsBooleanString);
+        newlyAddedParams.push(feature);
+      }
+    });
   }
 
+  removeParams(newlyAddedParams);
```
seve, line 26: We need to remove read params, so we aren't running the above loop over and over

tihuan (commented):
Thanks for the detailed comments, @seve ! Overall looks really good!

I created a [PR](https://github.com/chanzuckerberg/corpora-data-portal/pull/1119) for some changes. PTAL and let me know what you think 🙏 

Thank you!

tihuan (approved):
LGTM!! 🚀 🚀 🚀  Thanks Seve!!

PR #1016, opened 2021-03-24
thuang-update-localStorage-boolean-values

tihuan:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify
- `localStorage` values are changed from "`true"` and "`false"` to `"yes"` and `"no"` to match Explorer's convention

Note that feature flags will use `"yes"` and `"no"` from now on too, so please add the relevant feature flags in the URL params to overwrite the values again! E.g., `?auth=true&cc=true` (this part stays the same)

## Definition of Done (from ticket)

Screen Shot 2021-03-24 at 11 40 38 AM

## QA steps (optional)
1. Clear your localStorage for the site and try setting feature flags and cookie banner again!

PR #1119, opened 2021-04-22
thuang-feedback-1062

tihuan:
### Reviewers
**Functional:** 
@seve 

**Readability:** 

---

## Changes
1. Update `removeParams()` function to use web API more
2. Add collection `CURATOR_FEATURES` to host all feature flags we want to enable for curators
3. Split featureFlags/index.ts to have getter and setter (`checkFeatureFlags()`)
4. Check for feature flags at app initiation time `checkFeatureFlags()` in `_app.tsx`. The setter will NOT set `rubber-chicken` to `true` in localStorage, so other feature flags remain the single source of truth. This is to prevent the weird state where in localStorage `cc=true` and `rubber-chicken=false`
5. Remove removing the feature flag params, since getter and setter are now separated

1. feature flags should work -- TESTED
2. Redirect after login should work -- NOT TESTED PLEASE HELP 😆 

## Known Issues
N/A

### DIFF
--- a/frontend/src/common/featureFlags/features.ts
+++ b/frontend/src/common/featureFlags/features.ts
@@ -4,3 +4,5 @@ export enum FEATURES {
   GENE_SETS = "gs",
   CURATOR = "rubber-chicken",
 }
+
+export const CURATOR_FEATURES = [FEATURES.CREATE_COLLECTION, FEATURES.AUTH];
--- a/frontend/src/common/featureFlags/features.ts
+++ b/frontend/src/common/featureFlags/features.ts
@@ -4,5 +4,3 @@ export enum FEATURES {
   GENE_SETS = "gs",
   CURATOR = "rubber-chicken",
 }
-
-export const CURATOR_FEATURES = [FEATURES.CREATE_COLLECTION, FEATURES.AUTH];
--- a/frontend/src/common/featureFlags/index.ts
+++ b/frontend/src/common/featureFlags/index.ts
@@ -1,9 +1,11 @@
 import { get as getLocalStorage } from "src/common/localStorage/get";
 import { BOOLEAN, set as setLocalStorage } from "src/common/localStorage/set";
 import { isSSR } from "../utils/isSSR";
-import { CURATOR_FEATURES, FEATURES } from "./features";
+import { FEATURES } from "./features";
 
 const FEATURE_FLAG_PREFIX = "cxg-ff-";
+
+const allowedKeys = Object.values(FEATURES) as string[];
 
 export function get(key: string): string | null {
   if (isSSR()) return null;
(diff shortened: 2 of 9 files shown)


frontend/src/common/featureFlags/features.ts
```diff
@@ -4,3 +4,5 @@ export enum FEATURES {
   GENE_SETS = "gs",
   CURATOR = "rubber-chicken",
 }
+
+export const CURATOR_FEATURES = [FEATURES.CREATE_COLLECTION, FEATURES.AUTH];
```
seve, line 8: We're planning to deprecate the cc and auth flags, which is why I had them both in code for the moment while Brian communicates the deprecation

frontend/src/common/featureFlags/index.ts
```diff
@@ -1,29 +1,39 @@
...
+    }
+  });
+}
+
+function setFeatureFlag(key: string, value: string) {
+  const URLValueAsBooleanString =
+    value === "true" ? BOOLEAN.TRUE : BOOLEAN.FALSE;
+
+  setLocalStorage(FEATURE_FLAG_PREFIX + key, URLValueAsBooleanString);
+}
```
seve, line 39: There's no check to ee if the key is a valid feature flag, don't we risk dirtying the local storage?

frontend/src/components/CreateCollectionModal/index.tsx
```diff
@@ -41,10 +40,7 @@ const CreateCollection: FC<{
   const [isOpen, setIsOpen] = useState(shouldModuleOpen);
```
seve, line 40: Turns out, this has been broken since #1016

seve: We need to define a new BOOLEAN enum. https://github.com/chanzuckerberg/corpora-data-portal/pull/762#pullrequestreview-546520583

seve (changes requested):

frontend/src/components/CreateCollectionModal/index.tsx
tihuan: ahh great catch! I changed it to `href={`${API_URL}${API.LOG_IN}?redirect=?${QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT}=${BOOLEAN.TRUE}`}`!

frontend/src/common/featureFlags/features.ts
tihuan: Ooh I see! I'll add the checks back at callsites 👍

frontend/src/components/CreateCollectionModal/components/CTA/index.tsx
```diff
@@ -28,7 +29,7 @@ const CTA: FC<Props> = ({ onClose }) => {
           </Button>
           <AnchorButton
             intent={Intent.PRIMARY}
-            href={`${API_URL}${API.LOG_IN}?redirect=?${QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT}=true`}
+            href={`${API_URL}${API.LOG_IN}?redirect=?${QUERY_PARAMETERS.LOGIN_MODULE_REDIRECT}=${BOOLEAN.TRUE}`}
```
tihuan, line 32: use `BOOLEAN.TRUE` instead of hard coded `true`

tihuan:
@seve just updated the PR! PTAL again thank you!!

PR #1121, opened 2021-04-22
increase container memory

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- modify memory from 230000 -> 250000

### DIFF
--- a/.happy/terraform/envs/prod/main.tf
+++ b/.happy/terraform/envs/prod/main.tf
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.cziscience.com"
   backend_url                  = "https://api.cellxgene.cziscience.com"
   stack_prefix                 = ""
-  batch_container_memory_limit = 230000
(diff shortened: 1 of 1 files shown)


.happy/terraform/envs/prod/main.tf
```diff
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.cziscience.com"
   backend_url                  = "https://api.cellxgene.cziscience.com"
   stack_prefix                 = ""
-  batch_container_memory_limit = 230000
+  batch_container_memory_limit = 250000
```
maniarathi, line 16: I would crank it up to 300GB (300000). It's a limit so no need to be shy I think.

maniarathi (approved):

.happy/terraform/envs/prod/main.tf
Bento007:
```suggestion
  batch_container_memory_limit = 300000
```

PR #1123, opened 2021-04-22
Add happy dev environment

mbarrien:


.happy/config.json
```diff
@@ -11,6 +11,14 @@
             "terraform_directory": ".happy/terraform/envs/rdev",
             "log_group_prefix": "/dp/rdev"
         },
+        "stage": {
```
seve, line 14: Is this supposed to be dev?

seve (changes requested):

.happy/config.json
mbarrien: Yes. Thanks for the catch.

seve (approved):
LGTM, but I'd wait for @jgadling to triple check TF

PR #1124, opened 2021-04-22
fix(fe-e2e): wait to see collection on my collections page don't do manual navigation

seve:
### Reviewers
**Functional:** @Bento007 

---

## Changes
- add
  - Wait for automatic navigation
- remove
  - no more manual navigation
- modify
  - more attempts to check that collection is gone 

# Notes

We were navigating the collection page before the browser could send out the delete collection request


tihuan (approved):
LGTM!! Thanks for making the test more robust 💪 🤘

PR #1126, opened 2021-04-22
Make scale test point at staging

mbarrien:
Make output filename match what's read as input below, change step name to match staging.


Bento007 (commented):
Is change mostly superfical?

mbarrien:
Not completely. The filename I changed in an earlier PR didn't match the filename a few lines below the change. This fixes that.

Bento007 (approved):

MDunitz (approved):
lgtm

PR #1130, opened 2021-04-23
feat: Script to get dataset download metrics and output results weekly in Slack

maniarathi:
### Reviewers

**Functional:** @MDunitz  

**Readability:** @Bento007 

---

#### Additions

- Run a script every Sunday at 5pm to generate download metrics about datasets. Example successful run is [here](https://chanzuckerbergteam.slack.com/archives/CV7PFVABC/p1620751087236100).


.github/workflows/dataset-download-metrics.yml
```diff
@@ -0,0 +1,37 @@
...
+          role-duration-seconds: 900
+      - name: Run metrics aggregation script
+        run: |
+          python3 scripts/athena.py | tee dataset_metrics.txt
+      - name: Write to Slack on success
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+        run: |
+          DOWNLOAD_STATS=$(cat dataset_metrics.txt)
+          METRICS_MSG="\`\`\`PROD DATASET DOWNLOAD METRICS: \n${DOWNLOAD_STATS}\`\`\`"
```
MDunitz, line 35: I think slack limits the size of posted messages and this ends up getting split up. To maintain the formatting you may want to split it into ~30 datasets per post request

MDunitz (approved):
lgtm

.github/workflows/dataset-download-metrics.yml
maniarathi: I actually did a few other things to shorten this:
- Consolidated the ID to be the collection id + dataset id.. it was separating out by file type before.
- Removed private collections and private datasets from getting counted
- Set a one week time frame instead of all time.

These changes should make it such that it is less than 30 datasets for now. I'll revisit if it becomes more later. Is this OK with you?

PR #1131, opened 2021-04-23
feat(revision): Edit my-collections page to start and view revisions

seve:
### Reviewers
**Functional:** @tihuan 

**Readability:** @Bento007 

---
currently working on populating description
---

## Changes
- add
  - Revision flag (`rc`: revise-collection)
  - Revision status tag on my collections view
  - option to start or continue revision for published collections
- remove
  - Remove DOI and author information from collection**s** view 
- modify
  - Collection DOI name defaults to doi.org path

## QA steps (optional)

* RDEV: https://revisions-frontend.rdev.single-cell.czi.technology?rubber-chicken=true&rc=true (url has feature flags included)
* Log in and navigate to my-collections
* Revisions are only enabled on published collections

---
## Screenshots
image
image


frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -42,29 +52,44 @@ const CollectionsGrid: FC<Props> = ({
...
-): CollectionResponse[] {
+  collections: RevisionResponse[]
+): RevisionResponse[] {
   return collections?.sort((a, b) => b.created_at - a.created_at) || [];
 }
 
+function findRevisions(collections: CollectionResponse[]): RevisionResponse[] {
+  const revisionMap = new Map<CollectionResponse, number>();
+  collections.forEach((collection) => {
+    revisionMap.set(collection, revisionMap.get(collection) ?? 1);
```
tihuan, line 88: Probably need help understand how this piece of code works in terms of finding whether a collection has a revision or not, since I currently don't have time to test this further 😆 

So far it seems to be looping through `collections` to create `revisionMap`, where key is `collection` and value is always `1`? 

```
// seems like revisionMap.get(collection) can only return `null` or 1?
revisionMap.set(collection, revisionMap.get(collection) ?? 1); 
``` 

And then:

```
// revision will always be false, because number is always 1?
return { ...collection, revision: number > 1 } 
```

seve: Total bug, I haven't been able to test it since there's no backend route yet.  I think it should be:
```suggestion
    revisionMap.set(collection, (revisionMap.get(collection) ?? 0) + 1);
```

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -42,29 +52,44 @@ const CollectionsGrid: FC<Props> = ({
...
 function sortByCreatedAtDescending(
-  collections: CollectionResponse[]
-): CollectionResponse[] {
+  collections: RevisionResponse[]
+): RevisionResponse[] {
   return collections?.sort((a, b) => b.created_at - a.created_at) || [];
 }
 
+function findRevisions(collections: CollectionResponse[]): RevisionResponse[] {
+  const revisionMap = new Map<CollectionResponse, number>();
```
seve, line 86: Revisions don't have the same visibility

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -42,29 +52,44 @@ const CollectionsGrid: FC<Props> = ({
...
-        <CollectionRow
-          id={id}
-          key={id + visibility}
-          visibility={visibility}
-          accessType={accessType}
-        />
-      );
+  const revisions = revisionsEnabled ? findRevisions(collections) : false;
+  sortByCreatedAtDescending(revisions || collections).forEach(
+    ({ id, visibility, revision }) => {
```
seve, line 61: revision can be true/false/undefined

seve: true: means there is a revision
false: means there could but isn't
undefined: we're on 

- the my collections page, no revision logic

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
tihuan: Got it! 

Since the key is a Collection response, which is a pointer to a POJO, it's gonna be unique for every item in `collections`? In other words, what props in a collection object do we use to identify an original and a revision belong to the same collection? Is it `collection.id` or something?

E.g.,

```ts
// do collection and its revision share the same `id`?
const collection = { id: 1, name: 1, ... }
const collectionRevision = { id: 1, name: 'revision', ... }


const collections = [collection, collectionRevision, ... ]

// false, since they are different pointers
console.log(collection === collectionRevision)

// true, since they share the same id, thus if `revisionMap`'s key is `id`, the count will aggregate correctly?
console.log(collection.id === collectionRevision.id) 
```

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -42,29 +52,44 @@ const CollectionsGrid: FC<Props> = ({
...
-    if (!displayVisibility || visibility === displayVisibility) {
-      collectionElements.push(
-        <CollectionRow
-          id={id}
-          key={id + visibility}
-          visibility={visibility}
-          accessType={accessType}
-        />
-      );
+  const revisions = revisionsEnabled ? findRevisions(collections) : false;
```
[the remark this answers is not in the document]
seve: switched to enum! h/t @MDunitz

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -42,29 +56,61 @@ const CollectionsGrid: FC<Props> = ({
...
+  // If revisions are disabled just return the array with objects' revisions status set to disabled
+  if (!revisionsEnabled) return newCollections;
+
+  const revisionMap = new Map<string, RevisionResponse>();
+
+  newCollections.forEach((collection) => {
+    const revisionObj = revisionMap.get(collection.id) || collection;
+    if (revisionObj.revision !== REVISION_STATUS.STARTED) {
+      revisionObj.revision =
+        revisionObj.revision === REVISION_STATUS.DISABLED
```
tihuan, line 107: ahh I see, so if a public collection the user owns has a revision, it will have the same id with `visibility === PRIVATE` , so the second time the loop encounters the same id, it will bump the revision enum from `NOT_STARTED` to `STARTED`? 

That works! But I wonder if there's a more explicit way of doing it. Not a blocker though!

tihuan (approved):
LGTM!! Thanks for the comments, Seve!!

tihuan (approved):
LGTM!! Thanks for the comments, Seve!!

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
seve: That's it! We could probably ask for a revision property in the future. I think a /user/collections endpoint would be higher priority though

PR #1148, opened 2021-04-28
Cleanup upload processing Dockerfile

mbarrien:
Several cleanups for the upload processing Dockerfile to shrink the size of the image and for correctness
* Make the apt-get install quiet, to make ci output less verbose
* Unconditionally install BiocManager, to simplify the following commands that install SingleCellExperiment and LoomExperiment
* Use the Ubuntu way of managing python/pip alternatives rather than hardcoding symlinks
* Disable the pip cache on installation to prevent caching downloaded pip wheels to shrink the image
* Do not install npm solely for installing cellxgene, since only its backend its used. The right solution would be to make cellxgene support a backend only installation, but in lieu of that for now, we make a dummy /usr/bin/npm that just generates the throwaway files needed by the make pydist-czi-hosted step.
* Deletes the checked out cellxgene Git repo after the binary build, since make install-dist will install it into a global location independent of the repo. Doing this along with avoiding the pip cache saves 1.6GB from the image.

These changes are in preparation for splitting the image into 2 parts (with the Bioconductor libraries pre-compiled), but that split is not performed here.


Bento007 (approved):
you're a wizard

PR #1152, opened 2021-04-28
refactor: Remove Chalice

mbarrien:
* Remove Chalice from the app
  * Instead run the web application directly behind gunicorn
* Converted lambdas except for the cloudfront invalidator to directly run without Chalice.
  * Remove all lambda Makefiles for deployment, rely completely on happy path script
* Moved all code to be consistently in the backend.corpora Python namespace, instead of sometimes in corpora as top level.
  * Removes need for sys.path manipulation
  * Run scripts like populate_db as Python modules instead of calling directly, to handle more consistent backend.corpora namespace properly.
* Require explicit calls to dbconnect function decorator on all functions instead of relying on db_session_manager context manager, which isn't compatible with being called from Flask directly (since there is no common location where we can insert a wrapper easily).
* Use Flask app directly instead of using flask test_request_context as an adapter wrapper around Chalice
  * Preserve the authentication with cxgpublic dummy header by using WSGI middleware to inject the header; Flask request header field is immutable.
* Make unit tests use Flask test_client directly instead of requests library wrapped around chalice.
  * Check for status_code directly instead of raise_for_status() (since raise_for_status() only available if using requests)
  * Ensure unittest super().setUp() and super().setUpClass() are properly called.
  * Make flask test_client instantiated at every unittest setUp() instead of at setUpClass() time.
  * Force flask test_client to not manage cookies, to preserve previous behavior.
  * Pass headers argument by kwarg instead of positional, to satisfy Flask test_client
* Use assertIn instead of assertTrue(x in y)
* Fix initialization script paths

Passes all unit tests and linters.


mbarrien:
I believe this is working now. Final tweaks needed were to increase the gunicorn timeout (5 seconds was too short, raised to gunicorn's default of 30 seconds) and fixing the PYTHONPATH when using alembic.

Bento007 (approved):

MDunitz:
Tweaked some failing tests due to merge conflicts and a breaking release from pandas.

PR #1153, opened 2021-04-28
Tsmith/1056 migrate

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:** @seve 

---

## Changes
- add publish and original_id to dataset table
- modify public datasets to have published=True
- modify private datasets to have published=False
- publish set to true when the collection is published
- The new columns original_uuid and published are in the dataset table
- All public datasets have published set to true while private collections are set to false.
- When a collection is published, all of its untombstoned datasets are set to published = True
- Verify the frontend displays all of the collections.


MDunitz (approved):
lgtm

PR #1165, opened 2021-05-04
fix: optimize environment variable lookup in corpora_config

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:** 

---

## Changes
- modify corpora_config to avoid excessive dictionary lookups for the same variable.


MDunitz (approved):
lgtm

PR #1168, opened 2021-05-05
feat: (part 2 of 3), Tombstone published, and delete unpublished datasets

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:** @seve 

---

## Changes
- Use DELETE /v1/datasets/{dataset_id} on a private published dataset to tombstone the dataset. Its children are deleted from the database, artifact bucket, and cellxgene bucket.
- Use DELETE /v1/datasets/{dataset_id} on a private unpublished dataset to delete the database entry, and files in the cellxgene and artifact bucket
- Tombstoning a collection also deletes the cellxgene bucket resources associated with the datasets.
- Adde help function for generating and verifying test data stored in s3 buckets.
- Move Delete code from cxg_admin to Dataset for reusability
- Move upload code from processing.py to DatasetAsset for reusability
- Broke down Dataset.dataset_and_asset_deletion into smaller function (Dataset.delete, Dataset.asset_delete) to handle more deletion use cases
- 
- This is part of the work required to delete a dataset when canceling a revision.


MDunitz (approved):
lgtm

PR #1170, opened 2021-05-06
fix: break test_datasets into more granular test files

MDunitz:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- modify the test_datasets file -- it was nearing 500 lines and getting fairly unruly. I broke the test out based on the general functionality they were testing


Bento007 (approved):
Thank you. We badly needed some test cleanup

Issue #1145, opened 2021-04-27
Update POST /dp/v1/collections/{collection_uuid}/publish for refreshed datasets


Bento007:
## Need
Curators may publish datasets in the private revision of a published collection

## Approach
- Update POST /dp/v1/collections/{collection_uuid}/publish for refreshed datasets to support publishing refreshed dataset.
- For clarity `dataset_v1` refers to the published dataset in the public collection, while `dataset_v2` refers to the unpublished dataset in the collection revision which is a revision of `dataset_v1`. 
- `dataset_v2` will have the columns `published==False` and `original_uuid==dataset_v1`
- Replace `dataset_v1` with `dataset_v2` by doing the following:
   1. Delete `dataset_v1` artifacts, deployments, and processing_status.
   2. In the explorer bucket, rename the cxg data from _/{dataset_v2.uuid}/.*_ to _/{dataset_v1.uuid/.*_
   3. Copy columns from `dataset_v2` to `dataset_v1`.
   4. Link all `dataset_v2` artifacts, deployments, and processing_status to `dataset_v1`
   5. Delete `dataset_v2`
- Do this for all dataset with `published==False` and `original_uuid!=None`

## Definition of Done
- A revision can publish a refreshed dataset
- The old dataset info will no longer be accessible
- The new dataset is one revision higher than the old dataset after publishing
- The new dataset has published=True after publishing
- The new dataset has original_uud=None after publishing

PR #1177, opened 2021-05-11
feat: Publish updates to datasets when publishing a collection

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:** @seve 

---
Work for #1145

rdev deployment https://tsmith-revision-frontend.rdev.single-cell.czi.technology 

## Changes
- moved revision related publishing tests to test_revision
- add logic to copy revision datasets to published dataset
- add logic to delete tombstones dataset when publishing a revision
- published tombstones are not returned in collection details.
- remove Dataset.delete_explorer_cxg_object_from_s3 and replacing with Dataset.asset_deletion()
- add check on backend to verify at least one dataset exists in a private collection before publishing.
- add commit flag to functions that commit to database. This gives us better control over when changes are committed to the database. This will make it easier to roll back changes if publish fails.
- add after session commit hook to delete s3_assets
- moving s3_client and s3_resource to s3_bucket
- deleting a refreshed dataset, restores the original dataset to the revision.

- A revision can publish a refreshed dataset
- The old dataset info will no longer be accessible
- The new dataset is one revision higher than the old dataset after publishing
- The new dataset has published=True after publishing
- The new dataset has original_uud=None after publishing

## Known Issues
- additional work https://github.com/chanzuckerberg/corpora-data-portal/issues/1249
  -  grey out publish button if no datasets are in a collection
  - add pop up before publishing a revision
  - add toast after publishing a revision.


backend/corpora/common/entities/collection.py
```diff
@@ -182,9 +182,35 @@ def publish(self):
...
-            if dataset.original_id:
-                "skip modified datasets"
-                continue  # TODO: expand to support tombstone and refresh corpora-data-portal/1177
+            ds = Dataset(dataset)
+            ds_original_id = ds.original_id
+            if ds_original_id:
+                original = Dataset.get(self.session, ds_original_id)
+                ds_tombstone = ds.tombstone
+                ds_revision = ds.revision
+                if (public_collection.id, public_collection.visibility) == (
```
MDunitz, line 191: There is a lot going on in this if statement, what if you pulled it out into a separate check function?

Bento007: that would be a lot of variables going into that functions. How about this 
```
if all([public_collection.id == original.collection_id,
          public_collection.visibility == original.collection_visibility,
          ds_tombstone or ds_revision > original.revision]):
```

tests/unit/backend/chalice/api_server/test_revisions.py
```diff
@@ -290,6 +294,35 @@ def verify_datasets(self, actual_body, expected_dataset_ids):
...
+        self.app.delete(f"/dp/v1/datasets/{rev_dataset_id}", self.headers)
+        self.publish_collection(self.rev_collection.id)
+        dataset = Dataset.get(self.session, pub_dataset.id, include_tombstones=True)
+        self.assertTrue(dataset.tombstone)
+        for s3_object in published_s3_objects:
+            self.assertS3FileDoesNotExist(*s3_object)
+
+    def test__with_revision_with_refreshed_datasets__OK(self):
+        """"publish a revision with refreshed datasets"""
+        """publish a revision with delete datasets"""
```
MDunitz, line 321:
```suggestion
```

Bento007 (commented):
- add commit flag to line 172 of collection.py to prevent early commit.
- don't delete the datset_assets until after the DB changes have been commited

seve (approved):
Readability ✅

MDunitz (approved):
lgtm - one thought it might be worth adding a test of the rollback functionality to make sure if something goes wrong at the end of publishing a revision it is able to rollback to the previous state

PR #1192, opened 2021-05-18
fix: deleting a dataset during a revision crashes collection page

Bento007:
### Reviewers
**Functional:** @MDunitz 

**Readability:** @seve 

---

## Changes
- modify deleting a published dataset in a revision to set the tombstone flag rather than delete the dataset children. This makes it easier for the frontend to track if the datasets have been deleted. It also avoids making refactoring changes to handle cases where the datasets do not have any assets or processing status.


seve (approved):

MDunitz (commented):
lgtm

PR #1205, opened 2021-05-21
fix: commit change in script

MDunitz:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes

- modify script to update ownership of collections


scripts/cxg_admin.py
```diff
@@ -104,7 +104,7 @@ def transfer_collections(ctx, curr_owner, new_owner):
 
     with db_session_manager() as session:
         collections = session.query(DbCollection).filter(DbCollection.owner == curr_owner).all()
-
+        new_owner_collections = session.query(DbCollection).filter(DbCollection.owner == new_owner).all()
         if collections is not None:
```
Bento007, line 108: Is this ever true, [all](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.all) never returns None. It returns a list.

scripts/cxg_admin.py
```diff
@@ -104,7 +104,7 @@ def transfer_collections(ctx, curr_owner, new_owner):
 
     with db_session_manager() as session:
         collections = session.query(DbCollection).filter(DbCollection.owner == curr_owner).all()
```
Bento007, line 106:
```suggestion
        collections = session.query(DbCollection).filter(DbCollection.owner == curr_owner)
```

scripts/cxg_admin.py
Bento007:
```suggestion
        if collections.all():
```

scripts/cxg_admin.py
```diff
@@ -116,7 +116,10 @@ def transfer_collections(ctx, curr_owner, new_owner):
                 .filter(DbCollection.owner == curr_owner)
```
Bento007, line 116: I think this needs to be unpacked             
```updated = (
                *session.query(DbCollection)
                .filter(DbCollection.owner == curr_owner)
                .update({DbCollection.owner: new_owner})
            )
```

It can be shorter:
```
update = (*collections.update({DbCollection.owner: new_owner})

File: scripts/cxg_admin.py [hunk 4]
```diff
@@ -104,7 +104,7 @@ def transfer_collections(ctx, curr_owner, new_owner):
 
     with db_session_manager() as session:
         collections = session.query(DbCollection).filter(DbCollection.owner == curr_owner).all()
-
+        new_owner_collections = session.query(DbCollection).filter(DbCollection.owner == new_owner).all()
```
Bento007, line 107:
```suggestion
        new_owner_collections_count = len(session.query(DbCollection).filter(DbCollection.owner == new_owner).all())
```

scripts/cxg_admin.py
```diff
@@ -116,7 +116,10 @@ def transfer_collections(ctx, curr_owner, new_owner):
                 .filter(DbCollection.owner == curr_owner)
                 .update({DbCollection.owner: new_owner})
             )
+            session.commit()
             if updated > 0:
+                collections = session.query(DbCollection).filter(DbCollection.owner == new_owner).all()
+                click.echo(f"{new_owner} previously owned {len(new_owner_collections)}, they now own {len(collections)}. {updated} collections were updated")
```
Bento007, line 122:
```suggestion
                click.echo(f"{new_owner} previously owned {new_owner_collections_count}, they now own {len(collections)}.)
```

Bento007: I would iterate over `updated` and print out the `collection.id` for all of the modified collections.

Bento007 (changes requested):
There were a few things I missed from last time.

scripts/cxg_admin.py
MDunitz: its shorter but I think it loses some clarity about what is actually happening

Bento007 (approved):

PR #1215, opened 2021-06-02
fix: use a different database for testing corpora_orm

### LINKED ISSUE #1214: Remove test tables from db, update unit tests to prevent recreation
## Need
We need to remove unnecessary test tables from the staging and prod databases and update the unit tests to prevent them from being recreated

## Approach
- create migration removing the tables
- no longer import Base in the corpora_orm unit tests
- create a new database when removing tests

## DOD
- There are no parent or child tables in the staging or prod databases

Bento007:
### Reviewers
**Functional:** @MDunitz 

---
Before we were using the same SQLAlchemy Base for corpora_orm and the tests. This cause tests tabled to be added to the staging and production databases. To fix this, the tests now create their own database and use their own SQLAlchmey base when running the tests. The database and data are cleaned up after the test.

## Changes
- add Test_Base to avoid confusion with corpora_orm.Base 
- add DBSessionMaker.reset to reset the DBSessionMaker
- Create a new database to run corpora_orm tests on.

- related to #1214. It prevents the tables from being created in the future.
-

### DIFF
--- a/backend/corpora/common/utils/db_session.py
+++ b/backend/corpora/common/utils/db_session.py
@@ -18,13 +18,20 @@ class DBSessionMaker:
 
     def __init__(self, database_uri: str = None):
         if not self.engine:
-            database_uri = database_uri if database_uri else CorporaDbConfig().database_uri
-            self.engine = create_engine(database_uri, connect_args={"connect_timeout": 5})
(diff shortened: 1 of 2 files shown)


MDunitz (approved):
lgtm

PR #1216, opened 2021-06-03
feat: search for dataset by  explorer url db changes

MDunitz:
### Reviewers
**Functional:** 
- @Bento007 
**Readability:** 
---

## Changes
- add fields explorer_url and explorer_s3_uri to dataset table
- remove deployment_directories and associated code
- modify geneset csv creation function and s3_object deletion functions to use the explorer_s3_uri as the base to get the object name


backend/corpora/common/corpora_orm.py
```diff
@@ -247,13 +247,12 @@ class DbDataset(Base, AuditMixin):
     tombstone = Column(Boolean, default=False, nullable=False)
     original_id = Column(String)
     published = Column(Boolean, default=False)
+    explorer_s3_uri = Column(String)
```
maniarathi, line 250: I'm curious -- in the upload processing step, we create a row for each artifact created (i.e. H5AD, RDS, and LOOM), but not CXG. And yet we have a CXG `DatasetArtifactFileType` type. Why was that decision made and why link the artifact s3_uri here instead of populating another row in the DatasetArtifact table with the CXG type? Linking it here seems to violate the semantics of that table's purpose.

maniarathi, line 250: I'm curious -- in the upload processing step, we create a row for each artifact created (i.e. H5AD, RDS, and LOOM), but not CXG. And yet we have a CXG `DatasetArtifactFileType` type. Why was that decision made and why link the artifact s3_uri here instead of populating another row in the DatasetArtifact table with the CXG type? Linking it here seems to violate the semantics of that table's purpose.

backend/database/versions/b65dad1f2a7e_add_explorer_field_to_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+down_revision = "78103925d379"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("dataset", sa.Column("explorer_s3_uri", sa.String(), nullable=True))
+    op.add_column("dataset", sa.Column("explorer_url", sa.String(), nullable=True))
+    op.create_index(op.f("ix_dataset_explorer_url"), "dataset", ["explorer_url"], unique=False)
```
maniarathi, line 23: What is this doing and why is it necessary?

backend/database/versions/b65dad1f2a7e_add_explorer_field_to_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+    op.execute("UPDATE dataset SET explorer_url=deployment_directory.url FROM deployment_directory WHERE deployment_directory.dataset_id=dataset.id;")  # noqa E501
+    op.execute("UPDATE dataset SET explorer_s3_uri=CONCAT('s3://hosted-cellxgene-', split_part(explorer_url, '.', 2), '/', split_part(explorer_url, '/', 5), '/') WHERE dataset.explorer_url is not null")  # noqa E501
```
maniarathi, line 25: This isn't going to work for prod no? There's no `prod` value in the cellxgene.cziscience.com.

backend/database/versions/b65dad1f2a7e_add_explorer_field_to_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+        sa.Column(
+            "created_at", postgresql.TIMESTAMP(), server_default=sa.text("now()"), autoincrement=False, nullable=False
+        ),
+        sa.Column(
+            "updated_at", postgresql.TIMESTAMP(), server_default=sa.text("now()"), autoincrement=False, nullable=False
+        ),
+        sa.ForeignKeyConstraint(["dataset_id"], ["dataset.id"], name="deployment_directory_dataset_id_fkey"),
+        sa.PrimaryKeyConstraint("id", name="deployment_directory_pkey"),
+    )
+    op.execute('CREATE EXTENSION IF NOT EXISTS "uuid-ossp";')
```
maniarathi, line 47: What's this doing?

backend/corpora/common/entities/dataset.py
```diff
@@ -169,12 +151,10 @@ def asset_deletion(self):
...
             asset.delete()
 
-    def deployment_directories_deletion(self):
-        for deployment_directory in self.deployment_directories:
-            object_names = get_cxg_bucket_path(deployment_directory)
-            logger.info(f"Deleting all files in bucket {cxg_bucket.name} under {object_names}.")
-            cxg_bucket.objects.filter(Prefix=object_names).delete()
-            self.session.delete(deployment_directory)
+    def delete_explorer_cxg_object_from_s3(self):
+        object_name = self.explorer_s3_uri.split("/")[3].split(".")[0]
```
maniarathi, line 155: Seems like the `get_cxg_bucket_path` function below is a bit more robust than the above written parsing of the string. Why not use it?

backend/corpora/common/entities/dataset.py
```diff
@@ -185,7 +165,10 @@ def new_processing_status() -> dict:
         }
 
     def copy_csv_to_s3(self, csv_file: str) -> str:
-        s3_file = f"{get_cxg_bucket_path(self.deployment_directories[0])}-genesets.csv"
+        cxg_object_name = self.explorer_s3_uri.split("/")[3].split(".")[
```
maniarathi, line 168: Seems like the `get_cxg_bucket_path` function below is a bit more robust than the above written parsing of the string. Why not use it?

backend/corpora/common/entities/dataset.py
```diff
@@ -169,12 +151,10 @@ def asset_deletion(self):
             asset.delete_from_s3()
             asset.delete()
 
-    def deployment_directories_deletion(self):
-        for deployment_directory in self.deployment_directories:
-            object_names = get_cxg_bucket_path(deployment_directory)
-            logger.info(f"Deleting all files in bucket {cxg_bucket.name} under {object_names}.")
-            cxg_bucket.objects.filter(Prefix=object_names).delete()
-            self.session.delete(deployment_directory)
+    def delete_explorer_cxg_object_from_s3(self):
```
maniarathi, line 154: Wouldn't you also want to clear the two fields `explorer_s3_uri` and `explorer_url` here as well if you were to delete the artifact which can occur beyond deleting a dataset (say the conversion of CXG failed).

backend/corpora/common/entities/dataset.py
```diff
@@ -151,13 +135,11 @@ def create_revision(self) -> "Dataset":
...
         if self.processing_status:
             self.session.add(clone(self.processing_status, dataset_id=revision_dataset.id))
-        for deployment in self.deployment_directories:
-            self.session.add(clone(deployment, dataset_id=revision_dataset.id))
         self.session.commit()
         return Dataset(revision_dataset)
 
     def tombstone_dataset_and_delete_child_objects(self):
-        self.update(tombstone=True, deployment_directories=[], artifacts=[], processing_status={})
+        self.update(tombstone=True, artifacts=[], processing_status={})
```
maniarathi, line 142: Same comment as below.. would you need to add the fields `explorer_s3_uri` and `explorer_url` here to update them to be null or the link to indicate that it no longer exists?

maniarathi (changes requested):

backend/database/versions/b65dad1f2a7e_add_explorer_field_to_dataset.py
MDunitz: Adding an index on the explorer_url so we can search by it in a performant way when looking up datasets based on the explorer url

backend/corpora/common/corpora_orm.py
Bento007: We do have some CXG artifacts in the database and in the data-portal s3. We use it early on but stopped. I don't see why we couldn't do that. It would require modifying the database to include a URL field. It would also require us to modify the backend or frontend code to know where to find the CXG URL.

Bento007: I think the natural step was to go from a list of `DbDeploymentDirectory` which generally only ever contained one entry and had no fields in common with `DbDatasetArtifact`, back to the DATASET table. Now that we added `explorer_s3_uri` we now have one field that overlaps with what in the `DbDatasetArtifact`. I'm on the fence about making it a `DbDatasetArtifact`, but it wouldn't be too much work to make it happen.

backend/corpora/common/entities/dataset.py
Bento007: This function is intended to only delete from s3. This is because of the requirement to not delete resources from S3 until after the changes are published.

backend/corpora/common/entities/dataset.py
```diff
@@ -227,15 +210,16 @@ def reprocess(self):
...
+            explorer_s3_uri=None,
+            explorer_url=None,
             artifacts=[],
         )
 
 
-def get_cxg_bucket_path(deployment_directory: DbDeploymentDirectory) -> str:
+def get_cxg_bucket_path(explorer_url: str) -> str:
     """Parses the S3 cellxgene bucket object prefix for all resources related to this dataset from the
     deployment directory URL"""
```
Bento007, line 221: This description should be updated

backend/corpora/common/entities/dataset.py
Bento007: I recommend moving this into its own function or using `get_cxg_bucket_path`

backend/database/versions/b65dad1f2a7e_add_explorer_field_to_dataset.py
MDunitz: to recreate the deployment directory (when rollbacking the migration) we need to give each row a uuid. This gives us access to a function that creates uuids

backend/corpora/common/corpora_orm.py
MDunitz: I think it makes sense to make it a DbDatasetArtifact instead of adding the explorer_s3_uri. I can update this pr to remove the field and create a ticket to migrate the data/create an artifact to represent the s3_uri

MDunitz: Discussed above https://github.com/chanzuckerberg/corpora-data-portal/pull/1216#discussion_r645315885

backend/corpora/common/entities/dataset.py
Bento007: Some combination of these fields needs to remain. When a user visits the tombstones dataset in explorer, we will need to find it in the database and display the right information about being deleted.

maniarathi: Got it, thanks!

backend/corpora/common/entities/dataset.py
maniarathi: Got it, thanks!

backend/database/versions/b65dad1f2a7e_add_explorer_field_to_dataset.py
maniarathi: Ah got it.Thanks!

backend/corpora/common/entities/dataset.py
MDunitz: updated

maniarathi (commented):
The backend looks good, but I wonder whether you're going to break some frontend code. For example, `frontend/src/common/entities.js` line 121 references the now deleted fields in DatasetDeployment. And that bit of code is also used in the frontend testing directory as well.

maniarathi (approved):

PR #1217, opened 2021-06-04
chore: Delete old defunct scripts

maniarathi:
### Reviewers

**Functional:** @MDunitz 

**Readability:** N/A

---

#### Deletions

- Removed old curation utilities which are now in the `single-cell-curation` repo.
- Removed old curation scripts that were used to generate the first remixed datasets.
- Removed `enloom.py` -- we are moving towards not supporting Loom files at all and this is very outdated.
- Removed `load_h5ad.py` -- this was an old script to manually input datasets into our database. We now support that in the Portal directly and the script is very outdated.


MDunitz (approved):
lgtm

PR #1222, opened 2021-06-05
fix: Remove max 122 dataset download metrics and email detailed metrics on top of Slack messages.

maniarathi:
### Reviewers
**Functional:** @MDunitz 

**Readability:** N/A

---

#### Additions
- Parametrize the query such that a user can input the starting date to override the default which is `today`.

#### Deletions
- Delete the arbitrary 123 maximum results from the query. This was truncating the actual number of dataset downloads. Once we surpass 1000 downloads per week, we will have to implement pagination but removing the max works for now.

#### Modifications
- Instead of regurgitating the entirety of the dataset download metrics on Slack which is not easy to read, package the fine-grained metrics as a CSV and email it to cellxgene@chanzuckerberg.com and only send the top level metrics on Slack.


MDunitz (approved):
lgtm

PR #1240, opened 2021-06-21
feat: Add code to directly support conversion from H5AD to CXG

maniarathi:
Add direct support to convert H5AD files to CXG to the dataset processing container instead of calling the cellxgene convert function. The convert function will be deleted in the next release of cellxgene.

~IMPORTANT NOTE: this PR cannot be deployed until a cellxgene CZI-hosted release has been deployed to all environments to upgrade tiledb to 0.9 or greater. This PR converts H5AD files to CXG using tiledb 0.9 which makes the generated files incompatible with a tiledb version that is less than 0.9. However, there is no problem reading CXGs that were generated with an older version of tiledb in tiledb 0.9.~

~In a nutshell:~
~- If CXG is generated using tiledb < 0.9 (e.g. 0.8), then it can be read (rendered in Explorer) using all versions of tiledb. However, slashes and special characters are not allowed in the H5AD and must be converted to underscores.~
~- If CXG is generated using tiledb >= 0.9, then it can be read (rendered in Explorer) ONLY in tiledb>=0.9. Slashes and special characters are handled.~

Based on conversations in Slack, I have pinned the version of tiledb to be less than 0.9. The upgrade to 0.9 will be handled by a separate PR and tracked separately.

### Testing

In addition to the unit tests, I deployed this to rdev where I was able to successfully upload an H5AD containing slashes (and one without slashes), then downloaded the CXG locally and was able to launch it in cellxgene using the desktop CLI. You can test it out here as well to verify: https://maniarathi-cxg-frontend.rdev.single-cell.czi.technology/

### Reviewers

**Functional:** @MDunitz 

**Readability:** @Bento007 

---

### Additions
- Adds code to convert H5AD files to CXG.

### Modifications
- Upgrades tiledb version to be greater than 0.9.


backend/corpora/dataset_processing/process.py
```diff
@@ -328,19 +329,20 @@ def make_seurat(local_filename):
...
-        subprocess.run(
-            ["cellxgene", "convert", "-o", cxg_dir, "-s", "10.0", local_filename],
-            capture_output=True,
-            check=True,
-        )
-    except subprocess.CalledProcessError as ex:
+        h5ad_data_file = H5ADDataFile(local_filename)
+        h5ad_data_file.to_cxg(cxg_output_container, 10.0)
+    except Exception as ex:
         msg = f"CXG conversion failed: {ex.output} {ex.stderr}"
```
Bento007, line 341:
```suggestion
        msg = f"CXG conversion failed."
```

Bento007: building the string will fail otherwise.

requirements.txt
```diff
@@ -18,6 +19,6 @@ pytest
 requests>=2.22.0
 s3fs==0.4.2
 scanpy
-click==7.1.2
+tiledb>=0.9
```
Bento007, line 22:
```suggestion
tiledb>=0.8
```

Bento007 (approved):

MDunitz (approved):
Lgtm

PR #1255, opened 2021-06-30
fix: the explorer url is not replaced when publishing a dataset revision

Bento007:
### Reviewers
**Functional:** @MDunitz 

## Changes
- not replacing the explorer_url when publishing a dataset revision. This will preserve URL reference to the explorer once the updated dataset is published.


MDunitz (approved):
lgtm

PR #1260, opened 2021-06-30
fix: hardcode clever canary in allowed_origins

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/chalice/api_server/app.py
+++ b/backend/chalice/api_server/app.py
@@ -42,7 +42,7 @@ def get_chalice_app(flask_app):
     if deployment not in ["prod"]:
         allowed_origins.extend(
             [
-                "https://*.dev.single-cell.czi.technology",  # remove once contract work is completed.
+                r"https://.*\.dev\.single-cell\.czi\.technology",  # remove once contract work is completed.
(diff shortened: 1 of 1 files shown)


backend/chalice/api_server/app.py
```diff
@@ -42,7 +42,7 @@ def get_chalice_app(flask_app):
     if deployment not in ["prod"]:
         allowed_origins.extend(
             [
-                r"https://.*\.dev\.single-cell\.czi\.technology",  # remove once contract work is completed.
+                "https://canary-cellxgene.dev.single-cell.czi.technology",  # remove once contract work is completed.
```
MDunitz, line 45: lgtm - do we need to include cellxgene dev as well?

MDunitz (approved):

backend/chalice/api_server/app.py
seve, line 45: Can you make an issue so we remember, thank you!

Bento007: not based on the problem canary is having at the moment

PR #1272, opened 2021-07-16
feat(revisions): detect difference between revision and published collection

seve:
### Reviewers
**Functional:** @tihuan (post-merge), @codemonkey800, @dcwither 

---

## Changes
- add
   - Frontend only `is_revision` + `revision_diff` Collection attributes
   - `frontend/src/common/utils/checkForRevisionChange.ts` utilities for detecting revision differences
- modify
   - Collections are now stored in the frontend as `Map<Collection["id"], Map<VISIBILITY_TYPE, Collection>>` to allow for easier retrieval by ID
   - Datasets are now stored in a Collection as `Map<Dataset["id"], Dataset>` where ID is `original_id` if present or `id`

## QA steps (optional)
 - Should see a print in console if a difference is detected between a revision and its published counterpart

## Known Issues
- Refetching on mutation has been disabled or modified for collection details and dataset upload/reupload as it was causing hard to diagnose runtime errors.  Will rework in seperate issue.


frontend/src/common/queries/collections.ts
```diff
@@ -37,12 +38,26 @@ export interface RevisionResponse extends CollectionResponse {
...
-  return json.collections;
+  const collectionsMap: CollectionResponsesMap = new Map();
+
+  for (const collection of json.collections as CollectionResponse[]) {
+    const collectionsWithId = collectionsMap.get(collection.id) || new Map();
+    collectionsWithId.set(collection.visibility, collection);
+    collectionsMap.set(collection.id, collectionsWithId);
+  }
+
+  return collectionsMap;
```
seve, line 60: Storing as `Map<CollectionResponse["id"], Map<VISIBILITY_TYPE, CollectionResponse>>` now for ease of access

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
+    throw json;
   }
 
-  return result;
+  let datasetMap = new Map() as Collection["datasets"];
+  for (const dataset of json.datasets) {
+    datasetMap.set(dataset.original_id || dataset.id, dataset);
+  }
+
+  collections.set(visibility, { ...json, datasets: datasetMap });
```
seve, line 109: Store datasets as`Map<Dataset["id"], Dataset>` instead of `Array` for easy comparison between published collection and private revision

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
+  if (visibility === VISIBILITY_TYPE.PRIVATE) {
```
seve, line 110: If private collection (potentially a revision), fetch the possible published counterpart and add to response

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
+  }
+
+  return collections;
+}
+
+function useCollectionFetch({ id = "", visibility = VISIBILITY_TYPE.PUBLIC }) {
+  const { data: collections } = useCollections();
+  const queryCache = useQueryCache();
+  return useQuery<Map<VISIBILITY_TYPE, Collection> | null>(
+    [USE_COLLECTION, id, visibility, "FETCH"],
```
seve, line 135: This is an internal query to fetch API data only, `onSuccess` we generate a full Collection object (with `is_revision` and `revision_diff` booleans) which is stored in the consumed query

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
+    fetchCollection,
+    {
+      onSuccess: (data) => {
+        if (collections && data) {
+          queryCache.setQueryData([USE_COLLECTION, id, visibility], () => {
```
seve, line 140: said exposed query

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
+            const newData = data.get(visibility) || ({} as Collection);
+            const collectionsWithID = collections.get(newData.id);
+            if (!collectionsWithID) return newData;
+            newData.is_revision = collectionsWithID.size > 1;
```
seve, line 144: If there are two collections with the same ID, the collections have a revision in progress

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
+
+            // check for diffs between revision and published collection
+            if (newData.is_revision && visibility === VISIBILITY_TYPE.PRIVATE) {
```
seve, line 147: Only for the actual private revision do we compute `revision_diff`

frontend/src/common/queries/collections.ts
```diff
@@ -173,8 +245,8 @@ export function useCollectionUploadLinks(
   const queryCache = useQueryCache();
 
   return useMutation(collectionUploadLinks, {
-    onSuccess: () => {
-      queryCache.invalidateQueries([USE_COLLECTION, id, visibility]);
+    onMutate: () => {
+      // queryCache.invalidateQueries([USE_COLLECTION, id, visibility]);
```
seve, line 249: these mutations were breaking the collection view, need to take a look at them in a separate issue.  Disabled for now

frontend/src/common/queries/collections.ts
```diff
@@ -65,36 +80,93 @@ async function fetchCollection(
...
 export function useCollection({
   id = "",
   visibility = VISIBILITY_TYPE.PUBLIC,
 }) {
-  return useQuery<Collection | null>(
-    [USE_COLLECTION, id, visibility],
-    fetchCollection
-  );
+  useCollectionFetch({ id, visibility });
+  return useQuery<Collection>([USE_COLLECTION, id, visibility]);
```
seve, line 169: Should calls of this query invalidate the internal query? `useCollectionFetch()` is only called when the `useCollection` hook is generated

frontend/src/common/queries/collections.ts
```diff
@@ -258,15 +330,16 @@ const editCollection = async function ({
...
-      return queryCache.setQueryData(
-        [USE_COLLECTION, collection.id, collection.visibility],
-        collection
-      );
+    onSuccess: () => {
+      queryCache.invalidateQueries([
+        USE_COLLECTION,
+        collectionID,
+        VISIBILITY_TYPE.PRIVATE,
+      ]);
```
seve, line 342: instead of optimistically updating the collection, just invalidate the query to recalculate the revision attributes

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -0,0 +1,104 @@
...
+] as Array<keyof Collection>;
+const IGNORED_DATASET_FIELDS = [
+  "created_at",
+  "updated_at",
+  "collection_visibility",
+  "original_uuid",
+  "id",
+  "processing_status",
+  "dataset_assets",
+] as Array<keyof Dataset>;
```
seve, line 24: These fields will be ignored when comparing collections/datasets

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -80,10 +78,20 @@ function renderCollections(
...
+  collectionsMap: CollectionResponsesMap
+): CollectionResponse[] {
+  return (
+    Array.from(collectionsMap.values())
+      .map((collectionsWithIDMap) => {
+        return (
+          collectionsWithIDMap.get(VISIBILITY_TYPE.PUBLIC) ||
+          collectionsWithIDMap.get(VISIBILITY_TYPE.PRIVATE) ||
+          ({} as CollectionResponse)
+        );
```
seve, line 91: Prioritize returning the published collection over private

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -24,7 +23,7 @@ interface Props {
   id: string;
   accessType?: ACCESS_TYPE;
   visibility: VISIBILITY_TYPE;
-  revisionStatus?: REVISION_STATUS;
+  revisionsEnabled?: boolean;
```
seve, line 26: The changes on this file are just swapping out the old `REVISION_STATUS` for the new collection attribute

frontend/src/common/queries/collections.ts
codemonkey800: What use cases do you have in mind that require query invalidation? Maybe you could export a function that invalidates the query so that you invalidate it for a specific use case.

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
```diff
@@ -80,10 +78,20 @@ function renderCollections(
...
+function sortByCreatedAtDescendingAndFormat(
+  collectionsMap: CollectionResponsesMap
+): CollectionResponse[] {
+  return (
+    Array.from(collectionsMap.values())
+      .map((collectionsWithIDMap) => {
+        return (
+          collectionsWithIDMap.get(VISIBILITY_TYPE.PUBLIC) ||
+          collectionsWithIDMap.get(VISIBILITY_TYPE.PRIVATE) ||
+          ({} as CollectionResponse)
```
codemonkey800, line 90: nit: this doesn't seem like a good idea since the properties for this object are all required. it would probably be better to change the return type to something like `Array<CollectionResponse | null>` or maybe make the return type partial `Partial<CollectionResponse>[]` for null safety.

maybe you could also filter out items that aren't in `collectionsWithIDMap`.

codemonkey800 (commented):
left some comments 🍜

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
seve: The map is guaranteed to have at least one of the two `VISIBILITY_TYPE`'s, the empty object is just for type safety since `.get` can return null.  Is there a better way to do this?

frontend/src/views/Collection/index.tsx
```diff
@@ -105,13 +87,13 @@ const Collection: FC = () => {
               "Your file is being uploaded which will continue in the background, even if you close this window.",
           });
 
-          queryCache.invalidateQueries(USE_COLLECTION);
+          // queryCache.invalidateQueries(USE_COLLECTION);
```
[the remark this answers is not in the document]
seve: Maybe I should replace all these with TODO's

frontend/src/components/Collections/components/Grid/components/CollectionsGrid/index.tsx
codemonkey800: > The map is guaranteed to have at least one of the two VISIBILITY_TYPE's

Hmm, what you have is fine then since I didn't know this. Although for readability, I would add a comment or change the code to:

```ts
return (
  collectionsWithIDMap.get(VISIBILITY_TYPE.PUBLIC) ||
  collectionsWithIDMap.get(VISIBILITY_TYPE.PRIVATE)
) as CollectionResponse
```

seve: ah that's a great idea.  I'll do that.  Thanks!

frontend/src/common/queries/collections.ts
```diff
@@ -258,14 +315,21 @@ const editCollection = async function ({
...
   return useMutation(editCollection, {
-    onSuccess: (collection: Collection) => {
-      return queryCache.setQueryData(
-        [USE_COLLECTION, collection.id, collection.visibility],
-        collection
+    onSuccess: (newCollection) => {
+      queryCache.setQueryData(
+        [USE_COLLECTION, collectionID, VISIBILITY_TYPE.PRIVATE],
+        () => {
+          return { ...collection, ...newCollection };
```
tihuan, line 329: ah can you remind me why we need to spread `...collection, ...newCollection` here instead of just using `newCollection`? Thank you!

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -0,0 +1,121 @@
...
+  });
+}
+
+export default function checkForRevisionChange(
+  revision: Collection,
+  publishedCollection: Collection
+): boolean {
+  // Check collection fields for differences
+  let collectionKey = "" as keyof Collection;
+  for (collectionKey in publishedCollection) {
```
tihuan, line 99: `for (let collectionKey in publishedCollection) { ... }` will do? Since we're not using `collectionKey` outside of the for loop?

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -0,0 +1,121 @@
...
+import xorWith from "lodash/xorWith";
+import { Collection, Dataset } from "../entities";
+
+const IGNORED_COLLECTION_FIELDS = [
+  "visibility",
+  "created_at",
+  "updated_at",
+  "is_revision",
+  "revision_diff",
+  "datasets",
```
tihuan, line 12: probably need more context here, but I was imagining we'd need to detect change when someone updates a dataset with a new dropbox link or something?

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -0,0 +1,121 @@
...
+export default function checkForRevisionChange(
+  revision: Collection,
+  publishedCollection: Collection
+): boolean {
+  // Check collection fields for differences
+  let collectionKey = "" as keyof Collection;
+  for (collectionKey in publishedCollection) {
+    if (
+      !IGNORED_COLLECTION_FIELDS.includes(collectionKey) &&
+      publishedCollection[collectionKey] !== revision[collectionKey]
```
tihuan, line 102: should we use lodash.isEqual here in case we're comparing objects, arrays, etc.?

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -0,0 +1,121 @@
...
+function checkDatasetsForChanges(
+  revisedDatasets: Map<Dataset["id"], Dataset>,
+  publishedDatasets: Map<Dataset["id"], Dataset>
+): boolean {
+  if (publishedDatasets.size !== revisedDatasets.size) return true;
+  // Check dataset fields for differences
+  return Array.from(publishedDatasets.values()).some((publishedDataset) => {
+    const revisedDataset = revisedDatasets.get(publishedDataset.id) as Dataset;
+    let datasetKey = "" as keyof Dataset;
+    for (datasetKey in publishedDataset) {
```
tihuan, line 75: `for (let datasetKey in publishedDataset) { ... }`

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (IGNORED_DATASET_FIELDS.includes(datasetKey)) {
+        continue; // ignore these fields (I hate using continue, but sonarjs considers this best)
```
tihuan, line 77: for my own education, what's the alternative for `continue` here? I only know to use `continue` to skip an iteration in a loop lol

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
```diff
@@ -247,7 +248,7 @@ const Content: FC<Props> = (props) => {
 
     setIsLoading(true);
 
-    await mutateEditCollection({
+    mutateEditCollection({
```
tihuan, line 251: ah I see `await` is removed here, which means we will run the code below before the promise resolves?

```ts
setIsLoading(false);
onClose();
```

tihuan (commented):
Overall looks super good! Just some minor feedback 🥇 Thanks so much for the massive work here, @seve 🎉 !

frontend/src/common/queries/collections.ts
seve: `collection` holds some locally calculated attributes that `newCollection` won't hold

frontend/src/common/utils/checkForRevisionChange.ts
seve: typescript was getting mad about indexing by `collectionKey` without explicitly stating its type

frontend/src/common/utils/checkForRevisionChange.ts
seve: We explicitly check datasets and other arrays/objects

frontend/src/common/utils/checkForRevisionChange.ts
seve: https://github.com/chanzuckerberg/corpora-data-portal/pull/1272#discussion_r675236199

frontend/src/common/utils/checkForRevisionChange.ts
seve: Wrapping it all in the inverse of the `continue` check

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
seve: reinstated the await

frontend/src/common/utils/checkForRevisionChange.ts
tihuan: ahh got it! Yeah early return/continue/skip is the [guard clause pattern](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) that prevents excessive nesting of code!

frontend/src/common/utils/checkForRevisionChange.ts
tihuan: geez, this turns out to be quite a discussion in TS community lol

https://github.com/microsoft/TypeScript/issues/32321

Lots of people think `Object.keys(publishedCollection` should just return type `keyof Dataset`, but TS says it can't be, because any superset of an interface qualifies as the interface, so the keys could be more than just `keyof Dataset`

Details here: https://stackoverflow.com/questions/55012174/why-doesnt-object-keys-return-a-keyof-type-in-typescript

I think your solution works for us!

tihuan (approved):
LGTM!! Thanks so much for the quick iterations, @seve ! And @codemonkey800 @dcwither for the reviews too 🙏 🙏 🙏

PR #1288, opened 2021-07-23
chore: Update standup rotation to remove Trent and add Timmy.

maniarathi:


tihuan (approved):
Yayy I'm officially back!! 🤣

PR #1293, opened 2021-07-26
fix: update processing status instead of upload status on failure

MDunitz:
remove message telling users to try again

### Reviewers
**Functional:** 
- @seve 
**Readability:** 

---

## Changes
- remove message telling users to try again on upload failure
- modify upload failure lambda to update the processing status instead of the upload status

### DIFF
--- a/frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/components/Tooltip/index.tsx
+++ b/frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/components/Tooltip/index.tsx
@@ -53,9 +53,7 @@ const ERROR_TO_CONTENT: { [key: string]: Content } = {
   },
   [FAILED_RETURN_TYPE.UPLOAD + UPLOAD_STATUS.FAILED]: {
     color: RED.C,
-    content: (
-      <span>There was a problem uploading your file.</span>
-    ),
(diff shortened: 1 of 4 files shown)


frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/components/Tooltip/index.tsx
```diff
@@ -53,9 +53,7 @@ const ERROR_TO_CONTENT: { [key: string]: Content } = {
   },
   [FAILED_RETURN_TYPE.UPLOAD + UPLOAD_STATUS.FAILED]: {
     color: RED.C,
-    content: (
-      <span>There was a problem uploading your file. Please try again.</span>
-    ),
+    content: <span>There was a problem uploading your file.</span>,
```
seve, line 56: Do we want to say something to communicate:
> CZI will take a look at it and reach out with any updates, here is our SLA

cc: @maniarathi

maniarathi: Let's leave it as is for now as we test out our SLA for a few on-call cycles before committing publicly.

seve (approved):

PR #1302, opened 2021-07-27
fix: only show revision state on the revision, rename confusing field

seve:
### Reviewers
**Functional:**  @tihuan 

---

## Changes
- modify
  -  actually check if a collection **is** a revision when showing revision state
  - rename `is_revision1 to `has_revision` to better communicate boolean state


tihuan (approved):
LGTM! `has_revision` looks better indeed 👍 Thanks for the update!

Issue #1296, opened 2021-07-26
add new fields to dataset entity


seve:
New or updated fields as a result of schema 2.0 should be stored in the DB and served to the collection endpoint `/dp/v1/collections/{id}`.

Default Values for Migration:
* `sex`, `cell_type`, `X_normalization`, `X_approximate_distribution`: null
* `is_primary_data`: False
* schema_version: Script should generate values for existing datasets

| Old UML | New UML | 
| ---- | ---- |
| ![image](https://user-images.githubusercontent.com/8716829/127538627-cde2410f-5837-4e05-9bde-f8fd8fac0396.png) | ![image](https://user-images.githubusercontent.com/8716829/128054433-1bb45a70-8a94-4800-8d7c-e77402387120.png) |

##### changes marked in blue

Note: Sex is marked as a change, but no change to the DB will be needed. The API will switch from https://github.com/chanzuckerberg/corpora-data-portal/blob/135214e9f423d815645fb3e953dc3a222b7fd252/backend/config/corpora-api.yml#L821-L824
to
```
sex:
  $ref: "#/components/schemas/ontology_element_array"
```

----

A few dataset attributes have been either added, changed, or removed, update our api yml to reflect those changes:
https://github.com/chanzuckerberg/corpora-data-portal/blob/44b98b49bd3a38bf19d9f73b9b75e14d2b4e0586/backend/config/corpora-api.yml#L688-L743

The frontend should also add the fields to the collection entity defined in `src/common/queries/entities.ts`

### Definition of Done
* DB Migration with new and modified DB columns
* A script to generate `schema_version` for existing datasets
* Updated UML published and chanzuckerberg/single-cell references updated
* Updated api yaml
* Update entities.ts (in the frontend)
* Update all the unit tests so the mocked data is on par with the new schema definition

seve:
Removing frontend aspect of this ticket and creating new ticket since a bit more work will be needed

seve:
All new fields can be null while we migrate 1.x.x datasets to 2.0.0 since there will be no immediate UX impact

maniarathi:
I wonder if it might be useful to add the `schema_version` in the database as well to log what schema version the dataset was validated against. It may not be useful in the API now, but could be useful data to have at some point (I'm thinking even as a gut check to verify all the datasets that are public facing are migrated to version 2.0 before we launch).

seve:
Store schema in db and write script to generate the version # for older datasets

PR #1342, opened 2021-08-12
feat: Schema 2.0.0 dataset migration (for Postgres)

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 
@seve 

**Readability:** 
@blrnw3 
---
Zenhub Ticket: [#1296](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/1296)

## Changes
- Adds schema 2.0.0 changes to the Postgres Dataset table
- Adds an alembic migration script
- Adds a script that updates `schema_version` for the existing datasets

DB Migration with new and modified DB columns
A script to generate schema_version for existing datasets
Updated UML published and chanzuckerberg/single-cell references updated
Updated api yaml
Update all the unit tests so the mocked data is on par with the new schema definition

## Known Issues
- Unit tests to be added/modified in a separate PR
- Database migration still to be tested


scripts/cxg_admin.py
```diff
@@ -185,6 +185,31 @@ def create_cxg_artifacts(ctx):
...
+    import requests as rq
+    with db_session_manager() as session:
+        click.confirm(
+            f"Are you sure you want to run this script? It will assign schema_version to all the "
+            f"datasets",
+            abort=True,
+        )
+        for record in session.query(DbDataset):
+            dataset_id = record.id
+            url = f"https://api.cellxgene.cziscience.com/cellxgene/e/{dataset_id}.cxg/api/v0.2/config"
```
MDunitz, line 207: annoyingly this will miss some datasets that were named early on with a different naming scheme (cellxgene.cziscience.com/cellxgene/e/{datasest_name}.cxg) but you should be able to do something like f"api.{record.explorer_url}/api/v0.2/config" getting the explorer url off the dataset

scripts/cxg_admin.py
```diff
@@ -185,6 +185,31 @@ def create_cxg_artifacts(ctx):
...
+@click.pass_context
+def migrate_schema_version(ctx):
+    """
+    Populates `schema_version` for each existing dataset. Since the schema version only exists
+    in the cxg file and we don't want to open them, we will call the cellxgene explorer endpoint
+    which contains the version. This is a one-off procedure since new datasets will have
+    the version already set.
+    """
+
+    import requests as rq
```
blrnw3, line 198: nit: move to global import

scripts/cxg_admin.py
```diff
@@ -185,6 +185,31 @@ def create_cxg_artifacts(ctx):
                     s3_uri=s3_uri,
                 )
 
+@cli.command()
+@click.pass_context
+def migrate_schema_version(ctx):
+    """
+    Populates `schema_version` for each existing dataset. Since the schema version only exists
+    in the cxg file and we don't want to open them, we will call the cellxgene explorer endpoint
```
blrnw3, line 193: where does Explorer get the version from? Does it not have to open the files?

scripts/cxg_admin.py
```diff
@@ -185,6 +185,31 @@ def create_cxg_artifacts(ctx):
...
+    which contains the version. This is a one-off procedure since new datasets will have
+    the version already set.
+    """
+
+    import requests as rq
+    with db_session_manager() as session:
+        click.confirm(
```
blrnw3, line 200: minor but should this be outside of the context manager?

scripts/cxg_admin.py
MDunitz: Yea, we are just hitting prod and making it do the work

scripts/cxg_admin.py
ebezzi: Honestly I just copied this from another method, but if you can't establish a connection to a database, I think the prompt would be redundant.

scripts/cxg_admin.py
ebezzi: Ah, I didn't know the endpoint had more formats. I will implement what you suggested!

backend/config/corpora-api.yml
```diff
@@ -851,8 +849,18 @@ components:
...
+        cell_type:
+          $ref: "#/components/schemas/ontology_element_array"
         cell_count:
           type: integer
+        X_normalization:
+          type: string
+        X_approximate_distribution:
+          $ref: "#/components/schemas/X_approximate_distribution"
+        schema_version:
+          type: string
```
blrnw3, line 863: should this be enum? We only have a few versions and they change rarely, right?

ebezzi: The specification requires this to be a `string`. Moving to an enum would require to change the specs and at this point I don't think it's worth it. That said, @MDunitz could have a stronger opinion on this.

MDunitz: We currently only have a few versions but its not clear to me what the cadence of change will be in the future. I dont think there is much to gain from making it an enum (but I could be wrong)

scripts/cxg_admin.py
MDunitz: Also if possible to allow for testing I would set it up to build the api based on the passed in deployment stage. For staging it should be `https://api.cellxgene.staging.single-cell.czi.technology`

backend/config/corpora-api.yml
blrnw3: If it's informational then a string is fine. If its use goes beyond that, an enum would help avoid certain bugs. Happy to leave it as a string for now if we don't know

scripts/cxg_admin.py
ebezzi: I added the hostname from the explorer_url, that should work.

backend/corpora/common/entities/dataset.py
```diff
@@ -199,6 +185,7 @@ def generate_tidy_csv_for_all_linked_genesets(self, csv_file_path: str) -> str:
     def reprocess(self):
         if not self.published:
             self.asset_deletion()
+        # TODO: any other schema 2.0 changes needed?
```
MDunitz, line 188: remove?

MDunitz (approved):
lgtm

PR #1345, opened 2021-08-13
fix: upload dataset stale download

tihuan:
RDEV: https://thuang-stale-download-frontend.rdev.single-cell.czi.technology/?rc=true&curator=true

### Reviewers
**Functional:** 
@seve 

---

## Changes
- modify
Looks like CXG is now part of `dataset.dataset_assets`, so we no longer need to rely on `dataset.dataset_deployments` to infer `CXG` has been converted. This is the root cause of the off by one issue that causes the bug!

## QA steps (optional)
1. Upload a dataset
2. Wait till the process completes
3. Click download icon to see if all three formats are available (should be)
4. Upload another dataset
5. Wait till the process completes
6. Click download icon to see if all three formats are available <--- this is where the bug was. But now all 3 formats should  be available!

## Known Issues
N/A


seve (approved):

tihuan:
Thanks for the quick review, Seve! Merging 🚀 !!

PR #1358, opened 2021-08-18
feat: Save mean_genes_per_cell to the database

ebezzi:
### Reviewers

**Functional:** 
@MDunitz 
@seve 
@brianraymor 

**Readability:**
@tihuan 
@leslieypark 

---

## Changes
- Saves `mean_genes_per_cell` to the database (as a Float/double precision field) after computing it (previously it was computed and discarded)
- Adds an additional filter so only variables that are marked with `feature_biotype == ' gene'` are computed in the count.

N/A

## QA steps (optional)
I would recommend to compute the count independently on a notebook and see if the computation (e.g. in staging or even local) matches.

## Known Issues
I think we should add a more specific unit test for this feature. Current unit tests do not test the computation logic at all.


brianraymor:
The presence of `feature_biotype` is dependent on the schema 2.0.0 update.

ebezzi:
@brianraymor yes, this is going to be merged to a feature branch that will only work with 2.0.0 compliant datasets.

pablo-gar:
Just want to double check, are you ensuring that a raw layer is being used to create this calculation? And that non-RNA data may give spurious results as there may not a be a raw layer and it may be hard to identify when a gene has no signal?

ebezzi:
@pablo-gar - the logic for extracting the raw layer has been changed as per [the ticket description](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/1251) and the new code logic lives [here](https://github.com/chanzuckerberg/corpora-data-portal/blob/f2f17896e78ba9895661c5083a6026756ac7d013/backend/corpora/dataset_processing/process.py#L258). Of course, a review of that would be greatly appreciated since my familiarity with AnnData is very limited! Also, based on your comment, it looks like you're suggesting that if raw data isn't available, then we shouldn't calculate this field at all? If that's the case, we can easily do that.

brianraymor:
@pablo-gar - see the note below from [mean_genes_per_cell is calculated and then thrown away](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/1251) and let me know if you disagree.

Per a July 29 slack conversation with @ambrosejcarr and @seve - the source of the matrix for the calculation:

1. If AnnData.raw.X is defined, then process is done -- this is the raw data.
2. If AnnData.uns['X_normalization'] is set to “none”, then process is done -- AnnData.X is the raw data.
3. Otherwise, there is no raw data because this is an Accessibility Assay such as ATAC-seq. In this case, default to AnnData.X.

backend/corpora/dataset_processing/process.py
```diff
@@ -224,7 +224,6 @@ def update_db(dataset_id, metadata=None, processing_status=None):
 
         if metadata:
             # TODO: Delete this line once mean_genes_per_cell is in the db
```
leslieypark, line 226: Also delete this comment

MDunitz (approved):
lgtm

pablo-gar:
>  it looks like you're suggesting that if raw data isn't available, then we shouldn't calculate this field at all? If that's the case, we can easily do that.

There could cases where no signal is represented by something other than zero, I'm not sure how frequent this will happen. @MaximilianLombardo has more experience working with non-RNA-seq data that has no raw layer.

Max have you seen this case?

ambrosejcarr:
> it looks like you're suggesting that if raw data isn't available, then we shouldn't calculate this field at all? If that's the case, we can easily do that.

Sorry I was slow on responding to this - what is supposed to happen is:

> Otherwise, there is no raw data because this is an Accessibility Assay such as ATAC-seq. In this case, **default to AnnData.X.**

brianraymor:
As documented [above](https://github.com/chanzuckerberg/single-cell-data-portal/pull/1358#issuecomment-901456543), but there seems to be some continuing confusion or ambiguity here that I do not understand.

MaximilianLombardo:
> There could cases where no signal is represented by something other than zero, I'm not sure how frequent this will happen. @MaximilianLombardo has more experience working with non-RNA-seq data that has no raw layer.

@pablo-gar yes this is the case for methylation data where interpretation is inverted (i.e. 0 methylation implies more transcription/activity of a gene and high positive values of methylation imply no gene activity). This case will break things for calculating where genes are found in a standardized manner.

On a side note, I'm thinking that the epigenetic (methylation and open chromatin) assays (we get inferred gene activity values rather than actually measured values) in general don't fit nicely into the suggested model so @pablo-gar 's suggestion for not calculating this metric if there is no raw layer checks out for me.

ambrosejcarr:
@pablo-gar @MaximilianLombardo and I have determined that there are issues with the calculation of n_genes_detected in methylC-seq datasets. **No changes are requested at present**. We have [filed an issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-curation/127) to address the problems at a later date. 

Briefly, in the single methylation dataset we have, the values are inverted with respect to all other data in our corpus (0 is high, 1 is low). This is not made clear to users; they need to discover it by exploring the dataset or reading the paper. Until we clarify how these datasets should be interpreted, leaving the n_genes_detected calculation as is will produce a very high value and we think that will help prompt users to think twice until we have a more robust fix in place.

PR #1362, opened 2021-08-19
feat: 1298 use feature_name column as var_index

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 
@seve 

**Readability:** 
@leslieypark 
---

## Changes
Specify `feature_name` as the `var` index column when creating the H5AD object file.

- H5ADDataFile is called with var_index arg when converting H5AD to CXG
- Test that var_index is correctly set on resulting H5ADDataFile obj

## QA steps (optional)
Run this on a real, 2.0.0 compliant H5AD file and inspect the resulting cxg.

## Known Issues
N/A


leslieypark (approved):
Readability 🚀

PR #1364, opened 2021-08-23
chore: pin numpy and numba to consistent versions

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

---

## Changes
`numba` (which is at least used by `scanpy`) requires `numpy` to be of version `<= 1.20`. On my dev environment 1.21 gets pulled and this causes the app to malfunction. This pins it to a safe version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ furl
 h5py<3.0.0
 jsonschema
 moto==1.3.14
-numpy
+numpy==1.20.3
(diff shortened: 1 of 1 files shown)


leslieypark (approved):
🚀

ebezzi:
I changed this so that we use newer versions. `numba` was causing issues and pinning this version fixes the problem.

PR #1365, opened 2021-08-25
feat: API fixes for schema 2.0.0

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@leslieypark 

---

Ticket [#1296 ](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1296)

## Changes
Final changes for the API to be compatible with both version.
- Modify the API so that `organism` is now an array of objects (using the ontology schema)
- Modify the code so that `organism` gets converted to an array if it was an object (i.e. if the dataset was 1.1.0)
- Modify the FE so that `organism` is now reading from an array and not an object
- Modify the API so that `sex` is now an array of objects (using the ontology schema). Note: no FE changes are needed since this field is not used
- Modify the code so that `sex` gets converted to an array of objects if it was a simple array (i.e. if the dataset was 1.1.0)


MDunitz (approved):
lgtm

PR #1367, opened 2021-08-26
feat: 1368 authorization

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

---

Issue [1368](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1368)

## Changes
Implements token based authorization. Adds a `requires_auth` and `requires_scope` decorators that can be used to restrict APIs. 
Notes: 
1. The current authentication flow isn't ideal. The proper Oauth2 flow requires the token not to be stored in the cookie, but rather in the FE local storage/memory and sent via the `Authorization` header to each request. The `id_token` is also probably useless for this application and should be replaced with a call to `userinfo`
2. `connexion` (which is currently use to manage secured access to the APIs via the swagger yaml file) does not support scopes unless you use the `Oauth2` flow (currently we're using a custom flow). Therefore, we need to do this manually using decorators. Fixing 1. would allow to do this properly
3. We need two different audiences: one for the `id_token` (currently in configuration) and another for the `access_token` (to be stored in config.audience2 - need to figure out how to do that).

### DIFF
--- a/backend/corpora/lambdas/api/v1/authentication.py
+++ b/backend/corpora/lambdas/api/v1/authentication.py
@@ -178,6 +178,8 @@ def check_token(token: dict) -> dict:
 
     :param token: a dictionary that contains the token information.
     """
+    if "access_token" not in token:
+        return {}
     try:
         payload = assert_authorized_token(token.get("access_token"))
     except ExpiredSignatureError:
@@ -209,6 +211,18 @@ def apikey_info_func(tokenstr: str, required_scopes: list) -> dict:
     token = decode_token(tokenstr)
     payload = check_token(token)
     return payload
+
+
+def apikey_info_func_lenient(tokenstr: str, required_scopes: list) -> dict:
+    """
+    Lenient version that allows endpoints to work even if authentication fails.
+    Use this for endpoints that also require public access, so if users end up with a bad token,
+    they won't be locked out.
+    """
+    try:
+        return apikey_info_func(tokenstr, required_scopes)
+    except Exception:
+        return {}
 
 
 def userinfo() -> Response:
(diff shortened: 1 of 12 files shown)


tihuan:
Oh yes totally! Thanks so much, Emanuele 🤩🚀

MDunitz (approved):
lgtm

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -47,7 +48,8 @@ def get_oauth_client(config: CorporaAuthConfig) -> FlaskRemoteApp:
         refresh_token_url=config.api_token_url,
         access_token_url=config.api_token_url,
         authorize_url=config.api_authorize_url,
-        client_kwargs={"scope": "openid profile email offline_access"},
+        client_kwargs={"scope": "openid profile email offline_access read:collections write:collections"},
```
MDunitz, line 51: did we want to get rid of read:collections?

oauth/api_scopes.json
```diff
@@ -1,5 +1,11 @@
 [
   {
     "Name": "test_scope"
+  },
+  {
+    "Name": "read:collections"
```
MDunitz, line 6: drop?

oauth/clients-config.json
```diff
@@ -12,7 +12,7 @@
             "http://backend.corporanet.local:5000/dp/v1/oauth2/callback"
         ],
         "AllowedScopes": [
-           "openid", "profile", "email", "offline_access"
+           "openid", "profile", "email", "offline_access", "read:collections", "write:collections"
```
MDunitz, line 15: drop read:collections?

backend/corpora/lambdas/api/v1/authentication.py
ebezzi: I was thinking to leave it for future implementations, but we can always add it back later. Will remove it altogether.

PR #1374, opened 2021-08-30
fix: 1221 also strip link metadata

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

---

## Changes
- Also strip link metadata since these also belong to user input.


MDunitz (approved):
lgtm

Issue #1368, opened 2021-08-26
Implement authorization for curators


ebezzi:
The website should have this behavior with respect to authorization:

1. Everyone who is signed in should be able to create a collection (you still need the `?curator=true` flag to see the `Create collection` button) (currently implemented)
2. A user should be able to view, edit and create revisions only for collections they own (currently implemented)
3. Curators (i.e. users with `write:permissions` scope) should be able to view, edit and create revisions for all collections (TODO)

The basic authorization infra has already been created as part of https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1284

ebezzi:
@brianraymor thanks for the corrections, I edited the description.

maesen:
TEAM COMMENTS:
* Auth0 permissions already set up. `write:permissions` scope will give this curation power.
* Endpoint to get collections should return all collections (public and private) under My Collections for users with this scope. ("Waive" the filter.)
* Scope of this ticket also includes adding test cases. Test user will need this scope.

PR #1378, opened 2021-09-01
feat: 1368 authorization for curators

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@seve 

---

Ticket [#1368 ](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1368)

## Changes
- Changes all the collections endpoints to behave differently depending on the user's permissions

- Curators can view all the private collections, even if they belong to other users
- Curators can delete any private collection
- Curators can update any private collection
- Curators can start revisions on every private collection.

## Known issues
Verify if this doesn't introduce new authentication issues that might impact non logged in users.

### DIFF
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -9,17 +9,19 @@ from ....common.entities import Collection
 from ....common.utils.exceptions import ForbiddenHTTPException, ConflictException
 from ....api_server.db import dbconnect
 
+
 def _is_user_owner_or_allowed(user, owner):
     """
     Check if the user has ownership on a collection, or if it has superuser permissions
     """
     return (user and user == owner) or (has_scope("read:collections"))
 
+
 def _owner_or_allowed(user):
     """
     Returns None if the user is superuser, `user` otherwise. Used for where conditions
     """
-    return (None if has_scope("read:collections") else user)
+    return None if has_scope("read:collections") else user
 
 
 @dbconnect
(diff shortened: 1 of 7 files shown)


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -1,3 +1,4 @@
+from backend.corpora.lambdas.api.v1.authorization import has_scope
```
leslieypark, line 1: Super nit: Can we move this import down, after the 3rd party imports?

tests/unit/backend/corpora/api_server/test_v1_collection.py
```diff
@@ -863,3 +863,54 @@ def test__update_collection_links__OK(self):
...
+        collection = self.generate_collection(
+            self.session, visibility=CollectionVisibility.PRIVATE.name, owner="another_test_user_id"
+        )
+
+        headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
+        test_url = furl(path=f"/dp/v1/collections/{collection.id}", query_params=dict(visibility="PRIVATE"))
+        response = self.app.get(test_url.url, headers=headers)
+
+        # This will pass even for non curators.
+        # Why are users allowed to view private collections that they don't own?
```
leslieypark, line 880: Is this a todo question?

leslieypark (approved):
🚀  Super nitpicks, but otherwise LGTM

tests/unit/backend/corpora/api_server/test_v1_collection.py
ebezzi: Yes, if you look at the code [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/corpora/lambdas/api/v1/collection.py#L39), there is no explicit filter for this endpoint that requires the user to be an owner. Only tombstone datasets and `access_type` are gated. @MDunitz do you know if this is as expected?

MDunitz: At the time of implementation we decided that private datasets shouldnt be displayed on the portal page but they should be available to anyone with the link to handle cases where the owner of the dataset wanted to share it with other collaborators

Issue #1243, opened 2021-06-21
Upgrade tiledb version to be greater than 0.9


maniarathi:
## Need

tiledb 0.9 introduces native handling of special characters that allows us to bypass having to convert all the special characters into underscores. This issue simply upgrades the version to be 0.9 or greater.

## Definition of Done

- [ ] Upgrade `requirements.txt` to say `tiledb>=0.9`.
- [ ] Verify on rdev that the CXG conversion was successful.

PR #1382, opened 2021-09-07
fix: upgraded TileDB to 0.9+ (#1243)

metakuni:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@ebezzi 

---

## Changes
- modify: tiledb version: `<0.9` -> `==0.9.5`

- [x] Upgrade `requirements.txt` to say `tiledb>=0.9`.
- [x] Verify on rdev that the CXG conversion was successful.

## QA steps (optional)
- verified on rdev that CXG conversion of an h5ad file with slashes was successful

## Known Issues
- N/A

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,7 @@ pytest
 requests>=2.22.0
 s3fs==0.4.2
 scanpy
-tiledb<0.9
+tiledb>=0.9
 click==7.1.2
 -r ./backend/corpora/api_server/requirements.txt
 rsa>=4.7 # not directly required, pinned by Snyk to avoid a vulnerability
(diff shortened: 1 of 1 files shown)


maniarathi:
I'm assigning Madison to review instead of me, but just one thing I saw -- it looks like we've pinned a bunch of old versions of packages in Portal... I think it might be prudent to upgrade them so that we don't fall too far behind so you could create an issue for that.

MDunitz:
> I'm assigning Madison to review instead of me, but just one thing I saw -- it looks like we've pinned a bunch of old versions of packages in Portal... I think it might be prudent to upgrade them so that we don't fall too far behind so you could create an issue for that.

There should be an issue to upgrade dependencies in portal (and in the other repos). I created it last week based on the monthly reminder

MDunitz (approved):
lgtm

PR #1400, opened 2021-09-10
feat: 1297 use cellxgene-schema with labels

ebezzi:
### Reviewers
**Functional:** 
@leslieypark @atolopko-czi @pablo-gar 

**Readability:** 
@seve 
---

## Changes
- Uses `cellxgene-schema` with `--add-labels` to generate a new h5ad file with labels
- Uses the aforementioned file in the rest of the pipeline
- Cleans up all the old h5ad files

### DIFF
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -14,14 +14,15 @@ ENV PIP_NO_CACHE_DIR=1
 RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
  && echo '#!/bin/sh\nmkdir -p build/static && touch build/static/.gitignore build/index.html build/csp-hashes.json' > /usr/bin/npm && chmod 755 /usr/bin/npm \
  && cd single-cell-explorer \
- && make pydist-czi-hosted \
+ && make pydist-server \
  && make install-dist \
- && make dev-env-czi-hosted \
+ && make dev-env-server \
  && cd .. \
  && rm -rf single-cell-explorer client build/client /usr/bin/npm
 
 # Install python dependencies
-RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 tiledb==0.7.4 cellxgene-schema awscli
+RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 tiledb==0.7.4 awscli
+RUN pip3 install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple cellxgene-schema==1.0.2
 
 ADD requirements.txt requirements.txt
 ADD backend/corpora/api_server/requirements.txt backend/corpora/api_server/requirements.txt
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -496,23 +496,19 @@ def process(dataset_id, dropbox_url, cellxgene_bucket, artifact_bucket):
         "local.h5ad",
     )
 
-    try:
-        file_with_labels = validate_h5ad_file_and_add_labels(dataset_id, local_filename)
-    finally:
-        clean_up_local_file(local_filename)
+    # No file cleanup needed due to docker run-time environment.
+    # To implement proper cleanup, tests/unit/backend/corpora/dataset_processing/test_process.py
+    # will have to be modified since it relies on a shared local file
 
-    try:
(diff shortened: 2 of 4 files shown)


Dockerfile.processing_image
```diff
@@ -14,14 +14,15 @@ ENV PIP_NO_CACHE_DIR=1
...
  && make install-dist \
- && make dev-env-czi-hosted \
+ && make dev-env-server \
  && cd .. \
  && rm -rf single-cell-explorer client build/client /usr/bin/npm
 
 # Install python dependencies
-RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 tiledb==0.7.4 cellxgene-schema awscli
+RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 tiledb==0.7.4 awscli
+RUN pip3 install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple cellxgene-schema==1.0.2
```
atolopko-czi, line 25: TODO: Update to prod pip & cellxgene-schema version when ready

backend/corpora/dataset_processing/process.py
```diff
@@ -486,17 +496,20 @@ def process(dataset_id, dropbox_url, cellxgene_bucket, artifact_bucket):
...
 
     # create artifacts
-    process_cxg(local_filename, dataset_id, cellxgene_bucket)
-    create_artifacts(local_filename, dataset_id, artifact_bucket)
+    process_cxg(file_with_labels, dataset_id, cellxgene_bucket)
+    create_artifacts(file_with_labels, dataset_id, artifact_bucket)
     update_db(dataset_id, processing_status=dict(processing_status=ProcessingStatus.SUCCESS))
 
+    clean_up_local_file(file_with_labels)
+    clean_up_local_file(local_filename)
```
atolopko-czi, line 511: may do in a `try..finally` block to ensure cleanup

tests/unit/processing_container/test_process.py
```diff
@@ -56,9 +56,10 @@ def test_make_seurat(self):
     def test_make_loom(self):
         make_loom(str(self.real_h5ad_filename))
 
-    # TODO: Remove mocking of validate_h5ad_file after validation process is updated to be 2.0 compliant.
-    @patch("backend.corpora.dataset_processing.process.validate_h5ad_file")
+    # TODO: Remove mocking of validate_h5ad_file after validation process is updated to be 2.0 compliant.`
```
atolopko-czi, line 59: is it time to address this TODO? Seems like we should also verify that labels were added to the newly written file. The original test only appears to be ensuring that `process()` runs w/o raising an error, but we could add some assertions here I think.  Unless this is level of testing is being handled elsewhere?

atolopko-czi (changes requested):
Code changes look good, but we should probably wait to merge until cellxgene-schema 2.0.0 is up on pypi.org.  Also think we should add extend tests for the new functionality, if possibly.

Dockerfile.processing_image
ebezzi: Shall we upload to prod pip today, or do we want to at least run a full E2E testing before we do that? @pablo-gar

tests/unit/processing_container/test_process.py
ebezzi: Additional testing won't be part of this ticket, as per ticket comments (see https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1297#issuecomment-914721648). That said, I agree this should be done on a follow up ticket. It will be tricky since with the current test, we're downloading an existing h5ad file in a staging dataset (therefore, with labels), while we should change it to use a "raw" h5ad (without labels). That said, downloading it from a Dropbox folder instead of using S3 should be easy enough and will also make the test independent of the staging environment.

atolopko-czi: Great, maybe just make that ticket soon (with your above insights) so we don't forget. :)

atolopko-czi (approved):
Approving pre-maturely, since only blocked by the pip server/version update.

Dockerfile.processing_image
pablo-gar: We should do the testing before

pablo-gar (commented):
Looks good to me and I second Andrew's comment on the version of `cellxgene-schema`

PR #1414, opened 2021-09-15
fix: Bump rdev and Next.js build memory to 2048MB

tihuan:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.happy/terraform/modules/service/variables.tf
+++ b/.happy/terraform/modules/service/variables.tf
@@ -1,116 +1,116 @@
-variable vpc {
+variable "vpc" {
   type        = string
   description = "The VPC that the ECS cluster is deployed to"
 }
 
-variable custom_stack_name {
+variable "custom_stack_name" {
   type        = string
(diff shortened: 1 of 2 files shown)


ebezzi (approved):

.happy/terraform/modules/service/variables.tf
```diff
@@ -1,116 +1,116 @@
...
   type        = number
   description = "Listener rule priority number within the given listener"
 }
 
-variable memory {
+variable "memory" {
   type        = number
   description = "Amount of memory to allocate to each task"
-  default     = 768
+  default     = 2048
```
metakuni, line 110: Oh haha, not sure if 2048 is an appropriate value. I just took a stab in the dark. 😄

tihuan: I was using 8192, so I think yours is way more realistic than mine 😂

PR #1418, opened 2021-09-16
fix: patch FE vulnerabilities

tihuan:
Ran `npm audit fix` and `npm dedupe` to fix some vulnerabilities. `node-sass` fix is not included, since we still need Next.js to fix on their end!

Note that this is only patching `package-lock.json`, because `npm audit fix` only handles non breaking changes. `node-sass` upgrade to v6 is a breaking change, and Next.js' peerDependencies requires `4.x || 5.x` or `dart-sass`. But Blueprint is not compatible with `dart-sass`(they plan to [fix in v4](https://github.com/palantir/blueprint/pull/4820), so we're stuck on `node-sass` v5 until either Next.js updates its peer dependency or upgrade to Blueprint 4.x when it comes out

Also tested `npm run build && npm run serve` to make sure prod build works as expected 😆 

PTAL thank you!
---

## Changes
- add
- remove
- modify

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -3708,9 +3708,9 @@
       }
     },
     "node_modules/autoprefixer/node_modules/postcss": {
-      "version": "7.0.35",
-      "resolved": "https://registry.npmjs.org/postcss/-/postcss-7.0.35.tgz",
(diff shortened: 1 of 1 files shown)


tihuan:
Setting this as draft for now, since the package upgrade is not running the smoke tests 😆

leslieypark (approved):
Saw the updated description 👍 
LGTM!

tihuan:
Thanks so much, Leslie!

Issue #1426, opened 2021-09-20
Investigate inconsistency between dataset_id and explorer_url for a specific dataset


ebezzi:
This dataset:

https://cellxgene.staging.single-cell.czi.technology/e/7bfd5873-840b-4e42-882a-841df2f4be49.cxg/

has a different dataset_id when queried in the portal (`dfb825b4-28b7-4f8a-8e2f-68230b90a330`). 

1. Investigate why these don't match
2. Figure out if this behavior can be replicated somehow (e.g. through a revision flow)

PR #1431, opened 2021-09-21
fix: 1426 set new explorer_url for cloned datasets post revision creation

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@leslieypark 

Ticket [#1426](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1426)

---

## Changes
- Sets `explorer_url=None` when cloning a dataset after a revision. This is important because otherwise looking up for `explorer_url` (which is done by the meta endpoint) will yield multiple results and might not return the correct one.

## QA steps (optional)
QA this in staging by: 
1. Create a collection, add a dataset, publish
2. Create a revision
3. Open the old dataset in the explorer
4. Verify that the `dataset_identifier` in the `config` payload is the same as its id

## TODO
Add a condition to the revision functional test


tests/unit/backend/corpora/common/entities/datasets/test_revision.py
```diff
@@ -19,7 +19,9 @@ def test__create_dataset_revision(self):
         with self.subTest("deployment is correctly created and points to correct s3 uri "):
             rev_deployment = rev_dataset.pop("explorer_url")
             original_deployment = dataset.pop("explorer_url")
-            self.assertEqual(rev_deployment, original_deployment)
+            self.assertIsNotNone(original_deployment)
+            # explorer_url isn't set until a new dataset is uploaded
+            self.assertIsNone(rev_deployment)
```
leslieypark, line 24: So `rev_deployment` being None is the correct test for "deployment is correctly created and points to correct s3 uri"?

MDunitz (approved):
lgtm

tests/unit/backend/corpora/common/entities/datasets/test_revision.py
MDunitz: prior to uploading a dataset the explorer_url should be set to None to avoid issues identifying a dataset based on its explorer_url. Once a dataset is uploaded for the revision an explorer_url will be created as part of the upload process

leslieypark (approved):
🚀

MDunitz (approved):
lgtm

PR #1434, opened 2021-09-22
test: add tiledb regression test

metakuni:
### Reviewers
**Functional:** 
@bkmartinjr @MDunitz 

**Readability:** 
@leslieypark 

---

## Changes
- add: tiledb regression test
- remove: superfluous tiledb version spec in `Dockerfile.processing_image`

## QA steps (optional)
* ran `test__slash_in_attribute_name` locally using a known version of tiledb that would fail (tiledb==0.8.0)
* also ran via CI to demonstrate: https://github.com/chanzuckerberg/single-cell-data-portal/pull/1434/checks?check_run_id=3678016085#step:5:3314

## Known Issues
* This added unit test is a regression test of tiledb and contains no cellxgene code, so arguably doesn't belong in our unit tests, but it was requested, so here it is 😄

### DIFF
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
  && rm -rf single-cell-explorer client build/client /usr/bin/npm
 
 # Install python dependencies
-RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 tiledb==0.7.4 awscli
+RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 awscli
 RUN pip3 install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple cellxgene-schema==2.0.0rc1
 
 ADD requirements.txt requirements.txt
--- a/tests/unit/backend/corpora/dataset_processing/test_h5ad_data_file.py
+++ b/tests/unit/backend/corpora/dataset_processing/test_h5ad_data_file.py
@@ -174,6 +174,27 @@ class TestH5ADDataFile(unittest.TestCase):
         # Clean up
         remove(sparse_with_column_shift_filename)
 
+    def test__slash_in_attribute_name(self):
+        # tiledb failure repro code from https://github.com/TileDB-Inc/TileDB-Py/issues/294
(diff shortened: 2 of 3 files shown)


Dockerfile.processing_image
```diff
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
  && rm -rf single-cell-explorer client build/client /usr/bin/npm
 
 # Install python dependencies
-RUN pip3 install loompy==3.0.6 scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 tiledb==0.7.4 awscli
```
metakuni, line 24: I missed this previously when I was upgrading tiledb versions since I had only searched in `requirements.txt`.

In any case, this tiledb version is replaced below on line 29 from the version specified in `requirements.txt`.

tests/unit/backend/corpora/dataset_processing/test_h5ad_data_file.py
```diff
@@ -174,6 +174,28 @@ def test__to_cxg__with_sparse_column_encoding(self):
         # Clean up
         remove(sparse_with_column_shift_filename)
 
+    def test__slash_in_attribute_name(self):
```
metakuni, line 177: This bug repro code lifted from the [TileDB-Py bug that Bruce posted](https://github.com/TileDB-Inc/TileDB-Py/issues/294)

Using a known-bad version of tiledb (e.g. 0.8.0) line 194 will throw a `TileDBError` and cause the test to fail.

tests/unit/backend/corpora/dataset_processing/test_h5ad_data_file.py
```diff
@@ -174,6 +174,28 @@ def test__to_cxg__with_sparse_column_encoding(self):
...
+        # tiledb failure repro code from https://github.com/TileDB-Inc/TileDB-Py/issues/294
+        # this fails (throws a TileDBError) for 0.8.0 and below but works for 0.8.1+
+        import numpy as np
```
leslieypark, line 180: Quick question:
Is there a general practice reason for importing within the test vs. in the module? These imports already exist on lines 8 and 10.

metakuni: Ah, good point! Copy pasta 😄

metakuni: Fixed

MDunitz (approved):
lgtm -- thanks for adding the comments about failures

PR #1438, opened 2021-09-23
fix: revision test supports redirect from the explorer

ebezzi:
### Reviewers
**Functional:** 
@leslieypark @MDunitz 

**Readability:** 

---

## Changes
- This test now fails because we changed the behavior of those endpoints


tests/functional/backend/corpora/test_revisions.py
```diff
@@ -166,10 +166,11 @@ def test_revision_flow(self):
             res = requests.get(f"{self.api}/dp/v1/datasets/meta?url={self.create_explorer_url(deleted_dataset_id)}")
             self.assertEqual(res.status_code, 404)
 
-            # Endpoint is eventually consistent
-            (final_status_code, desired_status_code) = (None, 404)
-            for i in range(20):
-                res = requests.get(f"{self.api}/cellxgene/e/{original_dataset_id}.cxg/api/v0.2/schema")
+            # Endpoint is eventually consistent. This redirects to the collection page, so the status we want is 302
+            (final_status_code, desired_status_code) = (None, 302)
```
MDunitz, line 170: Would it be possible to also check the redirect link? 
Also this test is basically checking explorer functionality at this point. Its not a bad thing to be checking but I'm not sure it should be a part of the portal code base

tests/functional/backend/corpora/test_revisions.py
```diff
@@ -166,10 +166,11 @@ def test_revision_flow(self):
             res = requests.get(f"{self.api}/dp/v1/datasets/meta?url={self.create_explorer_url(deleted_dataset_id)}")
```
MDunitz, line 166: I dont think this test should work/if it is working we may an issue. This endpoint (`/datasets/meta`) shouldnt be 404ing for a (previously published) deleted dtaset. It should just have the tombstone status set to true and no s3_uri. To check that the dataset was removed from the data portal we probably want to check the collection details endpoint and make sure the dataset is no longer

ebezzi: You're right, but I think this test passes nevertheless. Let me verify that and I will add a comment.

tests/functional/backend/corpora/test_revisions.py
ebezzi: Yeah, this was previously checking for a 404. I think checking the redirect link shouldn't belong here, especially because if we change it in the explorer, this will break and the portal doesn't depend on the explorer. I would leave this as is.

metakuni (approved):

PR #1439, opened 2021-09-23
feat: remove loom references on frontend

seve:
Dropped all loom references on the frontend code

Also snuck in email validation by accident


frontend/src/views/Collection/utils.tsx
```diff
@@ -119,7 +119,7 @@ export function getIsPublishable(datasets: Array<Dataset>): boolean {
 
       return (
         numOfDeployments === 1 &&
-        numOfAssets === Object.keys(DATASET_ASSET_FORMAT).length
+        numOfAssets >= Object.keys(DATASET_ASSET_FORMAT).length
```
seve, line 122: numOfAssets will be > data asset formats until the backend is updated

tihuan: Should we wait to merge BE's change into this PR? Thanks!

frontend/tests/features/collection.test.ts
```diff
@@ -103,11 +104,21 @@ async function createCollection(
     testCollection.contactName,
     BLUEPRINT_SAFE_TYPE_OPTIONS
   );
+
+  if (tryBadEmail) {
+    // eslint-disable-next-line sonarjs/no-duplicate-string
```
tihuan, line 109: IMO it's still worth lifting `#contact-email` as a constant even in test, otherwise our tests would be harder to maintain long term?

frontend/src/views/Collection/utils.tsx
tihuan: Thinking about this a little more, maybe it's worth explicitly checking for expected formats? Feels like the total count is not reliable anymore 🤔

seve: @tihuan What makes you think that?

seve: I'm about to start working on backend changes, and I can swap back to `===` with them. Curious on why the number of assets isn't reliable outside of that case

tihuan: I was thinking about existing datasets having loom formats, so unless we clean up the database, `numOfAssets === Object.keys(DATASET_ASSET_FORMAT).length` wouldn't work in revision, since 4 > required 3

And `numOfAssets >= Object.keys(DATASET_ASSET_FORMAT).length` could cause false positive by enabling the publish button sooner, if `loom` was counted towards the threshold before the final required format becomes available. E.g.,
1. Upload dataset
2. cxg, h5ad, loom become available
3. publish button enabled <-- bug, since seurat format is not available yet
4. Thus needing to wait for BE change to be done

Something like that!

seve: What sort of workaround do you propose in the time between this PR and the second half of the ticket?

`numOfAssets > ASSET_KEYS.length` ?

seve: Once the backed half is in, we will no longer be serving the loom assets as part of the API response

tihuan: ahh I see, so loom won't show up in the API response 💡 That simplifies the check!

If that's the case, then we could use this PR as feature branch and merge the BE change in here when it's ready before merging everything, and keep the check as ` ===`?

seve: I separated the two parts of the ticket to help unblock launch if something goes off track. It seems like we're on path to start looking to partition tickets to have multiple deliverables, I thought this would be a good place to start

tihuan: Ahh I see! If we want to ship FE first without BE, I think we'll need to check for formats explicitly to avoid the edge cases discussed [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/1439#discussion_r717882472)

And then optionally change it back to `===` once BE guarantees no `loom` in response?

seve: Good idea!

frontend/src/views/Collection/utils.tsx
```diff
@@ -114,13 +114,20 @@ export function getIsPublishable(datasets: Array<Dataset>): boolean {
...
     datasets?.length > 0 &&
     datasets.every((dataset) => {
-      const numOfAssets = dataset.dataset_assets.length;
+      // const numOfAssets = dataset.dataset_assets.length;
       const numOfDeployments = dataset.dataset_deployments.length;
 
-      return (
-        numOfDeployments === 1 &&
-        numOfAssets >= Object.keys(DATASET_ASSET_FORMAT).length
+      // return (
```
tihuan, line 120: nit: Maybe add TODO for other readers to know why it's commented out 😄 ? Thanks!

tihuan (approved):
GO SEVE!! 🚀 🚀 🚀

PR #1447, opened 2021-09-24
chore: 1210 Add Plausible

tihuan:
### Reviewers
**Functional:** 
@seve 

**Readability:** 
@dimple-jethani 

---

## Changes
1. Add Plausible script
2. Update CSP to allow Plausible domain
3. Update config files to include Plausible env. Only prod.js is prod env, everything else uses staging env
4. NOTE: Plausible script ignore `localhost` event, so you'll only see `console.warn()` in the console when running locally

Screen Shot 2021-09-24 at 1 02 19 PM

Live link: https://thuang-plausible-frontend.rdev.single-cell.czi.technology/

PTAL thank you! 🎉 

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1210

## QA steps (optional)
1. Access Plausible Dashboard for staging env: https://plausible.io/cellxgene.staging.single-cell.czi.technology
2. Log in credentials in 1Password `cellxgene` vault
3. Check the current active user count
4. Visit: https://thuang-plausible-frontend.rdev.single-cell.czi.technology/
5. Check current active user count again, should increase by 1 (YOU!)


frontend/src/pages/_app.tsx
```diff
@@ -1,8 +1,10 @@
 import { AppProps } from "next/app";
+import Script from "next/script";
 import { QueryCache, ReactQueryCacheProvider } from "react-query";
 import { ReactQueryDevtools } from "react-query-devtools";
 import { checkFeatureFlags } from "src/common/featureFlags";
 import CookieBanner from "src/components/CookieBanner";
+import configs from "src/configs/configs";
 import "src/global.scss";
```
dimple-jethani, line 8: is there a certain way we order import statements on our team?

frontend/next.config.js
```diff
@@ -5,7 +5,9 @@ const { createSecureHeaders } = require("next-secure-headers");
 
 const isProdBuild = process.env.NODE_ENV === nodeEnv.PRODUCTION;
 
-const SCRIPT_SRC = ["'self'"];
+const PLAUSIBLE_URL = "https://plausible.io";
+
+const SCRIPT_SRC = ["'self'", PLAUSIBLE_URL];
```
dimple-jethani, line 10: why do we need to pass `PLAUSIBLE_URL` in the `SCRIPT_SRC`?

dimple-jethani (changes requested):
I see we're using the string `"cellxgene.staging.single-cell.czi.technology"` in a lot of places in this PR. In the case we accidentally mistype it/we shouldn't depend on typing this string correctly, we should set it as a const and use the const across files instead.

Please add a header to your PR title to get rid of the linter error. Otherwise, looks good!

frontend/src/pages/_app.tsx
tihuan: Yes! We use `prettier` to sort the imports for us 🙆‍♂️ 

Also in case you don't have it, we actually have `pre-commit` to lint both BE and FE at commit time:

`brew install pre-commit && pre-commit install`!

tihuan:
> I see we're using the string `"cellxgene.staging.single-cell.czi.technology"` in a lot of places in this PR. In the case we accidentally mistype it/we shouldn't depend on typing this string correctly, we should set it as a const and use the const across files instead.
> 
> Please add a header to your PR title to get rid of the linter error. Otherwise, looks good!

Ooh great idea! Done 🎉 !

frontend/next.config.js
tihuan: Great question!

script source to tell browser that it's okay to download and execute the script from Plausible
connect source to tell browser it's okay to connect to Plausible!

frontend/src/configs/common.js
```diff
@@ -0,0 +1,6 @@
+const PLAUSIBLE_DATA_DOMAIN_STAGING =
```
dimple-jethani, line 1: would we expect `common.js` file to hold all future strings as consts used across our codebase?

Is it a better idea to define this file to be more specific to just Plausible consts? I think it's fine for now but in the future, if this file becomes larger, we could consider splitting up more relevant consts.

dimple-jethani (commented):
this looks good to me, but I would like @seve to do the final approval!

frontend/src/configs/common.js
tihuan: Thanks for the great question, Dimple!

Since the file is inside `src/configs/`, the intention for `common.js` is only for hosting commonly used stuff within this directory

For files used across the whole codebase, we have a different `common` directory for it in `src/common/`, following [Dependency Inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle#Implementations) principle in SOLID!

Hope this makes sense 🙆‍♂️ Thank you!

dimple-jethani: thanks for the explanation! That makes sense.

seve (approved):

tihuan:
Thanks again everyone! Merging now 🎉

PR #1456, opened 2021-09-28
Chore: clean up old h5ad_data_file code

### LINKED ISSUE #1393: Clean up after updating cellxgene-schema for 2.0.0 changes

blrnw3:
fixes #1393

### Reviewers
**Functional:** 
Could do with someone super knowledgeable of the code here since I deleted some things which looked legacy but can't be 100%. Especially the `project_links` prop in the schema.
Git blame leads to @maniarathi 🙂 

**Readability:** 
@leslieypark 

---

## Changes
* Lots of unused code in h5ad_data_file - scrubbed
* Same in corpora_constants.py
* Fixed conversion of `batch_condition` by casting the ndarray to a list

## QA steps (optional)
* unit tests
* Converted the test h5ad file to cxg successfully

### DIFF
--- a/backend/corpora/common/utils/corpora_constants.py
+++ b/backend/corpora/common/utils/corpora_constants.py
@@ -1,106 +1,18 @@
-import collections
 
 
 class CorporaConstants(object):
-    # Namedtuples used to encapsulate information about a piece of metadata for verification purposes
-    TypedMetadata = collections.namedtuple(
-        "TypedMetadata", "field_name required_type valid_alternative", defaults=[None]
-    )
-    Ontology = collections.namedtuple("Ontology", "ontology_name s3_uri")
-
-    # Constants related to type checking ontology fields
-    EFO_ONTOLOGY = Ontology(ontology_name="efo", s3_uri="corpora-data-portal-ontologies/efo.txt")
-    CELL_ONTOLOGY = Ontology(ontology_name="cl", s3_uri="corpora-data-portal-ontologies/cl.txt")
-    HANCESTRO_ONTOLOGY = Ontology(ontology_name="hancestro", s3_uri="corpora-data-portal-ontologies/hancestro.txt")
-    HSAPDV_ONTOLOGY = Ontology(ontology_name="hsapdv", s3_uri="corpora-data-portal-ontologies/hsapdv.txt")
-    MONDO_ONTOLOGY = Ontology(ontology_name="mondo", s3_uri="corpora-data-portal-ontologies/mondo.txt")
-    NCBI_TAXON_ONTOLOGY = Ontology(ontology_name="ncbi_taxon", s3_uri="corpora-data-portal-ontologies/ncbitaxon.txt")
-    UBERON_ONTOLOGY = Ontology(ontology_name="uberon", s3_uri="corpora-data-portal-ontologies/uberon.txt")
-
-    CORPORA_ONTOLOGIES = [
-        EFO_ONTOLOGY,
-        CELL_ONTOLOGY,
-        HANCESTRO_ONTOLOGY,
-        HSAPDV_ONTOLOGY,
-        MONDO_ONTOLOGY,
-        NCBI_TAXON_ONTOLOGY,
-        UBERON_ONTOLOGY,
-    ]
-
-    # Constants related to type checking enum fields
-    SEX_VALID_ENUM_VALUES = ["male", "female", "na", "unknown"]
-
-    # Constants related to processing layers
-    LAYER_DESCRIPTIONS = "layer_descriptions"
-    X_DATA_LAYER_NAME = "X"
-    RAW_DATA_LAYER_NAME = "raw.X"
-
-    # Constants related to the metadata requirements of submitting a dataset to Corpora. Note that these metadata
-    # fields are CASE-SENSITIVE.
-    REQUIRED_OBSERVATION_METADATA_FIELDS = [
-        TypedMetadata(field_name="tissue", required_type=str),
-        TypedMetadata(field_name="assay", required_type=str),
-        TypedMetadata(field_name="disease", required_type=str),
-        TypedMetadata(field_name="cell_type", required_type=str),
-        TypedMetadata(field_name="sex", required_type=SEX_VALID_ENUM_VALUES),
-        TypedMetadata(field_name="ethnicity", required_type=str),
-        TypedMetadata(field_name="developmental_stage", required_type=str),
-    ]
-    REQUIRED_OBSERVATION_ONTOLOGY_METADATA_FIELDS = [
-        TypedMetadata(field_name="tissue_ontology", required_type=UBERON_ONTOLOGY),
-        TypedMetadata(field_name="assay_ontology", required_type=EFO_ONTOLOGY),
-        TypedMetadata(field_name="disease_ontology", required_type=MONDO_ONTOLOGY, valid_alternative="PATO:0000461"),
-        TypedMetadata(field_name="cell_type_ontology", required_type=CELL_ONTOLOGY, valid_alternative=""),
-        TypedMetadata(field_name="ethnicity_ontology", required_type=HANCESTRO_ONTOLOGY, valid_alternative="na"),
-        TypedMetadata(
-            field_name="developmental_stage_ontology", required_type=HSAPDV_ONTOLOGY, valid_alternative="EFO:0000399"
-        ),
-    ]
-    REQUIRED_DATASET_METADATA_FIELDS = [
-        TypedMetadata(field_name="organism", required_type=str),
-        TypedMetadata(field_name="organism_ontology", required_type=NCBI_TAXON_ONTOLOGY),
-        TypedMetadata(field_name=LAYER_DESCRIPTIONS, required_type=dict),
-    ]
-    REQUIRED_DATASET_PRESENTATION_METADATA_FIELDS = [
-        TypedMetadata(field_name="title", required_type=str),
-    ]
-
-    REQUIRED_DATASET_PRESENTATION_HINTS_METADATA_FIELDS = [
-        TypedMetadata(field_name="color_map", required_type=dict),
-        TypedMetadata(field_name="tags", required_type=list),
-        TypedMetadata(field_name="default_field", required_type=str),
-        TypedMetadata(field_name="default_embedding", required_type=str),
-    ]
-    OPTIONAL_COLLECTION_LEVEL_METADATA_FIELDS = [
-        TypedMetadata(field_name="collection_name", required_type=str),
-        TypedMetadata(field_name="collection_description", required_type=str),
-        TypedMetadata(field_name="collection_protocol_links", required_type=list),
-        TypedMetadata(field_name="collection_raw_data_links", required_type=list),
-        TypedMetadata(field_name="collection_other_links", required_type=list),
-    ]
-
     REQUIRED_SIMPLE_METADATA_FIELDS = [
         "schema_version",
         "title",
         "X_normalization",
     ]
 
-    # The Corpora specification requires some values encoded as JSON due to the inability of AnnData to store complex
-    # types.
-    OPTIONAL_JSON_ENCODED_METADATA_FIELD = ["contributors", "project_links"]
-
-    OPTIONAL_SIMPLE_METADATA_FIELDS = [
-        "preprint_doi",
-        "publication_doi",
-        "default_embedding",
-        "default_field",
-        "tags",
-        "project_name",
-        "project_description",
-        "X_approximate_distribution",
+    # Lists are encoded as ndarrays in h5ad/anndata, so must be cast to lists during cxg conversion
+    OPTIONAL_LIST_METADATA_FIELDS = [
+        "batch_condition",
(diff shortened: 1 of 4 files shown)


backend/corpora/common/utils/corpora_constants.py
```diff
@@ -1,106 +1,18 @@
...
 
-    # The Corpora specification requires some values encoded as JSON due to the inability of AnnData to store complex
-    # types.
-    OPTIONAL_JSON_ENCODED_METADATA_FIELD = ["contributors", "project_links"]
+    # Lists are encoded as ndarrays in h5ad/anndata, so must be cast to lists during cxg conversion
+    OPTIONAL_LIST_METADATA_FIELDS = [
+        "batch_condition",
+    ]
 
     OPTIONAL_SIMPLE_METADATA_FIELDS = [
```
blrnw3, line 15: removed anything not listed in the schema doc (same for json-encoded stuff)

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -29,21 +30,14 @@ class H5ADDataFile:
     def __init__(
         self,
         input_filename,
-        backed=False,
```
blrnw3, line 32: removed since never called by code in this repo

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -29,21 +30,14 @@ class H5ADDataFile:
...
-        dataset_title=None,
-        dataset_about=None,
-        obs_index_column_name=None,
-        vars_index_column_name=None,
-        use_corpora_schema=True,
+        var_index_column_name=None,
```
blrnw3, line 33: fixed typo in this param

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -146,7 +140,7 @@ def generate_cxg_metadata(self, convert_anndata_colors_to_cxg_colors):
 
         cxg_group_metadata = {
             "cxg_version": CxgConstants.CXG_VERSION,
-            "cxg_properties": json.dumps({"title": self.dataset_title, "about": self.dataset_about}),
```
blrnw3, line 149: unused, remove from here too?

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -187,41 +174,30 @@ def validate_anndata(self):
...
-        self.corpora_properties = self.get_corpora_properties() if self.use_corpora_schema else None
-
-        if self.corpora_properties is None and self.use_corpora_schema:
+        if self.corpora_properties is None:
             # If the return value is None, this means that we were not able to figure out what version of the Corpora
             # schema the object is using and therefore cannot extract any properties.
             raise ValueError("Unknown source file schema version is unsupported.")
 
         # The title and about properties of the dataset are set by the following order: if they are explicitly defined
-        # then use the explicit value. If the dataset is a Corpora-schema based schema, then extract the title and about
```
blrnw3, line 211: we don't have project_links anymore, right?

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -273,13 +249,14 @@ def get_corpora_properties(self):
...
 
-        for key in CorporaConstants.OPTIONAL_JSON_ENCODED_METADATA_FIELD:
+        for key in CorporaConstants.OPTIONAL_LIST_METADATA_FIELDS:
             if key not in self.anndata.uns:
                 continue
             try:
-                corpora_props[key] = json.loads(self.anndata.uns[key])
-            except json.JSONDecodeError:
-                raise json.JSONDecodeError(f"Corpora schema field {key} is expected to be a valid JSON string")
+                corpora_props[key] = self.anndata.uns[key].tolist()
```
blrnw3, line 256: new support for lists encoded as ndarray

leslieypark (approved):
Readability LGTM!

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -29,21 +30,14 @@ class H5ADDataFile:
...
+        var_index_column_name=None,
     ):
         self.input_filename = input_filename
-        self.backed = backed
-        self.dataset_title = dataset_title
-        self.dataset_about = dataset_about
-        self.obs_index_column_name = obs_index_column_name
-        self.vars_index_column_name = vars_index_column_name
-
-        self.use_corpora_schema = use_corpora_schema
```
maniarathi, line 46: Looking ahead a bit; the future of this code is that it will be moved out of the Portal repo and into the Explorer repo. It is worth noting that we may want to host datasets that do not adhere to the schema, but we want to manually transform them to CXG and put them up on the `/d/` dataroot. Because of this, I would recommend not removing this flag.

backend/corpora/dataset_processing/h5ad_data_file.py
maniarathi: Correct. Even if we were to add project links to the dataset, we probably shouldn't surface them through the CXG which is immutable.

backend/corpora/dataset_processing/h5ad_data_file.py
atolopko-czi: Is `about` unused in explorer?  The current CXG 0.2.0 spec does [include it](https://github.com/chanzuckerberg/single-cell-explorer/blob/main/dev_docs/cxg.md#cxg_properties-field) still, so would leave in the property for now, even if it has no value.

backend/corpora/dataset_processing/h5ad_data_file.py
atolopko-czi: @maniarathi Why would this be moved to the Explorer repo?  single-cell-explorer (hosted)?  I was thinking we might create a small, standalone CXG library, that encapsulates both the read and writing of the CXG format, and would also managed the CXG spec document.  Keeping all the logic in one place, and dealing with versioning compatibility seamlessly. Transforming from H5AD files to CXG could be part of that library as well.

atolopko-czi (approved):
A few comments, but nothing blocking approval. Thanks for the cleanup! I spent too much time during earlier efforts mentally eliding the unused code.

backend/corpora/dataset_processing/h5ad_data_file.py
maniarathi: @atolopko-czi Yes, you're right. I had paused on reviewing this, but `title` and `about` are ones that we'd use as well. 

I'd do a quick check with Product to see if we have any potential datasets coming through the near term pipeline (say in the next month or so). If there is, I would recommend keeping it as is and opening a new issue to modify this to clean up the design during the Q4 cool down period. If not, then I would go ahead and get rid of it and let Product know that it'll take a touch more of time to convert non-compliant datasets the next time we want to. Does that sound like a reasonable compromise?

maniarathi: Also @atolopko-czi -- creating a small standalone CXG library works too! A while back, Madison, Bruce, and I had discussed keeping it in the repo in which it serves (i.e. CXG serves the single-cell-explorer tool therefore should live there). I can't recall the exact details but worth discussing with them as well.

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -187,41 +174,30 @@ def validate_anndata(self):
 
     def extract_anndata_elements_from_file(self):
         logging.info(f"Reading in AnnData dataset: {path.basename(self.input_filename)}")
-        self.anndata = anndata.read_h5ad(self.input_filename, backed="r" if self.backed else None)
+        self.anndata = anndata.read_h5ad(self.input_filename)
```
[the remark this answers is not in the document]
maniarathi: Yep, this was there for memory reasons but I agree with your assessment in that it's easy to add back in if necessary. We use some pretty big EC2 instances so don't think we need it for now.

backend/corpora/dataset_processing/h5ad_data_file.py
maniarathi: It is currently unused but will be used by the new info drawer that the CC team is working on now.

backend/corpora/dataset_processing/h5ad_data_file.py
```diff
@@ -187,41 +174,30 @@ def validate_anndata(self):
...
-            )
-        else:
-            corpora_about_link = {}
+        # then use the explicit value. Otherwise, use the input filename (only for title, about will be blank).
 
         filename = path.splitext(path.basename(self.input_filename))[0]
-
-        self.dataset_title = self.dataset_title if self.dataset_title else corpora_about_link.get("link_name", filename)
-        self.dataset_about = self.dataset_about if self.dataset_about else corpora_about_link.get("link_url")
+        self.dataset_title = self.dataset_title or filename
```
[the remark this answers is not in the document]
maniarathi: Checked the frontend code -- if there is a title in `corpora_properties` then it prefers that title. Otherwise it uses the top-level title. This is a bit of a legacy design where the CXG conversion was built before we had a "corpora schema" and then once we had a schema, we heavily reduced the number of non-compliant H5ADs to convert.

maniarathi (approved):
LGTM! Left a few comments, but feel comfortable with you merging in the changes regardless of how you choose to resolve it! Thank you and sorry about the wait!

backend/corpora/dataset_processing/h5ad_data_file.py
atolopko-czi: If we need to manually create a CXG, we'd use the `single-cell-explorer` repo's [CLI](server/cli/convert_to_cxg.py), right? And since that repo has its own `h5ad_data_file.py` file, it will still support all the non-schema compliant H5AD->CXG conversion attributes (title, about, etc.).  So I think we should be able to make this repo only work for schema-compliant H5AD->CXG file.

backend/corpora/dataset_processing/h5ad_data_file.py
atolopko-czi: [Asked Brian about `about` on Slack](https://czi-sci.slack.com/archives/C023Q1APASK/p1633441275138600), in case CC is expecting to it have a value.

backend/corpora/dataset_processing/h5ad_data_file.py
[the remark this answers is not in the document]
atolopko-czi: Got it, thx.  For now, I've noted in my arch vision notes that we might consider moving away from having any "Corpora Schema"-awareness in CXG.  It's added a lot of confusion to what a valid CXG file should contain, and where it should be stored. The issue is also now noted in the [CXG spec](https://github.com/chanzuckerberg/single-cell-explorer/blob/main/dev_docs/cxg.md#corpora-schema-encoding).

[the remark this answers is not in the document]
maniarathi: ^^ this is definitely the way to go in the future. Thanks for noting it!

backend/corpora/dataset_processing/h5ad_data_file.py
maniarathi: Ah shoot -- I had forgotten that the CLI in the `single-cell-explorer` still existed. This is actually duplicated code. The code here was pulled straight from there and then we just removed the CLI wrapper. 

I actually think that in the long term, we shouldn't have two copies of the code. The CLI in the `single-cell-explorer` is currently not used by anything. I see two options here: either consolidate the conversion code to a standalone repo or consolidate to the `single-cell-explorer` repo. The code should be able to handle both compliant and non-compliant H5ADs since (as you've stated below) any metadata that is schema-specific should probably not be in the CXG itself but rather surfaced perhaps through a microservice/API (imo).

Does that align with what you're think/does that make sense?

atolopko-czi: Yup, 100%! The CLI, if needed, could tag along in the library repo (or subdir). And agree decisions need to be made re: attributes that are surfaced in CXG vs API.  I need to clean up my Arch notes so I can share & use as forum for ongoing discussion. :)

backend/corpora/common/utils/corpora_constants.py
blrnw3: see https://github.com/chanzuckerberg/single-cell-explorer/pull/117

backend/corpora/dataset_processing/h5ad_data_file.py
blrnw3: Will do re: commit msg 👍
We didn't create any story for this piece, just for the db/api side https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1459

blrnw3:
Thanks for all your comments, both.
I think the best way forward is to address the currently-unused fields at a later time. Agree with thoughts that 
a) conversion logic should live in one place. The ingestion pipeline seems the right place to me but I may be missing some use cases.
b) cxg is not the best place for a lot of this metadata anyway; should be stored in the db and leave cxg for the core data

PR #1464, opened 2021-09-29
chore: script to backfill missing processing status

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

---

## Changes
- Run this script to backfill the processing status.


MDunitz (approved):
This looks good to me. My only concern is setting the conversion status for a dataset that doesnt have the underlying artifacts. Do you know how many datasets are missing a processing status? Would it be possible to check that they have the artifacts stored in s3?

PR #1510, opened 2021-10-07
fix: properly check links for differences

seve:
### Reviewers
**Functional:** @tihuan 

**Readability:** @leslieypark 

---

Sometimes I don't like the ES6 Array methods


tihuan (approved):
Ah makes sense! Thanks for the fix!

PR #1516, opened 2021-10-07
fix: use list diffing instead of incorrect object equality check

seve:
### Reviewers
**Functional:** @tihuan 

---

There were two bugs, last PR only caught the returning true in `Array.forEach()` this catches the `object === object`


tihuan (approved):
Ahh `links` are objects! Glad that's caught 👍 LGTM!

PR #1523, opened 2021-10-08
fix: use frontend url for plausible data domain

seve:
### Reviewers
**Functional:** @ebezzi 

Revert all our attempts to fix config to adhere to plausible use.  Instead, just use a trimmed frontend URL when working with containerized frontend

NOTE: This assumes that it is safe to have invalid domains in the data-domain field (frontend.corporanet.local, rdev urls)

### DIFF
--- a/frontend/src/views/Collection/components/Banner/constants.ts
+++ b/frontend/src/views/Collection/components/Banner/constants.ts
@@ -1,16 +1,10 @@
 export const INCOMPLETE_GENE_MIGRATIONS_COLLECTIONS = [
+  "b9fc3d70-5a72-4479-a046-c2cc1ab19efc",
+  "6e8c5415-302c-492a-a5f9-f29c57ff18fb",
+  "d0e9c47b-4ce7-4f84-b182-eddcfa0b2658",
   "5d445965-6f1a-4b68-ba3a-b8f765155d3a",
   "a3c37598-b8e8-47db-92a9-eb91a77c9529",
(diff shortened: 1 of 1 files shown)


frontend/container_init.sh
```diff
@@ -9,7 +9,7 @@ if [ ! -e ./node_modules ]; then
 fi
 ln -sf /opt/node_app/node_modules/* /opt/node_app/node_modules/.bin ./node_modules/.
 if [ ! -z "$API_URL" ]; then
-  cat src/configs/$DEPLOYMENT_STAGE.js | envsubst >src/configs/configs.js
```
seve, line 12: Didn't end up going with this solution because it assumes a lot of hardcoded values, most importantly API_URL

frontend/src/configs/rdev.js
```diff
@@ -1,8 +1,7 @@
-const { PLAUSIBLE_DATA_DOMAIN_STAGING } = require("./common");
-
 const configs = {
   API_URL: "$API_URL",
-  PLAUSIBLE_DATA_DOMAIN: PLAUSIBLE_DATA_DOMAIN_STAGING,
+  // plausible doesn't expect  a protocol in the domain
+  PLAUSIBLE_DATA_DOMAIN: "$FRONTEND_URL".replace(/^https?:\/\//, ""),
```
seve, line 4: For production applications, this approach feels the most DRY and less of a headache should anything change upstream

Issue #1466, opened 2021-09-30
Delete Molecular characterization of foveal versus peripheral human retina by single-cell RNA sequencing


brianraymor:
The curators have requested the deletion of the [Molecular characterization of foveal versus peripheral human retina by single-cell RNA sequencing](https://cellxgene.cziscience.com/collections/7edef704-f63a-462c-8636-4bc86a9472bd) collection.

**The script must be reusable for future requests.**

maniarathi:
Is this issue to delete the collection or the dataset? We currently haven't implemented the toast that should pop up when a collection is deleted so I'm wondering if this should be blocked by the epic to delete a published collection.

brianraymor:
We will delete the entire collection and not wait for the delete collection feature. When it's finally implemented, then the toast should appear. Until then, if it's linked, it will return a 404. We also discussed other options like hiding collections, toggling collection private/pubic state, etc.

metakuni:
For posterity:
```
$ AWS_PROFILE=single-cell-prod ./scripts/cxg_admin.py --deployment prod tombstone-collection 7edef704-f63a-462c-8636-4bc86a9472bd
```

PR #1527, opened 2021-10-08
chore: added `tombstone-collection` command to `cxg_admin.py` (#1466)

metakuni:
### Reviewers
**Functional:** 
@MDunitz @ebezzi 

**Readability:** 
@blrnw3 

---

## Changes
- added: `tombstone-collection` command

### DIFF
--- a/frontend/container_init.sh
+++ b/frontend/container_init.sh
@@ -9,7 +9,7 @@ if [ ! -e ./node_modules ]; then
 fi
 ln -sf /opt/node_app/node_modules/* /opt/node_app/node_modules/.bin ./node_modules/.
 if [ ! -z "$API_URL" ]; then
-  cat src/configs/rdev.js | envsubst >src/configs/configs.js
+  cat src/configs/$DEPLOYMENT_STAGE.js | envsubst >src/configs/configs.js
 else
   cp src/configs/local.js src/configs/configs.js
 fi
--- a/frontend/src/configs/dev.js
+++ b/frontend/src/configs/dev.js
@@ -1,7 +1,7 @@
 const { PLAUSIBLE_DATA_DOMAIN_STAGING } = require("./common");
 
 const configs = {
-  API_URL: "$API_URL",
+  API_URL: "https://api.cellxgene.dev.single-cell.czi.technology",
   PLAUSIBLE_DATA_DOMAIN: PLAUSIBLE_DATA_DOMAIN_STAGING,
 };
 


scripts/cxg_admin.py
```diff
@@ -76,6 +76,51 @@ def delete_dataset(ctx, uuid):
             exit(0)
 
 
+@cli.command()
+@click.argument("uuid")
+@click.pass_context
+def tombstone_collection(ctx, uuid):
```
blrnw3, line 82: nit: type hints on the parameters

scripts/cxg_admin.py
```diff
@@ -76,6 +76,51 @@ def delete_dataset(ctx, uuid):
...
+    """
+    Tombstones the collection specified by UUID.
+
+    You must first SSH into the target deployment using `make db/tunnel` before running.
```
blrnw3, line 86: I know this is used a lot but it always confused me. For me, `SSH`  when used as a verb means to literally log in to the remote machine, not make a tunnel :D

scripts/cxg_admin.py
```diff
@@ -76,6 +76,51 @@ def delete_dataset(ctx, uuid):
...
+    """
+
+    with db_session_manager() as session:
+        collection = Collection.get_collection(session, uuid, include_tombstones=True)
+
+        if not collection:
+            click.echo(f"Collection:{uuid} not found!")
+            exit(0)
+        else:
+            print(
```
blrnw3, line 100: nit: `click.echo` for consistency. Pretty sure they do the same thing so you could make everything a print if you like :D

blrnw3 (approved):
Readability comments inline. Functionality looks right to me too 🤷

scripts/cxg_admin.py
metakuni: Fixed

scripts/cxg_admin.py
metakuni: Ah right, fixed.

Dockerfile.processing_base
```diff
@@ -20,5 +20,8 @@ RUN apt-get update && apt-get -yq install \
 RUN R -e "install.packages('BiocManager')"
 RUN R -e "BiocManager::install('SingleCellExperiment')"
 RUN R -e "install.packages(c('devtools', 'igraph'));"
+# note: Loom format has been deprecated from cellxgene but cellgeni/sceasy has a dependency on LoomExperiment
+#       See https://github.com/cellgeni/sceasy#installation
+RUN R -e "BiocManager::install('LoomExperiment')"
```
metakuni, line 25: This has nothing to do with tombstoning of collections, but this PR's build was broken by removing this dependency in #1465 , so re-adding it here.

metakuni: ... and now, the bioconductor.org site appears to be down causing dependency installation to fail: 🤦 
```
...
#8 490.1   cannot open URL 'https://bioconductor.org/packages/3.12/bioc/src/contrib/LoomExperiment_1.8.0.tar.gz': HTTP status was '504 Gateway Timeout'
...
```

MDunitz (approved):
lgtm

PR #1555, opened 2021-10-14
feat: added explain new tab toast. (#135)

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
- Added hook to check for `explainNewTab` query string param and pop work-in-progress toast if present.


tihuan:
Hurray! Thanks so much for the PR, @MillenniumFalconMechanic 🎉 !

Can you help check with @brianraymor to see if this needs to go into the hotfix branch please?

if so, we probably just need to update this branch to be based on the `hotfix` branch and merge against `hotfix`, in case it pulls in commits from `main` we don't want 🙏 

Thank you!!

PR #1558, opened 2021-10-15
fix: allow popover to handle repositioning of popouts

seve:
### Reviewers
**Functional:** @tihuan 

---

## Changes
- modify
   - Change popover options to allow for repositioning based on viewport


tihuan (approved):
Great solution indeed! LGTMMM 🚀 🚀 🚀

Issue #1577, opened 2021-10-25
Endpoint to delete published collection


seve:
Backend needs an endpoint to allow for the deletion of published collections and return collection tombstones to allow for proper redirect/toast behavior on frontend.

@Bento007 has completed most of the work needed for this ticket, remaining work is testing and tombstone response

Bento007:
Will the user be using the revision flow to delete a collection?

PR #1633, opened 2021-11-01
feat: delete published collection

Bento007:
### Reviewers
**Functional:** @seve 

**Readability:** 

---

## Changes
- modify DELETE collection/{collection_uuid} endpoint to allow the owner of a collection to tombstone a published collection. If an active revision exists for the collection, it will be deleted.
- modified GET collection/{collection_uuid} to return a gone<410> response with the contact_name and contact_email in the response body.

https://github.com/chanzuckerberg/single-cell-data-portal/issues/1577

### DIFF
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -53,15 +53,20 @@ def get_collections_list(from_date: int = None, to_date: int = None, user: Optio
 @dbconnect
 def get_collection_details(collection_uuid: str, visibility: str, user: str):
     db_session = g.db_session
-    collection = Collection.get_collection(db_session, collection_uuid, visibility)
+    collection = Collection.get_collection(db_session, collection_uuid, visibility, include_tombstones=True)
     if not collection:
         raise ForbiddenHTTPException()
-    get_tombstone_datasets = (
-        _is_user_owner_or_allowed(user, collection.owner) and collection.visibility == CollectionVisibility.PRIVATE
-    )
-    result = collection.reshape_for_api(get_tombstone_datasets)
+    if collection.tombstone == True and visibility == CollectionVisibility.PUBLIC.name:
+        result = ""
+        response = 410
+    else:
+        get_tombstone_datasets = (
+            _is_user_owner_or_allowed(user, collection.owner) and collection.visibility == CollectionVisibility.PRIVATE
+        )
+        result = collection.reshape_for_api(get_tombstone_datasets)
+        response = 200
     result["access_type"] = "WRITE" if _is_user_owner_or_allowed(user, collection.owner) else "READ"
-    return make_response(jsonify(result), 200)
+    return make_response(jsonify(result), response)
 
 
 @dbconnect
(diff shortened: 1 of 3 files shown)


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -53,15 +53,20 @@ def get_collections_list(from_date: int = None, to_date: int = None, user: Optio
...
-    collection = Collection.get_collection(db_session, collection_uuid, visibility)
+    collection = Collection.get_collection(db_session, collection_uuid, visibility, include_tombstones=True)
     if not collection:
         raise ForbiddenHTTPException()
-    get_tombstone_datasets = (
-        _is_user_owner_or_allowed(user, collection.owner) and collection.visibility == CollectionVisibility.PRIVATE
-    )
-    result = collection.reshape_for_api(get_tombstone_datasets)
+    if collection.tombstone == True and visibility == CollectionVisibility.PUBLIC.name:
+        result = dict(contact_name=collection.contact_name, contact_email=collection.contact_email)
```
seve, line 60: Brian got back to me on this, we're going to return just the tombstone with no data attached.  A bit naive, but safe for now until we revisit later with the dataset tombstone response as well

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -110,23 +115,43 @@ def get_collection_dataset(dataset_uuid: str):
...
+            db_session,
+            collection_uuid,
+            CollectionVisibility.PRIVATE.name,
+            owner=_owner_or_allowed(user),
+            include_tombstones=True,
+        )
+
+        if pub_collection:
+            if not pub_collection.tombstone:
+                pub_collection.tombstone_collection()
```
seve, line 137: Is there a needed error in the case that the collection has already been tombstoned?

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -110,23 +115,43 @@ def get_collection_dataset(dataset_uuid: str):
...
-        return "", 204
+    if visibility == CollectionVisibility.PUBLIC.name:
+        pub_collection = Collection.get_collection(
+            db_session,
+            collection_uuid,
+            visibility,
+            owner=_owner_or_allowed(user),
+            include_tombstones=True,
+        )
+        priv_collection = Collection.get_collection(
```
seve, line 127: Will this catch if there have been more than one attempted revision of the collection?
i.e.
1. Active revision
1. Tombstoned revision
1. Tombstoned revision

Is it possible hat `.get_collection()` could grab one tombstoned revision and skip:
```
if not priv_collection.tombstone:
                    priv_collection.delete()
```

seve (changes requested):
Let's drop the FE commits from the PR, I'll test them and put them out in a seperate PR to the same feat branch

backend/corpora/lambdas/api/v1/collection.py
Bento007: Can you expand on "just the tombstone with no data". What would that response look like?

backend/corpora/lambdas/api/v1/collection.py
Bento007: You'll see on line 152 that private collections are deleted not tombstoned. So in this case there is only ever one active revision, and never any tombstoned revisions. So to answer your question, is it possible that `.get_collection()` could grab one tombstone revision... I don't see how it could since no other revisions should exist in the database.

backend/corpora/lambdas/api/v1/collection.py
Bento007: No if the collection is already tombstoned, then we return 204 to indicate to the user that the deletion was and still is successful. This was an old decision made before I left.

backend/corpora/lambdas/api/v1/collection.py
seve: Actually I guess that would just be empty contents and just the 410

seve (approved):

PR #1635, opened 2021-11-01
fix: pin packaging version in dockerfile.processing_image to fix smoke tests

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---
This is the original failing [smoke test](https://github.com/chanzuckerberg/single-cell-data-portal/runs/4072903438?check_suite_focus=true#step:5:339). Here is the error in our smoke test.
> ERROR: packaging 21.2 has requirement pyparsing<3,>=2.0.2, but you'll have pyparsing 3.0.4 which is incompatible.

The python library Packaging is failing during install do to conflicting requirements. Packaging's version 21.2 locks down the version of pyparsing to resolve this https://github.com/pypa/packaging/pull/471. 

There is also an error installing the requirements for Owlready2==0.34. This is pinned in [chanzuckerberg/single-cell-curation](https://github.com/chanzuckerberg/single-cell-curation/blob/37d8a199c4017d743694ad9ff47f8ccca555efd0/cellxgene_schema_cli/requirements.txt#L7)

## Changes
- pin packaging version in dockerfile.processing_image to one that does not lock the pyparsing library version to >3.

### DIFF
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -19,7 +19,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
  && rm -rf single-cell-explorer client build/client /usr/bin/npm
 
 # Install python dependencies
-RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 cython==0.29.24 awscli
+RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 pyparsing<3,>=2.0.2 awscli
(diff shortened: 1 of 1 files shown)


tihuan (approved):
rubber stamping assuming this is a safe change 😆 Thanks so much, @Bento007 !

maniarathi:
One question and one ask: why not pin the `packaging` version instead of creating an explicit dependency on a  downstream dependency?

And the ask: can you please make sure to write a comment in code as to why it was pinned? We should remove it in the future once all the packages fast forward in compatibility.

PR #1636, opened 2021-11-02
chore: Update FE deps

tihuan:
### Reviewers
**Functional:** 
@seve 

**Readability:** 
@Bento007 

---

## Changes
1. Update most packages in minor and patch versions
2. Only a few safe major version upgrades that I've tested their corresponding features still work. Such as clip board, jest tests, etc.
3. I didn't upgrade Next.js to 12, since we use Styled Components, which Next.js is still in the process of porting its transpiler [from Babel to SWC](https://nextjs.org/docs/upgrading), and since we use Styled Components, which they don't have a plugin for yet, I think we can wait a few more months before migrating to v12

Test out live link: https://thuang-deps-frontend.rdev.single-cell.czi.technology/
Everything should work as expected!


.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: v2.4.0
     hooks:
       - id: prettier
-        args: [--write, ./frontend]
+        args: ["--ignore-path ./frontend/.prettierignore --write ./frontend"]
```
tihuan, line 17: Pass all args as one string to avoid bugs. Previously we were linting the whole frontend repo every time, which is why every pre-commit was slow. This change also passes `.prettierignore` file

frontend/src/components/Collections/components/Dataset/components/DownloadDataset/components/Content/index.tsx
```diff
@@ -62,7 +63,10 @@ const Content: FC<Props> = ({ onClose, name, dataAssets }) => {
 
       try {
         const result = await (
-          await fetch(`${API_URL}${url}`, { method: "POST" })
+          await fetch(`${API_URL}${url}`, {
+            ...DEFAULT_FETCH_OPTIONS,
```
tihuan, line 67: Add `cookies` for the POST call

frontend/tests/utils/helpers.ts
```diff
@@ -13,28 +13,26 @@ export async function goToPage(url: string = TEST_URL): Promise<void> {
 }
 
 export async function login(): Promise<void> {
-  goToPage();
+  await goToPage();
```
tihuan, line 16: Missing this await was causing memleak and test failures

frontend/tests/utils/helpers.ts
```diff
@@ -13,28 +13,26 @@ export async function goToPage(url: string = TEST_URL): Promise<void> {
...
-  try {
-    await expect(page).toHaveSelector(getText("My Collections"), {
-      timeout: TIMEOUT_MS,
-    });
-  } catch (error) {
-    const username = TEST_USERNAME;
+  const cookies = await context.cookies();
 
-    await page.click(getText("Log In"));
+  if (cookies.length) return;
```
tihuan, line 22: Instead of checking if `Log In` button is in the DOM, which can be flaky, we just check if the browser already has cookies

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: v2.4.0
     hooks:
       - id: prettier
-        args: [--write, ./frontend]
+        args: [--write]
```
tihuan, line 17: remove `./frontend`, so prettier will only lint staged files instead of linting the whole frontend directory

frontend/jest/e2e.config.js
```diff
@@ -9,7 +9,4 @@ module.exports = {
   setupFilesAfterEnv: ["expect-playwright", "./jest/playwright.setup.js"],
   testMatch: ["**/tests/**/?(*.)(spec|test).[jt]s?(x)"],
   testRunner: "jest-circus/runner",
-  transform: {
```
tihuan, line 12: transform doesn't seem needed with jest 27!

seve (approved):

PR #1638, opened 2021-11-03
feat: datasets index endpoint

ebezzi:
### Reviewers
**Functional:** 
@MillenniumFalconMechanic @NoopDog 

**Readability:** 

---

## Changes
WIP for the /datasets endpoint. Need to : 
1. Refine the schema 
2. Add tests
3. Assess performance

Don't merge until this is refined and tested in rdev.


backend/Makefile
```diff
@@ -53,6 +53,12 @@ db/connect:
 	# ARGS usage: AWS_PROFILE=single-cell-dev DEPLOYMENT_STAGE=dev make db/connect ARGS="-c \"select * from dataset_artifact where filetype='CXG'\""
 	PGPASSWORD=${DB_PW} psql --dbname corpora_${DEPLOYMENT_STAGE} --username corpora_${DEPLOYMENT_STAGE} --host 0.0.0.0 $(ARGS)
 
+db/rdev-connect:
```
ebezzi, line 56: Not needed by this PR, but this is a useful command if you want to connect to the rdev database.

Bento007 (commented):
Looks good so far.

backend/corpora/common/entities/dataset.py
```diff
@@ -124,6 +124,62 @@ def get_asset(self, asset_uuid) -> typing.Union[DatasetAsset, None]:
...
+            DbDataset.cell_type,
+            DbDataset.sex,
+            DbDataset.ethnicity,
+            DbDataset.development_stage,
+            DbDataset.schema_version,
+            DbDataset.tombstone,
+        ]
+        table = cls.table
+
+        def to_dict(db_object):
```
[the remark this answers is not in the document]
ebezzi: Yeah I saw that but that's doing a lot more stuff. I think inlining the method here isn't a terrible idea.

tests/unit/backend/corpora/api_server/test_v1_dataset.py
```diff
@@ -112,6 +112,14 @@ def test__minimal_status__ok(self):
             response = self.app.get(test_url.url, headers=headers)
             self.assertEqual(json.loads(response.data)["upload_status"], status.name)
 
+    def test__get_all_datasets_for_index(self):
+        test_url = furl(path="/dp/v1/datasets/index")
+        headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
+        response = self.app.get(test_url.url, headers=headers)
+        self.assertEqual(200, response.status_code)
+        # Note: testing the payload is difficult because this will retrieve datasets generated by previous
+        #  tests.
```
atolopko-czi, line 121: could _add_ new dataset and then at least assert that it is included and and correctly represented.  could also compare the length of the json array to the actual db table row count.

backend/corpora/common/entities/dataset.py
```diff
@@ -124,6 +124,62 @@ def get_asset(self, asset_uuid) -> typing.Union[DatasetAsset, None]:
...
+    @staticmethod
+    def transform_organism_for_schema_2_0_0(dataset):
+        # If organism is an object (version 1.1.0), wrap it into an array to be 2.0.0 compliant
+        if "organism" in dataset and dataset.get("schema_version") != "2.0.0":
+            dataset["organism"] = [dataset["organism"]]
+
+    @classmethod
+    def list_for_index(cls, session) -> typing.List[typing.Dict]:
+        """
+        Return a list of all the datasets. For efficiency reasons, this only returns the fields
```
atolopko-czi, line 142:
```suggestion
        Return a list of all the datasets and associated metadata. For efficiency reasons, this only returns the fields
```

backend/Makefile
atolopko-czi: does `db/tunnel` with with this or do we need an `db/rdev-tunnel` command too?

backend/corpora/common/entities/dataset.py
```diff
@@ -124,6 +124,62 @@ def get_asset(self, asset_uuid) -> typing.Union[DatasetAsset, None]:
...
+            DbDataset.disease,
+            DbDataset.assay,
+            DbDataset.organism,
+            DbDataset.cell_count,
+            DbDataset.cell_type,
+            DbDataset.sex,
+            DbDataset.ethnicity,
+            DbDataset.development_stage,
+            DbDataset.schema_version,
+            DbDataset.tombstone,
```
atolopko-czi, line 160: tombstone field needed if we're always filtering out tombstoned (below)?  Should also document in method comment and API spec that tombstoned are excluded.

backend/corpora/common/entities/dataset.py
```diff
@@ -124,6 +124,62 @@ def get_asset(self, asset_uuid) -> typing.Union[DatasetAsset, None]:
...
+        ]
+        table = cls.table
+
+        def to_dict(db_object):
+            _result = {}
+            for _field in db_object._fields:
+                _value = getattr(db_object, _field)
+                if _value is None:
+                    continue
```
atolopko-czi, line 169: will this cause any difficulty for the consumer of the json? (not having keys for which values are null) probably not, but worth quick consideration

backend/config/corpora-api.yml
```diff
@@ -604,6 +604,54 @@ paths:
         "403":
           $ref: "#/components/responses/403"
 
+  /dp/v1/datasets/index:
+    get:
+      tags:
+        - datasets
+      summary: List datasets for indexing purposes
```
atolopko-czi, line 611:
```suggestion
      summary: List non-tombstoned datasets for indexing purposes
```

atolopko-czi: `for indexing purposes` isn't entirely clear to me; maybe `for client-side user filtering`?  (though DP API is used by some collaborators, I think it's okay to continue to think of the current API as a backend specifically for the UI; a public API will probably be different...more RESTful, etc.).

backend/config/corpora-api.yml
```diff
@@ -604,6 +604,54 @@ paths:
...
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    id:
+                      type: string
+                    name:
+                      type: string
+                    collection_id:
```
atolopko-czi, line 632: Will this be a filterable field in the UI? Since it's an ID, guessing not. And if not, is it needed?

atolopko-czi (changes requested):
Looks good.  Just requesting a better test assertion, and a few other minor comments for your consideration.

tests/unit/backend/corpora/api_server/test_v1_dataset.py
ebezzi: The latter would be a good idea but I don't think we have access to the raw database session here. Ideally we would make the test independent although that will require some research (moving it to a separate class doesn't help). I think the idea of adding a dataset and assert that the id is there, as well as the relevant fields, is a good idea.

backend/corpora/common/entities/dataset.py
ebezzi: I don't believe so, generally the frontend is lenient on missing fields. This is mostly required because Swagger doesn't play well with converting `None` values to null (in particular, a test will fail since `cell_count` isn't defined). I think it's safe to leave this here.

backend/config/corpora-api.yml
ebezzi: Yeah, I will rewrite this. I think the key point to convey is that this endpoint isn't meant for a general usage because it lacks some fields that aren't required by the index we're trying to create, nor by displaying the dataset in the compact view.

backend/config/corpora-api.yml
ebezzi: This is required because the frontend will have to join each dataset with its collection to retrieve (at minimum) the collection name.

backend/corpora/common/entities/dataset.py
ebezzi: Actually, I think it's probably better to move the tombstone filtering to a `where` condition, as it will improve the transfer time between the database and the server. Therefore this can be removed.

tests/unit/backend/corpora/api_server/test_v1_dataset.py
Bento007: You can reset the state of the DB using this https://github.com/chanzuckerberg/single-cell-data-portal/blob/18021ed87f7365b6061b3036f34ea4fab12a87e6/tests/unit/backend/fixtures/data_portal_test_case.py#L23

backend/Makefile
ebezzi: Not sure because I keep my tunnels in a separate folder. I'll give it a try. Also this requires `DB_PW` to be provided externally since the way the secret is encoded is different from the other environments. I will try to improve this a bit.

atolopko-czi: I'd say it's a "nice to have".  But if we don't have a team-wide mechanism for tunneling to rdev, we should probably just remove this for now.

tests/unit/backend/corpora/api_server/test_v1_dataset.py
```diff
@@ -112,6 +112,36 @@ def test__minimal_status__ok(self):
...
+        self.assertEqual(actual_dataset["collection_id"], dataset.collection_id)
+        self.assertEqual(actual_dataset["assay"], dataset.assay)
+        self.assertEqual(actual_dataset["tissue"], dataset.tissue)
+        self.assertEqual(actual_dataset["disease"], dataset.disease)
+        self.assertEqual(actual_dataset["sex"], dataset.sex)
+        self.assertEqual(actual_dataset["ethnicity"], dataset.ethnicity)
+        self.assertEqual(actual_dataset["organism"], dataset.organism)
+        self.assertEqual(actual_dataset["development_stage"], dataset.development_stage)
+        self.assertEqual(actual_dataset["cell_count"], dataset.cell_count)
+        self.assertEqual(actual_dataset["cell_type"], dataset.cell_type)
```
atolopko-czi, line 143: 👏

atolopko-czi (approved):
Approving, but still one comment re:client-server API call count.

backend/Makefile
ebezzi: I removed it for now as the tunnel doesn't work either. Will repackage this as a script that can be shared among devs.

tests/unit/backend/corpora/api_server/test_v1_dataset.py
ebezzi: I tried it but I couldn't get it to work. I refactored the test so that it will work regardless, but I'd like to come back to this at some point.

backend/config/corpora-api.yml
ebezzi: We discussed this with @MillenniumFalconMechanic and @NoopDog and we decided for this approach. I think the rationale is that a list of collections is needed regardless for the other view. Adding the collection_name here isn't free because it needs another query, so I'd leave it out for now. We can add it back later if required.

PR #1647, opened 2021-11-08
fix: skip Seurat conversion if dataset will cause it to crash

metakuni:
### Reviewers
**Functional:** 
@danieljhegeman 

**Readability:** 
@ebezzi 

---

## Changes
- modify: skipping Seurat conversion if the validator detects that the dataset will cause it to crash (due to R limitation). (note, the actual changes were implemented in #1658)

### DIFF
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
 # Install python dependencies
 # Remove packaging dependency once pyparser>3 is supported.
 RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
-RUN pip3 install cellxgene-schema==2.0.4
+RUN pip3 install cellxgene-schema==2.0.5
 
 ADD requirements.txt requirements.txt
 ADD backend/corpora/api_server/requirements.txt backend/corpora/api_server/requirements.txt
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
 # Install python dependencies
 # Remove packaging dependency once pyparser>3 is supported.
 RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
-RUN pip3 install cellxgene-schema==2.0.3
+RUN pip3 install cellxgene-schema==2.0.5
 
 ADD requirements.txt requirements.txt
 ADD backend/corpora/api_server/requirements.txt backend/corpora/api_server/requirements.txt
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
 # Install python dependencies
 # Remove packaging dependency once pyparser>3 is supported.
 RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
-RUN pip3 install cellxgene-schema==2.0.5
+RUN pip3 install cellxgene-schema==2.1.0
 
 ADD requirements.txt requirements.txt
 ADD backend/corpora/api_server/requirements.txt backend/corpora/api_server/requirements.txt
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
 # Install python dependencies
 # Remove packaging dependency once pyparser>3 is supported.
 RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
-RUN pip3 install cellxgene-schema==2.0.5
+RUN pip3 install cellxgene-schema==2.0.4
 
 ADD requirements.txt requirements.txt
 ADD backend/corpora/api_server/requirements.txt backend/corpora/api_server/requirements.txt
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -194,7 +194,7 @@ def create_artifact(
         raise e
 
 
-def create_artifacts(local_filename: str, dataset_id: str, artifact_bucket: str, can_convert_to_seurat: bool):
+def create_artifacts(local_filename: str, dataset_id: str, artifact_bucket: str, can_convert_to_seurat: bool = False):
     bucket_prefix = get_bucket_prefix(dataset_id)
     logger.info(f"Creating artifacts for dataset {dataset_id}...")
     # upload AnnData
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -451,11 +451,12 @@ def validate_h5ad_file_and_add_labels(dataset_id, local_filename):
     # for now, expecting a dictionary with 3 items: is_valid (bool), errors (list), can_convert_to_seurat (bool)
     result = validate.validate(local_filename, output_filename)
 
-    if not result["is_valid"]:
-        logger.error(f"Validation failed with {len(result['errors'])} errors!")
+    if not result.get("is_valid", False):
+        errors = result.get("errors", [])
+        logger.error(f"Validation failed with {len(errors)} errors!")
         status = dict(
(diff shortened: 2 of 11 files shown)


backend/corpora/dataset_processing/process.py
```diff
@@ -194,22 +194,27 @@ def create_artifact(
         raise e
 
 
-def create_artifacts(local_filename, dataset_id, artifact_bucket):
+def create_artifacts(local_filename: str, dataset_id: str, artifact_bucket: str, can_convert_to_seurat: bool):
     bucket_prefix = get_bucket_prefix(dataset_id)
-    logger.info("Creating Artifacts.")
+    logger.info(f"Creating artifacts for dataset {dataset_id}...")
```
ebezzi, line 199: This is a bit redundant since only one dataset will be processed at a time (per cloudwatch group), but I guess it doesn't hurt

backend/corpora/dataset_processing/process.py
```diff
@@ -441,13 +446,16 @@ def validate_h5ad_file_and_add_labels(dataset_id, local_filename):
...
     output_filename = "local.h5ad"
-    is_valid, errors = validate.validate(local_filename, output_filename)
 
-    if not is_valid:
-        logger.error(f"Validation failed with {len(errors)} errors!")
+    # TODO: remove this comment after the interface to validate.validate() is decided
+    # for now, expecting a dictionary with 3 items: is_valid (bool), errors (list), can_convert_to_seurat (bool)
+    result = validate.validate(local_filename, output_filename)
+
+    if not result["is_valid"]:
```
ebezzi, line 454: Replace it with `result.get("is_valid")` to make it more resilient.

backend/corpora/dataset_processing/process.py
metakuni: Ah right, forgot about that.

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -590,3 +591,59 @@ def test__cxg_not_created_when_metadata_extraction_fails(
...
+        mock_download_from_dropbox_url,
+        mock_validate_h5ad_file_and_add_labels,
+        mock_extract_metadata,
+        mock_process_cxg,
+        mock_create_artifact,
+        mock_make_seurat,
+    ):
+        # given
+        mock_validate_h5ad_file_and_add_labels.return_value = dict(
+            labeled_filename=mock.ANY, can_convert_to_seurat=False
```
metakuni, line 614: If dataset isn't convertible to Seurat...

backend/corpora/dataset_processing/process.py
```diff
@@ -459,7 +465,7 @@ def validate_h5ad_file_and_add_labels(dataset_id, local_filename):
             validation_status=ValidationStatus.VALID,
         )
         update_db(dataset_id, processing_status=status)
-        return output_filename
+        return dict(labeled_filename=output_filename, can_convert_to_seurat=can_convert_to_seurat)
```
danieljhegeman, line 468: Personally, when the use case is fairly simply, I strongly prefer we use the implicit tuple return of `one, two, three` rather than a dictionary that we then have to use keys on later. Kind of like how `validate.validate` just returns the tuple on ln 450

danieljhegeman: https://github.com/chanzuckerberg/single-cell-data-portal/pull/1654 @metakuni

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -590,3 +591,55 @@ def test__cxg_not_created_when_metadata_extraction_fails(
         dataset = Dataset.get(self.session, dataset_id)
         processing_status = dataset.processing_status
         self.assertEqual(None, processing_status.cxg_status)
+
+    @patch("backend.corpora.dataset_processing.process.make_seurat")
```
metakuni, line 595: Why so many mocks? 😄 

None of these functions matter to the outcome of the behavior being tested in this unit test. 

e.g. whether `process_cxg` runs or not is irrelevant to Seurat conversion. Likewise, for these 2 unit tests, we don't care about writing to the database (`update_db`). In fact, writing to a real database is undesirable in a unit test since it makes them run much slower.

The fact that so many mocks are required to test something this simple is probably an indication that the code needs to be refactored. But that will be left as an exercise for another day.

backend/corpora/dataset_processing/process.py
metakuni: Now I remember why I added this... I thought it might be handy for tracing via Log Insights queries across log streams.

e.g. something like:
```
fields @timestamp, @message
| filter @message like /084311bf-678d-43a2-9890-d6e3c6bae13a/
| sort @timestamp desc
| limit 20
```

to show everything that happened to dataset `084311bf-678d-43a2-9890-d6e3c6bae13a`. Of course, this will require adding the dataset ID to other log statements to be useful.

tests/unit/backend/corpora/dataset_processing/test_process.py
```diff
@@ -573,11 +574,17 @@ def url(self):
...
+        mock_extract_metadata,
+        mock_validate_h5ad_file_and_add_labels,
+        mock_download_from_dropbox_url,
+        mock_make_cxg,
     ):
-        mock_metadata_extraction.side_effect = RuntimeError("metadata extraction failed")
-        mock_dropbox_download.return_value = self.h5ad_filename
-        mock_cxg.return_value = str(self.cxg_filename)
+        # given
+        mock_validate_h5ad_file_and_add_labels.return_value = (mock.ANY, False)
```
metakuni, line 584: Fixed the return signature

Dockerfile.processing_image
```diff
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
 # Install python dependencies
 # Remove packaging dependency once pyparser>3 is supported.
 RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
-RUN pip3 install cellxgene-schema==2.0.3
+RUN pip3 install cellxgene-schema==2.0.5
```
metakuni, line 24: Note: This will fail until this PR is merged and released to PyPI:
* https://github.com/chanzuckerberg/single-cell-curation/pull/172

backend/corpora/dataset_processing/process.py
```diff
@@ -436,12 +441,19 @@ def process_cxg(local_filename, dataset_id, cellxgene_bucket):
     update_db(dataset_id, metadata)
 
 
-def validate_h5ad_file_and_add_labels(dataset_id, local_filename):
+def validate_h5ad_file_and_add_labels(dataset_id: str, local_filename: str) -> typing.Tuple[str, bool]:
+    """
+    Validates and labels the specified dataset file and updates the processing status in the database
+    :param dataset_id: UUID of the dataset to update
+    :param local_filename: file name of the dataset to validate and label
+    :return: (str, bool) - file name of labeled dataset, boolean indicating if Seurat conversion is possible
```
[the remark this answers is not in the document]
metakuni: I don't disagree, but then how/where do you document the meaning of the returned tuple elements? (e.g. what the bool means, etc.)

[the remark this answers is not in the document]
blrnw3: comments yes, just not types :). Although I see your point that this would get out of sync too. 🤷

metakuni:
Note: The smoke-test and unit-test are currently failing since `cellxgene-schema` 2.0.5 hasn't been released to PyPI yet. This will happen early today.

tests/unit/backend/corpora/dataset_processing/test_process.py
metakuni: Much simpler 😄 

Fixed

Dockerfile.processing_image
```diff
@@ -21,7 +21,7 @@ RUN git clone https://github.com/chanzuckerberg/single-cell-explorer.git \
 # Install python dependencies
 # Remove packaging dependency once pyparser>3 is supported.
 RUN pip3 install scanpy==1.6.0 python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
-RUN python3 -m pip install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ cellxgene-schema==2.1.0rc0
+RUN pip3 install cellxgene-schema==2.1.0
```
metakuni, line 24: Replacing the Test PyPI version of `cellxgene-schema` with the one released on PyPI.

metakuni: `cellxgene-schema` 2.1.0 was released to PyPI this morning:
* https://pypi.org/project/cellxgene-schema/2.1.0/

danieljhegeman (approved):

Issue #1283, opened 2021-07-20
feat(withdraw): Add withdrawn collection toasts 


seve:
## Need
When a collection is tombstoned and user visits a direct link to it they should be directed to the collections page with a toast indicating the resource has been removed. We also need to edit the copy on the publish collection dialog that states that collections cannot be deleted

## Approach
The toasts should reflect this [figma](https://www.figma.com/file/jZzZF9oO0YRMxmVRDxDHp3/Public-Revisions?node-id=803%3A0). 

For a collection, the withdrawn collection toast should appear when a user attempts to visit the collection page of a tombstone collection. GET collection/{collection_id} will return the information need to populate this page. 

**Additional context**
Originally a part of #1237

## Definition of Done
	- Navigating to a tombstoned collection results in a withdrawn collection toast.

maesen:
Note: passed in refinement 8/24 because we don't have the ability to delete a collection yet.

PR #1651, opened 2021-11-09
feat: delete public collection button and collection withdrawn toast

seve:
### Reviewers
**Functional:** @tihuan 

**Readability:** @Bento007 

---

## Changes
- add
   - HttpStatusCode enum, ty @scokmen
   -  Handle tombstoned collections elegantly when possible to receive
   - Delete collection button for public collections
   - Redirect on tombstoned collection
   - Withdrawn collection toast
- modify
  - Alter collection query to allow for a informative 410 error with new `TombstonedCollection` return type
  - Add collection visibility parameter to `deleteCollection` mutation

## QA steps (optional)
   - https://delete-collection-frontend.rdev.single-cell.czi.technology/

## Known Issues
 - rdev incorrectly redirects off of removed explorer assets, needs manual change of url from `delete-collection-explorer` to `delete-collection-frontend`

#1283


frontend/src/common/constants/httpStatusCode.ts
```diff
@@ -0,0 +1,381 @@
+// https://gist.github.com/scokmen/f813c904ef79022e84ab2409574d1b45
```
tihuan, line 1: Since enum is constant, let's use UPPER_CASE for both name and file!

e.g.,

`frontend/src/common/constants/HTTP_STATUS_CODE.ts`

```ts
enum HTTP_STATUS_CODE { ... }
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#basic_const_usage

frontend/src/common/entities.ts
```diff
@@ -76,6 +76,7 @@ export interface Collection {
   updated_at: number;
   has_revision: boolean;
   revision_diff: boolean;
+  tombstone?: boolean;
```
tihuan, line 79: readability: `isTomstoned`

frontend/src/views/Collection/index.tsx
```diff
@@ -143,9 +150,13 @@ const Collection: FC = () => {
   const handleOnChange = function (newTabId: TABS) {
     setSelectedTab(newTabId);
   };
+  const hasWriteAccess = collection.access_type === ACCESS_TYPE.WRITE;
+  const shouldShowPrivateWriteAction = hasWriteAccess && isPrivate;
 
-  const shouldShowPrivateWriteAction =
-    collection.access_type === ACCESS_TYPE.WRITE && isPrivate;
+  const handleDeleteCollection = () => {
+    deleteMutation({ collectionID: id, visibility: VISIBILITY_TYPE.PUBLIC });
```
tihuan, line 157: maybe await here before redirect, so the user doesn't see the deleted collection in My Collections while delete is in flight?

```ts
await deleteMutation(...);
```

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
```diff
@@ -82,11 +83,13 @@ const Content: FC<Props> = (props) => {
 
   const formEl = useRef<HTMLFormElement>(null);
 
-  const { data } = useCollection({
+  let { data } = useCollection({
     id: props.id,
     visibility: VISIBILITY_TYPE.PRIVATE,
   });
 
+  if (isTombstonedCollection(data)) data = {} as Collection;
```
tihuan, line 91: instead of assigning a new POJO every time, maybe `null` would work here? Otherwise `{} !== {}` would cause unnecessary re-rendering

Also reassigning variable to a different value impacts readability and is a slippery slope if future authors continue to follow the pattern to reassign the value through out the file

Maybe we can change it to:

```ts
const { data } = useCollection({
    id: props.id,
    visibility: VISIBILITY_TYPE.PRIVATE,
  });

const collection = isTombstonedCollection(data) ? null : data;

// and then use `collection` for the rest of the file?
```

tihuan (approved):
LGTM! Thanks so much for adding this, @seve 🤩 👏 

Just some minor nits otherwise LGTM!!

frontend/src/common/entities.ts
Bento007: or is `isTombstone`. This avoid using our made up word "tombstoned"

seve: this is received from backend

tihuan: ahh I see, nvm!

frontend/src/views/Collection/index.tsx
seve: Nice catch. Even better, let's optimistically remove the collection!

713ba0e

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
seve: Sounds good, should I change line 96 too?

https://github.com/chanzuckerberg/single-cell-data-portal/blob/723b97039f34c503f97e0f7bacb08781d1f4968b/frontend/src/components/CreateCollectionModal/components/Content/index.tsx#L96

frontend/src/common/queries/collections.ts
```diff
@@ -248,20 +263,46 @@ async function deleteCollection(collectionID: Collection["id"]) {
...
+    onMutate: async () => {
+      await queryCache.cancelQueries([USE_COLLECTIONS]);
+
+      const previousCollections = queryCache.getQueryData([
+        USE_COLLECTIONS,
+      ]) as CollectionResponsesMap;
+
+      const newCollections = new Map(previousCollections);
+      // If we're deleting a public collection, nuke it from the cache
+      if (visibility === VISIBILITY_TYPE.PUBLIC) {
```
seve, line 289: Actually there's a second case where we nuke it, when there is only a private collection. Fixing

frontend/src/common/constants/httpStatusCode.ts
```diff
@@ -0,0 +1,381 @@
...
+
+  /**
+   * The client needs to authenticate to gain network access.
+   * Intended for use by intercepting proxies used to control access to the network (e.g., "captive portals" used
+   * to require agreement to Terms of Service before granting full Internet access via a Wi-Fi hotspot).
+   */
+  NETWORK_AUTHENTICATION_REQUIRED = 511,
+}
+
+export default HttpStatusCode;
```
tihuan, line 381:
```suggestion
export default HTTP_STATUS_CODE;
```

frontend/src/common/queries/collections.ts
```diff
@@ -3,7 +3,9 @@ import { Collection, VISIBILITY_TYPE } from "src/common/entities";
 import { apiTemplateToUrl } from "src/common/utils/apiTemplateToUrl";
 import { API_URL } from "src/configs/configs";
 import { API } from "../API";
+import HttpStatusCode from "../constants/httpStatusCode";
```
tihuan, line 6: let's use UPPER_CASE for constants at callsites too!

frontend/src/common/queries/collections.ts
```diff
@@ -248,20 +263,49 @@ async function deleteCollection(collectionID: Collection["id"]) {
...
+    onMutate: async () => {
+      await queryCache.cancelQueries([USE_COLLECTIONS]);
+
+      const previousCollections = queryCache.getQueryData([
+        USE_COLLECTIONS,
+      ]) as CollectionResponsesMap;
+
+      const newCollections = new Map(previousCollections);
+      const collectionsWithID = newCollections.get(id);
+      // If we're deleting a public collection or there is no revision, nuke it from the cache
```
tihuan, line 289: nice comments!

frontend/src/views/Collection/index.tsx
```diff
@@ -65,7 +73,14 @@ const Collection: FC = () => {
...
 
-  const { data: collection, isError, isFetching } = collectionState;
+  const { data, isError, isFetching } = collectionState;
+
+  const [deleteMutation] = useDeleteCollection(id, visibility);
+
+  if (data?.tombstone === true) {
+    router.push(ROUTES.HOMEPAGE + "?tombstoned_dataset_id=" + id);
+  }
+  const collection = data as CollectionType;
```
tihuan, line 83: just curious -- is there a reason to do this than keeping `const { data: collection, isError, isFetching } = collectionState;` ?

frontend/src/views/Collection/index.tsx
tihuan: Ahh great idea 💡 !

I think we still need to await `deleteMutation()` at line 157 😆

tihuan (approved):
Thanks for taking care of the feedback, @seve ! LGTM after last round of small nits. Thank you!! 🚀🚀🚀

frontend/src/views/Collection/index.tsx
seve: I thought it better expressed that once we get past the tombstone check we actually know its a collection

tihuan: ahh I see! I was thinking tombstoned collection is still a collection, and we're already extracting `data` from `collectionState`, so the association is already there, so we could probably save an assignment in a new line!

frontend/src/views/Collection/index.tsx
```diff
@@ -65,7 +73,14 @@ const Collection: FC = () => {
     visibility,
   });
 
-  const { data: collection, isError, isFetching } = collectionState;
+  const { data, isError, isFetching } = collectionState;
+
+  const [deleteMutation] = useDeleteCollection(id, visibility);
+
+  if (data?.tombstone === true) {
+    router.push(ROUTES.HOMEPAGE + "?tombstoned_dataset_id=" + id);
```
tihuan, line 81: Ah since this is a side effect, we should do it in `useEffect()`, so we don't `router.push()` at every re-render

seve: Done!

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
tihuan: Great catch! Yes please!

frontend/src/common/constants/httpStatusCode.ts
tihuan: ah last thing is updating the file name! `frontend/src/common/constants/HTTP_STATUS_CODE.ts` thank you!

frontend/src/components/Collections/components/DeleteCollection/index.tsx
```diff
@@ -28,7 +28,7 @@ const DeleteCollection: FC<Props> = ({
 
   const handleDelete = async () => {
     setIsLoading(true);
-    await deleteMutation(id);
+    await deleteMutation({ collectionID: id });
     router.push(ROUTES.MY_COLLECTIONS);
```
tihuan, line 32: not sure if you care to move redirect into `{ onSuccess() { router.push(ROUTES.MY_COLLECTIONS); } }` here too

frontend/src/components/CreateCollectionModal/components/Content/index.tsx
seve: I'm actually going to stick with the original here.
https://github.com/chanzuckerberg/single-cell-data-portal/blob/723b97039f34c503f97e0f7bacb08781d1f4968b/frontend/src/components/CreateCollectionModal/components/Content/index.tsx#L96

We shouldn't really be getting to the case that the create collection modal is open on a tombstoned collection.  Too much of an edge case to justify a tsignore/ the refactor necessary.

image

tihuan: Hmm yeah that sounds great 👍 Thanks for investigating this approach!

Bento007 (approved):

tihuan (approved):
PR looks super awesome! Thanks so much, @seve 🚀 🎆 👏

PR #1652, opened 2021-11-09
feat: block publish of duplicates

Bento007:
### Reviewers
**Functional:** @ebezzi 

**Readability:** @seve 

---

## Changes
- Add a check for duplicates among datasets by comparing S3 Etags associated with artifacts before publishing a collection.
- Add 409 response during publish when duplicates are detected.
- A collection cannot be publish if it container duplicated datasets 

## Known Issues
A call to S3 must be made for every artifact in a dataset. If a collection has a lot of artifacts then this could potentially be slow. To speed this process up, we don't check for all duplicates. Once we find a duplication we return.


seve (approved):
Readability 🏁

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -51,7 +51,15 @@ def get_file_size(self) -> typing.Union[int, None]:
...
-            return response["ContentLength"]
+            return response
+
+    def get_file_size(self) -> typing.Union[int, None]:
+        """
+        Retrieves the asset content length from the S3 object.
+        :return: The content length in bytes.
+        """
+        metadata = self.get_s3_metadata()
+        return metadata["ContentLength"] if metadata else None
```
ebezzi, line 62:
```suggestion
        return metadata.get("ContentLength")
```

backend/corpora/lambdas/api/v1/collection_uuid/publish.py
```diff
@@ -1,14 +1,38 @@
...
+    :param collection:
+    :return: True if duplicates detected, otherwise return false.
+    """
+    etags = []
+    for dataset in collection.datasets:
+        if not dataset.tombstone:
+            for artifact in dataset.artifacts:
+                _artifact = DatasetAsset(artifact)
+                metadata = _artifact.get_s3_metadata()
+                if not _artifact.get_s3_metadata():
```
ebezzi, line 24:
```suggestion
                if not metadata:
```

backend/corpora/common/entities/dataset_asset.py
Bento007: There is a chance that `metadata == None` in which case `metadata.get("ContentLength")` would throw an error

ebezzi: You're absolutely right, discard this.

danieljhegeman: I though the point of using `.get` was to avoid throwing a missing key error...? 🤔

danieljhegeman:
> A call to S3 must be made for every artifact in a dataset. If a collection has a lot of artifacts then this could potentially be slow. To speed this process up, we don't check for all duplicates. Once we find a duplication we return.

I'm confused here. Is the speed-up simply the improvement of average runtime and not worst-case? Won't most collections not have a duplicate, and therefore iterate over all?

backend/corpora/lambdas/api/v1/collection_uuid/publish.py
```diff
@@ -1,14 +1,38 @@
...
 from backend.corpora.lambdas.api.v1.collection import _owner_or_allowed
 
 
+def check_for_duplicate_datasets(collection: Collection) -> bool:
+    """
+    Check if duplicate datasets are found. Return true on the first duplicate match.
+    :param collection:
+    :return: True if duplicates detected, otherwise return false.
+    """
+    etags = []
```
danieljhegeman, line 18: intentional micro-optimization here by using list instead of set? For small data structure (since we're talking < 100 items)?

tests/unit/backend/corpora/api_server/collection_uuid/test_publish.py
```diff
@@ -155,34 +155,49 @@ def test__not_owner__403(self):
...
+        path = f"/dp/v1/collections/{collection.id}/publish"
+        response = self.app.post(path, headers=self.headers_authed, data=json.dumps(self.publish_body))
+        self.assertEqual(409, response.status_code)
+
 
 class TestPublishCurators(BasicAuthAPITestCurator):
+    def setUp(self):
+        super().setUp()
+        self.publish_body = {"data_submission_policy_version": "1.0"}
+
```
danieljhegeman, line 188: unrelated?

danieljhegeman (approved):
minor comments, but no blockers ✅

Bento007:
> > A call to S3 must be made for every artifact in a dataset. If a collection has a lot of artifacts then this could potentially be slow. To speed this process up, we don't check for all duplicates. Once we find a duplication we return.
> 
> I'm confused here. Is the speed-up simply the improvement of average runtime and not worst-case? Won't most collections not have a duplicate, and therefore iterate over all?

Yes it is improves the case where there are duplicate. There isn't a lot we can do for the worst-case unless we can [batch our requests to s3 somehow](https://aws.amazon.com/s3/features/batch-operations/).

PR #1654, opened 2021-11-10
Return tuple instead of dict

danieljhegeman:
I think this is a little cleaner / more "pythonic"

### Reviewers
**Functional:** 
@metakuni 
**Readability:** 

---

## Changes
- add
- remove
- modify


seve:
# PYTHONIC

Issue #1657, opened 2021-11-15
Issue with Seurat conversion


ebezzi:
On recently built rdevs, Seurat conversion fails with this error:
```
Loading required package: reticulate
X -> data; raw.X -> counts
Error in vapply(X = keyed.objects, FUN = function(x) { :
  values must be length 1,
 but FUN(X[[1]]) result is length 0
Calls: <Anonymous> ... func -> [[<- -> [[<- -> Key -> Key.Seurat -> vapply
Execution halted
```

metakuni:
Deployed to prod yesterday.

PR #1659, opened 2021-11-15
fix: pin SeuratObject to version 4.0.2

ebezzi:
### Reviewers
**Functional:** 
@pablo-gar @maniarathi 

Ticket [#1657](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1657)

---

## Changes
- Pins `SeuratObject` to version 4.0.2. The latest version (4.0.3) is causing the issue documented in the ticket.

### DIFF
--- a/backend/corpora/dataset_processing/make_seurat.R
+++ b/backend/corpora/dataset_processing/make_seurat.R
@@ -1,5 +1,8 @@
 library(sceasy)
 
+require(devtools)
+install_version("SeuratObject", version = "4.0.2", repos = "http://cran.rstudio.com/")
+
 h5adPath <- commandArgs(trailingOnly = TRUE)[1]
 
 sceasy::convertFormat(h5adPath, from="anndata", to="seurat", outFile = gsub(".h5ad", ".rds", h5adPath), main_layer = "data")


metakuni (approved):

backend/corpora/dataset_processing/make_seurat.R
```diff
@@ -1,5 +1,8 @@
 library(sceasy)
 
+require(devtools)
+install_version("SeuratObject", version = "4.0.2", repos = "http://cran.rstudio.com/")
```
pablo-gar, line 4: Does it actually force re-installation? 

I think a better place for this may be  when the docker image is created.

https://github.com/chanzuckerberg/single-cell-data-portal/blob/c12018441a6037bc3b4c50d1381fd84e4065b0a1/Dockerfile.processing_base#L27

ebezzi: Yes, this forces reinstallation. I don't think we should put that in `processing_base`, since this is meant to be a temporary solution. Also, `processing_base` is pushed on a schedule and this fix needs to go in ASAP. Happy to move this around later.

pablo-gar: okay! SGTM then

pablo-gar (approved):

PR #1669, opened 2021-11-16
ci: three branches deployment draft

ebezzi:
### Reviewers
**Functional:** 
@atolopko-czi @maniarathi @Bento007 

**Readability:** 
@metakuni 

---

## Changes
Enables three branch based deployment. The flow will be as follows:
- release to dev: automatic (like now)
- release to staging: merge `main` into `staging` (do not squash)
- release to prod: merge `staging` into `prod` (do not squash)

At a lower level, this is how it works:
- Every time you merge into one of the 3 branches, `push-tests.yml` will:
1. Run tests
2. Build docker images
3. Push docker images (note: aside from the `upload-failures` images, we're currently using our dev ECR account even in prod, so the pushes just go there)
4. Trigger a GHA deployment

Once this happens, the deployment flow (`deploy-happy-stack.yml`) will start. This is unchanged as it can already deploy to all 3 environments. The variable that determines the environment is `github.event.deployment.environment`, which was passed by `happy-deploy.py` previously, and is now set at step 4 of the previous workflow.

## Testing plan
In lieu of the normal staging deployment, we should:
1. Clean up the `staging` and `prod` branches (they currently exist but are unused. I'd recommend to delete them and reset with the current main)
2. Merge this PR directly in the `staging` branch 
3. This should trigger a staging deployment. If there are errors, fix them in the PR and keep merging into staging until a successful deployment happens
4. Clean up the `staging` branch again to get rid of the spurious commits
5. Merge this into `main` and test the dev deployment

Once a deployment to staging is confirmed to work, prod should also be fine.


atolopko-czi:
Do we still intend to decide as a team whether we're going with git flow? Branch naming conventions would be different, for one.  Overall, with this PR's approach, how will hot fixes be handled?

ebezzi:
@atolopko-czi this is not a "fully fledged" gitflow but more of a standard "one branch per environment" approach. The development flow isn't really affected since developers will continue opening PRs against the `main` branch. (incidentally, the explorer is using the same approach although shadowed by some `happy` trickery).

Hotfixes will be done by creating a PR against the relative branch (most likely `prod`). The PR will then be merged and this will automatically trigger a deployment. After the hotfix is applied, we should merge `prod` back to `main` and that will align the two branches.

The only caveat about this approach is that we need to enable normal merge, since PRs opened from `main` to `staging` and from `staging` to `prod` should have a normal merge. Note that this isn't a strict requirement, but will help with keeping the branches clean. This might create some confusion among developers since they need to remember when to squash and when to do a normal merge. As far as I know GitHub doesn't allow different behaviors for the merge button by branch, which is unfortunate.

atolopko-czi:
> 3. Push docker images (note: aside from the `upload-failures` images, we're currently using our dev ECR account even in prod, so the pushes just go there)

So prod deploys pulls the images from our AWS ECR dev account somehow? Should we change this now to build all prod images to AWS ECR prod account, for sanity? Curious that we push `upload-failures` image to prod--this was just an erroneous inconsistency?

> 2. Merge this PR directly in the `staging` branch 

Instead of `dev`, because we want to test the manual invocation of the deployment?

> The only caveat about this approach is that we need to enable normal merge, since PRs opened from main to staging and from staging to prod should have a normal merge. Note that this isn't a strict requirement, but will help with keeping the branches clean. This might create some confusion among developers since they need to remember when to squash and when to do a normal merge.

Can be addressed with a script (for now, a make target), right?

---

We'll still have to deal with the issue of smoke tests failing intermittently on staging and prod deploys, right?  I think the team agreed that we should just make the tests more reliable. And I think it makes sense to run all tests on staging and prod branches, since we're allowing hot fixes to these branches.

seve:
Should we rename push-tests.yml while we're at it? I feel like it's a bit overloaded now.

seve:
> The only caveat about this approach is that we need to enable normal merge, since PRs opened from `main` to `staging` and from `staging` to `prod` should have a normal merge. Note that this isn't a strict requirement, but will help with keeping the branches clean. This might create some confusion among developers since they need to remember when to squash and when to do a normal merge. As far as I know GitHub doesn't allow different behaviors for the merge button by branch, which is unfortunate.

I think we'll be able to get around the confusion by requiring linear history on main which would require squashing

ebezzi:
> Should we rename push-tests.yml while we're at it? I feel like it's a bit overloaded now.

Yes I totally agree, any idea for the new name?

ebezzi:
> > The only caveat about this approach is that we need to enable normal merge, since PRs opened from `main` to `staging` and from `staging` to `prod` should have a normal merge. Note that this isn't a strict requirement, but will help with keeping the branches clean. This might create some confusion among developers since they need to remember when to squash and when to do a normal merge. As far as I know GitHub doesn't allow different behaviors for the merge button by branch, which is unfortunate.
> 
> I think we'll be able to get around the confusion by requiring linear history on main which would require squashing

That's a good point, I will try to enable it and see how it works. Thanks!

danieljhegeman (approved):
👍

PR #1671, opened 2021-11-17
fix: address feedback and fix double toast

seve:


frontend/src/views/Collection/index.tsx
```diff
@@ -78,21 +80,26 @@ const Collection: FC = () => {
   const [selectedTab, setSelectedTab] = useState(TABS.DATASETS);
 
   useEffect(() => {
-    if (!tombstoned_dataset_id) return;
+    if (
+      !tombstoned_dataset_id ||
+      !collection ||
```
tihuan, line 85: ah if we're still fetching `collection`, collection will be `undefined`, so we'll still the toast incorretly?

tihuan (approved):
LGTM! Thanks for the bug fix, Seve!

PR #1676, opened 2021-11-22
fix: Track skipped Seurat, do not wait on FE

danieljhegeman:
Mark the dataset_processing_status.rds_status as SKIPPED whenever
conversion to seurat cannot be performed and is skipped. On the front
end, accept 'SKIPPED' conversion status and proceed.

### Reviewers
**Functional:** 
@metakuni (BE)
@seve (FE)
**Readability:** 
@ebezzi 
---

## QA steps (optional)

~~Still need to test in rdev.~~ SKIPPED status is passed successfully to front end; making separate issue for continuously-spinning Processing element. Ready for merge.

## Known Issues

dockerized dataset processing unit tests `make local-unit-test-processing` not working (for me at least)


danieljhegeman:
Still need to add in the db migration to add `SKIPPED` to `conversionstatus`, then will be ready to review.

Makefile
```diff
@@ -20,7 +20,7 @@ lint:
 
 .PHONY: unit-test
 unit-test: local-unit-test
-	# Keeping old target name for reverse comatibility
+	# Keeping old target name for reverse compatibility
```
danieljhegeman, line 23: unrelated

Makefile
```diff
@@ -94,7 +94,7 @@ local-ecr-login:
 .PHONY: local-hostconfig
 local-hostconfig:
 	if [ "$$(uname -s)" == "Darwin" ]; then \
-	  sudo ./scripts/happy hosts install; \
+	  sudo -E ./scripts/happy hosts install; \
```
danieljhegeman, line 97: related-ish - preserves env variables, ensures venv python works for me (although this was not necessary for others -- would be nice if it were confirmed that this recipe still works for you)

backend/Makefile
```diff
@@ -92,7 +92,7 @@ SSH_SERVER_ALIVE_COUNT_MAX?=60
 # - consider adding -M and -S options on ssh command when needed
 # - add db/tunnel as a dependency for all targets so that a tunnel is automatically opened if not already
 db/tunnel:
-	$(eval endpoint=$(shell aws rds describe-db-cluster-endpoints --db-cluster-identifier corpora-${DEPLOYMENT_STAGE}-corpora-api | jq '.DBClusterEndpoints[] | select(.EndpointType | contains("WRITER")) | .Endpoint'))
+	$(eval endpoint=$(shell aws rds describe-db-cluster-endpoints --db-cluster-identifier corpora-${DEPLOYMENT_STAGE}-corpora-api | jq -r '.DBClusterEndpoints[] | select(.EndpointType | contains("WRITER")) | .Endpoint'))
```
danieljhegeman, line 95: should have the `-r` to remove the `"` marks flanking the output

scripts/happy
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
danieljhegeman, line 1: again this... should be fine... if people are using venvs and/or global python defaults appropriately -- we don't have anything running on Python 2, right?

ebezzi: I would roll back this change. It will break for people who have their default `python` command set to python 2.7.

backend/Makefile
```diff
@@ -48,14 +48,14 @@ db/new_migration_auto:
 
 db/connect:
 	# Assuming you've created a tunnel to the DB. Check the docs for information on how to do that
-	$(eval DB_PW = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/${DEPLOYMENT_STAGE}/database --region us-west-2 | jq -r '.SecretString | split(":") | .[-1] | split("@") | .[0]'))
+	$(eval DB_PW = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/${DEPLOYMENT_STAGE}/database --region us-west-2 | jq -r '.SecretString | match(":([^:]*)@").captures[0].string'))
```
ebezzi, line 51: What is this change about?

ebezzi (approved):

Makefile
metakuni: `make local-hostconfig` and `make local-nohostconfig` appear to work for me. I'm using `pyenv virtualenv`.

backend/Makefile
metakuni: 🤔 

So, Dan is becoming our Makefile expert.  🤠

scripts/happy
metakuni: I'm guessing that `happy` wouldn't even run on Python 2.

metakuni (commented):
Looks good so far. Will approve once it gets promoted from draft.

backend/corpora/dataset_processing/process.py
```diff
@@ -225,7 +226,7 @@ def cancel_dataset(dataset_id):
         logger.info("Upload Canceled.")
 
 
-def update_db(dataset_id, metadata=None, processing_status=None):
+def update_db(dataset_id, metadata: dict = None, processing_status: dict = None):
```
metakuni, line 229:
```suggestion
def update_db(dataset_id: str = None, metadata: dict = None, processing_status: dict = None):
```

frontend/src/components/Collections/components/Grid/components/Row/DatasetRow/utils.ts
```diff
@@ -143,19 +143,23 @@ export function useCheckCollectionFormatsPopulated({
...
+      numOfProcessedFormats += 1;
+    }
   }
 
   const numOfAvailableFormats = dataset.dataset_assets.length;
 
   useCheckCollection({
     invalidateCollectionQuery,
-    shouldFetch: numOfConvertedFormats > numOfAvailableFormats,
+    shouldFetch: numOfProcessedFormats > numOfAvailableFormats,
```
seve, line 162: I haven't tried running this, but since we're introducing the possibility that the number of process formats(converted or skipped) will be more than available.  There will be cases where `shouldFetch` will be true when it shouldn't be?

Makefile
danieljhegeman: yep, understood. This helps it work for more people such as myself; my setup requires it for some reason. But conversed with @ebezzi , going to set these changes aside and have others review.

backend/Makefile
danieljhegeman: This parsing was just wrong before. It's possible the db connection string changed formats. But the output was not capturing the password; try the two diff versions on your local machine

backend/corpora/dataset_processing/process.py
danieljhegeman: I don't think we want to have a default value of None, but the type hint would be fine

scripts/happy
danieljhegeman: given that this uses happy requirements, this should only ever be executed when in an active venv, no?

ebezzi (approved):

metakuni (approved):

Issue #1632, opened 2021-11-01
Add additional data to collections endpoint


MillenniumFalconMechanic:
#### Need
Currently there is an initial request to retrieve core collection information and then a subsequent request for each individual collection. This results in the server being hit with many requests and adds load to time to the user experience.

#### Approach
~~Draft, to be confirmed with @ebezzi:~~
~~* Add additional information required to display and sort collections view to existing `/collections` API endpoint (possibly with API version bump).~~
* Add new collections endpoint that returns the following for all public collections:
  *  id
  * name
  * published_at
  * revised_at

MillenniumFalconMechanic:
[Collections Figma](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=1670%3A25035).

PR #1691, opened 2021-11-30
feat: collection index endpoint

ebezzi:
### Reviewers
**Functional:** 
@Bento007 @MillenniumFalconMechanic @NoopDog 

**Readability:** 

Ticket [#1632](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1632)

---

## Changes
- Add a `/collections/index` endpoint. `GET`, no parameters
- Returns an array with all the public, non tombstoned collections and a subset of the fields.


MillenniumFalconMechanic (approved):

tests/unit/backend/corpora/api_server/test_v1_collection.py
```diff
@@ -603,6 +603,41 @@ def test__list_collection__check_owner(self):
             self.assertIn(private_owned, ids)
             self.assertNotIn(private_not_owned, ids)
 
+    def test__get_all_collections_for_index(self):
+        # dataset = self.generate_dataset(self.session, id="test_dataset_id_for_index", cell_count=42)
```
Bento007, line 607: remove commented out code

tests/unit/backend/corpora/api_server/test_v1_collection.py
```diff
@@ -603,6 +603,41 @@ def test__list_collection__check_owner(self):
...
+        collection_id_private = self.generate_collection(
+            self.session, visibility=CollectionVisibility.PRIVATE.name, owner="test_user_id"
+        ).id
+
+        collection_id_tombstone = self.generate_collection(
+            self.session, visibility=CollectionVisibility.PUBLIC.name, tombstone=True, owner="test_user_id"
+        ).id
+
+        test_url = furl(path="/dp/v1/collections/index")
+        headers = {"host": "localhost", "Content-Type": "application/json", "Cookie": get_auth_token(self.app)}
```
Bento007, line 624: Is an auth token need for this endpoint?

Bento007 (changes requested):

tests/unit/backend/corpora/api_server/test_v1_collection.py
ebezzi: Nope, good call. Thanks!

backend/config/corpora-api.yml
```diff
@@ -413,6 +413,39 @@ paths:
         "404":
           $ref: "#/components/responses/404"
 
+  /dp/v1/collections/index:
```
Bento007, line 416: I think `/dp/v1/collections:` could be modified to support this need. Why is a new endpoint needed?

backend/config/corpora-api.yml
```diff
@@ -413,6 +413,39 @@ paths:
...
+    get:
+      tags:
+        - collections
+      summary: >-
+        Returns all public, non tombstoned collections. This endpoint is meant to be used for displaying collections
+        in a compact view. It doesn't return all the fields, so it is not suitable for general usage.
+      security:
```
Bento007, line 423: This can be removed if the functionality is the name for both authenticated and unauthenticated.

Bento007 (commented):
What do we need a new endpoint for this?

backend/config/corpora-api.yml
ebezzi: If we want to reuse the existing endpoint, we'd have to add parameters to specify whether to only return public and not tombstoned collections, plus add new fields (collection name and a few dates). I am not opposed to it but I think it's faster/cleaner to create a new one. Also, @MillenniumFalconMechanic , once we have the new views running, will the existing endpoint be needed at all?

Bento007 (approved):

backend/config/corpora-api.yml
MillenniumFalconMechanic: The existing `/dp/v1/collections` endpoint will still be required for the my collections functionality.

PR #1699, opened 2021-12-01
chore: fix checks for processing_status

ebezzi:
### Reviewers
**Functional:** 
@metakuni @tihuan @seve 

**Readability:** 

---

## Changes
- Add `dataset.processing_status === "SUCCESS"` to the conditions for the Publish button to succeed. This is the condition that signals that the processing is completely finished. With the split pipeline + the optional Seurat conversion, we could be in a situation where the artifacts are available but the processing is still running. Also note that in order to update the dataset in a revision, this condition needs to be true (see upload.py:63)
- Adds the same check to the functional tests
- The functional check now has a 10 minutes timeout. Even if this isn't ideal, the split pipeline adds some overhead on uploading smaller datasets (mostly due to high memory reservations for all of our steps). Also, right now the final step (HandleSuccess) is a batch and not a lambda, and this will add even more overhead than necessary. We have plans to move it to a lambda which will hopefully cause the functional tests to run quicker.

### DIFF
--- a/frontend/src/views/Collection/utils.tsx
+++ b/frontend/src/views/Collection/utils.tsx
@@ -7,6 +7,7 @@ import {
   Dataset,
   DATASET_ASSET_FORMAT,
   Link,
+  PROCESSING_STATUS,
   VISIBILITY_TYPE,
 } from "src/common/entities";
 import { getUrlHost } from "src/common/utils/getUrlHost";
@@ -123,6 +124,7 @@ export function getIsPublishable(datasets: Array<Dataset>): boolean {
         assets.some((asset) => asset.filetype === DATASET_ASSET_FORMAT.H5AD)
       );
     }) &&
+    datasets.every( dataset => dataset.processing_status.processing_status === PROCESSING_STATUS.SUCCESS) &&
     datasets.some((dataset) => !dataset.tombstone)
   );
 }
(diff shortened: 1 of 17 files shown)


tihuan (approved):
FE LGTM! Thanks for adding this, Emanuele 😄 🎉

frontend/src/views/Collection/utils.tsx
```diff
@@ -123,6 +124,7 @@ export function getIsPublishable(datasets: Array<Dataset>): boolean {
         assets.some((asset) => asset.filetype === DATASET_ASSET_FORMAT.H5AD)
       );
     }) &&
+    datasets.every( dataset => dataset.processing_status.processing_status === PROCESSING_STATUS.SUCCESS) &&
```
tihuan, line 127:
```suggestion
     // (ebezzi): We need to ensure all dataset `processing_status` to be success, since creating a revision relies
     // on this status. Otherwise in the case of RDS is still WIP and a user publishes the collection and immediately
     // tries to create a revision, the revision creation will fail
    datasets.every( dataset => dataset.processing_status.processing_status === PROCESSING_STATUS.SUCCESS) &&
```

metakuni (approved):

tihuan (approved):
Ahh nice thanks for catching this 🙏 LGTM!

PR #1709, opened 2021-12-03
fix: add display state to prevent multiple renders of toast

seve:
### Reviewers
**Functional:** @tihuan 

**Readability:** @atolopko-czi / @Bento007 

---

React Router was preserving the query param even though we were deleting it off the URL.  The collection object would change, triggering a call of the effect hook which would pop the toast again.


tihuan (approved):
LGTM!

atolopko-czi (approved):
readable! :)

PR #1744, opened 2021-12-10
fix: Correctly update revised_at when publishing Collection revision

danieljhegeman:
This commit fixes a bug that allowed the revised_at column to sometimes
not be updated even when there were changes (revisions) to a
Collection's datasets.

### Reviewers
**Functional:** 
@ebezzi 
**Readability:** 
@metakuni 
---

## Changes
- add
- remove
- modify


danieljhegeman:
Previously, if the last dataset checked 1) was not new, and 2) did not have changes, then the Collection would not update its `revised_at` date.

metakuni:
Which story is this for? 🤔 
P.S. Your branch is out-of-date, but I'm not pushing that button! 😆

ebezzi (approved):

ebezzi (approved):

Issue #1702, opened 2021-12-02
Reimplement HandleSuccess task as a lambda


metakuni:
As currently implemented (see below), the decomposed step function has the potential to starve the `HandleSuccess` (AWS Batch) task if several datasets are ingested within a short time span. 

![Screen Shot 2021-12-02 at 10.42.02.png](https://images.zenhubusercontent.com/610b35d9d461c1554fe7e78c/d81916ba-1d0c-4c68-960c-b99bf3da648e)

To avoid this situation, `HandleSuccess` will be reimplemented as an AWS Lambda:

![Screen Shot 2021-12-13 at 14.39.52.png](https://images.zenhubusercontent.com/610b35d9d461c1554fe7e78c/59849e45-b546-4942-b578-6505fa959278)

### To do:
- [x] Move `handle-success` code out of `process.py` and into `lambdas/handle_success`
- [x] Create Dockerfile for new `HandleSuccess` lambda
- [x] Create a `requirements.txt` for the `HandleSuccess` lambda
- [x] Add a new `aws_lambda_function` in `.happy/terraform/modules/lambda/main.tf` (in infra and portal repos)
- [x] Modify step function in `.happy/terraform/modules/sfn/main.tf` to use the new lambda instead of batch for `HandleSuccess` (in infra and portal repos)
- [x] Add ECR URL for new `HandleSuccess` Docker image to AWS Secrets Manager in [`happy/env-{rdev,dev,stage,prod}-config`](https://us-west-2.console.aws.amazon.com/secretsmanager/home?region=us-west-2#!/listSecrets/search%3Fall=happy) (via infra repo)
- [x] Test in rdev
- [x] Test in dev

### Background info:
* https://czi-sci.slack.com/archives/C02JP4SALNQ/p1638391151030400

### Testing
1. From multiple browser windows, start uploads of several biggish datasets (i.e. datasets that will take at least several minutes to validate and convert)
1. Monitor the step function executions in [AWS console](https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/statemachines/view/arn:aws:states:us-west-2:699936264352:stateMachine:dp-staging-stagestack-sfn). Since the success handler lambda will never have to wait for a free Batch compute resource, it should execute as soon as the `Cxg` and `Seurat` tasks are finished. This also means that it will never be starved waiting around if another upload task (aka step function run) takes the Batch compute resource before the lambda execution.
[...]

PR #1751, opened 2021-12-13
fix: id bug (#1702)

metakuni:
### Reviewers
**Functional:** 
@ebezzi 

**Readability:** 
@danieljhegeman 

---

## Changes
- modify: used the proper ID for `processing_status_updater()`

### DIFF
--- a/backend/corpora/common/utils/db_helpers.py
+++ b/backend/corpora/common/utils/db_helpers.py
@@ -25,6 +25,14 @@ def clone(model: Base, primary_key: dict = None, **kwargs) -> Base:
 
 
 def processing_status_updater(session, uuid: str, updates: dict):
+    """
+    Updates the DbDatasetProcessingStatus object with the specified statuses
+    Note: The uuid parameter is the ID of the dataset_processing_status row, _NOT_ the dataset!
+    :param session: active SQLAlchemy session
+    :param uuid: UUID of the dataset_processing_status to update
+    :param updates: dictionary of statuses to update
+    :return: None
+    """
     session.query(DbDatasetProcessingStatus).filter(DbDatasetProcessingStatus.id == uuid).update(updates)
     session.commit()
     logger.debug("updating status", updates)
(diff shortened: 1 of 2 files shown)


backend/corpora/common/utils/db_helpers.py
```diff
@@ -25,6 +25,14 @@ def clone(model: Base, primary_key: dict = None, **kwargs) -> Base:
 
 
 def processing_status_updater(session, uuid: str, updates: dict):
+    """
+    Updates the DbDatasetProcessingStatus object with the specified statuses
```
metakuni, line 29: Added docstring so that no one else makes the same mistake!

backend/corpora/upload_success/app.py
```diff
@@ -17,9 +18,7 @@ def success_handler(event: dict, context) -> None:
...
-    try:
-        with db_session_manager() as session:
-            success_status = {DbDatasetProcessingStatus.processing_status: ProcessingStatus.SUCCESS}
-            processing_status_updater(session, dataset_uuid, success_status)
-    except Exception:
-        logger.exception(f"Dataset {dataset_uuid}: Failed to update processing_status to SUCCESS")
+    with db_session_manager() as session:
+        dataset = Dataset.get(session, dataset_uuid)
+        success_status = {DbDatasetProcessingStatus.processing_status: ProcessingStatus.SUCCESS}
+        processing_status_updater(session, dataset.processing_status.id, success_status)
```
metakuni, line 24: ID passed to `processing_status_updater()` should have been the ID of the `dataset_processing_status` row, _NOT_ the dataset. 🤦

backend/corpora/upload_success/app.py
```diff
@@ -17,9 +18,7 @@ def success_handler(event: dict, context) -> None:
     """
     dataset_uuid = event["dataset_uuid"]
 
-    try:
-        with db_session_manager() as session:
-            success_status = {DbDatasetProcessingStatus.processing_status: ProcessingStatus.SUCCESS}
-            processing_status_updater(session, dataset_uuid, success_status)
-    except Exception:
```
metakuni, line 24: Realized that I was swallowing all exceptions here, which is bad!

backend/corpora/upload_success/app.py
danieljhegeman: good catch

PR #1768, opened 2021-12-16
fix: remove side effect of importing s3_bucket

Bento007:
### Reviewers

**Readability:** @ebezzi  

---

## Changes
- fixes cxg_admin when `DEPLOYMENT_STAGE` is not set.


scripts/cxg_admin.py
```diff
@@ -320,7 +320,7 @@ def create_cxg_artifacts(ctx):
         for dataset in datasets:
             if dataset.explorer_url:
                 object_key = dataset.explorer_url.split("/")[-2]
-                s3_uri = f"s3://{cxg_bucket.name}/{object_key}/"
+                s3_uri = f"s3://{S3.cxg_bucket.name}/{object_key}/"
```
danieljhegeman, line 323: where is `S3` coming from?

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -39,7 +39,7 @@ def test__generate_file_url__OK(self):
         self.assertIn(file_name, url)
         self.assertIn("Expires=", url)
 
-    @mock.patch("backend.corpora.common.entities.dataset_asset.s3_client")
+    @mock.patch("backend.corpora.common.entities.dataset_asset.s3.client")
```
danieljhegeman, line 42: I must really have more python to learn--where is this coming from? I see no `s3` attribute/method/class in `DatasetAsset` or `Entity` 😅

danieljhegeman: ```make local-unit-test-backend path='tests.unit.backend.corpora.common.entities.test_dataset_asset.test__generate_file_url__ERROR'```

gives 

```AttributeError: module 'tests.unit.backend.corpora.common.entities.test_dataset_asset' has no attribute 'test__generate_file_url__ERROR'```

scripts/cxg_admin.py
Bento007: that's a refactor error. Good catch.

danieljhegeman (approved):
Looks good. 👍

Issue #1745, opened 2021-12-10
Curator names must be added to data_submission_policy_version 1.0 collections


brianraymor:
Missing `curator_names` must be added for all collections with a `data_submission_policy_version` of "1.0". 

When complete, the `data_submission_policy_version` for all published collections must be set to “2.0”.

Bento007:
All published collections have had curator name added and have been updated to data_submission_policy_version "2.0"

brianraymor:
I ran an API query. All published collections include a `curator_name` and are set to correct policy version.
LGTM.

PR #1770, opened 2021-12-16
fix: curator name backfill

Bento007:
### Reviewers
**Functional:** @seve 

**Readability:** @atolopko-czi 

---

## Changes
- add function to update curator_name in a collection based on the owner of a collection, and update all published collection to data_submission_policy_version 2.0
- connected to #1745 

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1745

### DIFF
--- a/scripts/cxg_admin.py
+++ b/scripts/cxg_admin.py
@@ -570,8 +570,12 @@ auth0_apis = {
 @click.argument("auth0_client_secret")
 @click.pass_context
 def update_curator_names(ctx, auth0_client_id, auth0_client_secret):
-    """
-    Add the curator name to all collection based on the owner of the collection.
+    """Add the curator name to all collection based on the owner of the collection.
+
+    AUTH0_CLIENT_ID: Retrieved from Auth0 console in the application section under Client ID. The application must be
+    authorized to access to the Auth0 Management API with the following permissions read:users read:user_idp_tokens.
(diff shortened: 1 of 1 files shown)


scripts/cxg_admin.py
```diff
@@ -558,6 +558,90 @@ def add_trailing_slash_to_explorer_urls(ctx):
...
+                collection.data_submission_policy_version = "2.0"
+
+    access_token = get_auth0_access_token()
+    unique_owners = get_owners_from_database()
+    owners = dict()
+    for owner in unique_owners:
+        try:
+            owner_name = get_owner_info_from_auth0(owner, access_token)
+            if owner_name:
+                owners[owner] = owner_name
```
seve, line 638: Should we do a check here that owner_name isn't an empty string?

Bento007: in python an empty string == False

seve: Sorry, should have elaborated, do we want to notify the user when there is an owner with no name attached to their account?

Bento007: We can run this to populate all of the curators and then run another query to see what collections have no curator.

seve (approved):

PR #1776, opened 2021-12-20
chore: pin base processing image

ebezzi:
### Reviewers
**Functional:** 
@metakuni 

**Readability:** 

---

## Changes
- Given that there are issues with [this version of Seurat](https://cran.r-project.org/web/packages/Seurat/index.html), we should force the portal to use the previous published version, which is contained in this (stable) image.

### DIFF
--- a/Dockerfile.processing_image
+++ b/Dockerfile.processing_image
@@ -1,6 +1,6 @@
 ARG BASE_TAG=branch-main
 
-FROM ghcr.io/chanzuckerberg/corpora-upload-base:$BASE_TAG
+FROM ghcr.io/chanzuckerberg/corpora-upload-base:20211215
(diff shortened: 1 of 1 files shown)


Dockerfile.processing_image
```diff
@@ -1,6 +1,6 @@
 ARG BASE_TAG=branch-main
 
-FROM ghcr.io/chanzuckerberg/corpora-upload-base:$BASE_TAG
+FROM ghcr.io/chanzuckerberg/corpora-upload-base:20211215
```
metakuni, line 3: We should probably add a comment here indicating why we're pinning to this magic version of the base image.

metakuni (approved):

PR #1779, opened 2021-12-20
ci: Prod Release, Dec 20th

ebezzi:


Dockerfile.processing_image
```diff
@@ -1,6 +1,8 @@
 ARG BASE_TAG=branch-main
 
-FROM ghcr.io/chanzuckerberg/corpora-upload-base:$BASE_TAG
+# Note: versions after 20211215 will have a Seurat version which is incompatible with our code
+# Do not change this until the underlying issue is fixed
```
atolopko-czi, line 4: Thanks for the callout on the dependency. Can you link the issue here as well?

ebezzi: I can't modify this anymore since this is a staging -> prod PR. I will submit another one with more clarifications to main.

PR #1784, opened 2021-12-21
fix: detect duplicate H5AD files

Bento007:
Only check h5ad File for duplication across datasets in a collection. Don't check derived files for duplication. This was added because smoke tests were failing with a 500 error when publishing a collection. This was caused by the way were are retrieving metadata from S3. When parsing the file name from the S3 URI we were not taking into account that the URI for the CXG artifact was to a directory. Thus when requesting metadata for a directory, S3 return 404. Rather than retrieving all of the files from the CXG directory and check for duplication it's faster and easier to check if the source H5AD file is duplicated across the datasets in the collection. Since the H5AD file is used to generate the other artifacts, it's safe to assume if the H5AD file are unique, then the derived artifacts must also be unique.

**This currently blocks publishing a collection.**

### Reviewers

**Readability:** @seve 

---

## Changes
- modify duplication detection during publication to only check H5AD files.
- add collection clean up for smoke-tests. 
- add a test that we are only duplication check H5AD files.

### DIFF
--- a/backend/corpora/lambdas/api/v1/collection_uuid/publish.py
+++ b/backend/corpora/lambdas/api/v1/collection_uuid/publish.py
@@ -19,7 +19,7 @@ def check_for_duplicate_datasets(collection: Collection) -> bool:
     for dataset in collection.datasets:
         if not dataset.tombstone:
             for artifact in dataset.artifacts:
-                if artifact.file_type == DatasetArtifactFileType.H5AD:
+                if artifact.filetype == DatasetArtifactFileType.H5AD:
                     _artifact = DatasetAsset(artifact)
                     metadata = _artifact.get_s3_metadata()
                     if not metadata:
--- a/backend/corpora/lambdas/api/v1/collection_uuid/publish.py
+++ b/backend/corpora/lambdas/api/v1/collection_uuid/publish.py
@@ -1,6 +1,6 @@
 from flask import make_response, g
 
-from .....common.corpora_orm import CollectionVisibility
+from .....common.corpora_orm import CollectionVisibility, DatasetArtifactFileType
 from .....common.entities import Collection, DatasetAsset
 from .....common.utils.exceptions import ConflictException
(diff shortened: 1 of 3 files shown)


ebezzi (approved):

backend/corpora/lambdas/api/v1/collection_uuid/publish.py
```diff
@@ -19,17 +19,18 @@ def check_for_duplicate_datasets(collection: Collection) -> bool:
...
-                if not metadata:
-                    raise ServerErrorHTTPException(
-                        "Failed to check datasets for duplications. Unable to find associated artifacts."
-                    )
-                etag = metadata["ETag"]
-                if etag not in etags:
-                    etags.add(etag)
-                else:
-                    return True
+                if artifact.file_type == DatasetArtifactFileType.H5AD:
```
ebezzi, line 22:
```suggestion
                if artifact.filetype == DatasetArtifactFileType.H5AD:
```

tests/unit/backend/corpora/api_server/collection_uuid/test_publish.py
```diff
@@ -180,6 +180,21 @@ def test__publish_with_Duplicate_dataset__409(self, mocked):
...
+        collection = self.generate_collection(self.session)
+        for i in range(2):
+            dataset = self.generate_dataset(
+                self.session, collection_id=collection.id, collection_visibility=collection.visibility
+            )
+            self.generate_asset(self.session, dataset.id, filetype=DatasetArtifactFileType.CXG)
+
+        path = f"/dp/v1/collections/{collection.id}/publish"
+        response = self.app.post(path, headers=self.headers_authed, data=json.dumps(self.publish_body))
+        self.assertEqual(201, response.status_code)
```
Bento007, line 196:
```suggestion
        self.assertEqual(202, response.status_code)
```

Issue #1803, opened 2022-01-07
Create a DOI normalization function


brianraymor:
@ebezzi commented on [Fri Jan 07 2022](https://github.com/chanzuckerberg/single-cell/issues/266)

PR #1814, opened 2022-01-10
feat: create DOI normalization function

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

Ticket [#1803](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1803)

---

## Changes
- Adds a function that returns a normalized DOI for a collection. This DOI doesn't have an URL format and will be used to call the Crossref API to retrieve the metadata.

### DIFF
--- a/backend/corpora/common/entities/collection.py
+++ b/backend/corpora/common/entities/collection.py
@@ -1,11 +1,12 @@
 import typing
 from datetime import datetime
+from urllib.parse import urlparse
 from sqlalchemy import and_
 
 from . import Dataset
 from .entity import Entity
 from .geneset import Geneset
-from ..corpora_orm import DbCollection, DbCollectionLink, CollectionVisibility
+from ..corpora_orm import CollectionLinkType, DbCollection, DbCollectionLink, CollectionVisibility
 from ..utils.db_helpers import clone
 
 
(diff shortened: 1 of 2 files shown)


backend/corpora/common/entities/collection.py
```diff
@@ -280,3 +281,18 @@ def delete(self):
...
+
+    @staticmethod
+    def _normalize_doi(doi_url) -> str:
+        try:
+            return urlparse(doi_url.strip()).path.strip("/")
+        except Exception:
+            return None
+
+    def get_normalized_doi(self) -> str:
+        """Returns a normalized DOI if it exists, None otherwise"""
```
MDunitz, line 293: Could you specify what normalized means in this context?

backend/corpora/common/entities/collection.py
```diff
@@ -280,3 +281,18 @@ def delete(self):
...
+        doi = [link for link in self.links if link.link_type == CollectionLinkType.DOI]
```
MDunitz, line 294: Is it every possible for this to return more than one link?

ebezzi: Publications should have a unique DOI so in theory, no. I don't think we're enforcing this though, so maybe we should do that when we add links.

MDunitz: Can you create a ticket to ensue it is enforced? And maybe to double check the db to make sure nothing slipped through?

MDunitz (approved):
lgtm

PR #1828, opened 2022-01-12
fix: updated staleTime for filter queries. (#1773)

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
* Updated `stateTime` of filter-related queries to never expire.
* Note: data refresh-management will not be handled in Q1 2022. `staleTime` could be replaced with `refetchOnWindowFocus` during update of filter state management in _[Remember filter while navigating within the Portal](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1811)_ (if data re-fetching without clearing selected filters is supported).

## Definition of Done (from ticket)
* Any selected filters remain selected on blur and refocus of browser tab.

## QA steps
1. View collections page.
1. Select at least one category value from the filter sidebar.
1. Click to different browser tab.
1. Click back to tab containing collections page.

 => Selected filter should still be selected.

PR #1796, opened 2022-01-03
fix: Revert #1784 and #1652

Bento007:
### Reviewers
**Functional:** @danieljhegeman 

---
Reverting PRs related to blocking publish of collection with duplicate datasets.
## Changes
- revert #1784 
- revert #1652

PR #1789, opened 2021-12-22
fix: rdev non-GHA builds

atolopko-czi:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
Add explicit --stack-name option for GHA-based builds of docker images. Replaces the use of --extra-tag to pass in the stack name. This allows the script to work in non-GHA build usages.

## QA steps (optional)

Tested:
* single-cell-infra happy explorer + data portal rdev *local* create
* single-cell-infra happy explorer + data portal rdev *GHA* create
* single-cell-data-portal happy *local* create

Issue #1773, opened 2021-12-17
Switching to another browser tab and returning to the cellxgene tab resets filter


brianraymor:
For both the Collections and Datasets tabs, if I filter results, switch to another tab in the browser, and then return to the cellxgene tab, the filter is cleared:

https://images.zenhubusercontent.com/228681195/86c378b8-310e-4760-80a4-f163cfc2f023/screen_recording_2021_12_17_at_1_49_45_pm.mov

brianraymor:
Verified. Closing.

PR #1858, opened 2022-01-19
chore: deploy staging -> prod

MDunitz:
deploying
#1796
#1749
#1773
#1828
#1768

This should have no effect on prod env
#1789


tihuan (approved):
in case rubber stamping helps lol

PR #1869, opened 2022-01-21
feat: ontology ancestors term expansion

ebezzi:
### Reviewers
**Functional:** 
@MillenniumFalconMechanic @NoopDog 
**Readability:** 
@MDunitz 

---

## Changes
- Adds ontology term expansion for `development_stage`


backend/corpora/common/utils/ontology_mapping.py
```diff
@@ -0,0 +1,1644 @@
+
+
+ontology_mapping = {
```
atolopko-czi, line 3: What is the process for generating and updating this data? If/when we have newer datasets that are using a newer version of the ontology, could we run into minor problems if we forget to update this hardcoded data? I'm not sure there's a way to be defensive here in the code here, but I think we should at least indicate what ontology version this is based off of. 

In fact, since Portal already has a dependency on single-cell-curation, and that repo manages ontology versioning, maybe we should import this data from a file stored in that repo instead? There are some scripts/docs in that repo where we could note (or enforce?) that this data be updated when needed.

atolopko-czi (changes requested):

backend/corpora/common/utils/ontology_mapping.py
ebezzi: I have added the notebook I used to generate the mapping file [to the single-cell-curation repo](https://github.com/chanzuckerberg/single-cell-curation/pull/189). If we decide to merge it, I can add a comment with the link at the top of this file.

Importing the data from an external file could be a good idea, we would need to load a json as a singleton object and keep it in memory to avoid loading it for each call. That said:
1. The generality of this mapping is debatable, as it's mostly meant for the specific filtering function. For instance, not all UBERON mappings are considered, but just a subset of them.
1. Updates to the ontology are unlikely or at least very sparse. Given that we probably want to add these mappings to the database (possibly generalizing them so that WMG can also make use of them), additional engineering might not be worth here.

backend/corpora/common/entities/dataset.py
```diff
@@ -188,6 +190,21 @@ def to_dict(db_object):
...
+            leaves = [e["ontology_term_id"] for e in result["development_stage"]]
+
+            if not leaves: 
+                continue
+
+            ancestors = [ontology_mapping.get(leaf) for leaf in leaves]
+            flattened_ancestors = [item for sublist in ancestors if sublist for item in sublist]
+            if ancestors:
+                result["development_stage_ancestors"] = flattened_ancestors
+
```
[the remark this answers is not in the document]
MDunitz: +1

backend/corpora/common/utils/ontology_mapping.py
MDunitz: I like the idea of storing ontoogies and the code used to generate them in the curation repo. Otherwise I worry that it will be difficult to keep everything in sync

backend/corpora/common/entities/dataset.py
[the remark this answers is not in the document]
ebezzi: Good idea.

ebezzi:
@atolopko-czi after a discussion with @atolopko-czi , we decided to go with this solution for now and go for a more engineered solution that involves using `cellxgene-schema` to compute nodes using the ontology tree directly (as opposed to a static map) and serializing the ancestors in the database later on. We'll also collect WMG requirements and see if we can develop the solution so that both projects will fit in.

atolopko-czi (approved):

Issue #1797, opened 2022-01-04
Hide "no results" during load of collections and datasets


MillenniumFalconMechanic:
"No results" is displayed during load of collections and datasets pages.

https://user-images.githubusercontent.com/2446424/148095026-018ce34d-99a4-481a-9d28-42e21a58033c.mov

#### Approach
* Use react-table's `loading` flag to determine the difference between the loading and no results state.

#### Definition of Done
* "No results" is only displayed if collections and datasets are loaded and there are no results matching the selected filter.

PR #1870, opened 2022-01-21
fix: removed flash of no results for filtered grids. (#1797)

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
- Updated state management of filter-related queries.
- Freebie: removed gene set-related functionality from collections page as per [Slack](https://clevercanary.slack.com/archives/C02CF7MQPGV/p1642706774046900?thread_ts=1642655185.046100&cid=C02CF7MQPGV) ([here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/1870/files#diff-cacbfc6dcdd6f563592e6b8dbe34c27d187e17be08ab7211320aaf4dd76cbe82L43), [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/1870/files#diff-cacbfc6dcdd6f563592e6b8dbe34c27d187e17be08ab7211320aaf4dd76cbe82L57) and [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/1870/files#diff-cacbfc6dcdd6f563592e6b8dbe34c27d187e17be08ab7211320aaf4dd76cbe82L202)).

## Definition of Done
* "No results" is only displayed if collections and datasets are loaded and there are no results matching the selected filter.

### Collections
1. View `/collections` with a throttled network (e.g. "Fast 3G").
=> Collections should load without briefly displaying "No Results" hero.

### Datasets
1. View `/datasets` with a throttled network (e.g. "Fast 3G").
=> Datasets should load without briefly displaying "No Results" hero.

### DIFF
--- a/frontend/src/common/queries/filter.ts
+++ b/frontend/src/common/queries/filter.ts
@@ -120,6 +120,10 @@ export function useFetchCollectionRows(): FetchCategoriesRows<CollectionRow> {
   // View model built from join of collections response and aggregated metadata of dataset rows.
   const [collectionRows, setCollectionRows] = useState<CollectionRow[]>([]);
 
+  // Fetch states
+  const [isError, setIsError] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
   // Fetch datasets.
   const {
     data: datasets,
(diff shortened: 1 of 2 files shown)


frontend/src/common/queries/filter.ts
```diff
@@ -188,17 +206,27 @@ export function useFetchDatasetRows(): FetchCategoriesRows<DatasetRow> {
...
   useEffect(() => {
+    setIsLoading(datasetsLoading || collectionsLoading);
+    setIsError(datasetsError || collectionsError);
+
     if (!datasets || !collectionsById) {
       return;
     }
     setDatasetRows(buildDatasetRows(collectionsById, datasets));
-  }, [datasets, collectionsById]);
+  }, [
```
MillenniumFalconMechanic, line 218: Hey @tihuan! 

I have updated React Query's `loading` and `error` flags to be state here in an attempt to have atomic updates of state and data where possible. 

Previously, when React Query completed the fetch, I was seeing:

1. `loading` updated from from `true` to `false`.
2. Calling component re-rendered with updated `loading` value.
3. `datasetRows` calculated in this effect.
4. Calling component re-rendered with updated `rows` value.

This flow was causing a brief flash of the "no results" hero in the calling component where `loading` is updated to `false` but the rows haven't yet been computed.

I was hoping to get your opinion on my fix here as I am wondering if there's a cleaner solution than adding React Query flags as state and having such a long dependency list on the hook.

tihuan: Thanks for the PR and describing the issue in great detail, @MillenniumFalconMechanic 🤩 🙏 

It seems like the root cause is `useEffect()` lagging behind by one render cycle, so I have a PR that switches it out to use `useMemo`

https://github.com/chanzuckerberg/single-cell-data-portal/pull/1871

I think the reason is because the states (row state, in this case) that we update inside a `useEffect()` are updated in the next render cycle, that's why when the loading states change, the rows state is one cycle behind

Using `useMemo()` solves the problem, because the calculated result is available immediately and doesn't need to wait till `setState()` to update the value at the next cycle!

Hope that makes sense 😄 !

MillenniumFalconMechanic: Perfect, makes sense, thanks so much Timmy!

MillenniumFalconMechanic:
Superseded by #1871.

PR #1876, opened 2022-01-25
refactor: wheres my gene

tihuan:
### Reviewers
**Functional:** 
@seve 

**Readability:** 
@MDunitz 
---

## Changes
- add
1. Add comments to explain things better
2. Clean up `HeatMap` component and extract echarts option creations into their own util functions. Also add type definitions
3. Add x-axis mask:
    1. Before:  
        Screen Shot 2022-01-25 at 2 53 22 PM

    2. After:
        demo

- remove
- modify

### DIFF
--- a/frontend/src/views/WheresMyGene/common/types.ts
+++ b/frontend/src/views/WheresMyGene/common/types.ts
@@ -1,11 +1,13 @@
 import { CSSProperties } from "react";
 
-/** This object holds a cell type and its genes. */
-export interface CellTypeAndGenes {
+/** This object holds a cell type and its gene expressions. */
+export interface CellTypeSummary {
+  /** cellType id */
   id: string;
+  /** cellType name */
   name: string;
-  expressions?: {
-    [geneName: string]: GeneExpression;
+  geneExpressions?: {
+    [geneName: string]: CellTypeGeneExpressionSummaryData;
   };
 }
 
@@ -15,30 +17,39 @@ export interface CellType {
 }
 
 /**
- * This object holds a gene and all the cell types that has this
- * genetic expression.
+ * This object holds a gene and all the cell types that express this gene.
  */
-export interface RawGeneExpression {
-  gene_name: string;
-  cell_types: GeneExpression[];
+export interface GeneExpressionSummary {
+  /** gene name */
+  name: string;
+  cellTypeGeneExpressionSummaries: CellTypeGeneExpressionSummaryData[];
+}
+
+/** This is the original data shape in API response for `CellTypeGeneExpressionSummaryData` */
+export interface RawCellTypeGeneExpressionSummaryData {
+  /** cellTypeId */
+  id: string;
+  /** percentage of the current subset of cells that express this gene */
+  pc: number;
+  /** mean expression of the current subset of cells that express this gene */
+  me: number;
 }
 
 /**
- * This object describes the cell type id and the cell's genetic
- * expression metadata
+ * This object describes the cell type id and its gene expression metadata given
+ * the current subset of cells.
  */
-export interface GeneExpression {
+export interface CellTypeGeneExpressionSummaryData {
   /** cellTypeId */
   id: string;
-  /** percentage of the cells that have this genetic expression */
-  pc: number;
-  /** mean expression of the cells that have this genetic expression */
-  me: number;
+  /** percentage of the current subset of cells that express this gene */
+  percentage: number;
+  /** mean expression of the current subset of cells that express this gene */
+  meanExpression: number;
 }
 
 export interface Gene {
   id: string;
   name: string;
   style?: CSSProperties;
-  cell_types?: GeneExpression[];
 }
--- a/frontend/src/views/WheresMyGene/common/types.ts
+++ b/frontend/src/views/WheresMyGene/common/types.ts
@@ -1,5 +1,6 @@
 import { CSSProperties } from "react";
 
+/** This object holds a cell type and its genes. */
 export interface CellTypeAndGenes {
   id: string;
   name: string;
@@ -13,15 +14,25 @@ export interface CellType {
   name: string;
 }
 
+/**
+ * This object holds a gene and all the cell types that has this
+ * genetic expression.
+ */
 export interface RawGeneExpression {
   gene_name: string;
   cell_types: GeneExpression[];
 }
 
+/**
+ * This object describes the cell type id and the cell's genetic
+ * expression metadata
+ */
 export interface GeneExpression {
-  // cellTypeId
+  /** cellTypeId */
   id: string;
+  /** percentage of the cells that have this genetic expression */
   pc: number;
+  /** mean expression of the cells that have this genetic expression */
   me: number;
 }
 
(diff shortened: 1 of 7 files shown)


frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -1,5 +1,6 @@
 import { CSSProperties } from "react";
 
+// This object holds a cell type and its genes.
```
seve, line 3: nit: Prefer multiline comment for typedoc support

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -13,15 +14,23 @@ export interface CellType {
...
+ * genetic expression.
+ */
 export interface RawGeneExpression {
   gene_name: string;
   cell_types: GeneExpression[];
 }
 
+// This object describes the cell type id and the cell's genetic
+// expression metadata
 export interface GeneExpression {
```
seve, line 28:
```suggestion
/**
* This object describes the cell type id and the cell's genetic
* expression metadata
*/
export interface GeneExpression {
```

seve: formatting might be off

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -13,15 +14,23 @@ export interface CellType {
...
   // cellTypeId
   id: string;
+  // percentage of the cells that have this genetic expression
   pc: number;
+  // mean expression of the cells that have this genetic expression
```
seve, line 33: same here

seve (approved):
LGTM, did you look into the React wrapper at all or was that out of scope for this ticket

frontend/src/views/WheresMyGene/common/types.ts
tihuan: Ah great call. Thank you! Done!

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -13,15 +14,25 @@ export interface CellType {
   name: string;
 }
 
+/**
+ * This object holds a gene and all the cell types that has this
```
MDunitz, line 18: I'm not sure I understand what `all the cell types that has this genetic expression` means. Is this a list of cell types that have non zero expression levels for the gene?

frontend/src/views/WheresMyGene/components/GeneSearchBar/index.tsx
```diff
@@ -134,8 +134,8 @@ export default function GeneSearchBar({ onGenesChange }: Props): JSX.Element {
   // DEBUG
```
MDunitz, line 134: Can these be removed or replaced with a todo?

MDunitz (approved):
lgtm with a few nits around phrasing/cleanup

tihuan:
> LGTM, did you look into the React wrapper at all or was that out of scope for this ticket

Ah yes! I looked into it and it seems like the [thin wrapper library](https://github.com/hustcc/echarts-for-react/commits/master) is not actively being maintained. And I also can't tell what benefits we get from using it, since instead of learning echarts' API, using that thin wrapper means we need to learn another layer of API

Also this [blogpost](https://dev.to/maneetgoyal/using-apache-echarts-with-react-and-typescript-353k) says they have trouble using the wrapper to resize echarts properly 

So I think using echarts directly should be fine?

Thank you!

frontend/src/views/WheresMyGene/common/types.ts
tihuan: Thanks Maidons! That's right!

The object shape is:

```ts
export interface RawGeneExpression {
  gene_name: string;
  cell_types: GeneExpression[];
}
```

So each gene will have one `RawGeneExpression` object

E.g.,

geneA:

```ts
{
  gene_name: "geneA",
  cell_types: [ { id: "cellTypeIdA, pc: 0.01, me: 0.01 }, { id: "cellTypeIdB", pc: 0.02, me: 0.02 }]
}
```

geneB:

```ts
{
  gene_name: "geneB",
  cell_types: [ { id: "cellTypeIdA, pc: 0.03, me: 0.03 }, { id: "cellTypeIdB", pc: 0.04, me: 0.04 }]
}
```

frontend/src/views/WheresMyGene/components/GeneSearchBar/index.tsx
tihuan: Ah good point! Just removed it 👌 Thank you!

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -13,15 +14,25 @@ export interface CellType {
...
 
+/**
+ * This object describes the cell type id and the cell's genetic
+ * expression metadata
+ */
 export interface GeneExpression {
-  // cellTypeId
+  /** cellTypeId */
   id: string;
+  /** percentage of the cells that have this genetic expression */
```
MDunitz, line 33:
```suggestion
  /** percentage of the current subset of cells that express this gene */
```

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -13,15 +14,25 @@ export interface CellType {
...
   pc: number;
+  /** mean expression of the cells that have this genetic expression */
```
MDunitz, line 35: Is this the mean expression level of the cells expressing this gene? or of all the cells (are zero values included in the average)

tihuan: CC: @atolopko-czi @pablo-gar please help here 🙏 Thank you!

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -1,10 +1,13 @@
 import { CSSProperties } from "react";
 
-export interface CellTypeAndGenes {
```
tihuan, line 3: Hey @seve ! Madison and I went through the type names again to make sure it conveys the meanings as efficiently as possible 😆 Can you please take another look at the new names and see if they look good to you too?

Thanks a lot!

frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
```diff
@@ -134,6 +134,14 @@ export default function HeatMap({
...
     });
 
+    xAxisChart.on("click", function (params) {
+      console.log(params);
+    });
+
+    yAxisChart.on("click", function (params) {
+      console.log(params);
+    });
+
```
seve, line 144: Debug?

seve (approved):
One comment

frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
tihuan: Oooh yes! Part of the delete genes/cellTypes stuff sneaked in 😆 Removed! Thank you!

frontend/src/views/WheresMyGene/common/types.ts
atolopko-czi: @MDunitz Since mean expression is averaging _rankit_ values, which are in the range (-inf, +inf), we filter out expr values where the original raw count is < 3.  In the code that was used to generate the prototype data, mean expression is computed as N/D, where N is number of cells with raw count >= 3, and D is the total number of cells across for the given cell_type, across all datasets.

Issue #1847, opened 2022-01-19
Update /wmg layout to be three panels


tihuan:

seve:
@tihuan Can you attach a mock of some sort? Even if it's just a screenshot of CC's Search/Filter feature.

tihuan:
Morning Seve! Harley is gonna finalize the V1 mock today, but I think the layout is likely staying the same 👌

https://www.figma.com/file/htICFl7zihfs1z9eIYVwzw/Gene-Search?node-id=2402%3A423263

Is that what you're looking for?

Thank you!

seve:
Right! Thank you, forgot you mentioned that yesterday afternoon!

Just to clarify/refine, the layout is more akin to explorer with a main visualization book-ended by two sidebars, yes? I'm being nitpicky just to establish the difference in priority between the two headspaces: three panels with equal thirds of importance, vs main single importance viz with two supplementing sidebars.

tihuan:
Ooh good point that Explorer has the three panel layout too 💡

That's a great question, from the mock it doesn't seem like three panels will be equal in size, but they will be collapsible! 

Let's confirm with @hthomas-czi on the actual specs just in case 🙏

And please feel free to reach out to him directly if he doesn't respond here, since I know some people don't check GitHub notif frequently 😄

Thanks a lot!

hthomas-czi:
@seve Yes, that's the right way to think about the interface (main workspace + 2 sidebars). It differs from Explorer in the ways @tihuan described — namely, they're asymmetrically and collapsible.

tihuan:
Hey @seve ! 

Harley finalized the panel layout in the [latest Figma](https://www.figma.com/file/htICFl7zihfs1z9eIYVwzw/Gene-Search?node-id=2402%3A423263) 🚀

What we discussed in the kickoff meeting this morning is that the left panel expand/collapse will reposition the middle panel (instead of an overlay), but won't resize the middle panel. And the right panel expand/collapse will overlay the middle panel, so it could cover the middle panel partially when expanded and that's fine

Hope this makes sense! 

Thank you!

tihuan:
I just updated the DoD, please feel free to edit it as you see fit! Thanks!

tihuan:
@seve also just to be super explicit, the three panel layout doesn't include contents in the panel! We have other tickets for those 👍

[...]

PR #1890, opened 2022-01-31
feat(wmg): Add sidebars to main view

seve:
### Reviewers
**Functional:**  @frano-m, @tihuan 

---

## Changes
- add
  - Add sidebars to the wmg page
- modify
  - Added a prop to sidebar component to allow for left/right positioning 

https://user-images.githubusercontent.com/8716829/151847774-a148a639-38d4-4af2-9a13-09a4d26923d3.mov

--- 
#1847


frontend/src/pages/wmg.tsx
```diff
@@ -46,4 +47,6 @@ const Page = (): JSX.Element => (
   </StylesProvider>
 );
 
+Page.Layout = SidebarLayout;
+
```
seve, line 51: This was really tricky for me to find/realize, a single short line with a large impact.  Is there anyway we can surface this better? Maybe somehow require layout dependent components to be nested inside a component where `Layout` is defined? Similar to a Context Provider

frano-m (approved):
Hey @seve, LGTM! 🌟 

I've noticed there are a couple of warnings in my console. Not sure if it is related to this PR, but this is what I'm seeing:

```
react_devtools_backend.js:4045 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
at HeatMap (webpack-internal:///./src/views/WheresMyGene/components/HeatMap/index.tsx:40:24)
```
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
at WheresMyGene (webpack-internal:///./src/views/WheresMyGene/index.tsx:80:66)
at ThemeProvider (webpack-internal:///./node_modules/@material-ui/styles/esm/ThemeProvider/ThemeProvider.js:42:24)
```
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at WheresMyGene (webpack-internal:///./src/views/WheresMyGene/index.tsx:80:66)
    at ThemeProvider (webpack-internal:///./node_modules/@material-ui/styles/esm/ThemeProvider/ThemeProvider.js:42:24)
```

frontend/src/pages/wmg.tsx
tihuan: Hmm thanks for raising this! Yeah, right now you'd need to look at how other pages use custom Layout and check out `_app.tsx` to see how `Page.Layout` is used 🤔 

I think maybe let's start with adding a comment above `Page.Layout = SidebarLayout;` to explain what it does, so it  brings attention to the reader? And then when we have time in cool time period, we can figure out a structural way to solve this

The reasoning is because we don't add pages frequently, so hopefully comment is enough as a patch for now

What do you all think?

Thank you!

tihuan (approved):
Niice looks super good! Thanks for updating the Layout component and adding the layout, Seve 🎉!!

frontend/src/pages/wmg.tsx
seve: You're right.  If we aren't creating pages often, if any this Q, it's not worth putting time into this just yet.

seve:
```suggestion
// This layout needs to be defined in order to correctly style the SideBars and their wrappers
Page.Layout = SidebarLayout;

```

PR #1905, opened 2022-02-03
fix: change Dropbox content length header

ebezzi:
### Reviewers
**Functional:** 
@seve @Bento007 

**Readability:** 

---

## Changes
Changes the Dropbox content length header. This is currently broken in all environments and causes the dataset upload to fail. Note that there is no documentation around these headers, so this might change any time. I can't think of a better solution though. We could make the size check optional but as @maniarathi pointed out, that would still break one of our features (block datasets that are too big) and it will do so silently.

### DIFF
--- a/backend/corpora/common/utils/dl_sources/url.py
+++ b/backend/corpora/common/utils/dl_sources/url.py
@@ -73,7 +73,7 @@ class DropBoxURL(URL):
         resp.raise_for_status()
 
         return {
-            "size": int(self._get_key(resp.headers, "content-length")),
+            "size": int(self._get_key(resp.headers, "x-dropbox-content-length")),
             "name": self._get_key(resp.headers, "content-disposition").split(";")[1].split("=", 1)[1][1:-1],
         }
 
--- a/backend/corpora/common/utils/dl_sources/url.py
+++ b/backend/corpora/common/utils/dl_sources/url.py
@@ -37,6 +37,16 @@ class URL(ABC):
                 f"{self.__class__.__name__}:URL({self.url}) failed request. '{key}' not present in the header."
             )
 
+    def _get_key_with_fallback(self, headers: dict, key: str, fallback_key: str) -> str:
+        try:
+            return headers.get(key) or headers[fallback_key]
+        except KeyError:
+            raise MissingHeaderException(
+                f"""{self.__class__.__name__}:URL({self.url}) failed request.
+                Neither '{key}' nor '{fallback_key}' are present in the header.
+                """
+            )
+
 
 class DropBoxURL(URL):
     """Supports download URLs from a DropBox share link."""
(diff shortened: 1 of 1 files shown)


seve (approved):

Bento007 (approved):

Bento007 (changes requested):
Still getting 400 on the unit tests

tihuan (approved):

ebezzi:
@Bento007 Looks like the header changes depending on whether you use HTTP2 or HTTP1.1. Since this would require more investigation, I edited the PR so that both headers are tried and failed only if neither can be found. This seems safe enough for a hotfix.

backend/corpora/common/utils/dl_sources/url.py
```diff
@@ -37,6 +37,16 @@ def _get_key(self, headers: dict, key: str) -> str:
                 f"{self.__class__.__name__}:URL({self.url}) failed request. '{key}' not present in the header."
             )
 
+    def _get_key_with_fallback(self, headers: dict, key: str, fallback_key: str) -> str:
```
Bento007, line 40: Why not pass in a list and put the keys in priority order. I can see this functions getting overly verbose with arguments at this rate.

ebezzi: This is a hotfix, I prefer a solution which is explicit and contains the least amount of logic. I'm happy to refactor it later and add some unit tests, but that should be done on the main branch.

backend/corpora/common/utils/dl_sources/url.py
```diff
@@ -72,8 +82,10 @@ def file_info(self) -> dict:
         resp = requests.head(self.url, allow_redirects=True)
         resp.raise_for_status()
 
+        print(resp.headers)
+
         return {
-            "size": int(self._get_key(resp.headers, "content-length")),
+            "size": int(self._get_key_with_fallback(resp.headers, "content-length", "x-dropbox-content-length")),
```
Bento007, line 88: where did you find `x-dropbox-content-length`?

ebezzi: I created a notebook, replicated the call and inspected the headers. ```curl --head --http1.1 -L "https://www.dropbox.com/s/j0s89tokxwz3h2p/2_0_0_raw_valid.h5ad/?dl=1"``` can also be used.

brianraymor:
> We could make the size check optional but as @maniarathi pointed out, that would still break one of our features (block datasets that are too big) and it will do so silently.

But sizeLimit is already enforced by the dropbox chooser on the portal FE:

[https://github.com/chanzuckerberg/single-cell-data-portal/blob/f450d5024cca2c3fe22[…]6844456fc8f1a5/frontend/src/components/DropboxChooser/index.tsx](https://github.com/chanzuckerberg/single-cell-data-portal/blob/f450d5024cca2c3fe22454a3f16844456fc8f1a5/frontend/src/components/DropboxChooser/index.tsx#L31)

https://www.dropbox.com/developers/chooser

```
    // Optional. A limit on the size of each file that may be selected, in bytes.
    // If specified, the user will only be able to select files with size
    // less than or equal to this limit.
    // For the purposes of this option, folders have size zero.
    sizeLimit: 1024, // or any positive number
```

tihuan:
Good point that FE checks this! Would curator API be able to check file size too?

PR #1939, opened 2022-02-08
feat: new doi support

ebezzi:
### Reviewers
**Functional:** 
@MillenniumFalconMechanic @atolopko-czi @MDunitz 

**Readability:** 

---

## Changes
- Extends DOI support so that a shortened format (aka DOI CURIE) can be sent as the payload
- Adds DOI normalization so that if a shortened format is used, it is transformed into an URL

### DIFF
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -999,7 +999,7 @@ components:
             type: string
           link_type:
             type: string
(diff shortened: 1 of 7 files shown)


atolopko-czi (approved):
Code LGTM, though I didn't study the tests.

PR #1941, opened 2022-02-08
fix: copy collections DO NOT PROMOTE TO STAGING/PROD

MDunitz:
### Reviewers
**Functional:** 
- @ebezzi 
**Readability:** 
- @Bento007 
---

## Changes

- modify dataset processing to skip validation step

DO NOT MERGE
DO NOT PROMOTE TO STAGING/PROD

This will allow us to copy collections from prod to dev for testing and should be reverted as soon as that process is complete. This PR should NOT be pushed to staging/prod

PR #1942, opened 2022-02-09
fix: Revert "fix: copy collections DO NOT PROMOTE TO STAGING/PROD"

MDunitz:
Reverts chanzuckerberg/single-cell-data-portal#1941


Bento007 (approved):

backend/corpora/dataset_processing/process.py
```diff
@@ -466,14 +466,12 @@ def validate_h5ad_file_and_add_labels(dataset_id: str, local_filename: str) -> t
     :param local_filename: file name of the dataset to validate and label
     :return: file name of labeled dataset, boolean indicating if Seurat conversion is possible
```
MDunitz, line 467:
```suggestion
    :return: file name of labeled dataset, boolean indicating if seurat conversion is possible
```

PR #1950, opened 2022-02-10
feat(wmg): Add delete/reset cell types and genes

tihuan:
### Reviewers
**Functional:** 
@seve 
**Readability:** 
@MDunitz 
---

## Changes
1. Update `deleteSelectedGenesAndSelectedCellTypeIds()` event to no longer take payload, since data is in the state already
2. Add `tissueCellTypesFetched()` and `resetTissueCellTypes()` events to support delete and reset cell types
3. Add type `CellTypeMetadata`, which is string in `${id}~${tissue}~${name}` format. Also added corresponding serializer(`getAllSerializedCellTypeMetadata()`) and deserializer(`deserializeCellTypeMetadata ()`)
4. Update heat map to highlight selected cell types and genes, different text style for tissues, and reset style for tissues
5. Update `frontend/src/views/WheresMyGene/index.tsx` data flow. In general, reducer state now stores `selectedGenes` and `selectedCellTypeIds`, and `selectedCellTypes` is **derived state**, thus NOT needing to live in the store. Same thing for `tissuesWithDeletedCellTypes`, which is also a derived state that we cache via `useMemo()`
6. BUG FIX: The issue we had with React memory leak is due to debounced function not being cancelled, so we have that now for all debounced calls
7. Perf improvement: Split updating charts separately instead of in one single `useEffect()`, so we no longer trigger updates too frequently and unnecessarily -- we now have `useUpdateChart()`, useUpdateXAxisChart()` and `useUpdateYAxisChart()`
8. 

## QA steps (optional)
1. You can delete cell types and genes via the heat map
2. You can reset cell types by clicking its tissue
3. You should see tissue "Lung" in the row now, indicating multiple tissues support

### DIFF
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -38,6 +38,8 @@ module.exports = {
     "@typescript-eslint/camelcase": 0,
     // Disable prop-types as we use TypeScript for type checking
     "@typescript-eslint/explicit-function-return-type": "off",
+    // (thuang): Allow args prefixed with `_`
+    // example: https://eslint.org/docs/rules/no-unused-vars#argsignorepattern
     "@typescript-eslint/no-unused-vars": [
       "error",
       {
--- a/frontend/.eslintrc.js
+++ b/frontend/.eslintrc.js
@@ -38,6 +38,15 @@ module.exports = {
     "@typescript-eslint/camelcase": 0,
     // Disable prop-types as we use TypeScript for type checking
     "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        args: "after-used",
+        argsIgnorePattern: "^_",
+        ignoreRestSiblings: false,
+        vars: "all",
+      },
+    ],
     camelcase: "off",
     "react/jsx-no-target-blank": 0,
     "react/prop-types": "off",
--- a/frontend/src/views/WheresMyGene/common/store/actions.ts
+++ b/frontend/src/views/WheresMyGene/common/store/actions.ts
@@ -3,22 +3,15 @@
  * in `./reducer.ts`.
  */
 
+import { CellTypeMetadata } from "../../components/HeatMap/utils";
+import { CellTypeSummary } from "../types";
 import { REDUCERS, State } from "./reducer";
 
-export function deleteSelectedGenesAndSelectedCellTypeIds({
-  genesToDelete,
-  cellTypeIdsToDelete,
-}: {
-  genesToDelete: string[];
-  cellTypeIdsToDelete: string[];
-}): GetActionTypeOfReducer<
+export function deleteSelectedGenesAndSelectedCellTypeIds(): GetActionTypeOfReducer<
   typeof REDUCERS["deleteSelectedGenesAndSelectedCellTypeIds"]
 > {
   return {
-    payload: {
-      cellTypeIdsToDelete,
-      genesToDelete,
-    },
+    payload: null,
     type: "deleteSelectedGenesAndSelectedCellTypeIds",
   };
 }
@@ -33,7 +26,7 @@ export function toggleGeneToDelete(
 }
 
 export function toggleCellTypeIdToDelete(
-  cellTypeIdToDelete: string
+  cellTypeIdToDelete: CellTypeMetadata
 ): GetActionTypeOfReducer<typeof REDUCERS["toggleCellTypeIdToDelete"]> {
   return {
     payload: cellTypeIdToDelete,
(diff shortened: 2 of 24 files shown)


frontend/.eslintrc.js
```diff
@@ -38,6 +38,15 @@ module.exports = {
...
     "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-unused-vars": [
+      "error",
+      {
+        args: "after-used",
+        argsIgnorePattern: "^_",
+        ignoreRestSiblings: false,
+        vars: "all",
+      },
+    ],
```
seve, line 49: This explicit rule is just to tweak an already active rule to ignore the pattern, correct?

Might be worth adding a comment with a few examples captured by the pattern

frontend/src/views/WheresMyGene/common/store/actions.ts
```diff
@@ -3,22 +3,15 @@
...
-}): GetActionTypeOfReducer<
+export function deleteSelectedGenesAndSelectedCellTypeIds(): GetActionTypeOfReducer<
   typeof REDUCERS["deleteSelectedGenesAndSelectedCellTypeIds"]
 > {
   return {
-    payload: {
-      cellTypeIdsToDelete,
-      genesToDelete,
-    },
+    payload: null,
```
seve, line 14: What was the payload, is now held in redux store, correct?

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -161,13 +175,12 @@ function toggleCellTypeIdToDelete(
...
 
-function deleteByIds<Item>(collection: Item[], collectionToDelete: Item[]) {
+function deleteByItems<Item>(collection: Item[], collectionToDelete: Item[]) {
   return collection.filter((item) => !collectionToDelete.includes(item));
 }
 
 function resetGenesToDeleteAndCellTypeIdsToDelete(
   state: State,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _: PayloadAction<null>
```
seve, line 184: I see now!

seve: Surprised this isn't already in the rule, is there a thread/issue on this?

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -1,7 +1,11 @@
 import { CSSProperties } from "react";
 
+export type Tissue = string;
+
 /** This object holds a cell type and its gene expressions. */
 export interface CellTypeSummary {
+  /** tissue name */
+  tissue: string;
```
seve, line 8:
```suggestion
  tissue: Tissue;
```
Right?

seve (commented):
Redux review 1st, looking at components next

frontend/src/views/WheresMyGene/common/store/reducer.ts
tihuan: Yeah I checked eslint doc and they want us to ignore `^_` via the config 🤔 

https://eslint.org/docs/rules/no-unused-vars#argsignorepattern

frontend/src/views/WheresMyGene/common/store/actions.ts
tihuan: That's right!

frontend/.eslintrc.js
tihuan: Great idea! Done 🔥

frontend/src/views/WheresMyGene/common/types.ts
tihuan: Oh thanks for catching this. I went ahead and update all types to use `Tissue` 🙏

seve (commented):
Couple more Q

frontend/src/views/WheresMyGene/components/HeatMap/hooks/useUpdateChart.ts
```diff
@@ -0,0 +1,62 @@
...
+      UPDATE_THROTTLE_MS,
+      // (thuang): Trailing guarantees that the last call to the function will
+      // be executed
+      { trailing: true }
+    );
+  }, []);
+
+  useEffect(() => {
+    return () => throttledUpdateChart.cancel();
+  }, [throttledUpdateChart]);
```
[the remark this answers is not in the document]
tihuan: we want to call `cancel()` in the useEffect clean up function, thus needing to return a function inside the useEffect callback: https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup

The goal is to ensure that throttle with trailing doesn't trigger a setState after the component is unmounted, thus presenting a memleak and react set state error!

seve (approved):

PR #1951, opened 2022-02-10
feat: added filter by publication metadata

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
- Added filter by publication date to collections and datasets.
- Added filter by publication author to collections and datasets.

### DIFF
--- a/frontend/src/common/entities.ts
+++ b/frontend/src/common/entities.ts
@@ -230,7 +230,7 @@ export interface GeneSet {
  * Collection publication metadata, populated from Crossref by collection publication DOI.
  */
 export interface PublisherMetadata {
-  authors: Author[];
+  authors: (Author | Consortium)[];
   journal: string;
   published_day: number;
   published_month: number;
--- a/frontend/src/common/entities.ts
+++ b/frontend/src/common/entities.ts
@@ -58,6 +58,15 @@ export enum IS_PRIMARY_DATA {
   BOTH = "BOTH",
 }
 
+/**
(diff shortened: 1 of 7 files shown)


frontend/src/common/hooks/useCategoryFilter.ts
```diff
@@ -166,6 +168,29 @@ function addEmptyCategoryValues(
   }
 }
 
+/**
```
MillenniumFalconMechanic, line 171: I added this function as a way to swap out the filter type per column. I initially anticipated using a `between` filter type for filter by publication date but ended up using the existing `includesSome`. I have kept the functionality as-is (that is, always using `includesSome`) but I expect we will need it - with modifications - for future filters such as cell count, gene count and possible development stage.

frontend/src/common/entities.ts
```diff
@@ -58,6 +58,15 @@ export enum IS_PRIMARY_DATA {
   BOTH = "BOTH",
 }
 
+/**
+ * Author of publication associated with a collection, populated from Crossref as part of collection publication
+ * metadata.
+ */
+export interface Author {
+  family: string;
```
tihuan, line 66: ooh I like this over first name last name, since different countries do first name last name differently! 💡

tihuan (approved):
Wow super amazing PR 😮 👏 Love your approach for filtering date by indexing it upfront 😆  Thanks so much for adding all the comments and tests too 🙏

MillenniumFalconMechanic:
Thank you for taking the time to review this PR @tihuan!

PR #1975, opened 2022-02-16
Add GET curator/collections

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/open_api_spec.yml
+++ b/backend/open_api_spec.yml
@@ -24,8 +24,62 @@ paths:
 
   /curation/collections:
     get:
-      description: Get all Collections metadata
+      description: Get all Collections metadata. Include authentication information to retrieve private collections as well. 
+      security:
+        - cxguserCookie: []
+      Parameters:
+        - $ref: "#/components/parameters/query_visibility"
       responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  access_type:
+                    type: string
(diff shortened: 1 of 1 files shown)


backend/open_api_spec.yml
```diff
@@ -212,16 +266,14 @@ components:
         - visibility
         - links
         - datasets
-        - created_at
-        - updated_at
```
danieljhegeman, line 216: don't we want to substitute in `revised_at` and `published_at`?

backend/open_api_spec.yml
```diff
@@ -24,8 +24,62 @@ paths:
 
   /curation/collections:
     get:
-      description: Get all Collections metadata
+      description: Get all Collections metadata. Include authentication information to retrieve private collections as well.
```
danieljhegeman, line 27: This doesn't square with the `security` mechanism. I thought we were restricting all curator api functionality to curators with access tokens.

danieljhegeman (approved):

backend/open_api_spec.yml
Bento007: that required field is only used for json that is in a request, not the response.

backend/open_api_spec.yml
Bento007: that required field is only used for json that is in a request, not the response. So we don't need to put revised and published here.

PR #1976, opened 2022-02-17
Adding API Token endpoints

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/open_api_spec.yml
+++ b/backend/open_api_spec.yml
@@ -4,24 +4,72 @@ servers:
 paths:
   /curation/auth/key:
     post:
-      summary: Regenerate the curator's API key
+      summary: Generate the curator's API key
       security: 
         - cxguserCookie: [] 
-		get:
-      summary: Return the identifier of the curator's active API key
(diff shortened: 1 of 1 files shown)


backend/open_api_spec.yml
```diff
@@ -4,24 +4,72 @@ servers:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                id:
+                  type: string
+      responses:
+        "202":
```
danieljhegeman, line 52: had meant to ask why you're quoting the error codes like `"<code>"`

danieljhegeman (approved):

backend/open_api_spec.yml
Bento007: Just following the spec https://swagger.io/docs/specification/describing-responses/

danieljhegeman: interesting, fair enough

PR #1981, opened 2022-02-18
feat: add dataset_preview schema

danieljhegeman:
### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/open_api_spec.yml
+++ b/backend/open_api_spec.yml
@@ -70,7 +70,7 @@ paths:
                   collections:
                     type: array
                     items:
-                      $ref: "#/components/schemas/collection"
+                      $ref: "#/components/schemas/collection_with_dataset_preview"
         "400":
           $ref: "#/components/responses/400"
         "401":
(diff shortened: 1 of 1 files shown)


backend/open_api_spec.yml
```diff
@@ -335,6 +335,57 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/gene_set"
+    collection_with_dataset_preview:
+      type: object
+      required:
```
Bento007, line 340: These are all fields in a response. You don't need required.

backend/open_api_spec.yml
```diff
@@ -410,6 +461,31 @@ components:
           $ref: "#/components/schemas/distribution"
         schema_version:
           type: string
+    dataset_preview:
+      type: object
+      required:
```
Bento007, line 466: This is only a response, you don't need required.

Bento007 (changes requested):
What Endpoint with these be used?

danieljhegeman:
outdated

Issue #1902, opened 2022-02-03
Specify publication DOI on create and update of collection


MillenniumFalconMechanic:
#### Approach
* Add updated functionality behind filter feature flag.

#### Definition of Done
* "DOI" displayed as "Publication DOI".
* Only single Publication DOI can be specified.
* Publication DOI field displays [placeholder](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=2738%3A29681).
* Validation moved to BE.

PR #1990, opened 2022-02-23
feat: added DOI curie to collection form. #1902

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan

---

## Changes
Updated collection create and edit form to handle a DOI CURIE instead of a complete URL. 

* "DOI" displayed as "Publication DOI", optional name field removed, helper text added, placeholder updated ([default state](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=2738%3A29681)).
* Only single Publication DOI can be specified.
* Validation moved to BE ([error state](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=2728%3A31307)).

## Known Issues
I believe we have modified the form functionality to match the requirements. We wanted to avoid a big refactor in this area and have sewn in the required functionality with hopefully minimal changes. I do however feel that the DOI CURIE has introduced complexity to the form implementation and that the form component hierarchy and architecture will need revisiting if future modifications to the form are required.


frontend/src/common/queries/collections.ts
```diff
@@ -407,28 +449,33 @@ export function useEditCollection(collectionID?: Collection["id"]) {
...
-            collection?.has_revision &&
-            publishedCollection
-          ) {
-            revision_diff = checkForRevisionChange(
-              newCollection,
+    onSuccess: ({ collection: newCollection }) => {
+      // Check for updated collection: it's possible server-side validation errors have occurred where the error has
+      // been swallowed (allowing error messages to be displayed on the edit form) and success flow is executed even
+      // though update did not occur.
+      if (newCollection) {
```
tihuan, line 456: Wow thanks so much for catching this bug! :pray: CC @seve for knowledge sharing!

Also for readability should we use a guard clause instead? This way we don't indent the whole code block one level deeper and no need to scan for the non-existing `else` block at the bottom :smile: 

```ts
if (!newCollection) return;

queryClient.setQueryData...
```

frontend/src/common/queries/collections.ts
```diff
@@ -502,3 +549,15 @@ export function useReuploadDataset(collectionId: string) {
...
+/**
+ * Determine if a submitted DOI has failed validation on the BE. Expected response for invalid DOI:
+ * {"detail": "DOI cannot be found on Crossref", "status": 400, "title": "Bad Request", "type": "about:blank"}
+ * TODO generalize beyond DOI link type once all links are validated on the BE (#1916).
+ * @param status - Response status returned from server.
+ * @param detail - Response error text, if any.
+ * @returns True if DOI has been identified as invalid by the BE.
+ */
+function isInvalidDOI(status: number, detail?: string): boolean {
+  return status === 400 && detail === INVALID_DOI_MESSAGE;
```
tihuan, line 562: We have status enums!

https://github.com/chanzuckerberg/single-cell-data-portal/blob/b29ac1d746bd414ac44fe1959b52c97061eb4401/frontend/src/common/constants/HTTP_STATUS_CODE.ts

frontend/src/components/CreateCollectionModal/components/Content/components/LinkInput/index.tsx
```diff
@@ -187,6 +280,49 @@ export function isDOILink(
   };
 }
 
+/**
+ * Determine if validation of field needs to be executed on change on link type.  This is only true if:
```
tihuan, line 284: Thanks for listing out all the possible conditions 🙏 So easy to follow!

tihuan (approved):
Looks super good! Thanks so much adding server side DOI validation here :tada: So neat!

frontend/src/common/queries/collections.ts
MillenniumFalconMechanic: Guard cause is much cleaner; I'll make the update. Thanks Timmy!

frontend/src/common/queries/collections.ts
MillenniumFalconMechanic: Got it, thanks Timmy!

PR #1991, opened 2022-02-24
feat(wmg): Incremental debounce

tihuan:
Instead of hard coding a set debounce time, we apply a sliding scale depending on how many genes are selected. So the long debounce time is only applied on large gene selection. This way for small amount of genes, we get snappy interaction like the users expect!

demo


seve (approved):
🎉

PR #1994, opened 2022-02-24
curation api spec

Bento007:
### Reviewers

**Readability:** 

---

## Changes
- add empty security for fields that can be accessed with or without authentication.
- add securitySchema curatorAccess to allow curators to pass their access token in a header field. This is easier to do than passing as a cookie.
- remove cxguserCookieLenient because it is not used.
- remove ApiKey from Post /token since we are passing the API key in the request body and don't want connexion to automagically try to handle it.

### DIFF
--- a/backend/open_api_spec.yml
+++ b/backend/open_api_spec.yml
@@ -75,7 +75,8 @@ paths:
       summary: Fetch Collections metadata.
       description: Fetch all Collections metadata. Authorization required for private collections.
       security:
-        - cxg
+        - cxguserCookie: []
+        - {}
       parameters:
         - $ref: "#/components/parameters/query_visibility"
       responses:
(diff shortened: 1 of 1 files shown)


backend/open_api_spec.yml
```diff
@@ -77,7 +75,8 @@ paths:
       summary: Fetch Collections metadata.
       description: Fetch all Collections metadata. Authorization required for private collections.
       security:
-        - cxguserCookieLenient: []
+        - cxguserCookie: []
```
danieljhegeman, line 78: why not also `curatorAccess`?

Bento007: It should actually be all three

danieljhegeman (approved):

PR #1995, opened 2022-02-24
chore: add publisher_metadata to API schema

ebezzi:
### Reviewers
**Functional:** 
@danieljhegeman 

**Readability:** 

---

## Changes
- Add the new `publisher_metadata` field to the API schema.

### DIFF
--- a/backend/config/corpora-api.yml
+++ b/backend/config/corpora-api.yml
@@ -548,7 +548,13 @@ paths:
                   properties:
                     id:
                       $ref: "#/components/schemas/collection_uuid"
-                    created_at:
+                    name: 
+                      type: string
+                    publisher_metadata:
+                      $ref: "#/components/schemas/publisher_metadata"
+                    published_at:
+                      type: number
+                    revised_at:
                       type: number
         "400":
           $ref: "#/components/responses/400"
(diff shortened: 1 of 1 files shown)


backend/config/corpora-api.yml
```diff
@@ -548,7 +548,13 @@ paths:
...
                     id:
                       $ref: "#/components/schemas/collection_uuid"
-                    created_at:
+                    name: 
+                      type: string
+                    publisher_metadata:
+                      $ref: "#/components/schemas/publisher_metadata"
+                    published_at:
+                      type: number
+                    revised_at:
```
danieljhegeman, line 557: why remove `created_at`?

danieljhegeman: FE no longer uses, I assume

danieljhegeman (approved):

danieljhegeman (approved):

backend/config/corpora-api.yml
ebezzi: I think that was a mistake in the first place, `created_at` isn't among the returned fields by this endpoint, so I just removed it.

ebezzi:
Needs https://github.com/chanzuckerberg/single-cell-data-portal/pull/2203 in order for tests to pass

PR #2004, opened 2022-02-25
chore: refresh preprint doi script

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@metakuni 

---

## Changes
- Add a script that checks, for all the collections that are preprints, if a published DOI exists, and if so, updates it
- Note: since there are currently 0 preprints with a published DOI, the script is untested. My recommendation is to dry run it and then test the update part in dev/staging before moving to prod.


MDunitz (approved):
lgtm

PR #2018, opened 2022-03-02
fix(wmg): Fix SDS types

tihuan:
This PR will address new type errors when we upgrade to the latest SDS version (PR not merged yet)


tihuan:
Doesn't seem needed anymore!

PR #2020, opened 2022-03-02
chore: bypass content length for dropbox downloads

ebezzi:
### Reviewers
**Functional:** 
@metakuni @atolopko-czi 

**Readability:** 

---

## Changes
- Bypass content length for dropbox downloads
- Sets a default of 100MB.

### DIFF
--- a/backend/corpora/common/utils/dl_sources/url.py
+++ b/backend/corpora/common/utils/dl_sources/url.py
@@ -3,6 +3,9 @@ import typing
 from abc import ABC, abstractmethod
 from urllib.parse import urlparse, ParseResult
 
+import logging
+logger = logging.getLogger(__name__)
+
 
 class MissingHeaderException(Exception):
     def __init__(self, detail: str = "", *args, **kwargs) -> None:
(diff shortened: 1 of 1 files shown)


backend/corpora/common/utils/dl_sources/url.py
```diff
@@ -82,10 +85,16 @@ def file_info(self) -> dict:
         resp = requests.head(self.url, allow_redirects=True)
         resp.raise_for_status()
 
-        print(resp.headers)
+        # Important: getting the content-length from Dropbox is unreliable, so we're setting a default value (100MB)
+        # to avoid breaking the upload pipeline.
+        try:
+            size = int(self._get_key_with_fallback(resp.headers, "content-length", "x-dropbox-content-length"))
```
atolopko-czi, line 91: just checking that it is the processing of the response header, rather than the request itself, that is failing?

atolopko-czi (approved):
A sanity check question for you, but code-wise LGTM, so approving

backend/corpora/common/utils/dl_sources/url.py
ebezzi: The request itself works. It's a HEAD request to the download URL, so if it fails, the download won't work either, so I think it makes sense to fail fast here.

backend/corpora/common/utils/dl_sources/url.py
```diff
@@ -82,10 +85,16 @@ def file_info(self) -> dict:
...
+        except Exception:
+            logger.warning("Cannot retrieve the content length from Dropbox.")
+            size = 1e8
 
         return {
-            "size": int(self._get_key_with_fallback(resp.headers, "content-length", "x-dropbox-content-length")),
+            "size": size,
             "name": self._get_key(resp.headers, "content-disposition").split(";")[1].split("=", 1)[1][1:-1],
```
metakuni, line 99: Can the `content-disposition` key be trusted to exist? 😅

metakuni (approved):

PR #2048, opened 2022-03-04
WMG processing infra

ebezzi:


requirements.txt
```diff
@@ -14,7 +14,6 @@ numpy==1.22.2
 owlready2
 pandas<1.3.0
 pyarrow>=1.0
-psutil
```
ebezzi, line 17: This is creating a conflict when creating the makefile. Not quite sure why this works with the current images.

.happy/terraform/modules/wmg-batch/main.tf
```diff
@@ -0,0 +1,53 @@
+# This is a batch job
```
ebezzi, line 1: An alternative to creating this batch job is to parameterized the one that we already have. In general, I think this solution is better, since they're two distinct jobs (this one will require less variables, for instance). Code duplication is quite unavoidable with Terraform anyway.

We should probably modify the other module to be called `upload-batch` though.

.happy/terraform/modules/wmg-batch/main.tf
```diff
@@ -0,0 +1,53 @@
...
+  type = "container"
+  name = "dp-${var.deployment_stage}-${var.custom_stack_name}-wmg-processing"
+  container_properties = <<EOF
+{
+  "jobRoleArn": "${var.batch_role_arn}",
+  "image": "${var.image}",
+  "memory": ${var.batch_container_memory_limit},
+  "environment": [
+    {
+      "name": "ARTIFACT_BUCKET",
```
ebezzi, line 18: These two buckets are probably not needed. We should replace these with a new variable that points to a stable bucket that contains the WMG cube.

.happy/terraform/modules/wmg-batch/main.tf
MDunitz: I think keeping them as 2 separate jobs makes sense to me. Agree re renaming the other job

.happy/terraform/modules/wmg-batch/main.tf
MDunitz: I'm not sure I understand. We will need access to the artifact bucket to get the datasets that make up the cube. From the tech spec it looks like we need to create another bucket s3://wmg-$ENV

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -162,6 +163,38 @@ module upload_batch {
   batch_container_memory_limit = local.batch_container_memory_limit
 }
 
+# TODO: arguments to the upload-cube job go here
```
MDunitz, line 166: remove?

MDunitz (approved):
lgtm

.happy/terraform/modules/wmg-batch/main.tf
ebezzi: Ok, I'll keep it then and add another variable for the WMG bucket. What about the cellxgene bucket?

.happy/terraform/modules/ecs-stack/main.tf
ebezzi: We'll need to register the other variables (namely the WMG bucket). I'll add that to the PR.

PR #2061, opened 2022-03-06
feat: migration - use unique id for Collections

danieljhegeman:
This migration abandons the composite primary key (id, visibility)
approach for the project / Collections table and enforces a unique id.
Going forward, revisions will be identified by their unique id rather
than by the id of their public-facing Collection coupled with
visibility=PRIVATE.

At the time of this commit, there are no revisions in progress in
the corpora_prod database.

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 
@atolopko-czi @metakuni 
---

## Changes
- add `project.revision_of` so that all Collections may have unique `id`s
- remove all instances of composite keys formed by collection id and visibility.

## QA steps (optional)
- Passes `make db/test_migration` test _when the `project.id_bk` backup column is excluded._
## Known Issues
- Downgrading from this migration will produce an extra column `project.id_bk` to save unique revision `project.id` values (where `project.revision_of` was not null). This is just a safety measure.


backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
```diff
@@ -0,0 +1,126 @@
...
+    op.drop_constraint(op.f("fk_geneset_collection_id_project"), "geneset", type_="foreignkey")
+    op.drop_constraint(op.f("fk_dataset_collection_id_project"), "dataset", type_="foreignkey")
+
+    # Drop project.id as primary key
+    op.drop_constraint(op.f("pk_project"), "project")
+
+    # Safety: save current project.id in extra column as backup -- BREAKS MIGRATION FLOW
+    op.add_column("project", sa.Column("id_bk", sa.String()))
+    op.execute("UPDATE project SET id_bk = id")
+    op.alter_column("project", "id_bk", nullable=False)
```
danieljhegeman, line 96: Is this step of backing up the unique id's necessary? I've added it in case of a rogue/accidental migration downgrade.

danieljhegeman:
@Bento007 @atolopko-czi @metakuni I've ditched the [previous PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2023) since making just the single column addition without all of the cascading changes didn't really make much sense.

backend/corpora/common/corpora_orm.py
```diff
@@ -479,5 +478,5 @@ class DbGenesetDatasetLink(Base, AuditMixin):
 
     __tablename__ = "geneset_dataset_link"
 
-    geneset_id = Column(String, ForeignKey("geneset.id"), index=True)
-    dataset_id = Column(String, ForeignKey("dataset.id"), index=True)
+    geneset_id = Column(String, ForeignKey("geneset.id"), index=True, nullable=False)
+    dataset_id = Column(String, ForeignKey("dataset.id"), index=True, nullable=False)
```
danieljhegeman, line 482: unrelated change; reflects state of remote db's

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
```diff
@@ -0,0 +1,126 @@
+"""make_collection_id_pkey
+
+Revision ID: e08b6b67f076
```
metakuni, line 3:
```suggestion
Revision ID: 25_e08b6b67f076
```

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
atolopko-czi: TLDR: honorable engineering, but probably not worth the effort.

If we have to downgrade a prod migration, I would say all bets are off as to the permanence of any private collection URLs users may have been acquired by users. And I think this would only affect private revisions that were curated after the "up" migration (since no revisions can be in progress at that point).  So it's a very limited blast radius for this problem, even if it takes a couple of days to catch a problem.

backend/corpora/common/corpora_orm.py
```diff
@@ -228,7 +229,8 @@ class DbCollection(Base, AuditMixin, TimestampMixin):
     # the tablename is "project" instead of "collection" to avoid migrating the database
     __tablename__ = "project"
 
-    visibility = Column(Enum(CollectionVisibility), primary_key=True, nullable=False)
+    id = Column(String, primary_key=True, default=generate_uuid)
```
Bento007, line 232: rm, this key is defined in Base.
```suggestion
```

backend/corpora/common/corpora_orm.py
```diff
@@ -456,13 +456,12 @@ class DbGeneset(Base, AuditMixin):
...
     genes = Column(JSONB)
     collection_id = Column(String, nullable=False)
-    collection_visibility = Column(Enum(CollectionVisibility), nullable=False)
     collection = relationship("DbCollection", uselist=False, back_populates="genesets")
     datasets = relationship("DbDataset", secondary="geneset_dataset_link", back_populates="genesets")
 
     __table_args__ = (
-        ForeignKeyConstraint([collection_id, collection_visibility], [DbCollection.id, DbCollection.visibility]),
-        UniqueConstraint("name", "collection_id", "collection_visibility", name="_geneset_name__collection_uc"),
+        ForeignKeyConstraint((collection_id,), (DbCollection.id,)),
```
Bento007, line 463:
```suggestion
```

backend/corpora/common/corpora_orm.py
```diff
@@ -456,13 +456,12 @@ class DbGeneset(Base, AuditMixin):
     description = Column(String)
     genes = Column(JSONB)
     collection_id = Column(String, nullable=False)
```
Bento007, line 458:
```suggestion
    collection_id = Column(String, nullable=False, ForeignKey("DbCollection.id"))
```

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
```diff
@@ -0,0 +1,126 @@
...
+    op.drop_column("geneset", "collection_visibility")
+    # Recreate unique key without collection_visibility
+    op.create_unique_constraint(op.f("_geneset_name__collection_uc"), "geneset", ["name", "collection_id"])
+
+    # Add project.revision_of
+    op.add_column("project", sa.Column("revision_of", sa.String(), nullable=True))
+
+    # Drop project composite primary key; add project.id as primary key
+    op.drop_constraint("project_pkey", "project")
+    op.create_primary_key(op.f("pk_project"), "project", ["id"])
```
Bento007, line 39: Does `["id"]` need to be a list?

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Add foreign keys for project.id
+    op.create_foreign_key(op.f("fk_dataset_collection_id_project"), "dataset", "project", ["collection_id"], ["id"])
```
Bento007, line 42: Is it required to have so many lists with a single value?

Bento007 (changes requested):

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
danieljhegeman: Ok, will remove.

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
```diff
@@ -0,0 +1,126 @@
...
+    op.add_column("project", sa.Column("id_bk", sa.String()))
+    op.execute("UPDATE project SET id_bk = id")
+    op.alter_column("project", "id_bk", nullable=False)
+
+    # Reassign id for revisions
+    op.execute(
+        """
+        UPDATE project
+            SET id = revision_of
+        WHERE revision_of IS NOT NULL
```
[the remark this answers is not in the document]
danieljhegeman: yes! good catch

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
danieljhegeman: indeed, per `columns: List[str]`

backend/database/versions/26_e08b6b67f076_make_collection_id_pkey.py
danieljhegeman: yeah it's the way this function accepts column arguments in order to accommodate composites

backend/corpora/common/corpora_orm.py
danieljhegeman: have to have here because otherwise Python sees `id` as the builtin function, not an attribute for this class. Super annoying. But I couldn't come up with another solution.

danieljhegeman: I should add a comment for this line though

backend/corpora/common/corpora_orm.py
danieljhegeman: This would change the expected behavior from before. Could you explain why you want this removed?

Bento007: this changes goes with https://github.com/chanzuckerberg/single-cell-data-portal/pull/2061#discussion_r821191412

Bento007: You can put this inline with `revision_of`

danieljhegeman: ah got it, with the comment above 👍

backend/corpora/common/corpora_orm.py
```diff
@@ -239,12 +240,16 @@ class DbCollection(Base, AuditMixin, TimestampMixin):
     data_submission_policy_version = Column(StrippedString, nullable=True)
     tombstone = Column(Boolean, default=False, nullable=False)
     publisher_metadata = Column(JSON, nullable=True)
+    revision_of = Column(String, ForeignKey(id), nullable=True)
 
     # Relationships
+    revision = relationship("DbCollection", cascade="all, delete-orphan", uselist=False)
```
danieljhegeman, line 246: This conflicts with `revision` method in the `Collection` class. I have moved `Collection.revision` method to `Collection.create_revision`. Separate PR.

backend/corpora/common/corpora_orm.py
```diff
@@ -239,12 +240,16 @@ class DbCollection(Base, AuditMixin, TimestampMixin):
     data_submission_policy_version = Column(StrippedString, nullable=True)
     tombstone = Column(Boolean, default=False, nullable=False)
     publisher_metadata = Column(JSON, nullable=True)
+    revision_of = Column(String, ForeignKey(id), nullable=True)
```
Bento007, line 243:
```suggestion
    revision_of = Column(String, ForeignKey("project.id"), nullable=True)
```
then you can remove the `id` definition

backend/corpora/common/corpora_orm.py
```diff
@@ -254,21 +259,14 @@ class DbProjectLink(Base, AuditMixin):
     # the tablename is "project_link" instead of "collection_link" to avoid migrating the database
     __tablename__ = "project_link"
 
-    collection_id = Column(String, nullable=False)
-    collection_visibility = Column(Enum(CollectionVisibility), nullable=False)
+    collection_id = Column(String, ForeignKey("DbCollection.id"), nullable=False)
```
Bento007, line 262:
```suggestion
    collection_id = Column(String, ForeignKey("project.id"), nullable=False)
```
 or 
```suggestion
    collection_id = Column(String, ForeignKey(DbCollection.id), nullable=False)
```

backend/corpora/common/corpora_orm.py
```diff
@@ -455,15 +446,11 @@ class DbGeneset(Base, AuditMixin):
     name = Column(String, nullable=False)
     description = Column(String)
     genes = Column(JSONB)
-    collection_id = Column(String, nullable=False)
-    collection_visibility = Column(Enum(CollectionVisibility), nullable=False)
+    collection_id = Column(String, ForeignKey("DbCollection.id"), nullable=False)
```
Bento007, line 449:
```suggestion
    collection_id = Column(String, ForeignKey("project.id"), nullable=False)
```
 or 
```suggestion
    collection_id = Column(String, ForeignKey(DbCollection.id), nullable=False)
```

backend/corpora/common/corpora_orm.py
```diff
@@ -296,8 +294,7 @@ class DbDataset(Base, AuditMixin, TimestampMixin):
     cell_count = Column(Integer)
     is_valid = Column(Boolean, default=False)
     is_primary_data = Column(Enum(IsPrimaryData))
-    collection_id = Column(String, nullable=False)
-    collection_visibility = Column(Enum(CollectionVisibility), nullable=False)
+    collection_id = Column(String, ForeignKey("DbCollection.id"), nullable=False)
```
Bento007, line 297:
```suggestion
    collection_id = Column(String, ForeignKey("project.id"), nullable=False)
```
 or 
```suggestion
    collection_id = Column(String, ForeignKey(DbCollection.id), nullable=False)
```

backend/corpora/common/corpora_orm.py
danieljhegeman: ah, clever

backend/corpora/common/corpora_orm.py
```diff
@@ -239,12 +239,16 @@ class DbCollection(Base, AuditMixin, TimestampMixin):
     data_submission_policy_version = Column(StrippedString, nullable=True)
     tombstone = Column(Boolean, default=False, nullable=False)
     publisher_metadata = Column(JSON, nullable=True)
+    revision_of = Column(String, ForeignKey("project.id"), nullable=True, unique=True)
```
danieljhegeman, line 242: added `unique=True`

Bento007:
The migration looks good

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -172,7 +172,7 @@ def build_cell_orderings(cell_counts_cube_dir_, cell_ordering_generator_fn) -> D
 def create_dataset(dataset_id_ordinal: int) -> str:
     coll_id = f"dataset_id_{dataset_id_ordinal}_coll_id"
     with db_session_manager() as session:
-        if coll := Collection.get(session, (coll_id, CollectionVisibility.PUBLIC)):
+        if coll := Collection.get(session, coll_id):
```
danieljhegeman, line 175: @atolopko-czi pls confirm this is an acceptable change; the tests pass, just want to make sure there isn't a hidden requirement to verify that a collection is public at this line

tests/unit/processing_container/test_process.py
```diff
@@ -63,9 +60,7 @@ def test_main(self, mock_download_from_dropbox):
         """
         mock_download_from_dropbox.return_value = self.h5ad_raw
 
-        dataset = self.generate_dataset(
-            self.session, collection_id="test_collection_id", collection_visibility=CollectionVisibility.PUBLIC.name
-        )
+        dataset = self.generate_dataset(self.session, collection_id="test_collection_id")
```
danieljhegeman, line 63: lint fix

backend/wmg/data/extract.py
```diff
@@ -28,11 +28,12 @@ def get_dataset_s3_uris():
         dataset_ids = []
         published_dataset_non_null_assays = (
             session.query(Dataset.table.id, Dataset.table.assay)
+            .join(Dataset.table.collection)
             .filter(
                 Dataset.table.assay != "null",
                 Dataset.table.published == "TRUE",
                 Dataset.table.is_primary_data == "PRIMARY",
-                Dataset.table.collection_visibility == "PUBLIC",
+                Collection.table.visibility == "PUBLIC",
```
danieljhegeman, line 36: @atolopko-czi more wmg-related changes (tests pass, but pls confirm 🆗 )

tests/unit/backend/wmg/fixtures/test_snapshot.py
atolopko-czi: This is good--dataset visibility is not a concern of any wmg API tests.

backend/wmg/data/extract.py
atolopko-czi: 🆗

backend/corpora/common/entities/collection.py
```diff
@@ -254,19 +326,19 @@ def tombstone_collection(self):
             ds.tombstone_dataset_and_delete_child_objects()
         self.session.commit()
 
-    def update(self, links: list = None, **kwargs) -> None:
+    def update(self, links: list = None, keep_links=None, **kwargs) -> None:
```
metakuni, line 329: Should `keep_links` have a default of `False` instead of none?
Also need to add the `keep_links` param in the docstring below.

Bento007, line 329: How about this:

```
def update(self, links: list = None, **kwargs) -> None:
        if isinstance(links, list):
            for link in self.links:
                self.session.delete(link)
       new_objs = [DbCollectionLink(collection_id=self.id, **link) for link in links]

File: backend/corpora/common/entities/geneset.py [hunk 20]
```diff
@@ -97,3 +97,18 @@ def update_links_for_a_dataset(
         except SQLAlchemyError:
             session.rollback()
             raise CorporaException
+
+
+class CollectionLink(Entity):
```
Bento007, line 102: move to collection.py or its own file.

backend/corpora/common/entities/collection.py
```diff
@@ -174,74 +244,76 @@ def publish(self, data_submission_policy_version):
...
         )
         self.session.add(revision_collection)
         for link in self.links:
-            self.session.add(clone(link, collection_id=self.id, collection_visibility=CollectionVisibility.PRIVATE))
-        self.session.commit()
+            self.session.add(clone(link, collection_id=revision_collection.id))
         for dataset in self.datasets:
-            Dataset(dataset).create_revision()
+            Dataset(dataset).create_revision(revision_collection.id)
+        self.session.commit()  # why is this happening here and not after creating the Dataset revisions
```
Bento007, line 316: rm b/c `Dataset(dataset).create_revision(revision_collection.id)` already calls commit.

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -197,42 +218,27 @@ def get_collection_dataset(dataset_uuid: str):
...
+        if collection.visibility == CollectionVisibility.PUBLIC:
+            revision = Collection.get_collection(
+                db_session,
+                revision_of=collection_uuid,
+                owner=_owner_or_allowed(user),
+            )
+            if revision:
+                revision.delete()
+            collection.tombstone_collection()
             return "", 204
```
Bento007, line 239:
```suggestion
```

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -197,42 +218,27 @@ def get_collection_dataset(dataset_uuid: str):
...
-            CollectionVisibility.PRIVATE.name,
-            owner=_owner_or_allowed(user),
-            include_tombstones=True,
-        )
-        if priv_collection:
-            if not priv_collection.tombstone:
-                priv_collection.delete()
+        else:
+            collection.delete()
             return "", 204
```
Bento007, line 242:
```suggestion
        return "", 204
```

backend/corpora/lambdas/api/v1/collection_uuid/publish.py
```diff
@@ -24,8 +24,8 @@ def post(collection_uuid: str, body: object, user: str):
         raise ForbiddenHTTPException()
     if all([dataset.tombstone for dataset in collection.datasets]):
         raise ConflictException(detail="The collection must have a least one dataset.")
-
+    collection_id = collection.revision_of if collection.revision_of else collection.id
```
Bento007, line 27: We are reusing this logic in a number of places. We should move it into its own function to simplify our lives and reduce the cognitive load on our brains.

backend/database/versions/29_e08b6b67f076_make_collection_id_pkey.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+    # Remove project.revision_of
+    op.drop_constraint(op.f("ck_project_private_revision"), "project", type_="check")
+    op.drop_column("project", "revision_of")
+
+    # Add project composite primary key
+    op.create_primary_key("project_pkey", "project", ["id", "visibility"])
+
+    # op.drop_index(op.f('ix_geneset_dataset_link_geneset_id'), table_name='geneset_dataset_link')
```
Bento007, line 109: rm

Bento007 (changes requested):
Almost there.

frontend/src/common/constants/routes.ts
```diff
@@ -4,7 +4,6 @@ export enum ROUTES {
   COLLECTION = "/collections/:id",
   COLLECTIONS = "/collections",
   DATASETS = "/datasets",
-  PRIVATE_COLLECTION = "/collections/:id/private",
```
metakuni, line 7: We should probably get someone from the FE team to sanity check the `.ts` changes.

backend/database/versions/29_e08b6b67f076_make_collection_id_pkey.py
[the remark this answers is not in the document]
danieljhegeman: Thanks for pointing this out! We don't need it, will remove the commented lines.

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -63,10 +82,12 @@ def get_collection(db_session, collection_uuid, visibility, **kwargs):
...
 
 @dbconnect
-def get_collection_details(collection_uuid: str, visibility: str, user: str):
+def get_collection_details(collection_uuid: str, user: str):
     db_session = g.db_session
-    collection = get_collection(db_session, collection_uuid, visibility, include_tombstones=True)
-    if collection.tombstone and visibility == CollectionVisibility.PUBLIC.name:
+    collection = Collection.get_collection(db_session, collection_uuid, include_tombstones=True)
+    if not collection:
+        raise ForbiddenHTTPException()
```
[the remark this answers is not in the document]
danieljhegeman: No, because a user should only be told 1) yes, this is a collection you are authorized to view, or 2) you are not authorized to view this, and there may or may not be a collection with that uuid

backend/corpora/common/entities/collection.py
danieljhegeman: I blame this on late-night coding a month ago.

backend/corpora/common/entities/collection.py
```diff
@@ -174,74 +244,76 @@ def publish(self, data_submission_policy_version):
...
-        for link in self.links:
-            link.collection_visibility = CollectionVisibility.PUBLIC
 
         has_dataset_changes = False
         for dataset in self.datasets:
-            revision = Dataset(dataset)
-            original = Dataset.get(self.session, revision.original_id) if revision.original_id else None
+            revised_dataset = Dataset(dataset)
+            original = Dataset.get(self.session, revised_dataset.original_id) if revised_dataset.original_id else None
             if original and public_collection.check_has_dataset(original):
```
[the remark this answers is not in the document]
danieljhegeman: Correct, but it is also never referenced unless `original` exists, in which case we didn't enter the `else` on ln 265.

backend/corpora/common/entities/collection.py
```diff
@@ -131,6 +150,57 @@ def to_dict(db_object):
 
         return results
 
+    @classmethod
+    def list_public_datasets_for_index(cls, session: Session) -> typing.List[typing.Dict]:
```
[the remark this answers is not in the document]
danieljhegeman: I was moving the class method for getting datasets from [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/a967b3cc157987c6feecf305f077209a9485d470/backend/corpora/common/entities/dataset.py#L164) where it already had that suffix.

It refers to the 'index' route, which ends up calling this method at some point.

backend/corpora/lambdas/api/v1/collection_uuid/publish.py
danieljhegeman: Concern noted but I only count two ✌️ at the moment

backend/corpora/common/entities/collection.py
danieljhegeman: Is that bad practice? Asking honestly 🤔

Bento007: it is bad practice to be using commits embedded in lower functions. We want to keep the transactions as tight as possible. At most, we should be doing a flush, but I was young and naive when I originally implemented it. If there is nothing to commit sqlalchemy will thrown an exception.

Bento007:
Our dev and staging environments all have private collections and revisions in progress. What will the migration do when it runs into these?

Bento007:
I ran migration in rdev and all of the collections were delete 

```
2022-04-11T23:20:25.601000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb # The create db script exits with a 1 status code if the db already exists.
2022-04-11T23:20:25.604000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb Creating DB...
2022-04-11T23:20:26.520000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb Database already exists
2022-04-11T23:20:26.630000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb make: [Makefile:15: db/init_remote_dev] Error 1 (ignored)
2022-04-11T23:20:26.630000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb PYTHONPATH=.. alembic -x db=rdev -c=./config/database.ini upgrade head
2022-04-11T23:20:27.569000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
2022-04-11T23:20:27.569000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb INFO  [alembic.runtime.migration] Will assume transactional DDL.
2022-04-11T23:20:27.592000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb INFO  [alembic.runtime.migration] Running upgrade 28_d7976d87cc2f -> 29_e08b6b67f076, make_collection_id_pkey
2022-04-11T23:20:27.732000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb make: Leaving directory '/corpora-data-portal/backend'
```

backend/corpora/common/entities/collection.py
```diff
@@ -174,74 +244,76 @@ def publish(self, data_submission_policy_version):
...
                     has_dataset_changes = True
             else:
                 # The dataset is new
-                revision.publish_new(now)
+                revised_dataset.publish_new(now)
                 has_dataset_changes = True
 
-        if is_existing_collection and has_dataset_changes:
-            public_collection.update(commit=False, remove_attr="revised_at", revised_at=now)
+        self.session.flush()
```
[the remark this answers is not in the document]
danieljhegeman: Otherwise the subsequent `delete` on ln 294 produces incorrect results.

backend/corpora/common/entities/collection.py
danieljhegeman: Right, I agree that we shouldn't have the commits in the lower functions, but those were already there, and I wasn't interested in doing a refactor for that work, which would be much more involved.

backend/corpora/common/entities/collection.py
```diff
@@ -174,74 +244,76 @@ def publish(self, data_submission_policy_version):
...
+                remove_relationships=True,
             )
             revision["data_submission_policy_version"] = data_submission_policy_version
             public_collection.update(
                 commit=False,
                 **revision,
             )
+            for link in self.links:
+                CollectionLink(link).update(collection_id=self.revision_of, commit=False)
+            public_collection.links = self.links
```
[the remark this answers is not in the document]
danieljhegeman: Yes. It reconstructs the 'entity' class relationships, whereas the update on ln 256 handles the db table relationships.

backend/corpora/common/entities/collection.py
danieljhegeman: I think this would require more work to make sure that in cases where we want to keep the links as they are (without modifying timestamps), then we'd have to make sure to remove `links` from the arguments that are passed in.

danieljhegeman:
> I ran migration in rdev and all of the collections were delete
> 
> ```
> 2022-04-11T23:20:25.601000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb # The create db script exits with a 1 status code if the db already exists.
> 2022-04-11T23:20:25.604000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb Creating DB...
> 2022-04-11T23:20:26.520000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb Database already exists
> 2022-04-11T23:20:26.630000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb make: [Makefile:15: db/init_remote_dev] Error 1 (ignored)
> 2022-04-11T23:20:26.630000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb PYTHONPATH=.. alembic -x db=rdev -c=./config/database.ini upgrade head
> 2022-04-11T23:20:27.569000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
> 2022-04-11T23:20:27.569000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb INFO  [alembic.runtime.migration] Will assume transactional DDL.
> 2022-04-11T23:20:27.592000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb INFO  [alembic.runtime.migration] Running upgrade 28_d7976d87cc2f -> 29_e08b6b67f076, make_collection_id_pkey
> 2022-04-11T23:20:27.732000+00:00 ed065fd428e78045c5b7745f6c96aba2dac1c0e3ab85d792027ce0b75d4d5feb make: Leaving directory '/corpora-data-portal/backend'
> ```

I'm looking into this. That did not happen in my hands previously when I ran the migration on a local instance.

backend/corpora/common/entities/collection.py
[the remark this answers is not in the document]
danieljhegeman: It is a suboptimal way to get around a circular import issue. Trying to figure out the right way to do package imports so that I can move it back...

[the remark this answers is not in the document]
danieljhegeman: Alright I think I got the import issue figured out so that I can move this code back into dataset.py

backend/corpora/common/entities/geneset.py
metakuni: Don't forget 👆

metakuni (approved):
2 things remaining:
* move `CollectionLink` out of `geneset.py` (@Bento007 's suggestion)
* should we not get an FE eng to sanity check the `.ts` changes?

danieljhegeman:
> 2 things remaining:
> 
> * move `CollectionLink` out of `geneset.py` (@Bento007 's suggestion)
Yep I'm addressing two remaining issues from Trent, including the circular import problem
> * should we not get an FE eng to sanity check the `.ts` changes?
Seve contributed to this PR and signed off on the FE changes

PR #2047, opened 2022-03-04
feat: wmg query endpoint impl with filter dim datasets metadata

atolopko-czi:
Part 4 of 4 of [story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1959).

Branched off of https://github.com/chanzuckerberg/single-cell-data-portal/pull/2045

### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@tihuan 

---

## Changes
* wmg query endpoint response now includes dataset metadata in the `filter_dimensions` section
* API spec changes (relative to prev pr):
    * `response_option` (enum) -> `include_filter_dims` (bool_)
    * `query` endpoint response: `filter_dimensions.datasets.collection_url` -> `collection_id`

PR #2045, opened 2022-03-04
feat: wmg query endpoint impl with filter dim terms

atolopko-czi:
Part 3 of 4 of [story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1959).

Branched off of https://github.com/chanzuckerberg/single-cell-data-portal/pull/2008

### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@tihuan 
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
* wmg query endpoint response now includes `filter_dimensions` section (except for `dataset` subsection, to be implemented next).

PR #2008, opened 2022-02-28
feat: wmg query endpoint impl with ontology term labels

atolopko-czi:
Part 2 of 4 of [story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1959).

Branched off of https://github.com/chanzuckerberg/single-cell-data-portal/pull/1984

### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@tihuan 

---

## Changes
* Return real ontology labels in WMG API query endpoint response
* Add cellxgene-schema==2.1.0 package as , as this is now used for obtaining ontology labels from the "system of record" (that repo)
* update test cube generation code to support of semi-real ontology terms (i.e. actual ontology terms, but not always ones that would be admissable w.r.t. to the cellxgene dataset schema).
  

## QA steps (optional)
* Unit tests
* Manually validate `query` endpoint response using local Docker with generated test cube, which now uses real ontology terms:
```
❯ curl -X POST "http://backend.corporanet.local:5000/wmg/v1/query" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"filter\":{\"organism_ontology_term_id\":\"NCBITaxon:10001\",\"tissue_ontology_term_ids\":[\"UBERON:0000002\", \"UBERON:0000003\"], \"gene_ontology_term_ids\": [\"ENSG00000000419\", \"ENSG00000000005\"]},\"response_option\":\"include_filter_dims_include_dataset_links\"}" | jq .

{
  "expression_summary": {
    "ENSG00000000005": {
      "UBERON:0000002": [
        {
          "id": "CL:0000000",
          "me": 0.01043697834432073,
          "n": 36816,
          "pc": 0,
          "tpc": 0
        },
        {
          "id": "CL:0000001",
          "me": 0.009833387294340086,
          "n": 35693,
          "pc": 0,
          "tpc": 0
        },
        {
          "id": "CL:0000002",
          "me": 0.009822686854067019,
          "n": 35165,
          "pc": 0,
          "tpc": 0
        }
      ],
      "UBERON:0000003": [
        {
          "id": "CL:0000000",
          "me": 0.010225023139072866,
          "n": 36477,
          "pc": 0,
          "tpc": 0
        },
        {
          "id": "CL:0000001",
          "me": 0.009920255662181486,
          "n": 36376,
          "pc": 0,
          "tpc": 0
        },
        {
          "id": "CL:0000002",
[...]

PR #1984, opened 2022-02-18
feat: wmg query endpoint impl

atolopko-czi:
Part 1 of 4 of [story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1959)

### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@ebezzi or @metakuni 

---

## Changes
- Implements core query for WMG data viz response
- Update tiledb and numpy lib versions. TileDB updated to gain latest performance improvements.
- Add cube schema definition, adapted from aurora-scratch repo, with some notable changes:
    - `cell_type_ontology_term_id` is now a TileDB Attribute, instead of a Dimension
    - renamed `feature_id` to `gene_ontology_term_id`, for consistency with WMG terminology
    - zstdfilter compression level fixed (was -22, now +22), for max compression (only affects write-time usage)
- Add wmg `query` module
- Add test fixture for wmg cube creation. Currently used for both testing and for real endpoint (to return dummy data from an actual cube that meets updates schema requirements)
- For test/dev Docker envs, build and store cube in localstack S3 bucket
- Renamed wmg api `tissue_type_terms` param to `tissue_terms`, to align with schema
- Renamed wmg api dimension params to include `*[_ontology]_term_id`, to align with the schema
- Fix lint problem on GHA by pip pkg ver changes to `click` and `black`

## Known Issues
This is not a complete implementation of the query endpoint, which will be done in subsequent branches/PRs. Want to get this code into main for sharing with the cube generation dev effort.

PR #2068, opened 2022-03-07
feat: wmg query endpoint impl - the whole enchilada

atolopko-czi:
Full implementation for [query endpoint story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1959).  Effectively replaces PRs:
* #1984
* #2008 
* #2045
* #2047 
But those PRs are useful for their change descriptions.


backend/wmg/api/v1.py
```diff
@@ -1,58 +1,157 @@
...
+                cell_types=build_ontology_term_id_label_mapping(cell_type_term_ids),
             ),
+            filter_dims=response_filter_dims_values,
+        )
+    )
+
+
+# TODO: It would be reasonable to fetch datasets via a REST API call to Data Portal, rather than coupling to the
+#  portal's db layer, considering that this is the only occasion where the backend.wmg.api package introduces a
+#  dependency on the db. There is no appropriate API call at this time and the lower of making an API call would
```
MDunitz, line 90: I think you might be missing a word here? Unless there is a massive issue with performance I think it makes more sense to create an endpoint in the data portal to get this information for a given list of datasets... or add it to the cube

tests/unit/backend/wmg/data/test_ontology_labels.py
```diff
@@ -0,0 +1,45 @@
+import unittest
+
+from backend.wmg.data.ontology_labels import ontology_term_label, gene_term_label
+
+
+class OntologyLabelTests(unittest.TestCase):
+    def test__ontology_term_label(self):
+        # A conservative, high-level check that all ontologies have been loaded, without checking explicit counts
+        # This test is fragile, in that changes to our ontologies may break these these tests in the future, but that
+        # should happen infrequently
```
MDunitz, line 10: I think it should only happen when we update to the cellxgene schema, we may want to add a note to that repo (or the reqs file) to make necessary updates here on updates

backend/wmg/api/v1.py
atolopko-czi: ah, yes: `and the lower [performance] of making` will fix

atolopko-czi: Good point regarding writing this out at generation-time. It's in keeping with "snapshotted" data and consistency across all the data sources.  I think we'd have to write out as a new data artifact, rather than putting that data into the cube (it would be highly repetitive data, i.e. denormalized).   And this provides even less coupling between WMG and Data Portal than even just making an API call.

tests/unit/backend/wmg/data/test_ontology_labels.py
atolopko-czi: yeah, I'll look for a spot to document.  It will just require a pip version update for cellxgene-schema, assuming the _set_ and _format_ of the files do not change. It's a little fragile, but since the cellxgene-schema is versioned, I'm okay with it.

backend/wmg/data/query.py
```diff
@@ -0,0 +1,98 @@
...
+        # As TileDB API does not yet support logical OR'ing of attribute values in query conditions, the best we can do
+        # for a single TileDB query is to have TileDB perform the filtering for only the attributes that have
+        # a single criterion value specified by the user. We can then perform the filtering client-side for the
+        # multi-valued attributes, albeit at the cost of retrieving extra data from the TileDB engine. Note, however,
+        # that this extra data is always being retrieved by TileDB from the filesystem (and so across the network), so
+        # the performance impact should not be as bad as it may at first appear.
+
+        # Filter single-valued attribute criterion using TileDB filtering, using QueryCondition
+        # TODO: Benchmark whether this is faster than just doing all the attribute filtering in Pandas
+        single_valued_attrs = {
```
MDunitz, line 45: This code might be more clear if you pulled the single valued attr query building and the multi valued attr query building into separate functions

tests/unit/backend/wmg/fixtures/cube.py
```diff
@@ -0,0 +1,154 @@
...
+
+        def dim_domain_values(i_dim: int, dim_size_: int) -> List[str]:
+            dim_name = cube_logical_dims[i_dim]
+            domain_values = dim_ontology_term_ids_generator_fn(dim_name, dim_size_)
+            assert len(set(domain_values)) == dim_size
+            return domain_values
+
+        all_dims_domain_values = [dim_domain_values(i_dim, dim_size) for i_dim in range(n_dims)]
+        # create all possible coordinate values (dim_size ^ n_dims)
+        coords = [
```
MDunitz, line 131: I would consider breaking this down for readability/clarity
Also why is dim_domain_values nested?

MDunitz (approved):
lgtm -- with a few nits around readability (which could be updated during cooldown)

backend/wmg/api/v1.py
atolopko-czi: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2085
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2086

tests/unit/backend/wmg/data/test_ontology_labels.py
atolopko-czi: I've added my TODO list, but will not address immediately.

tests/unit/backend/wmg/fixtures/cube.py
atolopko-czi: add TODO for refactoring

dim_domain_values nesting is just to show that's only be used by the enclosing method

PR #2070, opened 2022-03-08
fix: Additional focus border displays on range filters in staging (#2069)

frano-m:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
- Removal of focus border display on range filter.

### DIFF
--- a/frontend/src/components/common/Filter/components/FilterRange/style.ts
+++ b/frontend/src/components/common/Filter/components/FilterRange/style.ts
@@ -38,6 +38,9 @@ export const useSliderStyles = makeStyles({
       height: 14,
       marginLeft: -7,
       width: 14,
+      "&:focus": {
+        outline: "none",
+      },
       "&:hover": {
         boxShadow: "none",
       },


frontend/src/components/common/Filter/components/FilterRange/style.ts
```diff
@@ -38,6 +38,9 @@ export const useSliderStyles = makeStyles({
       height: 14,
       marginLeft: -7,
       width: 14,
+      "&:focus": {
```
tihuan, line 41: Ah Science Design System is currently having a [discussion](https://czi-sci.slack.com/archives/C032S43KKFV/p1646681775446979?thread_ts=1646438838.295509&cid=C032S43KKFV) around this now as well!

In short, for accessibility, if we don't like the default outlines we'll need to mitigate it instead of just removing it. The link below provides three alternatives that may be worth discussing with the PM and design about?

https://www.a11yproject.com/posts/never-remove-css-outlines/

Screen Shot 2022-03-07 at 11 34 10 AM

frano-m: Ok, great thanks @tihuan, I'll ask for direction from the team!

hthomas-czi: Thanks @tihuan! I agree with Maya that I would only expect the focus state styling for keyboard navigation. We have hover states and active states for mouse events already. Those aren't the exact same as a focus state but I'm not sure what need a focus state would address that hover and active states don't already. So, I'd vote in favor of option 3. @frano-m

tihuan:
Good news, @frano-m !

We can just change `focus` to `focus-visible` 🚀 !

https://css-tricks.com/keyboard-only-focus-styles/

SDS example: https://github.com/chanzuckerberg/sci-components/pull/137/files

Thanks so much 😄 !

frano-m:
Hey @tihuan that's a great suggestion (I'll pocket that for another time)! It looks like Mui Slider already has an equivalent classname `focusVisible` for keyboard interaction. The style applied when using the keyboard is a `box-shadow`. Here is a [video of the interaction](https://clevercanary.slack.com/archives/C02CF7MQPGV/p1646872535139429?thread_ts=1646421389.353799&cid=C02CF7MQPGV). The `:focus` css rule in my PR does not affect this keyboard focus style (because I've targeted the `outline` style). @tihuan are you ok to proceed with the interaction as per the video?

tihuan:
Ooh that's a great point! Yeah if there's already focus-visible style in place then nuking focus outlines SGTM! Thank you!!

tihuan (approved):
LGTM!!

PR #2090, opened 2022-03-09
fix: updated DOI form field value on edit.

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
- Updated create of link view model to only convert DOI URLs to path if filter state is enabled.

#### Steps to Reproduce
1. Remove filter feature flag.
2. Edit collection that has a DOI specified.

#### What Happened
DOI form field displays a DOI path (e.g. `/123/123`).

#### What I Expected
DOI form field displays the full DOI URL (e.g. `https://doi.org/123/123`).

## Known Issues
Timmy, this change is bigger than I had hoped as I had to move `createFn` in `useState<Link[]>(createFn)` to a `useEffect` so I could access the return value from the feature flag hook (`isFilterEnabled`). Let me know if you can think of a simpler fix than my mine!

### DIFF
--- a/frontend/src/components/CreateCollectionModal/components/Content/index.tsx
+++ b/frontend/src/components/CreateCollectionModal/components/Content/index.tsx
@@ -137,26 +137,37 @@ const Content: FC<Props> = (props) => {
   if (isTombstonedCollection(data)) data = null;
   const { name, description, contact_email, contact_name } = data || {};
 
-  const [links, setLinks] = useState<Link[]>(
-    data?.links.map((link, index) => {
-      const {
-        link_name: linkName,
-        link_type: linkType,
-        link_url: linkUrl,
-      } = link;
+  const [links, setLinks] = useState<Link[]>([]);
+  // TODO useEffect can be reverted back to useState once isFilterEnabled is removed (#1718). See 7210a51d18b2747ea79b47fcad28579bb5b514b6.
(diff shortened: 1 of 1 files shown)


frontend/src/components/CreateCollectionModal/components/Content/index.tsx
```diff
@@ -137,26 +137,37 @@ const Content: FC<Props> = (props) => {
...
-      return {
-        id: Date.now() + index,
-        isValid: true,
-        linkName,
-        linkType,
-        url,
-      };
-    }) || []
-  );
+  const [links, setLinks] = useState<Link[]>([]);
```
tihuan, line 140: would useMemo() work here? This way we get `links` one render cycle sooner each time!

tihuan (commented):
Thanks so much for the fix, Mim!

Do we need this to be against `staging` branch, so it will go into `prod`? 

And then we back merge `staging` into `main` for `main` to have the fix!

MillenniumFalconMechanic:
Closing and recreating against `staging`.

tihuan (approved):
NVM! `useEffect()` is better 😆 Thank you!

MillenniumFalconMechanic:
Superseded by [#2096](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2096).

PR #2095, opened 2022-03-09
make rankit values positive

pablo-gar:
Makes rankit values positive, addresses #2066


MDunitz (approved):
lgtm -- thanks for updating the doc string, your function description adds a lot of clarity

PR #2096, opened 2022-03-10
fix: updated DOI form field value on edit.

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
- Updated create of link view model to only convert DOI URLs to path if filter state is enabled.

#### Steps to Reproduce
1. Remove filter feature flag.
2. Edit collection that has a DOI specified.

#### What Happened
DOI form field displays a DOI path (e.g. `/123/123`).

#### What I Expected
DOI form field displays the full DOI URL (e.g. `https://doi.org/123/123`).


tihuan (approved):
LGTM! Thanks so much for the fix, @MillenniumFalconMechanic 🤩 🙏 🙏 🚀 !

tihuan:
@tihuan will create a merge PR from `staging` back to `main` after this is merged into `staging!

PR #2111, opened 2022-03-10
fix: convert days_to_live to int

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


danieljhegeman (approved):
I would maybe just `days_to_live = int(CorporaAuthConfig().days_to_live)` but perhaps it's important to keep as float when it presents initially as one

PR #2073, opened 2022-03-08
feat: wmg query api requires gene terms

atolopko-czi:
Sub-PR of #2071, as part of this [story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2051). This separate PR is intended to ease CR, since these changes are somewhat out-of-scope for the story.

### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@tihuan 

---

## Changes
* wmg `query` api now requires gene terms; this was originally an optional filter param, but the api design changed

PR #2071, opened 2022-03-08
feat: wmg api primary filter dims endpoint

atolopko-czi:
For [story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2051).

Branched off of PR #2068, so wait for that PR to be merged before CR'ing, for reduced diff.

### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@tihuan 

---

## Changes
* impl `primary_filter_dimensions` endpoint: update openapi spec, api impl method, WmgQuery class
* cube test fixture logic now supports exclusion of specific coords (rows) to facilitate more detailed testing of inter-related dimensions (e.g. an organism dim value implies an independent set of gene dim values).
* renamed WmgQuery.query() to expression_summary() for differentation from new method for primary_filter_dimensions().

## Testing
* Unit tests
* Localhost Docker API call (dummy data cube):
```
❯ curl -X GET "http://backend.corporanet.local:5000/wmg/v1/primary_filter_dimensions" -H "accept: application/json"  | jq .

{
  "gene_terms": {
    "NCBITaxon:10001": [
      {
        "ENSG00000000003": "TSPAN6"
      },
      {
        "ENSG00000000005": "TNMD"
      },
      {
        "ENSG00000000419": "DPM1"
      }
    ],
    "NCBITaxon:10002": [
      {
        "ENSG00000000003": "TSPAN6"
      },
      {
        "ENSG00000000005": "TNMD"
      },
      {
        "ENSG00000000419": "DPM1"
      }
    ],
    "NCBITaxon:1000278": [
      {
        "ENSG00000000003": "TSPAN6"
      },
      {
        "ENSG00000000005": "TNMD"
      },
      {
        "ENSG00000000419": "DPM1"
      }
    ]
  },
  "organism_terms": [
    {
      "NCBITaxon:10001": "Sciurus"
    },
    {
      "NCBITaxon:10002": "Sciurus aberti aberti"
    },
    {
      "NCBITaxon:1000278": "Acheilognathus tabira tohokuensis"
    }
  ],
  "snapshot_id": "4e70b2b4a8b145c48c488951abd7bec5",
  "tissue_terms": [
    {
      "UBERON:0000000": "processual entity"
    },
    {
      "UBERON:0000002": "uterine cervix"
    },
    {
      "UBERON:0000003": "naris"
    }
  ]
}
```

PR #2115, opened 2022-03-11
feat: WMG API (multiple stories)

atolopko-czi:
Contains all multiple stories' unmerged changes, all of which have been approved and now rdev tested, and thus are now all ready for merge to main:
* #2068
    * #1984
    * #2008 
    * #2045
    * #2047 
* #2071
    * #2073
* #2087
* #2114 

* ### Reviewers
**Functional:** 
none, all based upon pre-approved PRs

**Readability:** 
@MDunitz @ebezzi (just for sanity check)

---

## Changes
See PRs above

## QA steps (optional)
Manual smoke testing on rdev of both WMG API endpoints (via Swagger UI page) and Data Portal UI.

## Known Issues
TileDB-based unit test are commented out until we get a bug fix release, since GHA tests will fail otherwise.


requirements.txt
```diff
@@ -10,18 +9,19 @@ h5py<3.0.0
 jsonschema
 moto==1.3.14
 numba==0.53.1
-numpy==1.21.2
+numpy==1.22.2
```
atolopko-czi, line 12: This breaks numba, but we don't seem to _use_ numba in portal anymore

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -1,96 +1,404 @@
...
+from unittest.mock import patch
 
 from backend.corpora.api_server.app import app
-from backend.wmg.api import v1
+from backend.wmg.data.schema import cube_non_indexed_dims
 from tests.unit.backend.corpora.fixtures.environment_setup import EnvironmentSetup
+from tests.unit.backend.wmg.fixtures.cube import create_temp_cube, all_ones_attr_values
 
 
+@unittest.skip("TileDB bug (<=0.13.1) causing these to fail")
```
atolopko-czi, line 12: N.B.

tests/unit/backend/wmg/test_query.py
```diff
@@ -0,0 +1,504 @@
+import unittest
+from typing import Tuple
+
+from backend.wmg.data.query import WmgQueryCriteria, WmgQuery, build_dot_plot_matrix
+from backend.wmg.data.schema import cube_non_indexed_dims
+from tests.unit.backend.wmg.fixtures.cube import create_temp_cube, all_ones_attr_values
+
+
+@unittest.skip("TileDB bug (<=0.13.1) causing these to fail")
```
atolopko-czi, line 9: N.B.

Dockerfile
```diff
@@ -34,4 +34,6 @@ LABEL commit=${HAPPY_COMMIT}
 ENV COMMIT_SHA=${HAPPY_COMMIT}
 ENV COMMIT_BRANCH=${HAPPY_BRANCH}
 
-CMD gunicorn --worker-class gevent --workers 8 --bind 0.0.0.0:5000 backend.corpora.api_server.app:app --max-requests 10000 --timeout 30 --keep-alive 5 --log-level info
+# FIXME: REVERT! For now, changed `--workers` from 8 to 1 for now, for TileDB mem requirements on rdev.
```
atolopko-czi, line 37: This is probably fine to keep, actually, since worker count used in deployment envs is specified via TF

backend/config/corpora-api.yml
```diff
@@ -1475,9 +1475,11 @@ components:
           A single-element object with the ontology term id as the element's property name and the ontology term label
           as the element's property value.
         type: object
-        maxProperties: 1
-        additionalProperties:
-          type: string
+        # TODO: fix
```
MDunitz, line 1478: future pr?

requirements.txt
MDunitz: In that case should we just remove the req?

backend/wmg/data/tiledb.py
```diff
@@ -0,0 +1,48 @@
...
+from backend.corpora.common.utils.math_utils import MB, GB
+
+
+def create_ctx(tiledb_mem_gb: float = 1.0) -> tiledb.Ctx:
+    cfg = {
+        "py.init_buffer_bytes": int(float(tiledb_mem_gb) * GB),
+        "sm.tile_cache_size": virtual_memory_size(0.5),
+        "sm.consolidation.buffer_size": consolidation_buffer_size(0.1),
+        "sm.query.sparse_unordered_with_dups.non_overlapping_ranges": "true",
+        # "sm.query.sparse_unordered_with_dups.non_overlapping_ranges": "false",
```
MDunitz, line 15: delete?

backend/wmg/api/v1.py
```diff
@@ -1,58 +1,141 @@
...
+from backend.wmg.data.schema import cube_non_indexed_dims
 
 
+# TODO: add cache directives: no-cache (i.e. revalidate); impl etag
 def primary_filter_dimensions():
-    organism_terms = [dict(oid1="olbl1"), dict(oid2="olbl2")]
-    tissue_type_terms = [dict(ttid1="ttlbl1"), dict(ttid2="ttlbl2")]
-    result = dict(snapshot_id=DUMMY_SNAPSHOT_UUID, organism_terms=organism_terms, tissue_type_terms=tissue_type_terms)
+    cube, snapshot_identifier = load_cube()
+    qry = WmgQuery(cube)
```
MDunitz, line 24: nit -- can you name this something more informative?

backend/wmg/api/v1.py
```diff
@@ -1,58 +1,141 @@
...
+def fetch_datasets_metadata(dataset_ids: List[str]) -> List[Dict]:
+    # We return a DTO because the db entity can't access its attributes after the db session ends,
+    # and we want to keep session management out of the calling method
+
+    with db_session_manager() as session:
+
+        def get_dataset(dataset_id_):
+            dataset = Dataset.get(session, dataset_id_)
+            if dataset is None:
+                # Handle possible missing dataset due to db state evolving past wmg snapshot
```
MDunitz, line 89: This is a good call out re data lifecycle. If we run the cube creation batch job weekly there is a high chance things may get out of sync. Will this result in a blank spot/confusing data in the front end display?

MDunitz: Actually I think this should be solved by https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2085 and the corresponding data serving ticket

backend/wmg/api/v1.py
```diff
@@ -1,58 +1,141 @@
...
+            )
+        )
+    return structured_result
+
+
+def extract_filter_dims_values(query_result: DataFrame) -> dict:
+    """
+    Return unique values for each dimension in the specified query result
+    """
+    return {
```
MDunitz, line 139: nit I think this might be a little more readable if you decomposed it to something like this (assuming I'm groking the original function correctly)

some_dict = {}
for col in query_result.columns:
    some_keys = query_result.groupby(col).groups.keys()
    some_dict[col] = set(some_keys) & set(cube_non_indexed_dims)
    
return some_dict

backend/wmg/data/cube.py
```diff
@@ -0,0 +1,69 @@
...
+
+    latest_snapshot_identifier_s3obj.reload()  # necessary?
+    new_snapshot_identifier = latest_snapshot_identifier_s3obj.get()["Body"].read().decode("utf-8").strip()
+
+    if new_snapshot_identifier != latest_snapshot_identifier:
+        logger.info(f"detected snapshot update from {latest_snapshot_identifier} to {new_snapshot_identifier}")
+        latest_snapshot_identifier = new_snapshot_identifier
+        return True
+    else:
+        if logger.isEnabledFor(logging.DEBUG):
```
MDunitz, line 59: doesn't logger.debug only log when the log level is set to debug?

Dockerfile
atolopko-czi: rm'd FIXME, replacing w/comment to justify 1 worker

backend/config/corpora-api.yml
atolopko-czi: All TODOs I introduced are now linked to future stories.

backend/wmg/data/tiledb.py
atolopko-czi: done!

backend/wmg/api/v1.py
atolopko-czi: Agreed--that generated data artifact address the problem perfectly. Resolving.

backend/wmg/api/v1.py
atolopko-czi: Was trying to remain Pythonic in use of the dictionary-producing comprehension. Let me take a stab at reworking for clarity.

backend/wmg/data/cube.py
atolopko-czi: Sometimes just constructing the string is an expensive operation, so one wants to avoid doing that when it's not going to be logged, due to the log level. Unnecessary here--I'll remove it here!

requirements.txt
atolopko-czi: broke upload pipeline, reverted

Issue #2012, opened 2022-02-28
Create and serve new API to users.


Bento007:
## Approach
- Create the initial openapi spec for the curator API
- Update the routes for the portal backend to include the new curator API.

## Definition of Done
- The curator API documentation can be reached at app../curator/

Bento007:
complete

Issue #2011, opened 2022-02-28
Create token generation endpoint


Bento007:
Add an endpoint to retrieve and access token that can be used to make authenticated requests to the curator API.

## POST /curator/auth/token
to get an access token using the API key
### Parameters
header: authorization: Api key
### Response
Access token

## References
- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1862

## Definition of Done
- The endpoint generates an Access token using a generated API key
- The access token can be used to make authenticated requests to the curator API
- the access token can not be used to make authenticated requests to the portal api.

Bento007:
work completed

PR #2116, opened 2022-03-11
feat: API Key to Access Token

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add curator_api.yml
- Use blueprints to import connexion apis
- Use connexion to display openapi spec.
- Add POST curator/auth/token to retrieve an auth token for the curator API.
- add unit tests and smoke tests for curator auth token API.

- Change the path to the dataportal openapi spec to /
- add the path to the curator openapi spec to /curator/

#2012 
#2011


backend/corpora/api_server/app.py
```diff
@@ -19,8 +19,14 @@ def create_flask_app():
     connexion_app = connexion.FlaskApp(f"{APP_NAME}-{DEPLOYMENT_STAGE}", specification_dir="backend/config")
     # From https://github.com/zalando/connexion/issues/346
     connexion_app.app.url_map.strict_slashes = False
-    swagger_spec_path = f"{APP_NAME}.yml"
-    connexion_app.add_api(swagger_spec_path, validate_responses=True)
+    dataportal_api = super(connexion.FlaskApp, connexion_app).add_api(
```
danieljhegeman, line 22: why is this `super` call necessary?

backend/corpora/api_server/app.py
```diff
@@ -19,8 +19,14 @@ def create_flask_app():
...
+        f"{APP_NAME}.yml", validate_responses=True, options={"swagger_ui": True}
+    )
+    curator_api = super(connexion.FlaskApp, connexion_app).add_api(
+        "curator-api.yml", validate_responses=True, options={"swagger_ui": True}
```
danieljhegeman, line 26: I'm noticing that we didn't specify `swagger_ui: True` previously. It existed by default though, right? But we aren't publicly advertising the url...? 🤔

Bento007: Yes it wasn't advertised. I'm being explicit here for simplicity.

backend/corpora/api_server/app.py
```diff
@@ -3,8 +3,8 @@
 import json
 import logging
 import os
-from connexion import FlaskApi, ProblemException, problem
-from flask import g, Response, jsonify
+from connexion import FlaskApi, ProblemException, problem, Api
```
metakuni, line 6: `Api` is an unused import, claims lint.

metakuni (approved):
What is the empty file `backend/config/create_collection copy.txt` for? 🤔

tests/unit/backend/corpora/api_server/test_app.py
```diff
@@ -2,7 +2,11 @@
...
 class TestAPI(BaseAPITest):
-    def test_smoke(self):
+    def test_smoke_dataportal(self):
         """If this fails then the server does not work"""
-        response = self.app.get("/")
+        response = self.app.get("/ui/")
+        self.assertEqual(200, response.status_code)
+
+    def test_smoke_curator(self):
+        response = self.app.get("/curation/ui/")
```
danieljhegeman, line 11: `ui`?

backend/corpora/api_server/app.py
danieljhegeman: what about [this comment](https://github.com/spec-first/connexion/issues/512#issuecomment-570658527) on that issue's thread?

danieljhegeman: It also takes care of the blueprint registration below, it seems. Correct me if I'm reading it wrong.

backend/corpora/api_server/app.py
```diff
@@ -19,8 +19,15 @@ def create_flask_app():
...
     # From https://github.com/zalando/connexion/issues/346
     connexion_app.app.url_map.strict_slashes = False
-    swagger_spec_path = f"{APP_NAME}.yml"
-    connexion_app.add_api(swagger_spec_path, validate_responses=True)
+    options = {"swagger_ui": True, 'swagger_url': '/'}
+    dataportal_api = super(connexion.FlaskApp, connexion_app).add_api(
+        f"{APP_NAME}.yml", validate_responses=True, options=options
+    )
+    curator_api = super(connexion.FlaskApp, connexion_app).add_api(
+        "curation-api.yml", validate_responses=True, options=options
```
danieljhegeman, line 27: is there any reason to have the app come in as an env variable `APP_NAME` but not for the curation api? Should we hardcode both? 🤔

backend/corpora/api_server/app.py
danieljhegeman: Oh, but we have to have be able to specify the `url_prefix` when we make the call to `register_blueprint`. Nvm then.

danieljhegeman (approved):

PR #2098, opened 2022-03-10
chore: Merge staging into main for hotfix

tihuan:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

PR #2117, opened 2022-03-11
chore: staging deploy

atolopko-czi:
⚠️ Need to address https://github.com/chanzuckerberg/single-cell-data-portal/pull/2098 first (hotfix in staging, back to main)

This PR shows ~2x actual diff file count, probably due to 1d3a9936 being a squash merge. 

We may want to do this:
1. Merge #2098 hotfix to `main`
2. Recreate `staging` from `main` to get a pure commit history and easier diff'ing in the future.


atolopko-czi:
recreating staging branch instead of resolving merge conflicts; staging should only require a fast forward merge, if the branch's git commit history is clean; it is not due to squash merges into staging.

PR #2145, opened 2022-03-15
feat: pre compute cube pipeline skeleton

MDunitz:
### Reviewers
**Functional:** 
- @atolopko-czi  
**Readability:** 
- @ebezzi 
---

## Changes
- add
skeleton for cube pipeline and basic tests

This was previously in https://github.com/chanzuckerberg/single-cell-data-portal/pull/2105


backend/wmg/data/cube_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+    pass
+
+
+def load_data_and_create_cube(path_to_datasets: str, corpus_name: str, log_level):
+    """
+    Function to copy H5AD datasets (from a preconfiugred s3 bucket) to the path given then,
+    open, transform, normalize and concatenate them together as a tiledb object with a global gene index
+    under the given corpus name.
+    A cube of summary statistics across genes (but queryable by the given dimensions/data) is then generated
+    and stored under big-cube.
```
atolopko-czi, line 24: In API layer, I'm referring to this cube as the "expression summary" cube, to differentiate from the upcoming "cell counts" cube.

backend/wmg/data/cube_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+    ## TODO add function to get cell count totals
+    A per tissue mapping of cell ontologies is generated and the files are copied to s3 under a shared timestamp,.
```
atolopko-czi, line 26:
```suggestion
    A per-tissue mapping of cell ontologies is generated and the files are copied to s3 under a shared timestamp.
```

backend/wmg/data/load.py
```diff
@@ -0,0 +1,19 @@
+def update_s3_resources():
+    """
+    Copy cube and cube related files to s3 under the current timestamp
+    """
+    pass
+
+
+def remove_oldest_data():
```
atolopko-czi, line 8:
```suggestion
def remove_previous_snapshot():
```

atolopko-czi: Or maybe it takes a timestamp, since it might be safer if the caller decides which snapshot to remove. can then then rename to just `remove_snapshot(timestamp)`

backend/wmg/data/cube_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+    On success the least recent set of files are removed from s3
+    """
+    copy_datasets_to_instance(path_to_datasets)
+    load(path_to_datasets, corpus_name)
+    create_cube(corpus_name)
+    cell_type_by_tissue = get_cells_by_tissue_type(corpus_name)
+    generate_cell_ordering(cell_type_by_tissue)
+    update_s3_resources()
```
atolopko-czi, line 34: recommend adding logging statements here, for troubleshooting and auditing

backend/wmg/data/cube_pipeline.py
```diff
@@ -0,0 +1,40 @@
...
+    return True
```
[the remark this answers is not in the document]
atolopko-czi: If there is a failure, will the pipeline cleanup intermediate, unfinished output?

backend/wmg/data/load.py
```diff
@@ -0,0 +1,19 @@
+def update_s3_resources():
```
atolopko-czi, line 1: will this call the below methods? could stub that out now, if desired

backend/wmg/data/transform.py
```diff
@@ -0,0 +1,16 @@
...
+
+def get_cells_by_tissue_type(corpus_group: str) -> Dict:
+    """
+    Return a list of all associated cell type ontologies for each tissue contained in the
+    provided corpus
+    """
+    pass
+
+
+def generate_cell_ordering(cell_type_by_tissue: Dict):
```
atolopko-czi, line 12: does it also write out the result to file(s)?  `None` return type?

atolopko-czi (changes requested):
Looks straight forward! Will wait for your feedback on the few suggestions & nits I added before approving. I realize much of this may have already changed in subsequent PRs, so ignore any comments that may have been addressed by later changes elsewhere.

backend/wmg/data/cube_pipeline.py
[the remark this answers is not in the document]
MDunitz: It doesnt do that currently, since we are spinning up instances to run the pipeline and then terminiating when done it didnt seem necessary to add additional cleanup. Let me know if you think its worth adding clean up and I'll create a ticket

backend/wmg/data/load.py
MDunitz: Maybe we need a name for all of the data (I intended this to remove the expression summary cube, the cell count cube, the corpus and the cell ordering json file)

MDunitz: My understanding was that snapshot just referred to the expression summary cube -- let me know if thats incorrect

backend/wmg/data/load.py
MDunitz: it will call the below methods and copy the data stored under the name of the cube to s3 under the snapshot id

atolopko-czi (approved):
👏

PR #2108, opened 2022-03-10
feat: select and load datasets to be included in cube

MDunitz:
### Reviewers

**Functional:** 
@atolopko-czi 
**Readability:** 
@ebezzi 

---

## Changes
- add 
    - code to query relational db for dataset ids of datasets fitting the appropriate criteria
    - code to query dataset assets for s3 uris based on dataset ids
    - code to copy h5ad datasets from s3
    - small test datasets and a cube created from them in the test fixtures file

PR #2146, opened 2022-03-15
feat: wmg load datasets

MDunitz:
### Reviewers
**Functional:** 
- @atolopko-czi 
**Readability:** 
@ebezzi 
---

## Changes
- add
  - code to query relational db for dataset ids of datasets fitting the appropriate criteria
  - code to query dataset assets for s3 uris based on dataset ids
  - code to copy h5ad datasets from s3

Previously - https://github.com/chanzuckerberg/single-cell-data-portal/pull/2108


backend/wmg/data/extract.py
```diff
@@ -1,11 +1,56 @@
...
+        s3_uris = DatasetAsset.s3_uris_for_datasets(session, dataset_ids, DatasetArtifactFileType.H5AD)
+    return s3_uris.values()
 
 
 def copy_datasets_to_instance(s3_uris, dataset_directory):
     """Copy given list of s3 uris to the provided path"""
-    pass
+    for dataset in s3_uris:
+        sync_command = f"aws s3 sync {s3_uris[dataset]} ./{dataset_directory}/{dataset}/local.h5ad"
+        os.subprocess(sync_command)  # TODO parallelize this step
```
atolopko-czi, line 56: also, maybe this could help speed: https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-concurrent-requests

tests/unit/backend/corpora/common/entities/test_dataset_asset.py
```diff
@@ -155,3 +155,66 @@ def test_get_most_recent_artifact(self):
...
+            filename="some file",
+            filetype=DatasetArtifactFileType.CXG,
+            type_enum=DatasetArtifactType.REMIX,
+            user_submitted=True,
+            s3_uri=second_uri,
+        )
+        s3_uri_dataset_dict = DatasetAsset.s3_uris_for_datasets(self.session, [dataset_0.id, dataset_1.id])
+
+        self.assertIn(s3_uri_dataset_dict.keys(), [dataset_0.id, dataset_1.id])
+        self.assertNotIn(s3_uri_dataset_dict.keys(), dataset_2.id)
```
atolopko-czi, line 217: https://docs.python.org/3.9/library/unittest.html#unittest.TestCase.assertNotIn docs say usage is `assertNotIn(member, container)`, and member may need to be a scalar value. So can you confirm whether these are valid assertions?

tests/unit/backend/wmg/data/test_extract.py
```diff
@@ -1,6 +1,144 @@
...
+        ]
+        for dataset in not_expected_datasets:
+            dataset_assets = dataset.get_assets()
+            for asset in dataset_assets:
+                not_expected_s3_uris.append(asset.s3_uri)
+        for dataset in expected_datasets:
+            assets = dataset.get_assets()
+            for asset in assets:
+                if asset.filetype == DatasetArtifactFileType.H5AD:
+                    expected_s3_uris.append(asset.s3_uri)
```
atolopko-czi, line 133: nit: I'm partial to using Python comprehensions to build and filter lists like the two above

tests/unit/backend/wmg/data/test_extract.py
```diff
@@ -1,6 +1,144 @@
...
+
+        s3_uris = set(extract.get_s3_uris())
+        for uri in expected_s3_uris:
+            self.assertIn(uri, s3_uris)
+        for uri in not_expected_s3_uris:
+            self.assertNotIn(uri, s3_uris)
```
atolopko-czi, line 139: `self.assertEquals(set(expected_s3_uris), s3_uris)`

might this replace both loops? with this change, I don't think there is a need to check (or build) `non_expected_s3_uris`.

backend/wmg/data/extract.py
```diff
@@ -1,11 +1,56 @@
...
+                Dataset.table.published == "TRUE",
+                Dataset.table.is_primary_data == "PRIMARY",
+                Dataset.table.collection_visibility == "PUBLIC",
+                Dataset.table.tombstone == "f",
+            )
+            .all()
+        )
+        for dataset in published_dataset_non_null_assays:
+            if dataset[1]["ontology_term_id"] in included_assay_ontology_ids:
+                dataset_ids.append(dataset[0])
```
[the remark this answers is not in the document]
atolopko-czi: I'm also wondering if the ORM supports the SQL `IN` operator? Potentially could have the db filter down to the assay types we want to retrieve.

backend/wmg/data/extract.py
```diff
@@ -1,11 +1,56 @@
...
+
+        s3_uris = DatasetAsset.s3_uris_for_datasets(session, dataset_ids, DatasetArtifactFileType.H5AD)
+    return s3_uris.values()
 
 
 def copy_datasets_to_instance(s3_uris, dataset_directory):
     """Copy given list of s3 uris to the provided path"""
-    pass
+    for dataset in s3_uris:
+        sync_command = f"aws s3 sync {s3_uris[dataset]} ./{dataset_directory}/{dataset}/local.h5ad"
```
atolopko-czi, line 55: if doing a single object copy at a time, maybe `aws s3 cp` is better? or maybe we can just use boto3?

atolopko-czi (changes requested):
Functional changes and tests look solid.  Made some minor comments for your review. 

In addition:
1. Do we need to add the TileDB array fixture files? Will it be used? I find it easier if such files are dynamically generated by tests, when needed.  If intended for ad hoc testing, a script can similarly be provided to dynamically generate.  Make it easier to know what's inside and keep up-to-date.
2. Similarly, is the h5ad file used for testing?

backend/wmg/data/extract.py
```diff
@@ -1,11 +1,56 @@
...
     Retrieve list of s3 uris for datasets included in the wmg cube
     """
-    pass
+    with db_session_manager() as session:
+
+        dataset_ids = []
+        published_dataset_non_null_assays = (
+            session.query(Dataset.table.id, Dataset.table.assay)
+            .filter(
+                Dataset.table.assay != "null",
```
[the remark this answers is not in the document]
MDunitz: it seems to be working as is, without that line it errors when checking the ontology ids

backend/wmg/data/extract.py
```diff
@@ -1,11 +1,56 @@
...
+                Dataset.table.published == "TRUE",
+                Dataset.table.is_primary_data == "PRIMARY",
+                Dataset.table.collection_visibility == "PUBLIC",
+                Dataset.table.tombstone == "f",
```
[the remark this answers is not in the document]
MDunitz: I think both work bc they are stored as "f" or "t", but updated for consistency/clarity

backend/wmg/data/extract.py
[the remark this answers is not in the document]
MDunitz: tbh I hate list comprehensions, I find them to be more difficult to grok at a glance then writing out the full loop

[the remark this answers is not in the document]
MDunitz: I'm not 100% sure but I dont think we can use the IN operator. The assay data is a json blob that contains a list of dicts so I need to pull them out to read them/do a comparison. 
When doing it directly in SQL this works
```select id, items
from dataset,
     jsonb_to_recordset(dataset.assay) as items(ontology_term_id text)
where assay != 'null'
  and published = 'TRUE'
  and is_primary_data = 'PRIMARY'
  and collection_visibility = 'PUBLIC'
  and tombstone = 'f'
  and items.ontology_term_id in ('EFO:0008722','EFO:0010010', 'EFO:0010550', 'EFO:0010961', 'EFO:0030002', 'EFO:0009901', 'EFO:0011025', 'EFO:0009899', 'EFO:0009900', 'EFO:0009922', 'EFO:0030003', 'EFO:0030004', 'EFO:0008919');```
  but I couldnt find a good way to do it using the orm so I thought it made more sense to just split it into two steps for simplicity/to avoid rabbit holing when there was an easy workaround

File: tests/unit/backend/wmg/data/test_extract.py (hunk 3 above)
Comment (MDunitz) [reply tests/unit/backend/wmg/data/test_extract.py:133]: tbh I hate list comprehensions, I find them to be more difficult to grok at a glance then writing out the full loop

File: backend/wmg/data/extract.py (hunk 6 above)
Comment (MDunitz) [reply backend/wmg/data/extract.py:55]: cant use boto for sync
we could use boto to do a direct copy but using sync will make it easy to switch over to cxgs and will be a consistent usage for later opying the cube to s3
also using boto I'd need to break down the uris into the bucket and file name as opposed to directly using the uri which seems like it has the potential to add edge cases/bugs

File: Makefile [hunk 9]
```diff
@@ -172,11 +172,11 @@ local-unit-test-backend: # Run container-unittest target in `backend` Docker con
 			CI=true; \
 		fi; \
 	    docker-compose $(COMPOSE_OPTS) run --rm -e DEV_MODE_COOKIES= -e CI $$ci_env -T backend \
-	    bash -c "cd /corpora-data-portal && make container-unittest && if [ \"${CI}\" == "true" ]; then apt-get update && apt-get install -y git && bash <(curl -s https://codecov.io/bash) -cF backend,python,unitTest; fi"; \
+	    bash -c "cd /corpora-data-portal && make container-unittest && if [ \"${CI}\" == "true" ]; then apt-get update && apt-get install -y git && bash <(curl -s https://codecov.io/bash) -cF backend,python3,unitTest; fi"; \
```
atolopko-czi, line 175: was this required or just a recommended practice? when I'm in a virtualenv, python links to python3

backend/wmg/data/extract.py
```diff
@@ -1,11 +1,56 @@
...
+    "EFO:0030002": "microwell-seq",
+    "EFO:0009901": "10x 3' v1",
+    "EFO:0011025": "10x 5' v1",
+    "EFO:0009899": "10x 3' v2",
+    "EFO:0009900": "10x 5' v2",
+    "EFO:0009922": "10x 3' v3",
+    "EFO_0030003": "10x 3' transcription profiling",
+    "EFO:0030004": "10x 5' transcription profiling",
+    "EFO:0008919": "Seq-Well S3",
+}
```
atolopko-czi, line 21: We _could_ use `backend.wmg.data.ontology_labels.ontology_term_label()` to obtain the labels and guarantee correctness. If you think that's worthwhile, perhaps just add a TODO for now.

atolopko-czi (approved):

backend/wmg/data/extract.py
MDunitz: I looked at doing that but I'd rather create a function to look up the ontology ids based on the labels. If we have a list of the ids it doesnt actually tell us much at a glance, I figure the dict at least makes it clear what we are filtering on

Makefile
MDunitz: No this was an experiment to see if it was the cause of the failing tests... it wasnt so I'm going to drop this commit

PR #2163, opened 2022-03-15
WMG processing - additional steps

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

---

## Changes
- add
- remove
- modify


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -163,6 +165,40 @@ module upload_batch {
...
+  cmd               = ""
+  custom_stack_name = local.custom_stack_name
+  remote_dev_prefix = local.remote_dev_prefix
+  deployment_stage  = local.deployment_stage
+  artifact_bucket   = local.artifact_bucket
+  wmg_bucket        = local.wmg_bucket
+  batch_container_memory_limit = 300000
+}
+
+# TODO: this doesn't work yet
```
ebezzi, line 182: I think for now we can keep this on a manual schedule - thoughts @MDunitz? This didn't quite work but I don't think it will take me too long to fix it if we want.

PR #2189, opened 2022-03-18
fix: fix token test

Bento007:
### Reviewers

**Readability:** 

---

## Changes
- Cleaning up the openapi specs
- Clean up collection handlers
- fix token generation test
- add tests case where API key is valid but no longer stored in Auth0


backend/config/curation-api.yml
```diff
@@ -43,8 +43,7 @@ paths:
       summary: Fetch Collections metadata.
       description: Fetch all Collections metadata. Authorization required for private collections.
       security:
-        - cxguserCookie: []
-        - curatorAccess: []
+        - curatorAccessLenient: []
```
danieljhegeman, line 46: we need to let all requests through but we check with user before adding private collections to response, right

backend/config/curation-api.yml
```diff
@@ -123,28 +121,11 @@ paths:
...
-        - $ref: "#/components/parameters/path_collection_uuid"
-      responses:
-        "202":
-          description: Accepted
-          content:
-            application/json:
-              schema:
-                type: string
-        "401":
-          $ref: "#/components/responses/401"
```
danieljhegeman, line 141: and we'll add this back in later

backend/corpora/lambdas/api/v1/curation/auth/token.py
```diff
@@ -16,5 +16,7 @@ def post():
         raise UnauthorizedError("The API key is invalid")
     else:
         identity = auth0_management_session.get_user_api_key_identity(token["sub"])
+        if not identity:
+            raise NotFoundHTTPException("The API key is has is no longer valid.")
```
danieljhegeman, line 20: typo

danieljhegeman (approved):
see typo

backend/config/curation-api.yml
Bento007: when we decide to support it

PR #2196, opened 2022-03-22
fix: correct jwt attr spelling "ait" => "iat"

danieljhegeman:
The attribute for the Issued At claim on the Curator API Key jwt was
misspelled and would not be recognized as a result.

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- fix the attribute typo for "iat"


tihuan (approved):
LGTM!

Bento007 (commented):
LGTM

PR #2197, opened 2022-03-23
feat: remove visibility from url's and api requests

seve:
### Reviewers
**Functional:** @tihuan && (@danieljhegeman || @Bento007)

---

## Changes
- modify
  – Remove visibility where applicable 

## Notes:

Note that I retained the old collection map for revision ↔ published collection relation.  There are probably some performance and request number optimizations to be made. But would rather get a feature complete PR out quick and then address performance opportunities later, if ever.

rdev: https://dan-buggy-2-frontend.rdev.single-cell.czi.technology/


frontend/src/common/queries/collections.ts
```diff
@@ -281,44 +277,45 @@ async function collectionUploadLinks({
...
-  id = "",
-  visibility = VISIBILITY_TYPE.PRIVATE
-) {
+  collection: Collection | TombstonedCollection | null | undefined
+): UseMutationResult<
+  void,
+  unknown,
+  { collectionID: string },
+  { previousCollections: CollectionResponsesMap }
+> {
```
seve, line 313: @tihuan Is this appropriate typing here?

tihuan: discussion: Wondering if the interface could be the same primitive types, so the arguments are more predictable, and we can just set default `""` directly there? 

https://www.johndcook.com/blog/2011/07/19/you-wanted-banana/

The return type looks good!

E.g.,

```ts
function useDeleteCollection(id = "", visibility = ""): UseMutationResult<...>  { ... } 
```

tihuan (commented):
Code LGTM! Just one quick discussion on `useDeleteCollection()` arguments! Thank you 🙆‍♂️

frontend/src/common/queries/collections.ts
seve: agreed, was trying out something new, but I think it makes sense this way

seve: Note though, this introduced _a lot_ of prop drilling:
https://github.com/chanzuckerberg/single-cell-data-portal/pull/2197/commits/450bcb143c2e324a87798a21ed22408b4ce270f1

tihuan: Woah 7 levels lol The deepest I've seen so far ahaha

Maybe time to pass `id` and `visibility` as context?

frontend/src/components/Collections/components/Grid/components/Row/CollectionRow/index.tsx
```diff
@@ -85,10 +84,7 @@ const CollectionRow: FC<Props> = (props) => {
   return (
     <StyledRow data-test-id="collection-row">
       <StyledCell>
-        <Link
-          href={`/collections/${id}${isPrivate ? "/private" : ""}`}
-          passHref
-        >
+        <Link href={`/collections/${id}]`} passHref>
```
Bento007, line 87:
```suggestion
        <Link href={`/collections/${id}`} passHref>
```

Bento007 (changes requested):

frontend/src/common/queries/collections.ts
```diff
@@ -82,9 +84,10 @@ async function fetchCollections(): Promise<CollectionResponsesMap> {
   const collectionsMap: CollectionResponsesMap = new Map();
 
   for (const collection of json.collections as CollectionResponse[]) {
-    const collectionsWithId = collectionsMap.get(collection.id) || new Map();
-    collectionsWithId.set(collection.visibility, collection);
-    collectionsMap.set(collection.id, collectionsWithId);
+    const publicID = collection.revision_of || collection.id;
```
seve, line 87:
```suggestion
    const publicID = collection.visibilty === Visibility.PRIVATE?  collection.revision_of : collection.id;
```

Bento007:
When I start a revision I can immediately publish without making any changes
Screen Shot 2022-04-04 at 10 13 29 AM

Bento007:
When uploading a dataset, the upload spinner transition between processing and uploading. Is there is state where a dataset goes back to uploading from processing?

Bento007:
A refresh is required to show the deleted tag when deleting a dataset from a revision.

danieljhegeman:
> When I start a revision I can immediately publish without making any changes Screen Shot 2022-04-04 at 10 13 29 AM

@Bento007 this should no longer be a problem!

danieljhegeman (approved):

PR #2203, opened 2022-03-25
chore: remove null values from collections/index endpoint

ebezzi:
### Reviewers
**Functional:** 
@MillenniumFalconMechanic 

**Readability:** 
@Bento007 

---

## Changes
- Remove null values from the `collections/index` endpoint. 
- Note: this will also remove null values from other collection related endpoints, but I don't see how this would be a problem.

### DIFF
--- a/backend/corpora/common/entities/collection.py
+++ b/backend/corpora/common/entities/collection.py
@@ -116,6 +116,9 @@ class Collection(Entity):
         def to_dict(db_object):
             _result = {}
             for _field in db_object._fields:
+                _value = getattr(db_object, _field)
+                if _value is None:
+                    continue
                 _result[_field] = getattr(db_object, _field)
             return _result
 
--- a/backend/corpora/common/entities/collection.py
+++ b/backend/corpora/common/entities/collection.py
@@ -116,9 +116,6 @@ class Collection(Entity):
         def to_dict(db_object):
             _result = {}
             for _field in db_object._fields:
-                _value = getattr(db_object, _field)
-                if _value is None:
-                    continue
                 _result[_field] = getattr(db_object, _field)
             return _result
 
--- a/tests/unit/backend/corpora/api_server/test_v1_collection.py
+++ b/tests/unit/backend/corpora/api_server/test_v1_collection.py
@@ -769,6 +769,8 @@ class TestCollection(BaseAuthAPITest):
             visibility=CollectionVisibility.PUBLIC.name,
             owner="test_user_id",
             publisher_metadata={"journal": "Nature"},
+            published_at=datetime.now(),
+            revised_at=datetime.now(),
         )
 
         collection_id_private = self.generate_collection(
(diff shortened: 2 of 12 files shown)


metakuni (approved):
👍

tests/unit/backend/corpora/api_server/test_v1_collection.py
```diff
@@ -769,6 +769,8 @@ def test__get_all_collections_for_index(self):
             visibility=CollectionVisibility.PUBLIC.name,
             owner="test_user_id",
             publisher_metadata={"journal": "Nature"},
+            published_at=datetime.now(),
+            revised_at=datetime.now(),
```
atolopko-czi, line 773: intended change for this PR? (seems out of scope per the PR title)

backend/corpora/common/entities/collection.py
```diff
@@ -116,6 +116,9 @@ def list_attributes_in_time_range(
         def to_dict(db_object):
             _result = {}
             for _field in db_object._fields:
+                _value = getattr(db_object, _field)
+                if _value is None:
+                    continue
```
atolopko-czi, line 121: Not sure where else this `to_dict` is used, but it seems like an API-specific change is leaking into ORM-level code. I would consider putting this filtering logic in the API endpoint impl method, for safety.  I think other code that might use this would be surprised if a key/value pair was entirely missing just because the value is None.

tests/unit/backend/corpora/api_server/test_v1_collection.py
atolopko-czi, line 773: These props are no longer "required", yet the test is only _now_ asserting their value. I guess I would have expected the test change to be that setting the values to None would cause the response to exclude these props, since that's the nature of the change in the PR.

ebezzi: Not really since the `null` removal isn't really something we should test, IMHO (it's good to have but it won't break the API contract). These were added now because previously the assertion basically tested two `None` values and therefore passed even without dates (which was arguably wrong). Now they would be removed from the response so the assertion would raise `KeyError` if we checked for these fields.

backend/corpora/common/entities/collection.py
ebezzi: Normally I would agree with you but our ORM and API logic is already mixed (for instance, a `to_dict` method shouldn't belong to an ORM at all) and I figured that it would make sense to minimize the amount of data manipulation by putting this additional filter here. I am not particularly opinionated on this though, so I am also happy to do it the way you suggested.

tests/unit/backend/corpora/api_server/test_v1_collection.py
atolopko-czi: Unless we're doing this on all entities consistently, I'd move it to the method impl for which it was added.

backend/corpora/common/entities/collection.py
atolopko-czi: Unless we're doing this on all entities consistently, I'd move it to the method impl for which it was added.

ebezzi: I edited this accordingly.

Issue #2204, opened 2022-03-25
Added (but unsaved links) disappear when clicking away from window


jahilton:
## Describe the bug
When trying to add links to a private Collection (or revision), any unsaved links that you add disappear if you click to a new tab or window.

## To Reproduce
Steps to reproduce the behavior: (shown in video)
1. Go to a private collection (or revision)
2. "Edit Details"
3. "Add Link"
4. select any Link Type
5. Click to a new tab or window (say to copy the link)
6. back to the cxg window
7. see the link you've just added go away

## Expected behavior
The links should remain fixed unless the 'x' is clicked to remove them.

## Environment
**Desktop (please complete the following information):**
 - OS: macOS Catalina
 - Browser: chrome
 - Version: 99

## Additional context
If you add one or more links, then click away, all of those will go away upon clicking back (in the video)
If you have already added links & Saved, then redo the steps above to add more, the saved links will not go away in these steps (not shown in video)

https://user-images.githubusercontent.com/15098808/160214776-92914c44-4ec2-444a-bfaa-2831ec509b87.mov

tihuan:
in prod!

PR #2208, opened 2022-03-28
fix: set refetch to false for collection detail. #2204

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@seve 

---

## Changes
- Added `refetchOnWindowFocus: false` for collection detail-related React Query fetches. The general approach for handling of stale data on the FE is currently being reviewed (see [Determine approach for updating search results when backend data changes](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1809) and [Add Cloudfront caching+invalidation for datasets/index and collections/index endpoints](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1899).

### DIFF
--- a/frontend/src/common/queries/collections.ts
+++ b/frontend/src/common/queries/collections.ts
@@ -196,7 +196,17 @@ export function useCollection({
   return useQuery<Collection | TombstonedCollection | null>(
     [USE_COLLECTION, id, visibility, collections],
     () => queryFn(id, visibility),
-    { enabled: !!collections, refetchOnWindowFocus: false } // Refetch can cause loss of unsaved links during edit.
+    {
+      enabled: !!collections,
+      // TODO review use of refetch flag below on remove of filter flag (#1718). This is set to false as a fix for #2204
+      //  where unsaved links on the collection edit form are cleared if the user clicks away and back again. The root
+      //  cause of the defect is the useEffect that was added in /components/CreateCollectionModal/components/Content/index.tsx
+      //  to handle filter flag-related functionality is re-run after the refetch causing unsaved links to be removed.
+      //  The useEffect can be reverted to useState on remove of the filter flag and will therefore remove the side
+      //  effect of the links being cleared as reported in #2204. Options going forward are to remove the refetch
+      //  option below and allow it to default to true, or to review setting the refetch option to false for all queries.
+      refetchOnWindowFocus: false,
+    }
   );
 }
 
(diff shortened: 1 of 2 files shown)


seve:
I think a better solution for [this bug](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2204) would be to introduce a pending state that is overlaid on state received from the backend

seve (commented):
Not sure if this is the right fix for the problem

MillenniumFalconMechanic:
Hey @seve, thank you for your quick review. I have been discussing this PR with @NoopDog  and wanted to give you some background on what we’re thinking.

#### Current State
1. It looks like the application does not currently support the detection of concurrent modifications of the same record by users. It seems to be “the last update wins”.

1. The different sections of the form behave differently with respect to concurrent modification detection:

    The top portion of the form (e.g. name, description, contact) is not currently updated from or merged with refetched data when the user clicks away and back again.

    The bottom part of the form (e.g. links) is always reset to the refetched data when the user clicks away and back again. (This is the defect Jason reported that we’re trying to fix here.)


#### Going Forward
We think a longer-term “best” solution here is something along the lines:

1. Implement optimistic locking on the record on the BE.
1. Detect and reject the (rare) concurrent modifications with appropriate user messaging.
1. Avoid the detection/merging of these cases in the FE.

For now, we would like to go ahead and suppress the refetch as per the PR. This fixes the UI defect, adds consistent form functionality and is consistent with the current concurrent modification behavior (that is, it maintains “last update wins”). 

We feel merging or marking changed links adds complexity, requires input from product/UX and isn't in line with the longer-term "best" solution I am proposing above.

I feel it might be fastest to discuss this on a call (with @brianraymor) given this requires input from product and possibly a longer-term solution.

seve:
> For now, we would like to go ahead and suppress the refetch as per the PR. This fixes the UI defect, adds consistent form functionality and is consistent with the current concurrent modification behavior (that is, it maintains “last update wins”).

Taking a look at some context around the changes, it looks like the breaking change was the introduction of the `useEffect` hook on any changes to `data`.  If this change is only present until `isFilterEnabled` is removed, then I'm happy to green light this.

Maybe add a comment similar to https://github.com/chanzuckerberg/single-cell-data-portal/blob/635ebd0be4d5c3fb197b8c887f149cf396bb3a6a/frontend/src/components/CreateCollectionModal/components/Content/index.tsx#L141

PR #2211, opened 2022-03-29
chore: use tiledb.open to open sparse and dense arrays

metakuni:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- modify: unified code to always use `tiledb.open()` to open TileDB arrays regardless of type (`DenseArray`, `SparseArray`), 
 to use context managers

### DIFF
--- a/backend/corpora/common/utils/cxg_generation_utils.py
+++ b/backend/corpora/common/utils/cxg_generation_utils.py
@@ -9,7 +9,7 @@ from backend.corpora.common.utils.type_conversion_utils import (
 )
 
 
-def convert_dictionary_to_cxg_group(cxg_container, metadata_dict, group_metadata_name="cxg_group_metadata"):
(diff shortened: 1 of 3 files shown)


bkmartinjr (approved):
LGTM

backend/corpora/common/utils/cxg_generation_utils.py
```diff
@@ -27,7 +27,7 @@ def convert_dictionary_to_cxg_group(cxg_container, metadata_dict, group_metadata
     # array.
     tiledb.from_numpy(array_name, np.zeros((1,)))
 
-    with tiledb.DenseArray(array_name, mode="w") as metadata_array:
+    with tiledb.open(array_name, mode="w") as metadata_array:
```
atolopko-czi, line 30: out-of-scope for the refactoring, but should this specify an explicit `ctx` arg?

atolopko-czi (approved):

PR #2214, opened 2022-03-29
feat: wmg api cell ordering

atolopko-czi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 
@ebezzi 

---

## Changes
- add cell ordering to the WMG `query` endpoint response: the `term_id_labels`: `cell_types` array is now ordered per cell ordering file.
- update snapshot test fixture to allow management and customization of cell ordering for tests
- refactor "cube" package to "snapshot" package, as it now includes all data artifacts (2 TileDB arrays and cell ordering json file)
- add WmgConfig `data_path_prefix` param, to support rdev env multi-tenancy user of shared S3 bucket.

## QA steps (optional)
new unit test
rdev API calls checked

PR #2228, opened 2022-03-31
Cell count cube

ebezzi:
### Reviewers
**Functional:** 
@MDunitz @atolopko-czi 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/wmg/cube_creation/load_cube.py
+++ b/backend/wmg/cube_creation/load_cube.py
@@ -15,7 +15,7 @@ from backend.corpora.common.entities import Dataset, DatasetAsset
 from backend.corpora.common.utils.db_session import db_session_manager
 from backend.wmg.cube_creation.corpus_schema import create_tdb
 from backend.wmg.cube_creation.loader import load
-from backend.wmg.cube_creation.wmg_cube import create_cube
+from backend.wmg.cube_creation.wmg_cube import create_cell_count_cube, create_cube
 
 from backend.wmg.data.snapshot import CELL_TYPE_ORDERINGS_FILENAME
 
@@ -138,7 +138,8 @@ def generate_cell_ordering(cell_type_by_tissue):
 
 def update_s3_resources(group_name):
     timestamp = int(time.time())
-    upload_cube_to_s3(group_name, timestamp)
+    upload_cube_to_s3(group_name, "cube", timestamp)
+    upload_cube_to_s3(group_name, "cell-count-cube", timestamp)
     upload_cell_ordering_to_s3(timestamp)
     update_latest_snapshot_identifier(timestamp)
     remove_oldest_datasets(timestamp)
@@ -174,8 +175,8 @@ def remove_oldest_datasets(timestamp):
             object.delete()
 
 
-def upload_cube_to_s3(group_name, timestamp):
-    sync_command = ["aws", "s3", "sync", f"{group_name}/cube", f"{get_wmg_bucket_path()}/{timestamp}/cube"]
+def upload_cube_to_s3(group_name, suffix, timestamp):
+    sync_command = ["aws", "s3", "sync", f"{group_name}/{suffix}", f"{get_wmg_bucket_path()}/{timestamp}/{suffix}"]
     subprocess.run(sync_command)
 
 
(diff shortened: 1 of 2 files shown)


backend/wmg/cube_creation/load_cube.py
```diff
@@ -174,8 +175,8 @@ def enumerate_timestamps_and_objects(objects):
             object.delete()
 
 
-def upload_cube_to_s3(group_name, timestamp):
-    sync_command = ["aws", "s3", "sync", f"{group_name}/cube", f"{get_wmg_bucket_path()}/{timestamp}/cube"]
+def upload_cube_to_s3(group_name, suffix, timestamp):
+    sync_command = ["aws", "s3", "sync", f"{group_name}/{suffix}", f"{get_wmg_bucket_path()}/{timestamp}/{suffix}"]
```
atolopko-czi, line 179: can't we just sync the entire local parent dir to s3, rather than individually for each cube?

if we stick with this usage, note that `group` is a little confusing, since [TileDB also uses "group" term](https://docs.tiledb.com/main/how-to/groups).  I don't think we're using it in that sense, right?  

I think a signature like this would be simpler: `sync_dir_to_s3(local_dir, remote_s3_path)`. And then let the caller append the `suffix` to both args being passed.

backend/wmg/cube_creation/load_cube.py
```diff
@@ -201,6 +202,10 @@ def load_data_and_create_cube(path_to_datasets, group_name):
         create_cube(group_name)
     except Exception as e:
         logger.error(f"Issue creating the cube: {e}")
+    try:
+        create_cell_count_cube(group_name)
+    except Exception as e:
+        logger.error(f"Issue creating the cell count cube: {e}")
```
atolopko-czi, line 208: beyond the scope of this PR, but is there any point in running the rest of the pipeline if either of the 2 cube creations fail? shouldn't we rethrow errors and fail loudly?

backend/wmg/cube_creation/wmg_cube.py
```diff
@@ -253,3 +253,31 @@ def create_cube(tdb_group):
...
+                by=[
+                    "dataset_id",
+                    "cell_type_ontology_term_id",
+                    "tissue_ontology_term_id",
+                    "assay_ontology_term_id",
+                    "development_stage_ontology_term_id",
+                    "disease_ontology_term_id",
+                    "ethnicity_ontology_term_id",
+                    "sex_ontology_term_id",
+                    "organism_ontology_term_id",
```
atolopko-czi, line 276: let's replace this array with this [variable](https://github.com/chanzuckerberg/single-cell-data-portal/blob/3c2aba50f4883d306a1f0646df67ac9f7f103827/backend/wmg/data/schema.py#L73)

backend/wmg/cube_creation/wmg_cube.py
```diff
@@ -253,3 +253,31 @@ def create_cube(tdb_group):
...
+                ],
+                as_index=False,
+            )
+            .size()
+        )
+
+        tiledb.from_pandas(uri, df)
```
atolopko-czi, line 283: Nice!

backend/wmg/cube_creation/wmg_cube.py
```diff
@@ -253,3 +253,31 @@ def create_cube(tdb_group):
     logger.info("Cube creation complete")
     create_cube_sec = time.time() - start_time
     logger.info(f"Big cube: time to create {create_cube_sec}, uri={uri}")
+
+
+def create_cell_count_cube(tdb_group):
+    """
+    Create cell count cube and write to disk
+    """
+    uri = f"{tdb_group}/cell-count-cube"
```
atolopko-czi, line 262: Let's DRY up `cell-count-cube` literal, per earlier comment.

atolopko-czi (changes requested):
Looks functionally correct, but requesting some DRY'ing, which will mean merging this [PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2214) to main first and then updating this branch as needed.

backend/wmg/cube_creation/load_cube.py
ebezzi: That's a good point, it's probably worth removing the try/except here and let the pipeline fail. Not sure that the "naive" batch implementation allows for retries, but given that this only needs to run once per week, it's probably enough if we  have some kind of alerting in place so it can be retried manually.

ebezzi:
Replaced by https://github.com/chanzuckerberg/single-cell-data-portal/pull/2270

PR #2232, opened 2022-03-31
Low gene expression cell filtering

ebezzi:
### Reviewers
**Functional:** 
@atolopko-czi @MDunitz @pablo-gar

### DIFF
--- a/backend/wmg/cube_creation/loader.py
+++ b/backend/wmg/cube_creation/loader.py
@@ -7,6 +7,7 @@ import numpy as np
 import pandas as pd
 from scipy import sparse
 import tiledb
+import scanpy
 
 from .utils import get_all_dataset_ids
 from backend.wmg.data.config import create_fast_ctx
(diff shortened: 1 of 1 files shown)


backend/wmg/cube_creation/loader.py
```diff
@@ -55,6 +56,11 @@ def load_h5ad(h5ad, group_name, validate):
         return
 
     anndata_object = anndata.read_h5ad(h5ad)
+
+    # Apply a low expression gene cell filtering. 
+    # Will get rid of all cells that have less than 500 expressed genes
+    scanpy.pp.filter_cells(anndata_object, min_genes=500)
```
ebezzi, line 62: Should we put this number in a constant?

MDunitz: I think that would be good for clarity

ebezzi: Do we have a good candidate file for these constants? Or this file works?

atolopko-czi (approved):
one comment, approving early.

PR #2236, opened 2022-04-01
feat: cloudfront invalidation for index paths

ebezzi:
### Reviewers
**Functional:** 
@Bento007 @MDunitz 

**Readability:** 

---

## Changes
- Creates a cloudfront invalidation for the index paths each time a collection is published.
Note: This cannot be tested easily in rdev since it doesn't use cloudfront. I also don't fully understand how the python class for reading secrets work, so I'll have to play around with it a little to make sure it works.


backend/corpora/common/corpora_config.py
```diff
@@ -68,3 +68,10 @@ def get_defaults_template(self):
             template["issuer"].append(self.api_signin_url)
 
         return template
+
+
+class CorporaCloudfrontConfig(SecretConfig):
+    environ_source = "CORPORA_CLOUDFRONT_CONFIG"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__("backend", secret_name="cloudfront", **kwargs)
```
MDunitz, line 77: I think you need to add a new line to here and in the cloudfront file to pass the linter

backend/corpora/common/utils/cloudfront.py
```diff
@@ -0,0 +1,32 @@
+from importlib.metadata import distribution
```
Bento007, line 1: What is this for?

backend/corpora/common/utils/cloudfront.py
```diff
@@ -0,0 +1,32 @@
...
+
+def _create_invalidation(distribution: str, paths: list[str]):
+    return client.create_invalidation(
+        DistributionId=distribution,
+        InvalidationBatch={
+            'Paths': {
+                'Quantity': len(paths),
+                'Items': paths,
+                },
+            'CallerReference': str(uuid.uuid4())
```
Bento007, line 27: In the event we need to lookup an invalidation, prefixing this with some sort of timestamp, logging the value uuid, or logging the return response would be useful.

Bento007 (changes requested):

backend/corpora/common/utils/cloudfront.py
ebezzi: Bad import, thanks for noticing 👍

backend/corpora/common/utils/cloudfront.py
```diff
@@ -5,30 +5,35 @@
...
 # only trigger an invalidation if the distribution_id is defined in secrets manager.
 # Otherwise this will be a no-op
 def create_invalidation(paths: list[str]):
-    distribution = CorporaCloudfrontConfig().distribution_id
-    if distribution:
+    try:
+        distribution = CorporaCloudfrontConfig().distribution_id
         return _create_invalidation(distribution, paths)
-    
+    except Exception:
```
Bento007, line 19: It would be an attribute error if distribution_id is not found. Also, you should move `_create_invalidation(distribution, paths)` out of the try block based on the warning you are giving.

Bento007 (changes requested):

backend/corpora/common/utils/cloudfront.py
ebezzi: I don't get what you mean. This catches the case where `CorporaCloudfrontConfig().distribution_id` is not defined (which will only happen in rdev), and yield a no-op.

Bento007 (approved):

PR #2238, opened 2022-04-01
WMG pipeline - final steps

ebezzi:
### Reviewers
**Functional:** 
@MDunitz @atolopko-czi 

**Readability:** 

---


backend/wmg/data/extract.py
```diff
@@ -50,8 +50,5 @@ def get_dataset_s3_uris():
 def copy_datasets_to_instance(s3_uris, dataset_directory):
     """Copy given list of s3 uris to the provided path"""
     for dataset in s3_uris:
-        sync_command = f"aws s3 sync {s3_uris[dataset]} ./{dataset_directory}/{dataset}/local.h5ad"
-        os.subprocess(
-            sync_command
-        )  # TODO parallelize this step see https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#max-concurrent-requests # noqa: E501
-
+        sync_command = ["aws", "s3", "cp", s3_uris[dataset], f"./{dataset_directory}/{dataset}/local.h5ad"]
```
MDunitz, line 53: nit- rename copy_command

backend/wmg/data/load_cube.py
```diff
@@ -1,19 +1,82 @@
...
+    else:
+        return f"s3://{wmg_bucket_name}"
+
+
+def update_s3_resources(corpus_name):
     """
     Copy cube and cube related files to s3 under the current timestamp
     """
-    pass
+    timestamp = int(time.time())
```
MDunitz, line 25: we could refactor to create the timestamp at the beginning of the pipeline run and then store all the created artifacts under the timestamp and then we would only need to run the sync command once to copy everything over

backend/wmg/data/load_cube.py
```diff
@@ -1,19 +1,82 @@
...
+    timestamps = sorted(list(set([x[0] for x in candidate_to_delete])))
+
+    if len(timestamps) > 2:
+        timestamps_to_delete = list(timestamps)[:-2]
+    else:
+        timestamps_to_delete = []
+
+    for timestamp, object in candidate_to_delete:
+        if timestamp in timestamps_to_delete:
+            object.delete()
```
MDunitz, line 61: I think you can use something like this
`aws s3 rm s3://bucket/ --recursive  --include "timestamp*"` to delete everything with a similar prefix (I think that will simplify a lot of the removal code)

backend/wmg/data/load_cube.py
ebezzi: We can do that easily, my only concern is whether that's also gonna copy temporary files created during the process.

backend/wmg/data/load_cube.py
```diff
@@ -1,19 +1,82 @@
...
 
-def update_latest_snapshot_identifier(time_stamp):
+def update_latest_snapshot_identifier(timestamp):
     """
     Update static timestamp in s3 to match the latest
     """
-    pass
+    s3 = boto3.resource("s3")
+    file_key = "latest_snapshot_identifier"
+    file_path = f"{stack_name.strip('/')}/{file_key}" if stack_name else f"{file_key}"
```
MDunitz, line 80: is this line mainly to support rdev?

ebezzi: Yes, uniquely for rdev.

backend/wmg/data/load_cube.py
MDunitz: Just updated the pipeline code take in a path to make testing easier. If a path isnt given it will create one from the curr dir (which is what it currently uses) and a timestamp

PR #2240, opened 2022-04-02
fix: add `id` to ignored fields

danieljhegeman:
### Reviewers
**Functional:** 
@seve 
**Readability:** 

---

## Changes
- add `id` (collections) and `collection_id` (datasets) to the ignored fields so that they don't erroneously cause the Collection to be marked as changed when it isn't (during revision).


seve (approved):

seve:
😴

PR #2241, opened 2022-04-03
chore: removed DatasetArtifactType

metakuni:
### Reviewers
**Functional:** 
@Bento007 

**Readability:** 
@danieljhegeman 

---

## Changes
- remove: unused `DatasetArtifactType` enum, `dataset_artifact.type` column

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -80,12 +80,12 @@ else
 endif
 
 db/test_migration:
-	$(MAKE) db/dump_schema > /tmp/before
+	$(MAKE) db/dump_schema > /tmp/before_migration
 	$(MAKE) db/migrate
 	$(MAKE) db/dump_schema > /tmp/after_migration
 	$(MAKE) db/rollback
-	$(MAKE) db/dump_schema > /tmp/after
-	diff /tmp/{before,after} # No news is good news.
+	$(MAKE) db/dump_schema > /tmp/after_rollback
+	diff /tmp/{before_migration,after_rollback} # No news is good news.
 
 SSH_SERVER_ALIVE_INTERVAL_IN_SECONDS?=60
 SSH_SERVER_ALIVE_COUNT_MAX?=60
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -82,6 +82,7 @@ endif
 db/test_migration:
 	$(MAKE) db/dump_schema > /tmp/before
 	$(MAKE) db/migrate
+	$(MAKE) db/dump_schema > /tmp/after_migration
 	$(MAKE) db/rollback
 	$(MAKE) db/dump_schema > /tmp/after
 	diff /tmp/{before,after} # No news is good news.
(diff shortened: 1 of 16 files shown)


backend/Makefile
```diff
@@ -80,11 +80,12 @@ else
 endif
 
 db/test_migration:
-	$(MAKE) db/dump_schema > /tmp/before
+	$(MAKE) db/dump_schema > /tmp/before_migration
 	$(MAKE) db/migrate
+	$(MAKE) db/dump_schema > /tmp/after_migration
```
metakuni, line 85: Added this to make it easier to check the schema after migration.

backend/Makefile
```diff
@@ -80,11 +80,12 @@ else
...
 	$(MAKE) db/rollback
-	$(MAKE) db/dump_schema > /tmp/after
-	diff /tmp/{before,after} # No news is good news.
+	$(MAKE) db/dump_schema > /tmp/after_rollback
+	diff /tmp/{before_migration,after_rollback} # No news is good news.
```
metakuni, line 88: Renamed the before/after schema dump files.

atolopko-czi (approved):
LGTM, and thx for the cleanup.  Two ops related questions:
1. Will the migration be run _before_ the code is deployed? Or does do we need to break up the migration and code deploys into separated, ordered PR and deployments?
2. If we revert both the schema and the code, will the code be okay with null values in the restored column?

Bento007 (approved):

Bento007 (changes requested):
Also update corpora-api.yml

metakuni:
> LGTM, and thx for the cleanup. Two ops related questions:
> 
> 1. Will the migration be run _before_ the code is deployed? Or do we need to break up the migration and code deploys into separated, ordered PR and deployments?

Good question! (to which I did not know the answer 😄 ) It appears that the database migrations run _after_ the `happy` stack is created/updated:
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/scripts/happy#L935
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/scripts/happy#L987

So, technically, between the time that a stack is created/updated and the migration runs, portal may be _momentarily_ 'broken'. However, this will be the case even if we deploy the migration first in a separate PR followed by deployment of the code changes. The latter will result in a longer potential 'broken' time since more steps are involved.

> 2. If we revert both the schema and the code, will the code be okay with null values in the restored column?

Yes, the dataset artifact's type was never used, so the rollback with nulls should be harmless.

Bento007 (approved):

atolopko-czi:
> > LGTM, and thx for the cleanup. Two ops related questions:
> > 
> > 1. Will the migration be run _before_ the code is deployed? Or do we need to break up the migration and code deploys into separated, ordered PR and deployments?
>
>  It appears that the database migrations run after the happy stack is created/updated:
>

Good, I think this is the order we want! It would only cause a problem for writes if the column is non-null.  I'm also assuming sqlalchemy is okay with ignoring columns that exist in the db but are not specified in the model. Either way, very short time between code & migration running, so I'm expect we're okay in most any scenario.

> 
> > 2. If we revert both the schema and the code, will the code be okay with null values in the restored column?
> 
> Yes, the dataset artifact's type was never used, so the rollback with nulls should be harmless.

👍

PR #2246, opened 2022-04-04
feat(logging): Log incoming requests.

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---
I noticed the logs were sparse, and it was difficult to determine what request was causing an error. Now incoming request information is logged. Only the bare minimum of information is logged to maintain user privacy and avoid store large amounts of log data.

example: ```[2022-04-04 18:29:56 +0000] [8] [INFO] {"url": "/ui/", "method": "GET", "schema": "http"}```

## Changes
- add Log for incoming requests.
- limiting what's logged to URL, method, and schema.


atolopko-czi (approved):
Thanks for adding!  Might want to add your PR description comment re:limiting what's logged, to ensure then next engineer that comes along doesn't add more output without considering these points.

MDunitz (approved):
lgtm -- do we need to update any docs/pms re whats being logged?

PR #2252, opened 2022-04-04
feat: wmg - filter counts that are less than 3

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

---

## Changes
- add
- remove
- modify


backend/wmg/data/load_corpus.py
```diff
@@ -64,6 +64,15 @@ def load_h5ad(h5ad: str, group_name: str, validate: bool):
     if not validate_dataset_properties(anndata_object):
         return
 
+    # Set all values <=min_value to 0
+    min_value = 3
```
ebezzi, line 68: To be moved to a constant together with the other optimization (the 500 filter) - I'll consolidate them on the other PR.

MDunitz (changes requested):
Hey can you move this into save_raw and ensure that the updates are also applied to the rankit values?
Also can you add sone unit tests?

PR #2257, opened 2022-04-04
fix(wmg): Update secondary filter options

tihuan:
### Reviewers
**Functional:** 
@seve 
**Readability:** 

---

## Changes
Update `useFilterDimensions()` hook to include an option to `useSecondaryFilters` or not. Default to `true`, which will tell BE to return filtered secondary options. If `useSecondaryFilters` is `false`, BE will return all available secondary filter options, which is what we want for the Filters panel to show

As seen in the screenshot below, the left filter panel still shows all available datasets, but the right info panel's `Source Data` section only shows the filtered 2 datasets 

Screen Shot 2022-04-04 at 2 41 59 PM


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -477,7 +485,17 @@ function aggregateIdLabels(items: { [id: string]: string }[]): {
...
 }
 
-function useWMGQueryRequestBody() {
+const EMPTY_FILTERS = {
+  datasets: undefined,
+  developmentStages: undefined,
+  diseases: undefined,
+  ethnicities: undefined,
+  sexes: undefined,
+};
```
seve, line 494: Is there a type we can assign to this object to make sure it doesn't break if we add any filters in the future?

frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -488,7 +506,7 @@ function useWMGQueryRequestBody() {
   const { data } = usePrimaryFilterDimensions();
 
   const { datasets, developmentStages, diseases, ethnicities, sexes } =
-    selectedFilters;
+    includeAllFilterOptions ? EMPTY_FILTERS : selectedFilters;
```
seve, line 509: Can you comment explaining this ternary? Seems like it's the crux of these changes, but I don't seem to understand the implications

seve (changes requested):

frontend/src/common/queries/wheresMyGene.ts
tihuan: Ooh yes, really good point. Thank you! Done!

frontend/src/common/queries/wheresMyGene.ts
tihuan: Good call out thank you! Done!

tihuan:
PTAL again @seve 🙏 Thanks for the great feedback!

seve (approved):

PR #2274, opened 2022-04-05
chore: Added CEL-seq2 and drop-seq to assay list

pablo-gar:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add "CEL-seq 2" and "Drop=seq" to assays, careful examination of these methods indicate that they don't need gene-length normalization and that they could be integrated with the bulk of 10X data


MDunitz (approved):
lgtm

PR #2284, opened 2022-04-06
fix: workaround for WMG tiledb queries that fail on empty result

atolopko-czi:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


MDunitz (approved):
lgtm
thanks for including all the comments

PR #2285, opened 2022-04-07
fix: update call to test snapshot module in setup_dev_data.sh

danieljhegeman:
### Reviewers
**Functional:** 
@atolopko-czi
**Readability:** 

---

Dev data initialization needs to be updated after file renaming.


tihuan:
When there's only one commit, GH will use the commit message, so you'll have to update the commit message to follow semantic versioning format. Only multiple commits will GH use the PR title as the commit message for squash and merge!

PR #2286, opened 2022-04-07
feat: reject badly formatted DOIs

ebezzi:
### Reviewers
**Functional:** 
@MillenniumFalconMechanic 

**Readability:** 

---

## Changes
Rejects DOIs that aren't formatted properly.
Note: this uses the regex validation proposed in [this article](https://www.crossref.org/blog/dois-and-matching-regular-expressions/), which is quite strict but should work well for any newly published paper. This will also not allow using test DOIs anymore, which might be a concern (~even for smoke tests~ smoke tests are unaffected). Alternatively, we could use a weaker regex that only validates the `123/456` form. @brianraymor thoughts?


brianraymor:
@ebezzi - That's s nice find and _crossref approved_. And since the post is from 2015, I'd expect that all DOI(s) that the portal sees will be "modern". 

I support requiring a real DOI in all cases instead of something DOI-ish like `123/456`.

MillenniumFalconMechanic (approved):
LGTM, thanks Emanuele.

PR #2301, opened 2022-04-08
fix: cache primary filters

MDunitz:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- cache primary filter dimensions


ebezzi (approved):

backend/wmg/api/v1.py
```diff
@@ -21,6 +22,7 @@
 #  -portal/2132
 
 
+@cache(max_size=None)
```
ebezzi, line 25:
```suggestion
@cache
```

PR #2312, opened 2022-04-08
fix: database setup for rdev

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---
currently `db/init_remote_dev` and `db/delete_remote_dev` are failing because we aren't installing the requirements needed to run `scripts.populate_db`. This is causing 500 errors on the backend because the database is not created. It's also causing issues when tearing down an rdev because the database is not deleted.

## Changes
- removed extra click
- install requirements for scripts.populate_db

- `happy` correctly sets up and tears down the databases used by rdev environments


metakuni (approved):
Thanks for the fix!

PR #2318, opened 2022-04-11
fix: happy rdev auto_run_migrations=true

edulop91:
### Reviewers
Make sure we are running Happy migrations in rdev

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -9,6 +9,7 @@
             "aws_profile": "single-cell-dev",
             "secret_arn": "happy/env-rdev-config",
             "terraform_directory": ".happy/terraform/envs/rdev",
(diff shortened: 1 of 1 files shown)


MDunitz (approved):
lgtm

Bento007:
I think there was a good reason we didn't have migrate enable for rdev automatically. It may have been for testing migrations.

edulop91:
> I think there was a good reason we didn't have migrate enable for rdev automatically. It may have been for testing migrations.

https://github.com/chanzuckerberg/single-cell-data-portal/blob/109d981621a41ed05ef8ac781636c1f39044452c/scripts/happy#L938-L945

Looks like it used to default True when undefined

PR #2321, opened 2022-04-11
fix: check if expression summary is already an empty object

seve:
### Reviewers
**Functional:** @tihuan 

---

This doesn't include an error message if any condition causes this conditional to return true. We could split this conditional in to the expected, often, trues (loading), and the error states. Returning a broad warning for the others. Cut this PR now immediately  so we can unblock.


tihuan (approved):
Ahh I see, so `data` could be `{}` instead of `undefined` 💡 Thanks for finding this!

PR #2324, opened 2022-04-12
fix(wmg): 2272 FE Polish

tihuan:
### Reviewers
**Functional:** 
@hthomas-czi 
**Readability:** 

---

## Changes
- Tooltip labels should be `fontWeight: 400` and values should be `foregroundColor: .secondary`
- Less spacing between sections in the tooltip
- Add question mark icon next to Scaled, on hover add definition: "Expression is scaled to the range [0,1]. Scaling is done by assigning the minimum value in the current view to 0 and the max is assigned to 1." **<--- ⚠️ NOTE: Instead of question mark, I use info circle, since that's the closest available in SDS. Hopefully that's okay!**

Screen Shot 2022-04-11 at 6 55 21 PM

Screen Shot 2022-04-11 at 6 50 45 PM

demo

### DIFF
--- a/frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
@@ -283,6 +283,7 @@ export default memo(function Chart({
   return (
     <Wrapper height={heatmapHeight} width={heatmapWidth}>
       <Tooltip
+        width="wide"
         classes={tooltipClasses}
         title={tooltipContent || <>No data</>}
         leaveDelay={0}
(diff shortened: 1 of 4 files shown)


hthomas-czi:
Looks good to me! 

It'd be great if we could increase the width of the tooltip so the Expressed in Cells line doesn't wrap, but no problem if that's not easy to do.

tihuan:
Ahh yeah great idea! Let me increase the max width and play with it a little!

tihuan:
@hthomas-czi just updated the max table width to be `500px`, is that enough or should we make it even larger?

Thank you!

demo

hthomas-czi:
Thanks @tihuan! It looks good to me.

tihuan:
Awesome thanks, Harley! MERGINGG

PR #2327, opened 2022-04-12
fix: don't clear input on item select

seve:
### Reviewers
**Functional:** @tihuan 

---

https://user-images.githubusercontent.com/8716829/162878701-3465788a-76dd-40bf-87be-c305b8a74e7e.mov

I maintained the behavior where closing the popup clears the input

## Known Issues
Typing these input changes are a _**p a i n**_ feel free to commit a re-type if you choose

### DIFF
--- a/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
@@ -246,7 +246,6 @@ export default function QuickSelect<
             placeholder,
           }}
           inputValue={input}
-          // @ts-expect-error -- We ignore the first attribute and its a pain to type
           onInputChange={handleInputChange}
         />
       </Popper>


frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
```diff
@@ -230,13 +243,11 @@ export default function QuickSelect<
           renderOption={renderOption}
           onPaste={handlePaste}
           InputBaseProps={{
-            onChange: (
-              event: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>
-            ) => {
-              setInput(event.target.value);
-            },
```
seve, line 237: The `onInputChange` on the Autocomplete component passes more args than the raw `Input`

tihuan: Ah I see why it's hard to type now. Apparently when we added `onInputChange` on SDS `MenuSelect`, we didn't realize it's colliding with MUI Autocomplete's `onInputChange`. Let's not use `onInputChange` until SDS works with CZ GenEpi on removing this unintended API collision 👌 

Thanks for exposing the issue!

seve: In that case, how should we address this bug?

tihuan: PR out! https://github.com/chanzuckerberg/sci-components/pull/156

seve: Do I need to introduce any changes to address the sds changes?

tihuan: We can upgrade `sds` to the latest and that should get the fix!

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
```diff
@@ -230,13 +243,11 @@ export default function QuickSelect<
...
             placeholder,
           }}
+          inputValue={input}
+          // @ts-expect-error -- We ignore the first attribute and its a pain to type
```
tihuan, line 249: we should be able to remove this now too! Thank you!

seve: You'll have to explain later why the SDS changes remove the downstream typing issues 🤔

tihuan (approved):
Perfect! Thanks so much, Seve 🔥 🚀 !! Thanks for staying up late to get this done ✨ !!

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
tihuan: I think since SDS also declared `onInputChange`, TypeScript eventually kept both function signatures and treated them as function overloading. But somehow TS couldn't tell which function signature we wanted to use in our code, thus the typing issue

tihuan: Ah actually, TypeScript created an intersection type of both `onInputChange` interfaces instead of an overload, so the intersection is invalid 😆 

Screen Shot 2022-04-11 at 10 24 44 PM

PR #2328, opened 2022-04-12
refactor: use InputBaseProps

tihuan:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
Revert back to use `InputBaseProps` to add `onChange`, since `onInputChange` is not typed correctly and no longer needed, since we now expose `InputBaseProps`!


seve (changes requested):
see https://github.com/chanzuckerberg/single-cell-data-portal/pull/2327#discussion_r847941807

PR #2340, opened 2022-04-12
chore: bump black version in pre commit hook

danieljhegeman:
### Reviewers
**Functional:** 
@metakuni @seve 
**Readability:**


seve (approved):
Does it make sense to pin this? In both/either place.

danieljhegeman:
> Does it make sense to pin this? In both/either place.

I'm agnostic on this. Haven't spent enough time with linters and regular changes therein.

PR #2351, opened 2022-04-13
fix(local): Update port to 5005

Bento007:
### Reviewers
**Functional:** @metakuni 

---

## Changes
- modify the port used by dataportal backend from 5000 to 5005

- able to make a local deployment
- able to make an rdev

## QA steps (optional)
- `make local-rebuild` and verify you can access the backend and the frontend can access the backend.


metakuni (approved):
Thanks for fixing this properly!

PR #2354, opened 2022-04-13
chore: more wmg pipeline fixes

ebezzi:
### Reviewers
**Functional:** 
@atolopko-czi @MDunitz 

**Readability:**

### DIFF
--- a/.happy/terraform/envs/dev/main.tf
+++ b/.happy/terraform/envs/dev/main.tf
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.dev.single-cell.czi.technology"
   backend_url                  = "https://api.cellxgene.dev.single-cell.czi.technology"
   stack_prefix                 = ""
-  batch_container_memory_limit = 230000
+  batch_container_memory_limit = 60000
   backend_memory               = 50000
   frontend_memory              = 4096
   backend_instance_count       = 1
--- a/.happy/terraform/envs/stage/main.tf
+++ b/.happy/terraform/envs/stage/main.tf
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.staging.single-cell.czi.technology"
   backend_url                  = "https://api.cellxgene.staging.single-cell.czi.technology"
   stack_prefix                 = ""
-  batch_container_memory_limit = 60000
+  batch_container_memory_limit = 300000
   backend_instance_count       = 4
 
   wait_for_steady_state        = var.wait_for_steady_state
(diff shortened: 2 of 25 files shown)


.happy/terraform/envs/dev/main.tf
```diff
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.dev.single-cell.czi.technology"
   backend_url                  = "https://api.cellxgene.dev.single-cell.czi.technology"
   stack_prefix                 = ""
-  batch_container_memory_limit = 230000
+  batch_container_memory_limit = 60000
```
ebezzi, line 16: Should be enough for the dev cube

.happy/terraform/envs/stage/main.tf
```diff
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.staging.single-cell.czi.technology"
   backend_url                  = "https://api.cellxgene.staging.single-cell.czi.technology"
   stack_prefix                 = ""
-  batch_container_memory_limit = 300000
+  batch_container_memory_limit = 60000
```
ebezzi, line 16: Should be enough for the staging cube

atolopko-czi: If it's not too onerous, I would keep staging and prod settings in sync. We do intend to be able to mirror prod data to other envs at some point. Not sure if we're intending to do that for all or a subset of data though.

backend/wmg/data/snapshot.py
```diff
@@ -114,11 +114,23 @@ def _update_latest_snapshot_identifier() -> Optional[str]:
...
 
+def _build_data_path_prefix():
+    wmg_config = WmgConfig()
+    rdev_prefix = os.environ.get("REMOTE_DEV_PREFIX")
+    if rdev_prefix:
+        return rdev_prefix.strip("/")
+    elif "data_path_prefix" in wmg_config.config:
+        return wmg_config.data_path_prefix
+    else:
+        return ""
```
atolopko-czi, line 125: I'm wondering if this config-related complexity should be encapsulated in WmgConfig. We have a [precedent](https://github.com/chanzuckerberg/single-cell-data-portal/blob/1eca356857b5f7c1994cdfd1977523069cd26ed5/backend/corpora/common/corpora_config.py#L34) of doing that.

.happy/terraform/envs/stage/main.tf
ebezzi: Not too onerous, after all this is just a memory requirement and the staging batch environment has enough to satisfy 300G. If we decide to mirror the prod data, then it definitely makes sense to keep it at 300G.

ebezzi, line 16:
```suggestion
  batch_container_memory_limit = 300000
```

atolopko-czi (approved):

backend/wmg/data/snapshot.py
atolopko-czi: On 2nd thought, we probably need a better way overall to think about all the rdev-specific config hacks. This is well encapsulated here as well, since it's only for snapshot.

ebezzi: I'm happy to move this to `WmgConfig` if you prefer. I agree it's probably better.

Issue #2359, opened 2022-04-14
Use RankIt values with 3-cell count filtering


atolopko-czi:
WMG snapshot needs to use RankIt values instead of raw counts. After RankIt values are computed, any values that were derived from cell counts <=3 must be removed.

PR #2360, opened 2022-04-14
feat: wmg cube: use filtered rankit values

atolopko-czi:
#2359 

### Reviewers
**Functional:** 
@MDunitz  @ebezzi 

**Readability:** 

---

## Changes
- filter rankit values using raw expr count threshold
- use rankit values instead of raw counts when building cube
- rename "raw" corpus TileDB Array to "integrated"
- do not store raw count value anymore in "integrated" corpus Array (unused)
- introduce constants
- rename tdb_group -> corpus_path

## QA steps (optional)
Comparing tabula sapiens endothelial h5ad file from prod, our 3-cell count filtering matches Pablo's counts.

The code produces:
```
500-filtering: 91,686,560
3-cell filtering: 17,842,577
% filtered: 0.80539594
```

Pablo's test:
```
500-filtering: 91,696,343
% filtered: 0.8053559453292483
```

So we're quite close...not sure why 500 filtering is off by a touch, but I'm happy with this for now.


backend/wmg/data/wmg_cube.py
```diff
@@ -191,16 +193,16 @@ def reduce_X(tdb_group, start_time, cube_indices, *accum):
...
+            iterable = X.query(return_incomplete=True, order="U", attrs=["rankit"])
             future = None
             for i, result in enumerate(iterable.df[:]):
-                logger.info(f"reduce raw X, iter {i}, {time.time() - start_time}")
+                logger.info(f"reduce integrated expression data, iter {i}, {time.time() - start_time}")
                 if future is not None:
                     future.result()  # forces a wait
                 future = tp.submit(
                     coo_cube_pass1_into,
-                    result["data"].values,
```
atolopko-czi, line 203: one of the key fixes (note that we don't even store "data" anymore, so no possibility of mistakenly using it now)

backend/wmg/data/wmg_cube.py
```diff
@@ -252,7 +254,7 @@ def coo_cube_pass1_into(data, row, col, row_groups, sum_into, nnz_into, min_into
             cidx = col[k]
             grp_idx = row_groups[row[k]]
             sum_into[grp_idx, cidx] += val
-            nnz_into[grp_idx, cidx] += 1
```
atolopko-czi, line 255: the integrated corpus (for now) _can_ have zero values, so we must handle here

tests/unit/backend/wmg/data/test_load_corpus.py
```diff
@@ -65,8 +65,8 @@ def tearDown(self) -> None:
     def test__load_loads_all_datasets_in_directory(self, mock_load_h5ad, mock_vacuum, mock_consolidate):
         load(self.path_to_datasets, self.corpus_path)
         self.assertEqual(mock_load_h5ad.call_count, 2)
-        self.assertEqual(mock_vacuum.call_count, 4)
-        self.assertEqual(mock_consolidate.call_count, 4)
```
atolopko-czi, line 69: no longer storing "X" array (raw expression counts)

backend/wmg/data/cube_pipeline.py
```diff
@@ -19,26 +19,26 @@
...
-            file_path = f"{dataset_directory}/{dataset}/local.h5ad"
+            h5ad_file_path = f"{dataset_directory}/{dataset}/local.h5ad"
             load_h5ad(
-                file_path, group_name, validate
+                h5ad_file_path, corpus_path, validate
             )  # TODO Can this be parallelized? need to be careful handling global indexes but tiledb has a lock I think
             gc.collect()
 
         logger.info("all loaded, now consolidating.")
-        for arr_name in [f"{group_name}/{name}" for name in ["obs", "var", "raw", "X"]]:
```
atolopko-czi, line 36: We don't need "X" array anymore, AFAICT.  We only care about the `integrated` array.

backend/wmg/data/load_corpus.py
```diff
@@ -205,13 +215,20 @@ def save_raw(anndata_object: anndata.AnnData, group_name: str, global_var_index:
             )
             rankit_data = rankit_normalized_coo_sparse_raw_expression_matrix.data
 
-            array[rows, cols] = {"data": raw_data, "rankit": rankit_data}
+            # Filter out rankit values that were computed from expression values having raw count <= 3.
+
+            # TODO: Ideally, we would just *remove* these elements from rows,cols,rankit_data , but that would
+            #  require also adjusting the `obs` matrix and first_obs_idx, which are already updated. For now,
+            #  we will compute nnz without assuming all values are zero
+            rankit_data[rankit_data <= RANKIT_RAW_EXPR_COUNT_FILTERING_MIN_THRESHOLD] = 0
```
ebezzi, line 223: Should use:
```
nonzero_mask = np.array(A[A.nonzero()] <= min_value)[0]
rows = A.nonzero()[0][nonzero_mask]
cols = A.nonzero()[1][nonzero_mask]
A[rows, cols] = 0
A.eliminate_zeros()
```
as it is drastically more performant.

backend/wmg/data/cube_pipeline.py
atolopko-czi: If so, can also remove the empty array creation from `corpus_schema.py`

backend/wmg/data/load_corpus.py
atolopko-czi: even if we're _not_ eliminating the zeros?

backend/wmg/data/load_corpus.py
```diff
@@ -190,30 +198,47 @@ def save_raw(anndata_object: anndata.AnnData, group_name: str, global_var_index:
...
-            rankit_data = rankit_normalized_coo_sparse_raw_expression_matrix.data
+            raw_expression_coo_matrix = csr_sparse_raw_expression_matrix.tocoo(copy=False)
+            rows = raw_expression_coo_matrix.row + start + first_obs_idx
+            cols = global_var_index[raw_expression_coo_matrix.col]
+            raw_expr_counts_data = raw_expression_coo_matrix.data
+
+            # Compute RankIt
+            rankit_integrated_csr_matrix = rankit(csr_sparse_raw_expression_matrix)
+
+            zero_out_low_expression_count_values(rankit_integrated_csr_matrix, raw_expression_coo_matrix)
```
atolopko-czi, line 209: @MDunitz @ebezzi reinstated with updated implementation

backend/wmg/data/load_corpus.py
atolopko-czi: I've updated the implementation quite a bit now. For one, it seems to filter out ~80% of the rankit values, and thus it is cheaper to _build a new_ sparse matrix from the values that we intend to _keep_ rather than zeroing-out and removing the elements we want to filter out. In any case, both strategies are implemented, with the caller able to decide on the strategy. Though currently, the code is unconditionally using the "build new" strategy, since _making_ the decision is too expensive!  

Performance-wise,  a 5GB test dataset consumes ~1min total wall time for this filtering operation, which seems reasonable.  @pablo-gar is verifying whether the extent of the filtering is correct, but it does match what the WMG prototype code filtered out.

backend/wmg/data/load_corpus.py
```diff
@@ -177,43 +182,93 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
-            )
-            assert np.array_equal(
-                coo_sparse_raw_expression_matrix.col, rankit_normalized_coo_sparse_raw_expression_matrix.col
+            raw_expression_csr_matrix = sparse.csr_matrix(expression_matrix[start:end, :])
+
+            # Compute RankIt
+            rankit_integrated_csr_matrix = rankit(raw_expression_csr_matrix)
+
+            rankit_integrated_coo_matrix = filter_out_rankits_with_low_expression_counts(
+                rankit_integrated_csr_matrix, raw_expression_csr_matrix, expect_majority_filtered=True
```
atolopko-czi, line 204: the key addition: filtering out rankits

backend/wmg/data/load_corpus.py
```diff
@@ -177,43 +182,93 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
             )
-            rankit_data = rankit_normalized_coo_sparse_raw_expression_matrix.data
 
-            array[rows, cols] = {"data": raw_data, "rankit": rankit_data}
+            global_rows = rankit_integrated_coo_matrix.row + start + first_obs_idx
+            global_cols = global_var_index[rankit_integrated_coo_matrix.col]
```
atolopko-czi, line 208: now computing global_{rows,cols} _after_ rankit filtering is done, so that we can write out the reduced set of coords & data to the integrated corpus Array.

backend/wmg/data/load_corpus.py
```diff
@@ -177,43 +182,93 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
+        cols_to_keep = raw_counts.col[to_keep_mask]
+        rankits_to_keep = rankits.data[to_keep_mask]
+        rankits_filtered = coo_matrix((rankits_to_keep, (rows_to_keep, cols_to_keep)))
+    else:
+        to_filter_mask = ~to_keep_mask
+        rows_to_filter = raw_counts.row[to_filter_mask]
+        cols_to_filter = raw_counts.col[to_filter_mask]
+        rankits[rows_to_filter, cols_to_filter] = 0.0
+        rankits.eliminate_zeros()
+        rankits_filtered = rankits.tocoo()
```
atolopko-czi, line 263: this `else` block is currently unused, but I couldn't bring myself to delete it just yet. 😄

backend/wmg/data/load_corpus.py
```diff
@@ -22,21 +24,25 @@
 
 # Minimum number of expressed genes for a cell to be included in the corpus.
 # See the following document for further details:
-# https://github.com/chanzuckerberg/cellxgene-documentation/blob/pablo-gar/wheres-my-gene/scExpression/scExpression-documentation.md#removal-of-low-coverage-cells
-MIN_GENE_EXPRESSION_COUNT = 500
+# https://github.com/chanzuckerberg/cellxgene-documentation/blob/main/scExpression/scExpression-documentation.md#removal-of-low-coverage-cells
+GENE_EXPRESSION_COUNT_MIN_THRESHOLD = 500
```
MDunitz, line 28: nit UNIQUE_GENE_EXPRESSION_COUNT_MIN_THRESHOLD?

backend/wmg/data/load_corpus.py
pablo-gar: There's a slight discrepancy in numbers obtained from independent verification:

@atolopko-czi numbers from tabula sapiens endothelial:

filter <=3: 86% reduction
filter <=2: 77% reduction
filter <=1: 61% reduction

@pablo-gar  number from same data:
<=3 : 80.53559453292483 %
<=2 : 73.05934545284973 %
<=1 : 55.75715925770344 %

they are in the ballpark and this requires further investigation but it should not block this PR

atolopko-czi: Re-verified,  and we're in agreement! 👏  I may have been using a diff dataset originally.  🤦   Validation notes in PR summary now.

backend/wmg/data/wmg_cube.py
```diff
@@ -114,6 +115,7 @@ def load_data_into_cube(tdb_group, uri: str):
 
     cube_sum = np.zeros((n_groups, n_genes), dtype=np.float32)
     cube_nnz = np.zeros((n_groups, n_genes), dtype=np.uint64)
+    # TODO: These do not appear to be used. Remove?
     cube_min = np.zeros((n_groups, n_genes), dtype=np.float32)
```
MDunitz, line 119: these are used by coo_cube_pass1_into when it is called by reduce_x

backend/wmg/data/schemas/corpus_schema.py
```diff
@@ -19,6 +19,9 @@
 
 uint32_domain = (np.iinfo(np.uint32).min, np.iinfo(np.uint32).max - 1)
 
+# TODO: also define and use constants for obs and var array names
```
MDunitz, line 22: Do you mean turning obs and var into constants? or using a more informative constant name? do you think cells and genes are too specific as replacements?

backend/wmg/data/load_corpus.py
```diff
@@ -54,27 +60,27 @@ def validate_dataset_properties(anndata_object: anndata.AnnData) -> bool:
...
-    if is_dataset_already_loaded(dataset_id, group_name):
+    logger.info(f"Loading {h5ad_path}...")
+    dataset_id = get_dataset_id(h5ad_path)
+    if is_dataset_already_loaded(corpus_path, dataset_id):
         return
 
-    anndata_object = anndata.read_h5ad(h5ad)
+    anndata_object = anndata.read_h5ad(h5ad_path)
 
     # Apply a low expression gene cell filtering.
```
MDunitz, line 76:
```suggestion
    # Filter out cells with low coverage (less than GENE_EXPRESSION_COUNT_MIN_THRESHOLD unique genes expressed)
```

MDunitz (approved):
lgtm

backend/wmg/data/load_corpus.py
```diff
@@ -177,43 +182,93 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
+    # logger.info(f"filter_out_rankits_with_low_expression_counts(): use extract strategy={expect_majority_filtered}; "
+    #             f"decision time={end - start}")
+
+    start = time.time()
+
+    if expect_majority_filtered:
+        rows_to_keep = raw_counts.row[to_keep_mask]
+        cols_to_keep = raw_counts.col[to_keep_mask]
+        rankits_to_keep = rankits.data[to_keep_mask]
+        rankits_filtered = coo_matrix((rankits_to_keep, (rows_to_keep, cols_to_keep)))
```
atolopko-czi, line 256: @ebezzi This logic is what's now being used to perform the rankit filtering. I ran an ad hoc benchmark on a 1.3GB H5AD with ~91M expr values and it took ~1 sec to perform the filtering down to ~18M values. Unless there's some other data "profile" that I'm missing (dense matrix h5ads?), the performance is good. Let me know if you agree since you spent some time on understanding the optimizations needed for the earlier implementation. Thanks!

backend/wmg/data/schemas/corpus_schema.py
atolopko-czi: Those names certainly make sense! I did this refactoring on a new [branch/PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2377), but just saw this comment now. We can rename again there, so if you want to let's take the thread over to that PR. Resolving here.

backend/wmg/data/cube_pipeline.py
atolopko-czi: rm'ing on PR https://github.com/chanzuckerberg/single-cell-data-portal/pull/2377

backend/wmg/data/load_corpus.py
ebezzi: This seems pretty similar to the code in the else branch, so this looks good to me. The "naive" solution `sparse_matrix[sparse_matrix < 3]` would take several minutes and raise a warning (something like "comparison on sparse matrices should be avoided), so I believe this is fine.

ebezzi: The `eliminate_zeros()` was also needed for restoring the sparsity of the matrix, which I assume is equivalent to line 256 here.

ebezzi (approved):

backend/wmg/data/load_corpus.py
atolopko-czi: yup, exactly! thank you for reviewing!

PR #2380, opened 2022-04-18
fix: multiple assays support in wmg/extract.py

ebezzi:
### Reviewers
**Functional:** 
@MDunitz 

**Readability:** 

---


MDunitz (commented):
looks good, want to add a test for the multiple assay condition?

MDunitz (approved):
lgtm, thanks!

PR #2386, opened 2022-04-19
chore: set batch memory to 120G in dev for wmg

ebezzi:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.happy/terraform/envs/dev/main.tf
+++ b/.happy/terraform/envs/dev/main.tf
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.dev.single-cell.czi.technology"
   backend_url                  = "https://api.cellxgene.dev.single-cell.czi.technology"
   stack_prefix                 = ""
-  batch_container_memory_limit = 60000
+  batch_container_memory_limit = 120000
   backend_memory               = 50000
   frontend_memory              = 4096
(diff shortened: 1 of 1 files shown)


.happy/terraform/envs/dev/main.tf
```diff
@@ -13,7 +13,7 @@ module stack {
   frontend_url                 = "https://cellxgene.dev.single-cell.czi.technology"
   backend_url                  = "https://api.cellxgene.dev.single-cell.czi.technology"
   stack_prefix                 = ""
-  batch_container_memory_limit = 60000
+  batch_container_memory_limit = 120000
```
atolopko-czi, line 16: Usually prod resources need to be larger than dev!  It feels odd having to give in to dev's messy data state in this way.  Maybe comment why this is larger than prod? 😬

atolopko-czi (approved):
Comment suggestion, but if we have do it, we have to do it.

.happy/terraform/envs/dev/main.tf
ebezzi: Actually, prod has 300G!

atolopko-czi: Ha, ok! I feel a little better about this then. 😆

PR #2390, opened 2022-04-19
feat: added generic disabled tooltip to filters.

MillenniumFalconMechanic:
### Reviewers
**Functional:** 
@seve 

---

## Changes
- Add generic tooltip to disabled filters indicating filter has no values matching current filter criteria.


frontend/src/components/common/Filter/index.tsx
```diff
@@ -241,21 +236,6 @@ function filterCategoryValuesWithCount(
   return values.filter(({ count }) => count > 0);
 }
 
-/**
- * Returns true if category is not applicable.
- * @param categoryView
- * @returns true when category  is not applicable.
- */
-function isCategoryNA(categoryView: CategoryView): boolean {
```
MillenniumFalconMechanic, line 249: I moved the disabled checks from the filter component to the `useCategoryFilter` hook.

seve (approved):

seve (changes requested):
oops

frontend/src/common/hooks/useCategoryFilter.ts
```diff
@@ -821,14 +821,14 @@ function buildRangeCategoryView(
   rangeCategory: RangeCategory
 ): RangeCategoryView {
   // Build view model of range category.
-  const rangeView = {
+  const rangeView: RangeCategoryView = {
     key: categoryKey as CategoryKey,
```
seve, line 825: Don't we already have `categoryKey` typed as a result of the arg typing?

seve (approved):

frontend/src/common/hooks/useCategoryFilter.ts
MillenniumFalconMechanic: Good catch! I'll update here as well as the corresponding areas for select and ontology category views.

PR #2400, opened 2022-04-21
fix: remove memoization, remove visibility from query invalidations

danieljhegeman:
The usage of memoization for this particular function call appears to be
incorrect. Also, 'visibility' should have been removed from the key
array for the query invalidation call as part of the original primary
key migration.

### Reviewers
**Functional:** 
@seve @tihuan 
**Readability:** 

---

## Changes
- remove visibility from query invalidations, an instance of memoization

### DIFF
--- a/frontend/src/common/queries/datasets.ts
+++ b/frontend/src/common/queries/datasets.ts
@@ -6,11 +6,7 @@ import {
 } from "react-query";
 import { API_URL } from "src/configs/configs";
 import { API } from "../API";
-import {
-  DatasetAsset,
-  DatasetUploadStatus,
-  VISIBILITY_TYPE,
-} from "../entities";
+import { DatasetAsset, DatasetUploadStatus } from "../entities";
 import { apiTemplateToUrl } from "../utils/apiTemplateToUrl";
 import { USE_COLLECTION } from "./collections";
 import { DEFAULT_FETCH_OPTIONS, DELETE_FETCH_OPTIONS } from "./common";
@@ -67,11 +63,7 @@ export function useDeleteDataset(collection_uuid = "") {
 
   return useMutation(deleteDataset, {
     onSuccess: (uploadStatus: DatasetUploadStatus) => {
-      queryClient.invalidateQueries([
-        USE_COLLECTION,
-        collection_uuid,
-        VISIBILITY_TYPE.PRIVATE,
-      ]);
+      queryClient.invalidateQueries([USE_COLLECTION, collection_uuid]);
 
       queryClient.cancelQueries([USE_DATASET_STATUS, uploadStatus.dataset_id]);
 
--- a/frontend/src/views/Collection/components/DatasetTab/index.tsx
+++ b/frontend/src/views/Collection/components/DatasetTab/index.tsx
@@ -1,6 +1,5 @@
 import { Button, Intent, UL } from "@blueprintjs/core";
 import { IconNames } from "@blueprintjs/icons";
-import memoize from "lodash/memoize";
 import { FC, useState } from "react";
 import { useQueryClient } from "react-query";
 import { Collection, Dataset } from "src/common/entities";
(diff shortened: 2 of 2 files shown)


frontend/src/common/queries/datasets.ts
```diff
@@ -67,11 +63,7 @@ export function useDeleteDataset(collection_uuid = "") {
 
   return useMutation(deleteDataset, {
     onSuccess: (uploadStatus: DatasetUploadStatus) => {
-      queryClient.invalidateQueries([
-        USE_COLLECTION,
-        collection_uuid,
-        VISIBILITY_TYPE.PRIVATE,
-      ]);
+      queryClient.invalidateQueries([USE_COLLECTION, collection_uuid]);
```
danieljhegeman, line 66: removing visibility here fixes [2346](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2346)

frontend/src/views/Collection/components/DatasetTab/index.tsx
```diff
@@ -50,12 +49,9 @@ const DatasetTab: FC<Props> = ({
...
     datasets?.length > 0 || Object.keys(uploadedFiles).length > 0;
 
-  const invalidateCollectionQuery = memoize(
-    () => {
-      queryClient.invalidateQueries([USE_COLLECTION, collectionId, visibility]);
-    },
-    () => collectionId + visibility
-  );
+  const invalidateCollectionQuery = () => {
+    queryClient.invalidateQueries([USE_COLLECTION, collectionId]);
```
danieljhegeman, line 53: removing visibility here fixes [2381](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2381)

danieljhegeman:
@tihuan @seve I'm going to go ahead and merge this since rdev is having issues. But I would like you to evaluate the removal of the memoization in particular before we push to prod next week.

frontend/src/views/Collection/components/DatasetTab/index.tsx
tihuan: Ah maybe we can keep memoize and just remove `visibility`?

```ts
const invalidateCollectionQuery = memoize(
    () => {
      queryClient.invalidateQueries([USE_COLLECTION, collectionId]);
    },
    () => collectionId
  );
```

This way we don't create a new function every time we render, and forces `CollectionDatasetsGrid` and `DatasetsGrid` to re-render as well ♼!

tihuan (commented):
Thanks so much for the fixes, @danieljhegeman ! Just one tiny comment otherwise LGTM!

danieljhegeman:
> Thanks so much for the fixes, @danieljhegeman ! Just one tiny comment otherwise LGTM!

What is the intended purpose of that memoization? From my reading, that memoization would result in the `queryClient.invalidateQueries` function being called *only* on the very first invocation of `invalidateCollectionQuery`...

seve:
https://github.com/chanzuckerberg/single-cell-data-portal/pull/2400#discussion_r855401954
Aren't we not memorizing the function but the void response in this case?

If we wanted to memoize the function wouldn't it be:

```typescript
const invalidateCollectionQuery = memoize(
    () => () => {
      queryClient.invalidateQueries([USE_COLLECTION, collectionId]);
    },
    () => collectionId
  );
  ```

**Comment by seve** [comment]
Zoom Call Finding: let's use useCallback!

**Comment by seve** [comment]
instead this look something like:

```
const invalidateCollectionQuery = memoize(
    useCallback(() => {
      queryClient.invalidateQueries([USE_COLLECTION, collectionId]);
    },
   [collectionId, visibility]
  );
  ```

tihuan:
@danieljhegeman @seve ah we can drop `memoize` if we use `useCallback` directly!

```ts
const invalidateCollectionQuery = useCallback(() => {
      queryClient.invalidateQueries([USE_COLLECTION, collectionId]);
    },
   [collectionId])
```

seve:
> @danieljhegeman @seve ah we can drop `memoize` if we use `useCallback` directly!
> 
> 
> 
> ```ts
> 
> const invalidateCollectionQuery = useCallback(() => {
> 
>       queryClient.invalidateQueries([USE_COLLECTION, collectionId]);
> 
>     },
> 
>    [collectionId])
> 
> ```

Typo! Oops! Don't GH and zoom

danieljhegeman:
@tihuan yes, I will use `useCallback`!

tihuan:
Awesome!! Thanks so much again both ✋ 🔆 🤩

PR #2423, opened 2022-04-22
feat: cxg evolution infra

ebezzi:
### Reviewers
**Functional:** 
@metakuni @bkmartinjr 

**Readability:** 

---

## Changes
- Refactors the script so that the main function body is inside the backend package
- Adds a step function that can be called to trigger the reprocessing
- Builds on top of the existing dataset_processing infra (no new docker images/permissions are needed)

### DIFF
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -571,7 +571,7 @@ def main():
         elif step_name == "cxg_remaster":
             from backend.corpora.dataset_processing.remaster_cxg import process
 
-            process(f"{dataset_id}.cxg", os.environ["CELLXGENE_BUCKET"], dry_run=False)
+            process(dataset_id, os.environ["CELLXGENE_BUCKET"], dry_run=False)
         else:
             logger.error(f"Step function configuration error: Unexpected STEP_NAME '{step_name}'")
 
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -557,6 +557,10 @@ def main():
             from backend.corpora.dataset_processing.process_seurat import process
 
             process(dataset_id, os.environ["ARTIFACT_BUCKET"])
+        elif step_name == "cxg_remaster":
+            from backend.corpora.dataset_processing.remaster_cxg import process
+
+            process(dataset_id, os.environ["CELLXGENE_BUCKET"])
         else:
             logger.error(f"Step function configuration error: Unexpected STEP_NAME '{step_name}'")
 
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -560,7 +560,7 @@ def main():
         elif step_name == "cxg_remaster":
             from backend.corpora.dataset_processing.remaster_cxg import process
 
-            process(dataset_id, os.environ["CELLXGENE_BUCKET"])
+            process(dataset_id, os.environ["CELLXGENE_BUCKET"], dry_run=False)
         else:
             logger.error(f"Step function configuration error: Unexpected STEP_NAME '{step_name}'")
 
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -108,6 +108,17 @@ Once all conversion are complete, the conversion status for each file will be ei
     cxg_status = ConversionStatus.UPLOADED
     h5ad_status = ConversionStatus.UPLOADED
 }
+
+# Standalone processing steps
+
+## Seurat
+This is used to recompute the Seurat artifact in place, starting from the original h5ad. This is a state machine with a single
+state that mimics the Conversion step of the main step function.
+
+## CXG_Remaster
+This is used to migrate the cxg to a different, more performant format. This is a state machine with a single
+state that mimics the Conversion step of the main step function.
+
 """
 
 import logging
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -571,7 +571,7 @@ def main():
         elif step_name == "cxg_remaster":
             from backend.corpora.dataset_processing.remaster_cxg import process
 
-            process(dataset_id, os.environ["CELLXGENE_BUCKET"], dry_run=False)
+            process(f"{dataset_id}.cxg", os.environ["CELLXGENE_BUCKET"], dry_run=False)
         else:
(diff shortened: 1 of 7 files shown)


backend/corpora/dataset_processing/process.py
```diff
@@ -557,6 +557,10 @@ def main():
             from backend.corpora.dataset_processing.process_seurat import process
 
             process(dataset_id, os.environ["ARTIFACT_BUCKET"])
+        elif step_name == "cxg_remaster":
```
bkmartinjr, line 560: Maybe add to the state machine doc (comment) at top of file?

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+import psutil
+import tiledb
+
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def process(dataset_id: str, cellxgene_bucket: str):
+    """
```
bkmartinjr, line 21: this comment seems incorrect?

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+    var_extent = 2048
+    cell_order = "row"
+    tile_order = "col"
+    capacity = 128000
+    compression = 22
+    target_array = "X_new"
+    source_array = "X_old"
+    sparse_threshold = 25.0
+
+    compute(
```
bkmartinjr, line 57: stylistic nit - do not need to fix if you don't want to do so: I dislike long lists of positional args as it is easy to screw them up in a weakly typed language like Python.  Much prefer kwargs or passing a dict.

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+    compression,
+    cell_order,
+    tile_order,
+    capacity,
+):
+    X_extent = [obs_extent, var_extent]
+
+    ctx = create_fast_ctx(
+        {
+            "py.init_buffer_bytes": 32 * 1024**3,
```
bkmartinjr, line 91: what kind of environment (available RAM) will this run in?  This config requires in excess of 98GiB, which may be a lot.  I don't see anything in the queries that requires nearly that amount of pre-allocated memory.

ebezzi: The prod batch environment had 300G available, so this wouldn't be a concern. I agree that we shouldn't over-allocate though, I just copied this from the script. What do you think a reasonable size could be?

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+                capacity,
+            )
+            logger.info("created, starting to read...")
+            in_sparse = old_X.schema.sparse
+            with tiledb.open(f"{cxg}/{target_array}", "w") as new_X:
+                logger.info(f"sparsity analysis: {in_sparse} {out_sparse} {old_X.schema.sparse} {new_X.schema.sparse}")
+                if in_sparse and out_sparse:
+                    logger.info("sparse->sparse")
+                    i, chunk = 0, 200_000
+                    while i < old_X.shape[0]:
```
bkmartinjr, line 130: side note for future reference:  TileDB has incremental queries which obviate the need to do this kind of iterative & chunked reading.  No reason to change this code, but I wouldn't use this style for future/new code.  See the `return_incomplete` option on the tiledb.Array.query() function for more info.

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+    cell_order,
+    tile_order,
+    capacity,
+):
+    X_extent = [obs_extent, var_extent]
+
+    ctx = create_fast_ctx(
+        {
+            "py.init_buffer_bytes": 32 * 1024**3,
+            "sm.tile_cache_size": 2 * 1024**3,
```
bkmartinjr, line 92: is overriding with a fixed constant preferable to the default (which is 50% of the environments memory)?  See `create_fast_ctx` for the default.

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+
+def frac_mem(f):
+    mem_size = psutil.virtual_memory().total
+    return int(f * mem_size) // (1024**2) * (1024**2)
+
+
+def fast_config(config_overrides: dict = {}) -> dict:
+    # consolidation buffer heuristic to prevent thrashing: total_mem/io_concurrency_level, rounded to GB
+    io_concurrency_level = int(tiledb.Config()["sm.io_concurrency_level"])
+    consolidation_buffer_size = (
```
bkmartinjr, line 237: this was originally written for very large memory instances.  I don't know what environment it will run in, but if any have small available memory (eg, < 1GB), I suggest just using the default, or bumping the 10% to something larger.

bkmartinjr (approved):
Just a bunch of optional stylistic or comment issues - otherwise it LGTM.

backend/corpora/dataset_processing/remaster_cxg.py
bkmartinjr: Rough SWAG:  the queries are 3 columns, 200000 elements, 4 bytes each, so that is roughly 2.4MiB.  I think you could make it 1GiB and still have plenty of room to spare.

backend/corpora/dataset_processing/remaster_cxg.py
ebezzi: Honestly, I am not sure what's the rationale behind these values, I just got them from Ben's script. Since the default isn't used anywhere else, we should probably remove it and keep this value.

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -0,0 +1,250 @@
...
+            "py.init_buffer_bytes": 32 * 1024**3,
+            "sm.tile_cache_size": 2 * 1024**3,
+        }
+    )
+
+    if kind == "auto":
+        """
+        Determine target encoding (sparse or dense) from the data.
+        The "magic" sparsity threshold should in sync with
+        backend/corpora/dataset_processing/process.py::make_cxg()
```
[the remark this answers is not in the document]
ebezzi: Great point, I'll move it to a constant.

PR #2481, opened 2022-05-04
chore: add s3 upload lambda tests

metakuni:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


backend/corpora/common/utils/dl_sources/url.py
```diff
@@ -132,7 +132,6 @@ def file_info(self) -> dict:
         }
 
 
-# TODO: consider replacing S3URL with S3URI
```
metakuni, line 135: Keeping `S3URL` for testing purposes.

backend/corpora/dataset_submissions/app.py
```diff
@@ -12,15 +12,14 @@
 from backend.corpora.common.utils.exceptions import CorporaException
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
```
metakuni, line 15: Removed hardcoded log level so that we can change it via the `LOGLEVEL` environment variable.

backend/corpora/dataset_submissions/app.py
```diff
@@ -12,15 +12,14 @@
...
 UUID_REGEX = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
 
 
-def dataset_submissions_handler(s3_event: dict, context) -> None:
```
metakuni, line 19: Renamed for clarity.

Bento007 (approved):
Looks good. It would be nice if the source code was a part of this PR. Why weren't these tests included with the source code?

backend/corpora/dataset_submissions/app.py
metakuni: Oops, sorry, my mistake... I misread an article about this. In hindsight, no, we'd have to read such an environment variable and then set the log level using `logging.setLevel`. Would be a handy feature to have!

PR #2487, opened 2022-05-05
fix: flakey auth unit test

Bento007:
### Reviewers

**Readability:** @metakuni 

---

## Changes
- add comment to explain why the auth test might fail and how to fix it.
- TOKEN_EXPIRES increase


metakuni (approved):
Will one extra second always be enough?

Bento007:
A second in compute time is a lot. If not we ca n increase easily

PR #2492, opened 2022-05-06
feat: functional tests

MDunitz:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


tests/functional/backend/wmg/test_wmg_api.py
```diff
@@ -0,0 +1,80 @@
...
+                           "sex_ontology_term_ids": [],
+                           "tissue_ontology_term_ids": ["UBERON:0000178", "UBERON:0002048", "UBERON:0000029",
+                                                        "UBERON:0000970", "UBERON:0000362"]},
+                "include_filter_dims": True,
+                "snapshot_id": self.data["snapshot_id"]}
+        start_time = time.time()
+        res = requests.post(f"{self.api}/query", data=json.dumps(data), headers=headers)
+        end_time = time.time()
+        func_call_time_ms = end_time - start_time
+        self.assertGreater(MAX_RESPONSE_TIME_SECONDS, func_call_time_ms)
```
atolopko-czi, line 79: might also want to assert result size is "reasonably" large, and hasn't shrunk due to changed or bad data. This is more of a sanity check the func test itself is still valid, and not something the test is trying validate directly.

MDunitz: I added a check that the content isnt empty, but "reasonably large" is hard to define given that it will be different based on the environment/cube

atolopko-czi (changes requested):
LGTM! Just one Q/concern before approving: In which envs will these tests be run? I believe functional tests are _not_ run in prod, but this _could_ be (non-mutating), and probably should be since: 1) performance is specific to the data (snapshot), and we don't yet have reliable, automated env data mirroring in place; 2) can be affected by instance sizes, which could potentially get out of sync between staging and prod.

Non-critical suggestions:
* To avoid potentially frustrating transient failures, we could run queries multiple times and take averages (or median or P90 quantile).  
* Should we log the actual response times even when tests pass? Could be useful to have this data.

atolopko-czi:
> but I think we are starting to move away from functional tests? Maybe it would be better to pull those subtests out into a regression test? We can use the functional tests in dev/staging to ensure code changes dont break the api and then run the regression tests exclusively in prod to catch any changes in response time

Can you elaborate on "we are starting to move away from functional tests"?

It does seem reasonable to separate the testing goals into "API correctness" and "performance" (which I think you're calling "regression"?), and run the latter only in prod.

tests/functional/backend/wmg/test_wmg_api.py
```diff
@@ -0,0 +1,113 @@
...
+                ],
+            },
+            "include_filter_dims": True,
+            "snapshot_id": self.data["snapshot_id"],
+        }
+        with self.subTest("Test once"):
+            res = requests.post(f"{self.api}/query", data=json.dumps(data), headers=headers)
+            self.assertGreater(MAX_RESPONSE_TIME_SECONDS, res.elapsed.seconds)
+            self.assertEqual(res.status_code, requests.codes.ok)
+            self.assertGreater(len(res.content), 10)
```
atolopko-czi, line 105: Is this single-call test still needed if we're doing an avg-of-10 timing test below? The motivation for validating avg performance instead of single-call performance was to avoid test failures due to "bad luck" (e.g. cold cache). 

Also, if the  MAX_RESPONSE_TIME_SECONDS is currently set to a conservative (higher) value than needed to account for "bad luck" response times, maybe we can lower it if we're just doing a 10-call average.

MDunitz:
I think API correctness and performance is a good way of describing the split. In that case I was thinking that the 1- call would be used in staging/dev to quickly catch breaking changes to the api and then the 10 call would be run in prod to monitor performance

atolopko-czi (approved):
LGTM! Thanks for breaking up into separate func and perf tests.

Some optional suggestions:
I suppose the perf test could be further defensive using a `@unittest.skipIf(os.getenv('DEPLOYMENT_STAGE') == 'prod')`

Could be worth cross-linking the two related sets of tests (func & perf) with a comment.

tests/performance/test_wmg_apis.py
```diff
@@ -0,0 +1,68 @@
...
+
+from tests.functional.backend.wmg.fixtures import (
+    secondary_filter_common_case_request_data,
+    secondary_filter_extreme_case_request_data,
+)
+
+# Note that these tests share fixtures and general test paths with the wmg api functional tests
+
+
+@unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "prod")
```
seve, line 15: https://docs.python.org/3/library/unittest.html#unittest.skipIf

PR #2517, opened 2022-05-10
fix: exclude datasets with multiple organisms

MDunitz:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes

- modify dataset selection function to exclude datasets with multiple organisms

### DIFF
--- a/backend/wmg/data/extract.py
+++ b/backend/wmg/data/extract.py
@@ -29,16 +29,16 @@ def get_dataset_s3_uris():
         dataset_ids = []
         published_dataset_non_null_assays = (
             session.query(Dataset.table.id, Dataset.table.assay, Dataset.table.organism)
-                .join(Dataset.table.collection)
-                .filter(
+            .join(Dataset.table.collection)
+            .filter(
                 Dataset.table.assay != "null",
                 Dataset.table.published == "TRUE",
                 Dataset.table.is_primary_data == "PRIMARY",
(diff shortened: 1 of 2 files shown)


backend/wmg/data/extract.py
```diff
@@ -29,20 +28,23 @@ def get_dataset_s3_uris():
...
-            session.query(Dataset.table.id, Dataset.table.assay)
+            session.query(Dataset.table.id, Dataset.table.assay, Dataset.table.organism)
             .join(Dataset.table.collection)
             .filter(
                 Dataset.table.assay != "null",
                 Dataset.table.published == "TRUE",
                 Dataset.table.is_primary_data == "PRIMARY",
                 Collection.table.visibility == "PUBLIC",
                 Dataset.table.tombstone == "FALSE",
+                Dataset.table.organism != "null",
```
atolopko-czi, line 39: do the null and empty organisms cases also need tests?

atolopko-czi (approved):
LGTM, though possibly missing tests for edge cases (see comment)

backend/wmg/data/extract.py
MDunitz: our schema should check for that so its unlikely to make it to the db but I noticed that null cases could break the db query for assays so I figured I would include the check for organisms too.
added a test case just to be safe

ebezzi (approved):

PR #2546, opened 2022-05-17
Build frontend before deploying

Bento007:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/frontend/container_init.sh
+++ b/frontend/container_init.sh
@@ -4,15 +4,13 @@ echo " ====="
 echo "| starting frontend container"
 echo " ====="
 echo
-if [ ! -e ./node_modules ]; then
-  mkdir -p ./node_modules
-fi
-ln -sf /opt/node_app/node_modules/* /opt/node_app/node_modules/.bin ./node_modules/.
-if [ ! -z "$API_URL" ]; then
-  cat src/configs/$DEPLOYMENT_STAGE.js | envsubst >src/configs/configs.js
-else
-  cp src/configs/local.js src/configs/configs.js
-fi
+#mkdir -p ./node_modules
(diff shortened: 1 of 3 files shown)


danieljhegeman:
I thought this got merged a while back 😬 😅

tihuan:
@seve this PR is not working yet. Do you want to take a stab at it? This hope is that GHA has more memory to build the FE image successfully, so we don't need to build FE in ECS, which is constantly crashing with doc-site generation!

tihuan:
@ebezzi pointed out that since images are shared across different envs, we need to make sure build time variables get moved to runtime, and Nextjs consumes the variables at runtime as well!

frontend/container_init.sh
```diff
@@ -4,15 +4,6 @@ echo " ====="
...
 echo
-if [ ! -e ./node_modules ]; then
-  mkdir -p ./node_modules
-fi
-ln -sf /opt/node_app/node_modules/* /opt/node_app/node_modules/.bin ./node_modules/.
-if [ ! -z "$API_URL" ]; then
-  cat src/configs/$DEPLOYMENT_STAGE.js | envsubst >src/configs/configs.js
-else
-  cp src/configs/local.js src/configs/configs.js
-fi
```
seve, line 15: Erroneously removed?

Bento007: moved to config.sh

Bento007: and line 27 of frontend/dockerfile

seve:
When this is merged, we need to revert the memory increases we did as a temporary workaround

Issue #2338, opened 2022-04-12
scExpression: Secondary filters affect each other but not themselves


signechambers1:
A filter's available options should be limited only by the options selected in other filters.

Example ideal work flow:

Assume we have male and female data for African, female data only for European, and male data only for Asian.
1. User selects organism, tissue, genes
2. User selects "male" from Sex filter. User continues to see male and female in the dropdown for Sex.
3. User clicks on Ethnicity, they see African and Asian as options.
4. User selects Asian. User continues to see African and Asian as options.
5. User clicks back to Sex and only sees male as an option.
6. User clears male as a filter, user now sees African, European and Asian as options in Ethnicity.

From slack, we may need a separate internal query per filter. [Relevant slack thread](https://czi-sci.slack.com/archives/C0247APK621/p1650920787830669)

tihuan:
Slack discussion now for likely implementing this in BE:

https://czi-sci.slack.com/archives/C0247APK621/p1650920787830669

tihuan:
@signechambers1 I added backend label on this. Not sure if BE will commit to getting this done in Q2? Thank you!

PR #2556, opened 2022-05-18
feat: secondary filters

MDunitz:
#2338 
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes

- modify query api to not filter based on category for secondary filters

### DIFF
--- a/tests/unit/backend/wmg/api/test_v1.py
+++ b/tests/unit/backend/wmg/api/test_v1.py
@@ -632,9 +632,7 @@ class WmgApiV1Tests(unittest.TestCase):
                 expected_development_stage_terms = [
                     {"development_stage_ontology_term_id_0": "development_stage_ontology_term_id_0_label"},
                 ]
-                expected_ethnicity_term  = [
-                    {"ethnicity_ontology_term_id_0": "ethnicity_ontology_term_id_0_label"}
-                ]
+                expected_ethnicity_term = [{"ethnicity_ontology_term_id_0": "ethnicity_ontology_term_id_0_label"}]
                 ethnicity_1_request = dict(filter=ethnicity_1_filter, include_filter_dims=True)
                 response = self.app.post("/wmg/v1/query", json=ethnicity_1_request)
                 dev_stage_terms_no_dev_filter = json.loads(response.data)["filter_dims"]["development_stage_terms"]
--- a/tests/unit/backend/wmg/api/test_v1.py
+++ b/tests/unit/backend/wmg/api/test_v1.py
@@ -481,7 +481,7 @@ class WmgApiV1Tests(unittest.TestCase):
             # setup up API endpoints to use a mocked cube
             with tempfile.TemporaryDirectory() as cube_dir:
                 expression_summary_cube_dir = f"{cube_dir}/{EXPRESSION_SUMMARY_CUBE_NAME}"
-                tiledb.from_pandas(uri=expression_summary_cube_dir, dataframe=df)
+                tiledb.from_pandas(uri=expression_summary_cube_dir, dataframe=df, sparse=True)
                 with tiledb.open(expression_summary_cube_dir, ctx=create_ctx()) as expression_summary_cube:
                     snapshot.expression_summary_cube = expression_summary_cube
                     load_snapshot.return_value = snapshot
--- a/tests/unit/backend/wmg/api/test_v1.py
+++ b/tests/unit/backend/wmg/api/test_v1.py
@@ -474,7 +474,7 @@ class WmgApiV1Tests(unittest.TestCase):
             # check that rows were dropped
             self.assertGreater(snapshot.expression_summary_cube.df[:].shape[0], df.shape[0])
             # setup up API endpoints to use a mocked cube
-            snapshot.expression_summary_cube = df
+            snapshot.expression_summary_cube.df[:] = df
             load_snapshot.return_value = snapshot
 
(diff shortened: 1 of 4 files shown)


tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -555,19 +555,23 @@ def test__query_request_with_filter_dims__returns_elided_filter_dims(
                         "label": "dataset_id_0_name",
                         "collection_label": "dataset_id_0_coll_name",
                         "collection_id": "dataset_id_0_coll_id",
-                    },
```
atolopko-czi, line 558: ⚠️ This should also have `dataset_id_1`. This may just be a side-effect of the test setup, and line 531 may just need to mock the response to have both datasets returned.

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -515,7 +515,7 @@ def test__query_request_with_filter_dims__returns_valid_filter_dims(
     @patch("backend.wmg.api.v1.gene_term_label")
     @patch("backend.wmg.api.v1.ontology_term_label")
     @patch("backend.wmg.api.v1.load_snapshot")
-    def test__query_request_with_filter_dims__returns_elided_filter_dims(
+    def test__query_request_with_filter_dims__secondary_filters_dont_filter_themselves(
```
atolopko-czi, line 518: ⚠️ I think we will need to update tests more substantially to properly test the new logic. Note that the cube contains the cross-product of all dimension values. With self-filtering removed, the secondary filter dim values in the result are actually no different than if we had specified no secondary filtering at all. So this test can longer faithfully verify that secondary filter dims are being elided. We'll need a more "interesting" cube.  E.g., all rows where disease=X have only ethnicity=Y, .  Then we can filter on disease X and assert that 1) ethnicity secondary dim values are correctly elided and 2) disease secondary values contain all diseases.  These two assertions are better off as  separate test methods.  Let me get back to you on suggestions for how to setup the cube for such tests.

atolopko-czi (changes requested):
The implementation looks good! The tests do need to be extended for proper coverage, per comment.

tests/unit/backend/wmg/api/test_v1.py
atolopko-czi: The [create_temp_wmg_snapshot](https://github.com/chanzuckerberg/single-cell-data-portal/blob/7438d2de77ee7bd345e1fc576347bc5d70a3da99/tests/unit/backend/wmg/api/test_v1.py#L56) method  has an keyword param `exclude_logical_coord_fn`, which takes a function that allows one to prune rows from what would otherwise be a cube that contains one row for _every_ combination of dimension values (i.e. the rows are the cross-product all dimensions' values).  This _could_ be used to specify a function that to create the necessary cube. It's not a beautiful solution for cube data setup, but it's serviceable. We might want to put more effort in writing test fixture code for generating cubes that have specific data requirements. I can help further on this, if asked!

tests/unit/backend/wmg/api/test_v1.py
MDunitz: good catch. I think you are right that the value is just based on the mock on line 531. At this point it feels like I'm mostly testing the mock which def isnt ideal. I'm working on implementing a test as you described above to more throughly test the secondary filters. Do you think its even worth keeping this test?

atolopko-czi: If the new test is asserting both of the following, then yes, I'd say we can safely eliminate this test: 
1) when a secondary dimension has criteria, the remaining filter values for other secondary dimensions are properly restricted (preferably exercising _all_ the other secondary dimensions)
2) when a secondary dimension has criteria, it's own values are not restricted (preferably testing this for all secondary dimensions as well)

It's generally nice to break up different behavioral requirements into separate tests in order to capture the requirements explicitly, but a single test will suffice if it's explaining all the things it's testing.

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -441,137 +445,235 @@ def test__query_missing_tissue_request__returns_400(self):
     def test__query_request_with_filter_dims__returns_valid_filter_dims(
         self, load_snapshot, ontology_term_label, gene_term_label, fetch_datasets_metadata
     ):
+        # mock the functions in the ontology_labels module, so we can assert deterministic values in the
```
Bento007, line 448: I'm having trouble stepping through the code. Are you able to run this test under the debugger?

MDunitz: discussed over slack

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -100,6 +100,15 @@ def exclude_random_coords_75pct(_) -> bool:
     return random() > 0.75
 
 
+def exclude_dev_stage_and_ethnicity_for_secondary_filter_test(coord) -> bool:
+    dev_stages_to_exclude = ("development_stage_ontology_term_id_1", "development_stage_ontology_term_id_2")
+    ethnicity_terms_to_exclude = ("ethnicity_ontology_term_id_1", "ethnicity_ontology_term_id_2")
+    if coord.development_stage_ontology_term_id in dev_stages_to_exclude:
```
MDunitz, line 106: This and the following line could be put together with `and` but then the linter gets mad

seve (approved):

Issue #2501, opened 2022-05-09
A user should be able to select either ontological cell type ordering


tihuan:
JSON structure:

```ts
{tissue_id : [{id = "", name = "", depth = 10}, ... ], ...} 
```

tihuan:
@signechambers1 is it fair to say ontological cell type ordering is p1 or is it actually p0? Thank you!

signechambers1:
P1 sounds good! Thanks @tihuan!

tihuan:
Got it! Thanks so much, Signe! If we can get BE done this week, FE should be able to get it done the next week!

tihuan:
https://github.com/chanzuckerberg/single-cell-data-portal/pull/2600

PR #2560, opened 2022-05-18
feat: Add depths to cell types for dot plot

pablo-gar:
- #2501

## Changes
- modify backend to create a tree structure for cell type ordering of scExpression
- update `generate_cell_ordering` from `transform.py` to write a pandas dataframe into a json file with the following columns `"tissue_ontology_term_id", "cell_type_ontology_term_id", "depth", "order"]`. The column `depth` encodes the level in the tree where a cell type is
- update `build_ordered_cell_types_by_tissue` of `v1.py` to work with the file mentioned above, the api request now returns a json file of the following structure `{tissue_id : [{id = "", name = "", depth = 10}, ... ], ...} `

### DIFF
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from typing import Dict, List, Any, Iterable
+from math import isnan
 
 import connexion
 from flask import jsonify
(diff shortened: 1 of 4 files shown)


backend/wmg/data/transform.py
```diff
@@ -65,18 +65,25 @@ def compute_ordering(cells, root):
...
-        def recurse(node):
+        def recurse(node: Set[str], depth=0):
+
             if node in cells:
-                yield (node)
+
+                cells.remove(node)
+                yield {"id": node, "depth": depth}
+
+                if node != "CL:0000003":
```
MDunitz, line 75: why are we checking for this particular id?

backend/wmg/api/v1.py
```diff
@@ -152,18 +153,49 @@ def build_ordered_cell_types_by_tissue(
...
+                "cell_type_ontology_term_id": row.cell_type_ontology_term_id,
+                "cell_type": ontology_term_label(row.cell_type_ontology_term_id),
+                "depth": row.depth,
+            }
         )
 
     return structured_result
+
+
+def build_ordered_cell_types_by_tissue_fix_depths(x):
```
MDunitz, line 181: can you use a more informative param name?

pablo-gar: done

MDunitz (approved):

backend/wmg/data/transform.py
pablo-gar: "CL:0000003" is "naive cell", which we use for "unknown" cell types. However "naive cell" is the parent of virtually all other cell types, so we don't want put everything under naive cell.

In other words the CL structure looks like this:

-cell
---naive cell
------all other cells
------ [...]

but we want to make it look this way:
-cell
--naive celll
--all other cells
--[...]

PR #2564, opened 2022-05-19
feat(curation api): Add PUT /v1/collections/{collection_uuid}/datasets/upload-link Endpoint 

Bento007:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add new tests for curation api
- update existing test
- add function to get dataset from curation tag and collection uuid.
- check for duplicate curation tags before creating a new dataset.
- move check for collection permission to upload_from_link to simplify unit testing.
- update relink to support a link passed in as either url or link.
-

### DIFF
--- a/backend/corpora/common/entities/dataset.py
+++ b/backend/corpora/common/entities/dataset.py
@@ -97,6 +97,15 @@ class Dataset(Entity):
         return dataset
 
     @classmethod
+    def get_dataset_from_curator_tag(cls, session: Session, collection_id, curator_tag) -> "Dataset":
+        dataset = (
+            session.query(cls.table)
+            .filter(cls.table.collection_id == collection_id, cls.table.curator_tag == curator_tag)
+            .one_or_none()
+        )
+        return cls(dataset) if dataset else None
+
+    @classmethod
     def get_by_explorer_url(cls, session: Session, explorer_url):
         """
         Return the most recently created dataset with the given explorer_url or None
--- a/backend/corpora/common/upload.py
+++ b/backend/corpora/common/upload.py
@@ -16,6 +16,7 @@ from .utils.exceptions import (
     NonExistentCollectionException,
     InvalidProcessingStateException,
     NonExistentDatasetException,
+    DuplicateTagExistsException,
 )
 from .utils.math_utils import GB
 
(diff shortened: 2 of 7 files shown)


backend/corpora/common/entities/dataset.py
```diff
@@ -96,6 +96,15 @@ def get(cls, session: Session, dataset_uuid, include_tombstones=False) -> "Datas
                 return None
         return dataset
 
+    @classmethod
+    def get_dataset_from_curator_tag(cls, session: Session, collection_id, curator_tag) -> "Dataset":
+        dataset = (
+            session.query(cls.table)
+            .filter(cls.table.collection_id == collection_id, cls.table.curator_tag == curator_tag)
```
metakuni, line 103: This is what I _should_ have done for the [dataset submission lambda](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/corpora/dataset_submissions/app.py#L103). 🤦

backend/corpora/common/upload.py
```diff
@@ -89,6 +90,8 @@ def upload(
                 dataset.reprocess()
 
     else:
+        if curator_tag and Dataset.get_dataset_from_curator_tag(db_session, collection_uuid, curator_tag):
```
metakuni, line 93: The API doc should be updated to reflect this behavior.

metakuni (approved):

backend/corpora/common/upload.py
danieljhegeman: will do

danieljhegeman: This is the wrong behavior. If the tag already exists (i.e., the second clause of this statement evaluates to true), then we need to update/replace that dataset rather than throw an exception.

backend/corpora/lambdas/api/v1/collection_uuid/upload.py
```diff
@@ -17,22 +20,32 @@
...
     return make_response({"dataset_uuid": dataset_id}, 202)
 
 
 @dbconnect
-def upload_from_link(collection_uuid: str, token_info: dict, url: str, dataset_id: str = None):
+def upload_from_link(collection_uuid: str, token_info: dict, url: str, dataset_id: str = None, curator_tag: str = None):
     db_session = g.db_session
+    get_collection(
+        db_session, collection_uuid, visibility=CollectionVisibility.PRIVATE.name, owner=owner_or_allowed(token_info)
+    )
```
danieljhegeman, line 48: So we're using this to check ownership? I think we should add a comment here, otherwise it seems like this statement has no effect.

backend/corpora/common/upload.py
danieljhegeman: @Bento007 tagging to make sure you see

PR #2571, opened 2022-05-20
feat(docs): add doc navigation to left side bar

seve:
### Reviewers
**Functional:** @tihuan 

---

## Changes
- add
  - add left side bar navigation menu 
  - add unsafe-eval, signed off by arathi, awaiting convo with sec 
- modify
  - Re-enable docs 
  - Add new method of traversing and saving file structure to data struct

## Notes
* Currently not showing active page, will be adding that later


frontend/next.config.js
```diff
@@ -7,7 +7,8 @@ const isProdBuild = process.env.NODE_ENV === nodeEnv.PRODUCTION;
 
 const PLAUSIBLE_URL = "https://plausible.io";
 
-const SCRIPT_SRC = ["'self'", PLAUSIBLE_URL];
+// unsafe-eval is required for next-mdx-remote. Currently pending sec review
+const SCRIPT_SRC = ["'self'", "'unsafe-eval'", PLAUSIBLE_URL];
```
tihuan, line 11: @jakeyheath is it okay to use `unsafe-eval` here? My impression is that this opens up to XSS, but not sure if it's a valid concern in our context 😆 Thank you!

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -1,75 +1,158 @@
...
+  files: Array<string>;
+  slug: Array<string>;
+  subDirectories: Array<Directory>;
+}
+
+const CACHED_FILE_PATHS = new Map<Array<string>, Directory>();
+function filePaths(...root: Array<string>): Directory {
+  const cacheStore = CACHED_FILE_PATHS && CACHED_FILE_PATHS.get(root);
+  if (cacheStore) {
+    console.log("CACHE HIT");
```
Comment (tihuan) [inline frontend/src/pages/docs/[...slug].tsx]: remove!

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -1,75 +1,158 @@
...
+    props: {
+      filePath,
+      frontMatter,
+      mdxSource,
+      slug,
+    },
+  };
+};
+
+const RenderDirectory = ({ directory }: { directory: Directory }) => {
```
Comment (tihuan) [inline frontend/src/pages/docs/[...slug].tsx]: nit: Since this is a component, maybe just `Directory`?

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -1,75 +1,162 @@
...
+  return <RenderDirectory directory={filePath} />;
+};
+
+const StyledDocsLayout = styled.div`
+  display: flex;
+  flex-direction: row;
+`;
+const DocContent = styled.div`
+  margin: 16px;
+`;
```
Comment (seve) [inline frontend/src/pages/docs/[...slug].tsx:148]: I'm leaving these in here for a bit until we there's actual styles

tihuan (commented):
Love your Directory solution! 👍 ⭐ Let's wait to see what Jake says about CSP!

frontend/src/pages/docs/[...slug].tsx
Comment (seve) [reply frontend/src/pages/docs/[...slug].tsx]: you're right!

frontend/next.config.js
seve: https://chanzuckerbergteam.slack.com/archives/C02H8TRBPC7/p1652731472038849

jakeyheath: sorry I am not sure if yall got your answer here. last week, we were having our central all hands so I was a bit AFK. Ideally, I wouldn't want to see this, but sounds like there isn't a way around this library. They specifically call it out as a requirement 🤷🏻 . 

Generally, CSP is a defense-in-depth. You are already protected from XSS by using React.  Try not to use dangerouslySetInnerHTML and you should be good. Maybe periodically review that library to see if they remove that need for unsafe-eval as a requirement.

tihuan: Thanks @jakeyheath for your detailed answer! We do have another option to use (https://nextjs.org/docs/advanced-features/using-mdx), so maybe it's worth investigating it further?

@seve what's the eng overhead for using@next/mdx again vs. the hashicorp one?

So far I see it's the part that we need to inject doc site layout to every MDX files. Is there anything else?

Thank you!

tihuan: Also [Data Portal scExpression](https://cellxgene.cziscience.com/scExpression) uses a third-party library ([echarts](https://github.com/apache/echarts)) to render data viz DOM elements, so it's outside of React's control too!

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -1,75 +1,162 @@
...
+  directory: Directory;
+}) {
+  return (
+    <ul>
+      {directory.files.map((file) => {
+        let href = "/docs/";
+        if (directory.slug.length > 0) href += directory.slug.join("/") + "/";
+        href += file;
+        return (
+          <NextLink key={file} href={href}>
```
Comment (tihuan) [inline frontend/src/pages/docs/[...slug].tsx:112]: I think we need `passHref` prop to pass the URL, otherwise the `<a />` link doesn't get `href`!

Screen Shot 2022-05-23 at 2 42 54 PM

PR #2574, opened 2022-05-23
style: use proper terminology for s3 upload key prefix

danieljhegeman:
We were referring to what is techically a 'key prefix' as a 'path'.
Changing because it is surfaced to users as an attribute in the response
body.

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- modify var name

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -425,7 +425,7 @@ paths:
                         type: string
                       SessionToken:
                         type: string
-                  UploadPath:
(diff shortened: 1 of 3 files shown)


danieljhegeman:
kindly merge[ this PR](https://github.com/chanzuckerberg/single-cell-curation/pull/224) for the corresponding change in the notebook as well

Bento007 (approved):

Issue #2580, opened 2022-05-23
Links disappearing upon re-publish for some Collections


danieljhegeman:
An issue [was reported](https://czi-sci.slack.com/archives/C024HCSH9PT/p1653059553034979) with links disappearing from revisions of Collections upon publishing.

Issue likely stems from [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/f925286dcb8371821abdce4a9f85a0e5f05aab33/backend/corpora/common/entities/collection.py#L261-L263)

PR #2583, opened 2022-05-24
fix: missing links

danieljhegeman:
- #2580 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- modify the publishing step for revisions

## QA steps (optional)
- verified in rdev (bug and fix)


backend/corpora/common/entities/collection.py
```diff
@@ -260,7 +260,7 @@ def publish(self, data_submission_policy_version):
             )
             for link in self.links:
                 CollectionLink(link).update(collection_id=self.revision_of, commit=False)
-            public_collection.links = self.links
+            self.session.expire(public_collection)
```
danieljhegeman, line 263: @Bento007 you were right! I thought this code was correct initially for a different reason, but yeah this was overwriting the public collection's links with soon-to-be-deleted entities.

tests/unit/backend/corpora/api_server/collection_uuid/test_publish.py
```diff
@@ -151,6 +150,8 @@ def test__publish_collection_with_links__OK(self):
             ],
             published_at=self.mock_published_at,
         )
+        self.generate_dataset(self.session, collection_id=collection.id, published_at=self.mock_published_at)
+
```
danieljhegeman, line 154: tangentially related refactor

danieljhegeman:
@Bento007 still need to add tests to test this issue. Attempted to do so today but our test db is a bit of a mess and I couldn't seem to tune things just right. Will need to take another pass at it tomorrow, but we should get this into staging asap.

backend/corpora/common/entities/collection.py
danieljhegeman: Technically we don't, as the code currently stands. I did this to prevent mistakes from being made in the future, in case someone adds further link logic in this method: with all other entity modifications (such as the `update` call), updates will be available immediately, but with the link reassignment, those won't be available unless you `expire` this object. So trying to keep the behavior consistent.

Bento007: If you expire the object without committing the changes will the changes persist?

Bento007: I suggest removing it.

danieljhegeman: > If you expire the object without committing will the changes persist?

Yes, the `expire` just tells sqlalchemy to refetch the object (and its relations) in all subsequent usages of `public_collection`.

backend/corpora/common/entities/collection.py
```diff
@@ -260,7 +260,7 @@ def publish(self, data_submission_policy_version):
             )
             for link in self.links:
                 CollectionLink(link).update(collection_id=self.revision_of, commit=False)
-            public_collection.links = self.links
+            # self.session.expire(public_collection) <-- uncomment to update the public collection object with links
```
Bento007, line 263: Isn't this what the code above is doing?

Bento007 (approved):

PR #2593, opened 2022-05-24
chore(fmt): add lint to make fmt recipe

Bento007:
### Reviewers
**Functional:** @MDunitz 

---

## Changes
- Black doesn't always find everything flake8 is looking for. This should speed up development by reducing the number of times we commit code that fails the linter.
- Applied flake8 and black changed to the /scripts directory.


MDunitz (approved):
lgtm

PR #2598, opened 2022-05-25
fix: expand 400_multiple response specification for flexibility

danieljhegeman:
- #TICKET_NUMBER

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- modify 400_multiple error response so as to make the inclusion of link errors easier to fit in the specification


danieljhegeman:
Trent implemented separately.

PR #2609, opened 2022-05-26
fix: Revert "feat(wmg): 2501 Cell Grouping"

tihuan:
Reverts chanzuckerberg/single-cell-data-portal#2600


tihuan:
Actually, `dev` is working after we copy the WMG cubes to `dev` env in S3!

PR #2613, opened 2022-05-26
chore: Json Logging

Bento007:
- #TICKET_NUMBER

### Reviewers
**Functional:** @atolopko-czi @ebezzi 

**Readability:** @MDunitz 

---

## Changes
- use dictConfig to setup python logging.
- use python-json-logging to convert logs to json.
- log before request and after request.
- add line number, files path, and function name to log message

## QA steps (optional)
1. go to https://jsonlogs-frontend.rdev.single-cell.czi.technology
1. `happy logs jsonlogs backend` to see the json formatted logs.
## Notes for Reviewer
Typical log record output:
```
2022-05-27T12:32:35.017000+00:00 2dc72b2253584184e51f580997e11f1ff24b8b2ce53ba336797e18227317373c { 
"levelname": "INFO", 
"asctime": "2022-06-30T19:49:42.030Z", 
"name": "corpora-api-test", 
"message": "CORS allowed_origins: ['http://.*\\\\.corporanet\\\\.local:\\\\d+', '^http://localhost:\\\\d+']", 
"lineno": 74, 
"funcName": "configure_flask_app", 
"pathname": "/single-cell-data-portal/backend/corpora/api_server/app.py"
}
```


backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+
+        :param fields: an ordered list of fields to set `required_fields`
+        :return:
+        """
+        self._required_fields = fields
+        self._skip_fields = dict(zip(self._required_fields, self._required_fields))
+
+    def add_fields(self, log_record: dict, record: LogRecord, message_dict: dict) -> None:
+        """
+        Adds additional log information from `log_record` to `records. If a required field does not exist in the
```
atolopko-czi, line 47:
```suggestion
        Adds additional log information from `log_record` to `record`. If a required field does not exist in the
```

backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+        merge_record_extra(record, log_record, reserved=self._skip_fields)
+
+
+gunicorn_logger = logging.getLogger("gunicorn.error")
+
+
+def configure_logging():
+    dictConfig(
+        {
+            "version": 1,
```
atolopko-czi, line 69: what does version mean here? api version? log message format version?

backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+from logging import LogRecord
+from logging.config import dictConfig
+from typing import List
+
+from pythonjsonlogger import jsonlogger
+from pythonjsonlogger.jsonlogger import merge_record_extra
+
+
+# The fields to log using the json logger.
+LOGGED_FIELDS = ["levelname", "asctime", "name", "message"]
```
atolopko-czi, line 12: * `level` might be sufficient
* is `name` useful? it is always `corpora-api-<ENV>`? That seems verbose for the value it provides.

backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+LOG_FORMAT = " ".join([f"%({field})" for field in LOGGED_FIELDS])
+
+
+class JsonFormatter(jsonlogger.JsonFormatter):
+    default_time_format = "%Y-%m-%dT%H:%M:%S"
+    default_msec_format = "%s.%03dZ"
```
atolopko-czi, line 18: these are read by `jsonlogger.JsonFormatter`, I assume?

backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+from logging.config import dictConfig
+from typing import List
+
+from pythonjsonlogger import jsonlogger
+from pythonjsonlogger.jsonlogger import merge_record_extra
+
+
+# The fields to log using the json logger.
+LOGGED_FIELDS = ["levelname", "asctime", "name", "message"]
+LOG_FORMAT = " ".join([f"%({field})" for field in LOGGED_FIELDS])
```
atolopko-czi, line 13: Is this equivalent to `" ".join(LOGGED_FIELDS)`?

backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class JsonFormatter(jsonlogger.JsonFormatter):
+    default_time_format = "%Y-%m-%dT%H:%M:%S"
+    default_msec_format = "%s.%03dZ"
+
+    converter = time.gmtime
+
+    def add_required_fields(self, fields: List[str]) -> None:
```
atolopko-czi, line 22: It's not immediately clear to the reader how these methods are being called. I assume they are hooks called by `pythonjsonlogger`. Maybe a comment to explain?

backend/corpora/api_server/logger.py
```diff
@@ -0,0 +1,86 @@
...
+from pythonjsonlogger import jsonlogger
+from pythonjsonlogger.jsonlogger import merge_record_extra
+
+
+# The fields to log using the json logger.
+LOGGED_FIELDS = ["levelname", "asctime", "name", "message"]
+LOG_FORMAT = " ".join([f"%({field})" for field in LOGGED_FIELDS])
+
+
+class JsonFormatter(jsonlogger.JsonFormatter):
```
atolopko-czi, line 16: Instead of unioning `LOGGED_FIELDS` and additional fields, it might be cleaner to _nest_ the additional fields as the `message` and add a log msg `type` property. This would allows for different "flavors" of log messages to be recorded, beyond just API logging.

```
{
  "level": "<LEVEL>",
 "asctime": "<TIME>",
 "type": "REQUEST" | "METRIC" | "QUERY" | etc
 "msg": { <JSON> }
}
```
When parsing log programmatically, one could filter on `type`, and then know how to reliably parse the `msg` JSON content.

atolopko-czi (changes requested):
Great improvement! Some thoughts in comments.

backend/corpora/api_server/app.py
```diff
@@ -108,8 +106,21 @@ def apis_landing_page() -> str:
...
 def pre_request_logging():
-    message = json.dumps(dict(url=request.path, method=request.method, schema=request.scheme))
-    app.logger.info(message)
+    app.logger.info(
+        dict(
+            message="REQUEST",
+            url=request.path,
+            method=request.method,
+            schema=request.scheme,
+            content_length=request.content_length,
```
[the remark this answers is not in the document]
atolopko-czi: Yeah, definitely something we need to consider. There may be a select few API calls we might need to make PII-safe, logging-wise.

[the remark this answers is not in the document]
Bento007: > Maybe API calls be configured (via OpenAPI spec) to enable/disable logging per endpoint 

I don't see anything in openapi for this but we can figure something custom out.

backend/corpora/api_server/app.py
```diff
@@ -108,8 +106,21 @@ def apis_landing_page() -> str:
...
 @app.before_request
 def pre_request_logging():
-    message = json.dumps(dict(url=request.path, method=request.method, schema=request.scheme))
-    app.logger.info(message)
+    app.logger.info(
+        dict(
+            message="REQUEST",
+            url=request.path,
+            method=request.method,
+            schema=request.scheme,
```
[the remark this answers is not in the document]
Bento007: we can safely remove

backend/corpora/api_server/logger.py
Bento007: the python logger expects a format of "(field1) (field2)". This is wrapping them in `( )`

backend/corpora/api_server/logger.py
Bento007: It's the version of the python dictConfig. It's so different versions of the config can be supported in the future. Currently, only version 1 exists. I'll add a comment

backend/corpora/api_server/logger.py
Bento007: we import logging in a few other places so the name of logger would be different. For example, https://github.com/chanzuckerberg/single-cell-data-portal/blob/fa9c0d2fd4b263bd76aa833704e0f53af5b09490/backend/corpora/common/entities/dataset.py#L30

Bento007: I changed the above lines to use the root logger so the log name will be root

backend/corpora/api_server/logger.py
Bento007: I moved the log specific details to `details`

backend/corpora/common/entities/dataset_asset.py
```diff
@@ -10,8 +10,6 @@
 from ..corpora_orm import DbDatasetArtifact, DatasetArtifactFileType
 from ..utils.s3_buckets import buckets
 
-logger = logging.getLogger(__name__)
```
Bento007, line 13: Using the root logger so the json logging changes take effect.

backend/corpora/api_server/logger.py
MDunitz: I find this super useful for tracing where in the code the log messages are coming from

backend/corpora/common/entities/dataset_asset.py
MDunitz: Does this mean you'll no longer see what file the logging message is coming from?

MDunitz (approved):
lgtm 
I think this (particularly the request logging) will be extremely useful when debugging issues in the future. Thanks!

backend/corpora/api_server/logger.py
Bento007: we can add [pathname](https://docs.python.org/3/library/logging.html#logrecord-objects) to the log record.

backend/corpora/common/entities/dataset_asset.py
Bento007: It does. Before this would have said in the logs {"name": "dataset_asset"}, now it will be {"name":"root"}. We can add pathname to the log record. That will include the full path to the file that produced the log.

backend/corpora/api_server/logger.py
atolopko-czi: comment pending?

backend/corpora/api_server/logger.py
atolopko-czi: will pathname be added?

atolopko-czi: Mostly, I would like to understand how engineers can configure arbitrary "namespaces" for log messages to aid in filtering. If this PR is just going to focus on request logging (and not deal with logging coming from the underlying logic used to generate the response), that seems fine for now, but we should capture ideas for future improvements in a story or the Tech Works Ideas doc.

atolopko-czi (changes requested):
Just a couple pending comments/discussions.

I also noticed double-logging is occurring in `jsonlogs` stack rdev logs (is this the latest logging output?):
```
2022-06-23T19:54:36.035-04:00 | [2022-06-23 23:54:36,035] INFO in app: {'type': 'REQUEST', 'details': {'url': '/', 'method': 'GET', 'content_length': None}}
2022-06-23T19:54:36.035-04:00 | {"levelname": "INFO", "asctime": "2022-06-23T23:54:36.023Z", "name": "corpora-api-rdev", "message": null, "type": "REQUEST", "details": {"url": "/", "method": "GET", "content_length": null}}
```
[[cloudwatch](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Fdp$252Frdev$252Fjsonlogs$252Fbackend/log-events/67d2663e0e52bc45b0ba94c8bb41fd0c5d3fca32c5e63694ab0b68bb76af7277)]

backend/corpora/api_server/logger.py
Bento007: any other loggers that are created using `logging.get()` will use the root logger unless it get configure somewhere else. This means they will all be in JSON format.

atolopko-czi (approved):

PR #2616, opened 2022-05-27
fix: keep links on final update call during revision

danieljhegeman:
This commit fixes a bug present that deletes links for revisions of
collections in most scenarios when publishing

- #TICKET_NUMBER

### Reviewers
**Functional:** 
@atolopko-czi @Bento007 @brianraymor 
**Readability:** 

---

## Changes
- add
- remove
- modify


atolopko-czi (commented):
[not a code review]
Maybe we should remove the default value for `keep_links` to force the caller (i.e. engineer) to decide.

PR #2617, opened 2022-05-28
fix: streamline publish function to prevent loss of links

Bento007:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- Reorganize publish logic.
- Add __repr__ to make debugging easier
- Add additional publish revision test with refreshed dataset and new links.
- don't change published_at if tombstoned

### DIFF
--- a/backend/corpora/common/entities/collection.py
+++ b/backend/corpora/common/entities/collection.py
@@ -262,6 +262,8 @@ class Collection(Entity):
                     revised_dataset.publish_new(now)
                     has_dataset_changes = True
             self.session.flush()
+            # calling expire because the above code updates and deletes rows related to this object. If we don't expire
+            # the object, SqlAlchmey will try to modify rows that don't exists and thrown an error.
             self.session.expire(self.db_object)
             revision = self.to_dict(
                 remove_attr=(
--- a/backend/corpora/common/entities/collection.py
+++ b/backend/corpora/common/entities/collection.py
@@ -279,7 +279,9 @@ class Collection(Entity):
             )
             revision["data_submission_policy_version"] = data_submission_policy_version
             revision["revised_at"] = now if has_dataset_changes else self.revised_at
-            public_collection.update(**revision, commit=False)  # This function call deletes old links (keep_links param defaults to False)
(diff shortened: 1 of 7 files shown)


backend/corpora/common/entities/collection.py
```diff
@@ -245,23 +244,49 @@ def publish(self, data_submission_policy_version):
...
-                **revision,
-            )
+                **revision, commit=False
+            )  # This function call deletes old links (keep_links param defaults to False)
             for link in self.links:
-                CollectionLink(link).update(collection_id=self.revision_of, commit=False)
-            is_existing_collection = True
-
+                link.collection_id = self.revision_of
+            self.session.flush()
```
danieljhegeman, line 287: @Bento007 I just added this line. I am pretty sure this is necessary to make 100% sure that the links are moved to the published object rather than themselves deleted. I think we had an inverse of this situation with the original bug which meant that the code usually produced the bug but occasionally worked (as per Jason's experience); without this line, I think the code would usually work but could possibly produce the bug.

I think the combination of reassigning the links and then the subsequent `delete` call does not have deterministic behavior with SQLAlchemy.

Basically, the flush is required to make 100% sure that the db transaction first sees that the revision links now belong to the published collection _before_ the revision object is deleted (in which case the links would themselves be deleted).

danieljhegeman (approved):

backend/corpora/common/entities/collection.py
```diff
@@ -245,23 +244,49 @@ def publish(self, data_submission_policy_version):
...
+                if original and public_collection.check_has_dataset(original):
+                    dataset_is_changed = original.publish_revision(revised_dataset, now)
+                    if dataset_is_changed:
+                        has_dataset_changes = True
+                else:
+                    # The dataset is new
+                    revised_dataset.publish_new(now)
+                    has_dataset_changes = True
+            self.session.flush()
+            # calling expire because the above code updates and deletes rows related to this object. If we don't expire
```
Bento007, line 265: @danieljhegeman

PR #2634, opened 2022-06-02
feat(curator api): Delete a private collection

Bento007:
### Reviewers

**Readability:** 

---

## Changes
- Add openapi API definition
- Add collection delete handler
- Add test file stub
- Can delete a private collection or revision using the curation API. Not a public collection.
- 403 is returned when trying to delete a deleted or tombstones resource
- Reorganized curation API handlers.


backend/corpora/lambdas/api/v1/curation/collection_uuid/__init__.py
```diff
@@ -0,0 +1,26 @@
+from flask import g
```
danieljhegeman, line 1: Stepping back a bit, I'm not sure I understand why the file tree is organized as it is. For starters, `../collection/` and `../collection_uuid/`...? What's the difference, exactly?

And why are we putting this logic in this `__init__.py` file?

danieljhegeman: We don't seem to be using this as common code within the other files in this module

danieljhegeman: I'm not seeing the lines within which anything is being organized (not in this PR specifically, but generally)

`collection_uuid/` directory holds `revision.py` and `dataset.py` but the collection deletion code is going in the init file for that dir, and then there is a whole separate `collection/` dir?

backend/corpora/lambdas/api/v1/curation/collection_uuid/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    db_session = g.db_session
+    collection = get_collection(
+        db_session,
+        collection_uuid,
+        visibility=CollectionVisibility.PRIVATE,
+        owner=owner_or_allowed(token_info),
+        include_tombstones=True,
+    )
+    if collection.revision_of:
+        raise MethodNotAllowedException("Can only delete a private collection through curation API.")
```
danieljhegeman, line 21: A collection revision *is* a private collection. This needs to be allowed.

backend/corpora/lambdas/api/v1/curation/collection_uuid/__init__.py
```diff
@@ -13,14 +13,13 @@ def delete_collection(collection_uuid: str, token_info: dict):
...
         db_session,
         collection_uuid,
-        visibility=CollectionVisibility.PRIVATE,
         owner=owner_or_allowed(token_info),
         include_tombstones=True,
     )
-    if collection.revision_of:
-        raise MethodNotAllowedException("Can only delete a private collection through curation API.")
-    elif collection.tombstone:
+    if collection.tombstone:
```
danieljhegeman, line 19: If the collection is tombstoned, that means it _was_ a public Collection, in which case there should still be an error response, not 204.

As the already-deleted private collection returns 204, the already-deleted public collection should return an error.

danieljhegeman: So I think you should just remove this line, and change the `elif` on ln 21 to `if`

danieljhegeman: And obviously update the test below accordingly

backend/corpora/lambdas/api/v1/curation/collection_uuid/__init__.py
Bento007: your right I put some of the files in the wrong place. In general, the goal is the have the API handlers be in roughly the same location as the API path. For example: `Delete /curation/v1/collection/{collection_uuid}` handler should live in `/curation/v1/collection/collection_uuid.py`, `/curation/v1/collection/collection_uuid/__init__.py`, or `/curation/v1/collection/collection_uuid/delete.py`. This whole API directory needs reorganizing, but I don't plan on doing it in this PR.

danieljhegeman: > This whole API directory needs reorganizing, but I don't plan on doing it in this PR.

Agreed, we can leave to a later date.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/dataset.py
```diff
@@ -1,17 +1,34 @@
...
-from backend.corpora.lambdas.api.v1.common import get_collection
+from backend.corpora.common.entities import Dataset
+from backend.corpora.common.utils.http_exceptions import InvalidParametersHTTPException
 from backend.corpora.lambdas.api.v1.authorization import owner_or_allowed
+from backend.corpora.lambdas.api.v1.common import get_collection
+from backend.corpora.lambdas.api.v1.dataset import delete_dataset_common
+
+
+@dbconnect
+def delete_dataset(token_info: dict, collection_uuid: str, curator_tag: str = None, dataset_uuid=None):
```
danieljhegeman, line 18: nit format
```suggestion
def delete_dataset(token_info: dict, collection_uuid: str, curator_tag: str = None, dataset_uuid = None):
```

danieljhegeman:
@Bento007 ok now I'm understanding why you put the code in `__init__.py` initially for collection deletion. I def still think we need a reorg but that makes a bit more sense now. We can leave it in `actions.py` for the time being and move elsewhere later, perhaps F&F or as part of a larger tech debt reorg for dp in the coming months.

danieljhegeman (approved):

PR #2642, opened 2022-06-06
feat(curation api): add GET collections/collection_uuid/assets endpoint

Bento007:
- #TICKET_NUMBER

### Reviewers

**Readability:** 

---

## Changes
- Add Post collections/collection_uuid/assets endpoint 
- Return some asset links if some assets cannot be found.

## Notes for Reviewer
- Added a 202 response when only some assets are returned.
- Changed from a GET to a POST since the response changes due to generated s3 links.


backend/config/curation-api.yml
```diff
@@ -501,14 +501,15 @@ paths:
         "401":
           $ref: "#/components/responses/401"
 
-  /v1/collections/{collection_uuid}/asset:
-    get:
+  /v1/collections/{collection_uuid}/assets:
+    post:
```
danieljhegeman, line 505: I think this should remain a `GET` since the underlying resource is not being changed in any way. Yes, we're creating additional download links each time, but those are ancillary to the actual entity (asset) in question.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
```diff
@@ -0,0 +1,56 @@
...
+from backend.corpora.api_server.db import dbconnect
+from backend.corpora.common.entities import Dataset, DatasetAsset
+from backend.corpora.common.utils.http_exceptions import InvalidParametersHTTPException, NotFoundHTTPException
+
+
+@dbconnect
+def get(collection_uuid: str, curator_tag: str = None, dataset_uuid=None):
+    db_session = g.db_session
+    if dataset_uuid:
+        dataset = Dataset.get(db_session, dataset_uuid)
```
danieljhegeman, line 12: `collection_uuid` is not used here to fetch the Dataset; as a consequence, if `dataset_uuid` is used, then `collection_uuid` is not required to be correct in order to fetch the Dataset in question.

I think this operation should always _fail_ if the id for the Collection is wrong.

danieljhegeman: I don't think this represents a security issue as `dataset_uuid` can be considered a 'capability identifier', i.e. if you have it, then you should have the right to view this.

But we may wish to express that to curators as well.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
```diff
@@ -0,0 +1,56 @@
...
+        dataset = Dataset.get_dataset_from_curator_tag(db_session, collection_uuid, curator_tag)
+    else:
+        raise InvalidParametersHTTPException()
+    if not dataset:
+        raise NotFoundHTTPException("Dataset not found.")
+
+    # retrieve the artifact
+    assets = dataset.get_assets()
+    if not assets:
+        raise NotFoundHTTPException("No assets found not found. The dataset may still be processing.")
```
danieljhegeman, line 23:
```suggestion
        raise NotFoundHTTPException("No assets found. The dataset may still be processing.")
```

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
```diff
@@ -0,0 +1,56 @@
...
+            asset_list.append(result)
+            error_flag = True
+            continue
+        else:
+            result["file_size"] = file_size
+
+        # Generate pre-signed URL
+        presigned_url = asset.generate_file_url()
+        if not presigned_url:
+            result["presigned_url"] = "Not Found."
```
danieljhegeman, line 44: when does this happen, if `file_size` is nonzero?

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
```diff
@@ -0,0 +1,56 @@
...
+            asset_list.append(result)
+            error_flag = True
+            continue
+        else:
+            result["presigned_url"] = presigned_url
+        asset_list.append(result)
```
danieljhegeman, line 50: It seems a bit strange to use a `continue` directly above in concert with an `else` but then to also have this last line on the parent logic flow. 

Prefer removing `continue` and indenting last line to be under the `else` flow OR remove `else`.

danieljhegeman: ^ nit-adjacent 😄

tests/unit/backend/corpora/api_server/curator/collection/test_assets.py
```diff
@@ -0,0 +1,139 @@
...
+                    dataset_uuid=dataset_uuid,
+                    assets=[dict(file_name="test_filename", file_size=len(content), file_type="H5AD")],
+                )
+                if dataset_uuid == "test_curator_tag":
+                    expected_body["curator_tag"] = self.curator_tag
+                    response = self.app.get(
+                        "/curation/v1/collections/test_curator_tag_collection_id/assets",
+                        query_string=dict(curator_tag=self.curator_tag),
+                    )
+                else:
```
danieljhegeman, line 65: should add `expected_body["curator_tag"] = None`?

danieljhegeman (changes requested):
looks good, a few q's, most notably I think we should address the issue of what happens when the wrong collection id is passed but the dataset id still returns a dataset

backend/config/curation-api.yml
```diff
@@ -519,34 +520,40 @@ paths:
...
-                        presigned_url:
-                          type: string
-                        file_name:
-                          type: string
-                        file_size:
-                          type: number
-                        filetype:
-                          type: string
+                $ref: "#/components/schemas/asset_response"
+        "202":
```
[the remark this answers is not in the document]
Bento007: 206 is partial content returned. We returned all of the content we have so there is no more content to get from that endpoint. So no. I don't see a response for partial success.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
Bento007: if AWS fails to generate a pre-signed URL for some reason. It's not likely to happen.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
Bento007: Added check for collection

tests/unit/backend/corpora/api_server/curator/collection/test_assets.py
Bento007: curator_tag is not a required field in the response, so it's not returned if it doesn't exist.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/assets.py
danieljhegeman: Shouldn't the error message indicate that they should _try again_ in that scenario?

tests/unit/backend/corpora/api_server/curator/collection/test_assets.py
danieljhegeman: Got it - I think this gets back to the issue Andrew and Brian were discussing recently about the shapes of our responses (including null values or no). Speaking of which, I'm not sure we ever came to a clear resolution...

danieljhegeman (approved):

Issue #2414, opened 2022-04-21
Data Validation Tests


MDunitz:
## Need
- As a user I want to be confident the data I'm working with is accurate

## Approach
- [x] Walk through manual verification with @pablo-gar 
- [ ] Check that housekeeping genes are present at expected values (MALAT1 - high)
- [ ] cells with high expression values have high rankit values
- [ ] slice out subsection of data and compare to expected results

## DOD
- [ ] Validation tests run automatically after a cube is generated (before the snapshot id is updated in s3)

pablo-gar:
I don't know if you want to add the following qualitative checks along with MALAT1

In the human lung, the following genes should be observed as markers of the indicated cell types:

- FCN1: monocytes
- TUBB4B: ciliated cells, and this gene is expressed to lower values in most other cell types 
- CD68: macrophages
- AQP5: goblet cells and secreting cells

MDunitz:
@pablo-gar so if we slice on cell type we should see high expression levels for those genes (in the specified cell type)? 

Is it worth comparing to the expression value in a different type of cell?

MDunitz:
XIST is a gene that’s primarily expressed in women, as it is the gene responsible for silencing one of the two X chromosomes in women.
So expression of this gene disappears when selecting only “male”

MDunitz:
ACTB is should also be expressed in most cells (at a lower level than MALAT1)

MDunitz:
lung map dataset uuid 3de0ad6d-4378-4f62-b37b-ec0b75a50d94
genes ["MALAT1", "CCL5"]

PR #2643, opened 2022-06-06
feat: cube validation

MDunitz:
- #2414

### Reviewers
**Functional:** 
- @atolopko-czi 
- @ebezzi 
**Readability:** 
- @pablo-gar 
---

## Changes
- add validation of the data stored in the cube based on known biological patterns 
- validate a particular dataset/gene combo in the cube by comparing to a manually curated version (from @pablo-gar)

## QA steps (optional)
- ran against full cube in testing instance without env var set
- ran against full cube in testing instance with env var set to PROD

## Notes for Reviewer
@pablo-gar do these values look correct?
female avg xist expression 3.4171425456084803
male avg xist expression 2.914560236618371


backend/wmg/data/validation/validation.py
```diff
@@ -0,0 +1,384 @@
...
+        )
+        malat_expected = expected_values.obs.assign(MALAT1=expected_values.layers["rankit"].toarray()[:, 0])
+        ccl5_expected = expected_values.obs.assign(CCL5=expected_values.layers["rankit"].toarray()[:, 1])
+
+        expected_malat1_by_cell_type = malat_expected.groupby("cell_type_ontology_term_id").sum().MALAT1
+        expected_ccl5_by_cell_type = ccl5_expected.groupby("cell_type_ontology_term_id").sum().CCL5
+        # drop ccl5 cell types with expression value of zero (to match pipeline processing)
+        expected_ccl5_by_cell_type = expected_ccl5_by_cell_type[expected_ccl5_by_cell_type != 0]
+
+        # Todo actually compare once the rankit bug is fixed
```
atolopko-czi, line 370: Already noted this dependency in that issue?

atolopko-czi (commented):
Some high-level considerations, in addition to a few scattered comments:
* If this does fail, how are on-call engineers be notified?  Presumably, the batch job fails, but not if any alerts or notifications are generated. This is arguable out-of-scope for this PR, but the question comes to mind now.
* On failure, we are left with a partially-populated snapshot directory. Should it be marked as failed, maybe via a special file in the snapshot directory? (Could also delete, but that won't help with troubleshooting the failure.).  Backend API could check for validated status before trying to load, merely as a defensive measure (the latest_snapshot_id will keep things safe).
* Instead of failing fast, could ensure all checks are run and report all failures at the end. Maybe that would provide some insights into how "bad" the cube is when troubleshooting. And if there are multiple errors that need to be investigated, it might save a few re-runs.
* Is it worth mentioning this validation step exists somewhere in the Tech Overview WMG section?

MDunitz:
@atolopko-czi re the high level concerns you mentioned above

`If this does fail, how are on-call engineers notified? Presumably, the batch job fails, but not sure if any alerts or notifications are generated. This is arguably out-of-scope for this PR, but the question comes to mind now.`
- currently the engineers are not notified. I have a ticket to add notifications (via slack alerts) for any pipeline failure 

`On failure, we are left with a partially-populated snapshot directory. Should it be marked as failed, maybe via a special file in the snapshot directory? (Could also delete, but that won't help with troubleshooting the failure.). Backend API could check for validated status before trying to load, merely as a defensive measure (the latest_snapshot_id will keep things safe).`
- since validation is set to run before updating s3 if it fails there will be no orphaned assets.
- this does raise a good point re debugging issues. On failure it might be worth copying the cube to a different bucket to help with debugging but I think that's probably out of scope for this ticket

`Instead of failing fast, could ensure all checks are run and report all failures at the end. Maybe that would provide some insights into how "bad" the cube is when troubleshooting. And if there are multiple errors that need to be investigated, it might save a few re-runs.`
Thats a great idea. Do you have any suggestions for an error log? My first thought is to just do an array of strings and if the array is empty its valid and if not we print all the errors and then fail the cube.

Is it worth mentioning this validation step exists somewhere in the Tech Overview WMG section?
- yes!

backend/wmg/data/validation/validation.py
MDunitz: Just merged that fix. I'm rerunning the pipeline now and I'll test that the arrays match/update the validation code to do an explicit check when its done

atolopko-czi:
> Thats a great idea. Do you have any suggestions for an error log? My first thought is to just do an array of strings and if the array is empty its valid and if not we print all the errors and then fail the cube.

Yup, that's pretty much all I've ever done or seen done for handling multiple errors!

atolopko-czi:
> since validation is set to run before updating s3 if it fails there will be no orphaned assets

Ah, great! I missed that. 

> On failure it might be worth copying the cube to a different bucket to help with debugging but I think that's probably out of scope for this ticket

Yeah, unless we start failing a lot, not much need to help facilitate troubleshooting the failure case.  Maybe the simplest thing we could do is provide a flag that says "don't validate" which would allow manual re-run w/o validation and that _does_ copy to S3, and does _not_ update the latest snapshot.

backend/wmg/data/validation/validation.py
```diff
@@ -0,0 +1,384 @@
...
+            expected_ccl5_by_cell_type = ccl5_expected.groupby("cell_type_ontology_term_id").sum().CCL5
+            # drop ccl5 cell types with expression value of zero (to match pipeline processing)
+            expected_ccl5_by_cell_type = expected_ccl5_by_cell_type[expected_ccl5_by_cell_type != 0]
+
+            malat1_comparison = expected_malat1_by_cell_type.compare(malat1_expression_sum_by_cell_type)
+            ccl5_comparison = expected_ccl5_by_cell_type.compare(ccl5_expression_sum_by_cell_type)
+            logger.info(malat1_comparison)
+            logger.info(ccl5_comparison)
+
+            """
```
MDunitz, line 356: @pablo-gar can you take a look at 357 - 374 and just confirm it seems reasonable?

pablo-gar: looks good to me! less than 1 difference is very strict and I'm glad the method is working to this extent

backend/wmg/data/cube_pipeline.py
```diff
@@ -75,13 +81,19 @@ def load_data_and_create_cube(
...
     cell_type_by_tissue = get_cell_types_by_tissue(corpus_path)
     generate_cell_ordering(snapshot_path, cell_type_by_tissue)
     generate_primary_filter_dimensions(snapshot_path, corpus_name, timestamp)
     logger.info("Generated cell ordering json file")
-    update_s3_resources(snapshot_path, timestamp)
+    upload_artifacts_to_s3(snapshot_path, timestamp)
     logger.info("Copied snapshot to s3")
+    if validate_cubes:
+        update_s3_resources(timestamp)
+    logger.info(f"Updated latest_snapshot_identifier in s3. Current snapshot id is {timestamp}")
```
atolopko-czi, line 96: ⚠️ Indent this line or move it to `update_s3_resources`, if keeping `if` stmt.

backend/wmg/data/cube_pipeline.py
```diff
@@ -41,13 +42,18 @@ def load(dataset_directory: List, corpus_path: str, validate: bool = False):
...
             tiledb.vacuum(arr_name)
 
 
 def load_data_and_create_cube(
-    path_to_h5ad_datasets: str, corpus_name: str = "corpus_group", snapshot_path=None, extract_data=True
+    path_to_h5ad_datasets: str,
+    corpus_name: str = "corpus_group",
+    snapshot_path=None,
+    extract_data=True,
+    validate_cubes=True,
```
atolopko-czi, line 56: nit: If this method is never being called with `validate_cubes=False`, do we need the new param?

backend/wmg/data/cube_pipeline.py
```diff
@@ -75,13 +81,19 @@ def load_data_and_create_cube(
     logger.info("Loaded datasets into corpus")
     create_cubes(corpus_path)
     logger.info("Built expression summary cube")
-
+    if validate_cubes:
+        is_valid = Validation(corpus_path).validate_cube()
+        if is_valid is False:
+            return
```
atolopko-czi, line 87: should it return boolean? The pipeline `main` block could return as `sys.exit(1)` if validation fails. Not sure if that's helpful in the grand scheme of things, but maybe it marks the batch job as failed? _Is_ failed validation to be considered a job failure?

atolopko-czi (approved):
One log stmt out of place, and just nits other than that! Preemptively approving, since none of my comments are preventing world-ending events.

backend/wmg/data/cube_pipeline.py
MDunitz: This is based on your suggestion for making debugging validation failures easier to debug. This will skip the validation checks and copy to s3 without updating the snapshot id

backend/wmg/data/cube_pipeline.py
MDunitz: This will copy the resources to s3 to make them available for debugging but it will skip updating the snapshot if the cube wasnt validated

backend/wmg/data/cube_pipeline.py
atolopko-czi: Retracting this comment, since validation is only to be used for ad hoc runs

backend/wmg/data/cube_pipeline.py
MDunitz: ahh good catch. Thanks!

backend/wmg/data/load_cube.py
```diff
@@ -15,11 +15,10 @@ def _get_wmg_bucket_path():
         return f"s3://{wmg_bucket_name}"
 
 
-def update_s3_resources(snapshot_path, timestamp):
+def update_s3_resources(timestamp):
     """
-    Copy cube and cube related files to s3 under the current timestamp
+    Update snapshot pointer and remove older datasets
```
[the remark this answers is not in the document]
MDunitz: yes! I really like that idea. active snapshot is clearer

backend/wmg/data/cube_pipeline.py
MDunitz: Because that isnt super straightforward, I'm going to create a ticket to do that in a separate pr (in the interest of getting this merged in asap).

MDunitz: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2704

PR #2673, opened 2022-06-09
fix(curator-api): support adding new datasets using s3

Bento007:
### Reviewers

---

## Changes
- add additional testing that includes the get_dataset_info function.
- modify get_dataset_info to support new datasets uploaded directly to S3


danieljhegeman (approved):
Awesome, thanks Trent

nayib-jose-gloria (approved):

PR #2681, opened 2022-06-10
feat: GET /collections

danieljhegeman:
- #TICKET_NUMBER

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add `GET /v1/collections` endpoint
- include Collection-level `processing_status`
- Fix the function type for the leniant curator access security scheme: change from `x-apikeyInfoFunc` to `x-bearerInfoFunc`. This allows usage of access token.
- ~Add new schema `query_visibility_default_null` to Curation API which allows a null value to be passed rather than default value of `"PUBLIC"`.~
- Improve description for endpoint.


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -61,6 +61,32 @@ def get_collections_list(from_date: int = None, to_date: int = None, token_info:
...
+    for collection in collections:
+        owner = collection["owner"]
+        if is_user_owner_or_allowed(token_info, owner):
+            collection["access_type"] = "WRITE"
+        elif collection["visibility"] == "PRIVATE":
+            continue
+        else:
+            collection["access_type"] = "READ"
+        del collection["owner"]  # Don't actually want to return 'owner' in response
+        collection["collection_url"] = f"https://www.cellxgene.cziscience.com/collections/{collection['id']}"
```
Bento007, line 84: This domain-only works for prod.

backend/corpora/common/entities/collection.py
```diff
@@ -150,6 +152,108 @@ def to_dict(db_object):
...
+                if relation_name in collection:
+                    updated_relations = []
+                    for rel in collection[relation_name]:
+                        updated_relations.append(cls._copy_and_transform_columns(rel_cols, rel))
+                    collection[relation_name] = updated_relations
+
+        return collections
+
+    @staticmethod
+    def _copy_and_transform_columns(cols: typing.Iterable, from_obj: dict) -> dict:
```
Bento007, line 233: [to_dict](https://github.com/chanzuckerberg/single-cell-data-portal/blob/69505d5406796441556320c6a1fb9965d0735285/backend/corpora/common/corpora_orm.py#L61) does this not accomplish the same thing?

Bento007 (changes requested):
Here is my partial review. I look at it closer on monday.

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -61,6 +61,32 @@ def get_collections_list(from_date: int = None, to_date: int = None, token_info:
...
+    """
+    Collections index endpoint for Curation API. Only return Collection data for which the curator is authorized.
+    @param visibility: the CollectionVisibility in string form
+    @param token_info: access token info
+    @return: Response
+    """
+
+    collections = Collection.list_collections_for_curator_api(g.db_session, visibility)
+    allowed_collections = []
+    for collection in collections:
```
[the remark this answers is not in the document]
danieljhegeman: The actual compute time doesn't change. I wanted to keep the data retrieval logic separate from the authorization logic.

What is the primary concern?

I should rename `Collection.list_collections_for_curator_api` to `Collection.get_all_collections`

danieljhegeman:
@Bento007 I think I've addressed the concerns you laid out initially regarding the fact that I wasn't relying on the custom JSON encoder or the (also-custom) `to_dict` method on the base class for the db objects.

backend/corpora/lambdas/api/v1/collection.py
danieljhegeman: Fixed... except for the case of `test`... will address that

danieljhegeman: Addressed `test` as well

backend/config/curation-api.yml
```diff
@@ -44,12 +44,14 @@ paths:
...
+      operationId: backend.corpora.lambdas.api.v1.collection.get_collections_curation
       security:
         - curatorAccessLenient: []
-      parameters:
-        - $ref: "#/components/parameters/query_visibility"
+        - {}
       tags:
         - collection
+      parameters:
+        - $ref: "#/components/parameters/query_visibility_default_null"
```
danieljhegeman, line 54: I couldn't figure out how to override the default value of `"PUBLIC"` to make it null for `#/components/parameters/query_visibility`, so this is my solution

backend/corpora/common/entities/collection.py
```diff
@@ -150,6 +150,81 @@ def to_dict(db_object):
...
+        Return a subset copy of 'from_obj' which contains only the attributes found as strings in iterable 'attrs'
+        @param attrs: the keys to copy
+        @param from_obj: dict from which to copy key-value pairs
+        @return: copied subset of 'from_obj'
+        """
+        to_obj = {}
+        for a in attrs:
+            to_obj[a] = from_obj[a]
+        return to_obj
+
```
danieljhegeman, line 227: this helper method is still needed to do column selection. Prefer affirmative selection to specifying the columns that you want removed (i.e., via `remove_attrs`)

backend/corpora/common/entities/collection.py
```diff
@@ -150,6 +151,96 @@ def to_dict(db_object):
 
         return results
 
+    @classmethod
+    def list_collections(cls, session: Session, visibility: str = None) -> typing.List[dict]:
+        """
+        Get a subset of columns, in dict form, for all Collections with the specified visibility. If visibility is None,
+        return *all* Collections.
+        @param session: the SQLAlchemy session
```
Bento007, line 159: change @ to : for all paramters and return

backend/corpora/common/entities/collection.py
```diff
@@ -150,6 +151,96 @@ def to_dict(db_object):
...
+            "id",
+            "curator_tag",
+            "tissue",
+            "assay",
+            "disease",
+            "organism",
+            "tombstone",
+        ]
+
+        filters = []
```
Bento007, line 198: move this to the else clause of the bellow if statements.

Bento007 (changes requested):
Some things to fix

backend/corpora/common/entities/collection.py
```diff
@@ -150,6 +151,96 @@ def to_dict(db_object):
...
+
+            # Add a Collection-level processing status
+            status = ProcessingStatus.SUCCESS
+            for dataset in collection["datasets"]:
+                processing_status = dataset["processing_status"]
+                if processing_status["processing_status"] == ProcessingStatus.PENDING:
+                    status = ProcessingStatus.PENDING
+                elif processing_status["processing_status"] == ProcessingStatus.FAILURE:
+                    status = ProcessingStatus.FAILURE
+                    break
```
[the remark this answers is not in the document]
Bento007: Ah, it's because if there is a failure we don't care about the status of the remaining datasets. Maybe add a comment for clarity.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -128,6 +128,61 @@ def test__create_collection__InvalidParameters(self):
...
+        self.assertEqual(6, len(res_no_auth.json["collections"]))
+        [self.assertEqual("PUBLIC", c["visibility"]) for c in res_no_auth.json["collections"]]
+
+    def test__get_collections_no_auth_processing_status_PENDING__OK(self):
+        res_no_auth = self.app.get("/curation/v1/collections")
+        self.assertEqual(200, res_no_auth.status_code)
+        self.assertEqual(6, len(res_no_auth.json["collections"]))
+        for collection in res_no_auth.json["collections"]:
+            if collection["id"] == "test_collection_id":
+                self.assertEqual(collection["processing_status"], "PENDING")
```
Bento007, line 150: I assume this can only happen in testing because we should not be able to publish a collection with pending datasets.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -128,6 +128,61 @@ def test__create_collection__InvalidParameters(self):
...
 
 
+class TestGetCollections(BaseAuthAPITest):
+    def setUp(self):
+        super().setUp()
+        self.test_collection = dict(
+            name="collection", description="description", contact_name="john doe", contact_email="johndoe@email.com"
+        )
+
+    def test__get_collections_no_auth__OK(self):
```
Bento007, line 138: get publish collections

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -128,6 +128,61 @@ def test__create_collection__InvalidParameters(self):
...
+        res_no_auth = self.app.get("/curation/v1/collections")
+        self.assertEqual(200, res_no_auth.status_code)
+        self.assertEqual(6, len(res_no_auth.json["collections"]))
+        [self.assertEqual("PUBLIC", c["visibility"]) for c in res_no_auth.json["collections"]]
+
+    def test__get_collections_no_auth_processing_status_PENDING__OK(self):
```
Bento007, line 144: you could make this a subtest of the above test.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -128,6 +128,61 @@ def test__create_collection__InvalidParameters(self):
...
+        res_no_auth = self.app.get("/curation/v1/collections")
+        self.assertEqual(200, res_no_auth.status_code)
+        self.assertEqual(6, len(res_no_auth.json["collections"]))
+        for collection in res_no_auth.json["collections"]:
+            if collection["id"] == "test_collection_id":
+                self.assertEqual(collection["processing_status"], "PENDING")
+            else:
+                self.assertEqual(collection["processing_status"], "SUCCESS")
+        [self.assertEqual("PUBLIC", c["visibility"]) for c in res_no_auth.json["collections"]]
```
Bento007, line 153: We test this same condition on the same data in `test__get_collections_no_auth__OK`. Is there enough difference between the two tests to require two separate tests?

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -128,6 +128,61 @@ def test__create_collection__InvalidParameters(self):
...
+
+    def test__get_collections_with_auth__OK(self):
+        res_auth = self.app.get("/curation/v1/collections", headers=self.get_auth_headers())
+        self.assertEqual(200, res_auth.status_code)
+        self.assertEqual(6, len(res_auth.json["collections"]))
```
Bento007, line 158: shouldn't this be 7, because we get both the public and private collections?

tests/unit/backend/fixtures/test_db.py
```diff
@@ -482,3 +482,29 @@ def _create_test_dataset_processing_status(self):
         )
         self.session.add(dataset_processing_status)
         self.session.commit()
+        dataset_processing_status = DbDatasetProcessingStatus(
```
Bento007, line 485: we should probably stop adding test data here and instead add it to the test file where it's used. This will speed up testing overall, and make it clear what test data is needed for a test. Can you move these to `test_curation_collections.py`. The `generate_dataset` function should help with this.

Bento007 (changes requested):
Add a test that check for Write or Read fields are set correctly.

tests/unit/backend/fixtures/test_db.py
Bento007: If it's easy do it, otherwise we can do this in a larger refactor

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: No; originally I had it that way, but then I noticed I noticed [this](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2083#:~:text=When%20the%20visibility%20parameter%20is%20unspecified%20or%20set%20to%20%22PUBLIC%22%2C%20a%20list%20of%20all%20public%20collections%20is%20returned.) from Brian

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: great point. I should rework this to be private only.

backend/corpora/common/entities/collection.py
```diff
@@ -150,6 +153,75 @@ def to_dict(db_object):
...
+            "disease",
+            "organism",
+            "tombstone",
+            "processing_status",
+        ],
+        DbDatasetProcessingStatus: ["processing_status"],
+    }
+
+    @classmethod
+    def list_collections(cls, session: Session, visibility: str = None) -> typing.List[dict]:
```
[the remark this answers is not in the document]
danieljhegeman: I'll change it to `list_collections_curation_filter` or something like that for all of the above reasons

backend/corpora/common/entities/collection.py
danieljhegeman: is this a new standard for docstrings for python?

Bento007: It's a standard, not really new. It's sphinx docstring format.

Bento007: It's what we've been using, so let's keep things consistent for now.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: waiting for confirmation from Brian on this

backend/corpora/lambdas/api/v1/collection.py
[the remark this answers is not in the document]
danieljhegeman: Changed the function name to `list_collections_for_curation`

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: implicitly it gets public collections but it is named without "public" because the `visibility` is not specified

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: I like this idea. I've seen you use this pattern and I should be using it more.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: Addressed by combining with subTest

tests/unit/backend/fixtures/test_db.py
danieljhegeman: I'm kind of in favor of leaving in these additions because they complete/fill out other entities that are already in `test_db.py`.

But generally I would agree that we could benefit from a more modular way to initialize the db for each test suite. Probably need to review and decide on an ideal testing db "base" state.

backend/corpora/common/corpora_config.py
```diff
@@ -11,10 +11,24 @@ def __init__(self, *args, **kwargs):
...
+            collections_base_url = f"https://{RDEV_PREFIX.strip('/')}-frontend.rdev.single-cell.czi.technology"
+        else:
+            DEPLOYMENT_STAGE = os.environ.get("DEPLOYMENT_STAGE")
+            if DEPLOYMENT_STAGE == "test":
+                collections_base_url = "http://frontend.corporanet.local:3000"
+            elif DEPLOYMENT_STAGE == "prod":
+                collections_base_url = "https://cellxgene.cziscience.com"
+            else:
+                collections_base_url = f"https://cellxgene.{DEPLOYMENT_STAGE}.single-cell.czi.technology"
+
```
danieljhegeman, line 26: cc @Bento007

backend/corpora/common/corpora_orm.py
```diff
@@ -115,6 +115,48 @@ def to_dict(
...
+                    result[relation.key] = None
+                elif isinstance(value.__class__, DeclarativeMeta):
+                    result[relation.key] = value.to_dict_keep(keep=keep, backref=backref)
+                elif isinstance(value, list):
+                    result[relation.key] = [i.to_dict_keep(keep=keep, backref=backref) for i in value]
+                else:
+                    raise CorporaException(f"Unable to convert to dictionary. Unexpected type: {type(value)}.")
+        backref.pop()
+        return result
+
```
danieljhegeman, line 159: 👏

backend/corpora/common/corpora_config.py
```diff
@@ -11,10 +11,24 @@ def __init__(self, *args, **kwargs):
 
     def get_defaults_template(self):
 
+        # Set collections_base_url
+        RDEV_PREFIX = os.environ.get("REMOTE_DEV_PREFIX")
```
Bento007, line 15: these values should be lower case since they are not CONSTANTS

danieljhegeman: good point

PR #2707, opened 2022-06-15
feat: GET /v1/collections/{collection_uuid}

danieljhegeman:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


danieljhegeman:
Need to add a couple more tests still...

backend/corpora/lambdas/api/v1/common.py
```diff
@@ -35,3 +37,32 @@ def delete_dataset_common(db_session: Session, dataset: Dataset, token_info: dic
                 original.create_revision(dataset.collection.id)  # Restore the original dataset and S3 assets
             dataset.asset_deletion()  # Delete the S3 assets and database rows.
             dataset.delete()  # Delete the dataset row.
+
+
+def reshape_for_curation_api_and_is_allowed(collection, token_info, allow_access=False):
```
nayib-jose-gloria, line 42: an existing reshape_for_api exists as a Collections class method. Should this also be a Collections class method for consistency? Perhaps with a rename for clarity distinguishing the use case difference

backend/corpora/lambdas/api/v1/common.py
```diff
@@ -35,3 +37,32 @@ def delete_dataset_common(db_session: Session, dataset: Dataset, token_info: dic
...
+    owner = collection["owner"]
+    if is_user_owner_or_allowed(token_info, owner):
+        collection["access_type"] = "WRITE"
+    elif not allow_access and collection["visibility"] == CollectionVisibility.PRIVATE:
```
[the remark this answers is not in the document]
nayib-jose-gloria: actually, to emphasize that it's a check for the user providing the uuid, what about renaming the flag something like `user_has_uuid`? Non-blocking suggestion, it just took me a little bit to decipher allow_access at first

backend/corpora/lambdas/api/v1/common.py
nayib-jose-gloria: I'd like to retain the reusability of this function while also avoiding the messiness of having a function that is both an in-place modifier and also returns a bool evaluating auth, while maintaining a flag param to essentially ignore the auth evaluation. 

what if we separated the access_type evaluation into one function (returning either "WRITE", "READ", or None) and then handled the in-place collection reshape in a separate `reshape_for_curation_api`? In the non-UUID endpoint, we can skip over collections with access type None and in the UUID endpoint we can call the reshape anyway and pass in the access_type as a param to avoid rework. Thoughts?

backend/corpora/lambdas/api/v1/common.py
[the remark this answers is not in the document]
nayib-jose-gloria: ^ you can ignore the above if you like the idea of extracting this into two calls, which would obviate the flag entirely. otherwise, if you want to keep it one call, let me know what you think of this

backend/corpora/lambdas/api/v1/common.py
```diff
@@ -35,3 +37,32 @@ def delete_dataset_common(db_session: Session, dataset: Dataset, token_info: dic
...
+        collection["access_type"] = None
+
+    del collection["owner"]  # Don't actually want to return 'owner' in response
+    collection["collection_url"] = f"{CorporaConfig().collections_base_url}/collections/{collection['id']}"
+
+    if "datasets" in collection:
+        for dataset in collection["datasets"]:
+            if "artifacts" in dataset:
+                dataset["dataset_assets"] = dataset["artifacts"]
+                del dataset["artifacts"]
```
nayib-jose-gloria, line 63: very nit: you could do `dataset["dataset_assets"] = dataset.pop("artifacts")` (this is done in the `reshape_for_api` method currently)

backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -76,20 +76,29 @@ def get_collections_curation(visibility: str, token_info: dict):
...
+
+    import pprint
+
+    pprint.pprint(allowed_collections)
 
     return jsonify({"collections": allowed_collections})
 
 
+@dbconnect
+def get_collection_uuid_curation(collection_uuid: str, token_info: dict):
```
Bento007, line 90: remove curation from the name or change to `get`, and move this to _backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/actions.py_

backend/corpora/lambdas/api/v1/common.py
Bento007: Move this to the function handler that uses it. This function is only used in one place, and doesn't seem like something that will be resued.

>an existing reshape_for_api exists as a Collections method. Should this also be a Collections method for consistency? Perhaps with a rename for clarity distinguishing the use case difference

We should move the `Collection.reshape_for_api` out of the `Collection` class, but that can be done later. We want to move away from adding more functionality to the classes that inherit `Entity` and keep the code specific to the APIs closer to the handlers.

backend/corpora/lambdas/api/v1/common.py
```diff
@@ -35,3 +37,32 @@ def delete_dataset_common(db_session: Session, dataset: Dataset, token_info: dic
...
+
+
+def reshape_for_curation_api_and_is_allowed(collection, token_info, allow_access=False):
+    owner = collection["owner"]
+    if is_user_owner_or_allowed(token_info, owner):
+        collection["access_type"] = "WRITE"
+    elif not allow_access and collection["visibility"] == CollectionVisibility.PRIVATE:
+        # User neither provided the uuid for access nor are they authorized by their access token
+        return False
+    elif token_info:
```
Bento007, line 49: Remove line49-51. Default access-type is "READ". We don't need to check for the existence of token-info to set that.

backend/corpora/lambdas/api/v1/common.py
```diff
@@ -35,3 +37,32 @@ def delete_dataset_common(db_session: Session, dataset: Dataset, token_info: dic
...
+    del collection["owner"]  # Don't actually want to return 'owner' in response
+    collection["collection_url"] = f"{CorporaConfig().collections_base_url}/collections/{collection['id']}"
+
+    if "datasets" in collection:
+        for dataset in collection["datasets"]:
+            if "artifacts" in dataset:
+                dataset["dataset_assets"] = dataset["artifacts"]
+                del dataset["artifacts"]
+            if "processing_status" in dataset:
+                if type(dataset["processing_status"]) == dict:
```
Bento007, line 65: When is this not True?

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -185,6 +185,134 @@ def test__get_only_private_collections_with_auth__OK(self):
...
+            "visibility": "PUBLIC",
+        }
+
+        res_body = res.json
+        del res_body["created_at"]  # too finicky; ignore
+        self.assertDictEqual(expected_body, res_body)
+
+    def test__get_private_collection__OK(self):
+        res = self.app.get("/curation/v1/collections/test_collection_id_revision")
+        self.assertEqual(200, res.status_code)
```
Bento007, line 296: check access_type

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -185,6 +185,134 @@ def test__get_only_private_collections_with_auth__OK(self):
...
+        self.assertEqual("test_collection_id_revision", res.json["id"])
+
+    def test__get_nonexistent_collection__Not_Found(self):
+        res = self.app.get("/curation/v1/collections/test_collection_id_nonexistent")
+        self.assertEqual(404, res.status_code)
+
+    def test__get_public_collection_with_auth__OK(self):
+        # Collection with write access
+        res = self.app.get("/curation/v1/collections/test_collection_id", headers=self.get_auth_headers())
```
Bento007, line 305: These tests are independent. Move into their own tests or subtest.

Bento007 (changes requested):
Some code organization things to take care of.

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: It's used in two places, in the handlers for both `GET /collections` and `GET /collections/{collection_uuid}`

backend/corpora/lambdas/api/v1/common.py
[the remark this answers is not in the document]
danieljhegeman: Sure, I'll think about that. I want to keep it as a flag because the collection id doesn't actually get used in the function itself.

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: But it has to be set explicitly, right?

This is motivated by my most recent comment on the thread in [this issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2083)

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: Great pattern! Thank you

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: Ah, good catch. I had this leftover from when I was going to use the same logic to handle the collection-level aggregate processing status as well, and that is a string.

I'll remove this conditional.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: I do that in the last two tests. Here I'm just focused on private v public.

danieljhegeman: fixed to check access type on private collection call with auth provided ✅

backend/corpora/common/entities/collection.py
```diff
@@ -153,42 +154,85 @@ def to_dict(db_object):
 
         return results
 
+    collections_cols = [
```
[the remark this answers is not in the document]
danieljhegeman: moved out to class `EntityColumns` in `common.py`

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: we can revisit this particular issue once Brian lets us know what he wants

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/actions.py
```diff
@@ -16,3 +19,16 @@ def delete(collection_uuid: str, token_info: dict):
     else:
         collection.delete()
     return "", 204
+
+
+@dbconnect
+def get_collection_uuid(collection_uuid: str, token_info: dict):
```
nayib-jose-gloria, line 25: nit: this could just be `get`, the function is clear from the filepath. Would conform to collection uuid `delete` action's naming convention, too.

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -0,0 +1,121 @@
...
+    CollectionVisibility,
+    DbCollectionLink,
+    DbCollection,
+    DbDataset,
+    DbDatasetProcessingStatus,
+    DbDatasetArtifact,
+)
+
+
+def reshape_for_curation_api_and_is_allowed(collection, token_info, allow_access=False):
```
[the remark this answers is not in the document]
nayib-jose-gloria: that said, I also still think it might be a cleaner pattern to extract the access_type bit into a helper function, do the boolean eval in the handlers that need it, and do only in-place modification in the reshape function (and thus get rid of the flag altogether).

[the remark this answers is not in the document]
Bento007: @nayib-jose-gloria in a separate branch, if you'd like to demonstrate how that'd look, I'd be happy to review and give my opinion.

backend/corpora/common/entities/collection.py
```diff
@@ -153,50 +151,15 @@ def to_dict(db_object):
...
     @classmethod
-    def list_collections_for_curation(cls, session: Session, visibility: str = None) -> typing.List[dict]:
+    def list_collections_curation(
+        cls, session: Session, collection_columns: list, visibility: str = None
+    ) -> typing.List[dict]:
         """
         Get a subset of columns, in dict form, for all Collections with the specified visibility. If visibility is None,
         return *all* Collections.
         @param session: the SQLAlchemy session
+        @param collection_columns: the list of columns to be returned (see usage by TransformingBase::to_dict_keep)
```
Bento007, line 162:
```suggestion
        :param collection_columns: the list of columns to be returned (see usage by TransformingBase::to_dict_keep)
```

Bento007: update the rest of the parameters for consistency.

danieljhegeman: I'm confused,  you want to change this line back to `:`?

danieljhegeman: I thought you had asked to move to `@` recently

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -0,0 +1,121 @@
...
+        collection["access_type"] = "WRITE"
+    elif not allow_access and collection["visibility"] == CollectionVisibility.PRIVATE:
+        # User neither provided the uuid for access nor are they authorized by their access token
+        return False
+    elif token_info:
+        # Access token was provided but user is not authorized
+        collection["access_type"] = "READ"
+    else:
+        # No access token was provided
+        collection["access_type"] = None
```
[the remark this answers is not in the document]
danieljhegeman: I think I'm more in favor of keeping the response shape the same (did you reference the conversation that I linked to in which Andrew and Brian and I were discusssing this?)

But I'm fine removing this for now. We can revisit. Will change accordingly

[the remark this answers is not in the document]
danieljhegeman: done

backend/corpora/common/entities/collection.py
Bento007: everything else is using `:`. If we switch to `@` we need to do it all at once to avoid confusion. I don't think the migration to `@` is actually needed.

danieljhegeman: made the change

Bento007 (approved):

Issue #2516, opened 2022-05-10
Create resources for wmg testing


MDunitz:
## Need
- to support testing wmg in the local environment 

## Approach
- store all aspects of the snapshot in the local stack S3 bucket (add to [setup dev data script](https://github.com/chanzuckerberg/single-cell-data-portal/blob/7cdcf09a5bd276e0a8f63cc000a0b2337d7ab413/scripts/setup_dev_data.sh#L105-L104))
    - cell count cube
    - primary filter dimensions
    - cell ordering 

## DOD
- wmg works in the local env

atarashansky:
Progress prior to deprioritization in #2717:
- Added function `create_temp_wmg_snapshot_to_file` to save temporary WMG snapshots to disk for subsequent upload into the local AWS S3 endpoint.
- Added function `simple_ontology_terms_generator_local` to generate fake attribute IDs (with real cell type IDs) for the temporary snapshot

The frontend is capable of displaying the proper cell type information but:
- Heatmap is erroneously displaying `No data`
- The interface is extremely slow, even with a temporary snapshot with a tiny dimensionality of 2

Question for the future: Do we need a local AWS S3 endpoint via `localstack`? Can we hook up all the local `backend/frontend/processing/etc` containers to our actual AWS S3 buckets? That should fix a lot of the slowness.

PR #2717, opened 2022-06-16
chore: created resources for wmg testing in localstack (#2516)

atarashansky:
- #2516 

### Reviewers
**Functional:** 
@MDunitz 

## Changes
- Fixed the dummy wmg snapshot creation

### DIFF
--- a/backend/wmg/data/snapshot.py
+++ b/backend/wmg/data/snapshot.py
@@ -83,7 +83,11 @@ def _load_snapshot(new_snapshot_identifier) -> WmgSnapshot:
 
 
 def _open_cube(cube_uri) -> Array:
-    return tiledb.open(cube_uri, ctx=create_ctx(json.loads(WmgConfig().tiledb_config_overrides)))
+    config = WmgConfig()
+    tiledb_config_overrides = config.tiledb_config_overrides
+    if type(tiledb_config_overrides) is str:
+        tiledb_config_overrides = json.loads(tiledb_config_overrides)
+    return tiledb.open(cube_uri, ctx=create_ctx(tiledb_config_overrides))
 
 
 def _load_cell_type_order(snapshot_identifier: str) -> DataFrame:
--- a/scripts/setup_dev_data.sh
+++ b/scripts/setup_dev_data.sh
@@ -108,7 +108,7 @@ ${local_aws} s3api create-bucket --bucket ${wmg_bucket} &>/dev/null || true
 ${local_aws} s3 sync --delete --quiet ${tmp_snapshot_dir} s3://${wmg_bucket}/$snapshot_identifier/
 echo $snapshot_identifier | ${local_aws} s3 cp --quiet - s3://${wmg_bucket}/latest_snapshot_identifier
 ${local_aws} secretsmanager create-secret --name ${wmg_config_secret_name} &>/dev/null || true
-${local_aws} secretsmanager update-secret --secret-id ${wmg_config_secret_name} --secret-string '{"bucket": "wmg-test", "tiledb_config_overrides": "{\"py.init_buffer_bytes\": 536870912, \"sm.tile_cache_size\": 134217728, \"sm.mem.total_budget\": 1073741824, \"sm.memory_budget\": 536870912, \"sm.memory_budget_var\": 1073741824}"}'
+${local_aws} secretsmanager update-secret --secret-id ${wmg_config_secret_name} --secret-string '{"bucket": "wmg-test"}'
 
 echo
 echo "Dev env is up and running!"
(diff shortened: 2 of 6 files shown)


atarashansky:
@MDunitz Current problem with the test data: the expression summary and cell count cubes are populated when reading them with tiledb directly, but querying them with `WmgQuery` yields empty dataframes.

atarashansky:
> @MDunitz Current problem with the test data: the expression summary and cell count cubes are populated when reading them with tiledb directly, but querying them with `WmgQuery` yields empty dataframes.

The query is entering the hack of the century bit:
https://github.com/chanzuckerberg/single-cell-data-portal/blob/18e4b93421e1d03cfa42b6e5e1ecac2f75c5863d/backend/wmg/data/query.py#L74

Still poking around to see why the query is returning empty...

atarashansky:
Ah, silly me. The ontology term IDs used for the test are not the same as the ones generated for the test snapshot. Fixing now.

atarashansky:
Ok, query is completing but frontend is hanging on "Loading..." and not displaying expressions or cell types...Investigating further.

scripts/setup_dev_data.sh
```diff
@@ -103,11 +103,12 @@ snapshot_identifier='dummy-snapshot'
 wmg_bucket="wmg-test"
 wmg_config_secret_name="corpora/backend/test/wmg_config"
 python3 -m tests.unit.backend.wmg.fixtures.test_snapshot ${tmp_snapshot_dir}
+
 ${local_aws} s3api create-bucket --bucket ${wmg_bucket} &>/dev/null || true
 ${local_aws} s3 sync --delete --quiet ${tmp_snapshot_dir} s3://${wmg_bucket}/$snapshot_identifier/
 echo $snapshot_identifier | ${local_aws} s3 cp --quiet - s3://${wmg_bucket}/latest_snapshot_identifier
 ${local_aws} secretsmanager create-secret --name ${wmg_config_secret_name} &>/dev/null || true
-${local_aws} secretsmanager update-secret --secret-id ${wmg_config_secret_name} --secret-string "{\"bucket\": \"${wmg_bucket}\"}" || true
+${local_aws} secretsmanager update-secret --secret-id ${wmg_config_secret_name} --secret-string '{"bucket": "wmg-test"}'
```
MDunitz, line 111: Why is this hardcoded rather than using the wmg_bucket var defined on line 103?

tests/unit/backend/wmg/fixtures/test_primary_filters_local.py
```diff
@@ -0,0 +1,42 @@
+test_organism_terms = [
+    {"organism_ontology_term_id_0": "Mus musculus"},
```
MDunitz, line 2: If possible I would use the correct ontology ids
Many of them are predefined in `tests/unit/backend/wmg/fixtures/test_primary_filters.py`

backend/wmg/data/snapshot.py
```diff
@@ -83,7 +83,11 @@ def _load_snapshot(new_snapshot_identifier) -> WmgSnapshot:
 
 
 def _open_cube(cube_uri) -> Array:
-    return tiledb.open(cube_uri, ctx=create_ctx(json.loads(WmgConfig().tiledb_config_overrides)))
+    config = WmgConfig()
+    tiledb_config_overrides = config.tiledb_config_overrides
+    if type(tiledb_config_overrides) is str:
```
MDunitz, line 88: Why do we need a type check here? Is it an issue with the default template? I think it would make more sense to change it there for consistency

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -28,11 +28,31 @@
...
 
 
 def simple_ontology_terms_generator(dimension_name: str, n_terms: int) -> List[str]:
     return [f"{dimension_name}_{i}" for i in range(n_terms)]
 
+def simple_ontology_terms_generator_local(dimension_name: str, n_terms: int) -> List[str]:
+    if dimension_name == "cell_type_ontology_term_id": 
+        # everything can be fake but cell types need to be real
+        # because they are queried against the ontology for the display labels
+        import backend.wmg.data.ontology_labels as ontology_labels
```
MDunitz, line 41: Is there a reason this is a local import?

MDunitz (changes requested):
see comments

scripts/setup_dev_data.sh
atarashansky: Because I'm silly - thanks for the catch.

backend/wmg/data/snapshot.py
atarashansky: I tried changing the empty dict in the defaults template to "{}" but it broke the API elsewhere. I'll investigate further.

atarashansky:
Backburnered for a future CD period.

Issue #2169, opened 2022-03-16
Curators can get the processing status of an uploaded Dataset


danieljhegeman:
Implement `GET` `/curation/collections/{collection_uuid}/datasets/status` to allow a curator to retrieve the processing status for any Dataset for which they are authorized.

[See the spec](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2148)

PR #2719, opened 2022-06-16
feat: GET /datasets/status

Bento007:
### Reviewers

---

## Changes
- add get dataset status handler
- change get_collection to authorize_get_collection
- #2169


backend/corpora/lambdas/api/v1/common.py
```diff
@@ -2,11 +2,15 @@
...
+from backend.corpora.common.utils.http_exceptions import (
+    ForbiddenHTTPException,
+    MethodNotAllowedException,
+    InvalidParametersHTTPException,
+)
 from backend.corpora.lambdas.api.v1.authorization import owner_or_allowed
 
 
-def get_collection(db_session, collection_uuid, **kwargs):
+def authorize_get_collection(db_session, collection_uuid, **kwargs):
```
danieljhegeman, line 13: I would call this `get_collection_if_authorized` or `get_collection_authorized`

backend/corpora/lambdas/api/v1/common.py
```diff
@@ -35,3 +39,13 @@ def delete_dataset_common(db_session: Session, dataset: Dataset, token_info: dic
                 original.create_revision(dataset.collection.id)  # Restore the original dataset and S3 assets
             dataset.asset_deletion()  # Delete the S3 assets and database rows.
             dataset.delete()  # Delete the dataset row.
+
+
+def validate_dataset_identifier(db_session, collection_uuid, dataset_uuid, curator_tag) -> Dataset:
```
danieljhegeman, line 44: this doesn't really describe (to me) what this function does.

How about `get_dataset_by_identifier`?

danieljhegeman: Good refactor though to consolidate the shared logic 👍

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: `get_authorized_collection_or_error`

tests/unit/backend/corpora/api_server/curator/collection/datasets/test_status.py
```diff
@@ -0,0 +1,34 @@
...
+        curator_tag = "tag.h5ad"
+        processing_status = {
+            "processing_status": ProcessingStatus.PENDING.name,
+        }
+        queries = ["dataset_uuid", "curator_tag"]
+        auth_credentials = [
+            (self.make_super_curator_header, "super", 200),
+            (self.get_auth_headers, "owner", 200),
+            (None, "none", 401),
+            (self.make_not_owner_header, "not_owner", 403),
```
danieljhegeman, line 16: This should be a 404 (with `404` added to the API `responses` section of the connexion yml as well)

danieljhegeman: Well, actually, I'm unclear on this. I know we decided to use 404 for certain scenarios where the resource exists but the person isn't authorized... and I can't remember what those were...

But since we're allowing collection uuid's to serve as capability entities, _should_ we allow someone to get the processing status of a dataset if they're not the owner but they _do_ have the collection id? 🤔

danieljhegeman (commented):
@Bento007 some q's / comments

backend/corpora/lambdas/api/v1/common.py
nayib-jose-gloria: I think if this is meant to be an auth-gated call for collections, `token_info` should be a required parameter ala `delete_dataset_common` below. And at that point, I think the required param would speak for itself and we could stick to a consistent, simpler `get_collection_common`. 

Alternatively, if we need to keep the auth parameters optional for some use case, we probably shouldn't reference auth in the name (and maybe consider making separate function for unauth and auth use cases)

nayib-jose-gloria: thinking about it more, I do see the value of being descriptive with a name like `get_authorized_collection` so I could be convinced to go that route. but I do think it should require `token_info` if so

backend/corpora/lambdas/api/v1/common.py
nayib-jose-gloria: +1 on both of dan's comments

backend/corpora/lambdas/api/v1/dataset.py
```diff
@@ -63,17 +63,12 @@ def get_status(dataset_uuid: str, token_info: dict):
...
         dataset.collection.id,
-        dataset.collection.visibility,
         owner=owner_or_allowed(token_info),
     )
-    if not collection:
-        raise ForbiddenHTTPException()
-    status = dataset.processing_status.to_dict(remove_none=True)
-    for remove in ["dataset", "created_at", "updated_at"]:
-        status.pop(remove)
+    status = dataset.processing_status.to_dict(remove_none=True, remove_attr=["dataset", "created_at", "updated_at"])
```
nayib-jose-gloria, line 71: we are also removing `id` and `dataset_id` in the call in status.py--is it intentional that they are not removed in this similar call?

nayib-jose-gloria (commented):
added a couple of comments/questions but overall this looks good!

tests/unit/backend/corpora/api_server/curator/collection/datasets/test_status.py
Bento007: Is there anything a user should not be able to read?

backend/corpora/lambdas/api/v1/common.py
Bento007: change name to get_dataset_else_error

backend/corpora/lambdas/api/v1/common.py
Bento007: change name to get_collection_else_forbidden

tests/unit/backend/corpora/api_server/curator/collection/datasets/test_status.py
Bento007: changed to 404.

backend/corpora/lambdas/api/v1/common.py
danieljhegeman: @Bento007 should we change the name of `delete_dataset_common` to `delete_dataset_else_forbidden_common`? It's not crucial, and at some point these names become cumbersome, but it is worth considering that it uses the same pattern as `get_collection_else_forbidden`. 

Generally-speaking I think we have more refactoring / tech debt to take care of wrt standards for how our methods act on our various entities, but not something we necessarily need to address before we wrap up this project, so it can wait

danieljhegeman: also... I'm a little confused: how did this all come up in this PR? None of this relates to getting the status for a dataset, does it... 😅

tests/unit/backend/corpora/api_server/curator/collection/datasets/test_status.py
```diff
@@ -0,0 +1,73 @@
...
+        for query in self.queries:
+            for auth, auth_description in self.auth_credentials:
+                with self.subTest(f"{query} {auth_description} 404"):
+                    headers = auth() if callable(auth) else auth
+                    query_string = {query: "fake_id"} if query == "dataset_uuid" else {query: "fake_tag"}
+                    response = self.app.get(self.test_url, headers=headers, query_string=query_string)
+                    self.assertEqual(404, response.status_code)
+
+    def test__get_status_wrong_collection__404(self):
+        """Return 404 when the dataset is not part of the collection."""
```
danieljhegeman, line 50:
```suggestion
        """Return 404 when the dataset exists but is not part of the collection."""
```

danieljhegeman:
Just thinking about this again: why exactly are we requiring `collection_uuid` for this endpoint? 🤔 cc @Bento007

backend/config/curation-api.yml
```diff
@@ -469,8 +469,7 @@ paths:
       description: >-
         Get the processing status for a Dataset. MUST indicate the Dataset by providing *one* of either the
         **curator_tag** (non-null) OR the **dataset_uuid** parameters.
-      security:
-        - curatorAccess: []
```
danieljhegeman, line 473: 👍

nayib-jose-gloria (approved):
looks good to me wrt readability!

backend/corpora/lambdas/api/v1/common.py
Bento007: it seems like a small, uncontroversial, and beneficial change. I'm going to leave things as is for now. I don't want to introduce any additional controversial changes in this PR.

Bento007:
It's not required. The old API didn't use it, but this is the API we agreed on with our PM.

Issue #2697, opened 2022-06-13
scExpression populate cell type names when a user selects a tissue


signechambers1:
Right now when a user selects a tissue, the tissue name appears with no additional information. The backend API requires both a tissue and a gene to return cell types.

Ideally, when a user selects a tissue the visualization would update with the tissue name, cell types, and [cell counts](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2696).

This work builds on improvements made in [#2336](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2336)

atarashansky:
Thanks for making this ticket - I'm copying my proposed solution from this (now duplicate) issue: #2686

To further elaborate, I believe users should be able to see what cell types from the ontology are present in the tissue without needing to query a gene. Currently, selecting tissues without selecting any genes displays the following:

![image](https://user-images.githubusercontent.com/16548075/173465538-5db38f58-0b7f-4e76-8454-05629289e45c.png)

The motivation for omitting the cell types is due to concerns about performance issues when querying the backend without a gene specified. 

My suggestion would be to skip the following lines of code if genes are not specified and return undefined values to the frontend:
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/f41fb6f2c1aaf42e3f68d948592467418171681b/backend/wmg/api/v1.py#L36
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/f41fb6f2c1aaf42e3f68d948592467418171681b/backend/wmg/api/v1.py#L38
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/f41fb6f2c1aaf42e3f68d948592467418171681b/backend/wmg/api/v1.py#L39
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/f41fb6f2c1aaf42e3f68d948592467418171681b/backend/wmg/api/v1.py#L42
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/f41fb6f2c1aaf42e3f68d948592467418171681b/backend/wmg/api/v1.py#L49
- https://github.com/chanzuckerberg/single-cell-data-portal/blob/f41fb6f2c1aaf42e3f68d948592467418171681b/backend/wmg/api/v1.py#L47

[...]

PR #2729, opened 2022-06-17
feat: enable empty gene queries to retreive cell type ID labels

atarashansky:
- Solves #2697 

### Reviewers
**Functional:** 
@seve 

## Changes
- removed a couple checks that returned early if genes were not specified
- if no genes selected, pass a dummy gene (".") to the backend to get the cell type IDs

image


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -599,19 +591,13 @@ function useWMGQueryRequestBody(options = { includeAllFilterOptions: false }) {
...
-    ) {
+    if (!data || !selectedOrganismId || !selectedTissues.length) {
       return null;
     }
-
     const gene_ontology_term_ids = selectedGenes.map((geneName) => {
       return organismGenesByName[geneName].id;
     });
-
+    if (!gene_ontology_term_ids.length) gene_ontology_term_ids.push(".");
```
MillenniumFalconMechanic, line 600: Passing "." as a gene param value here is a workaround that:

1. prevents the backend from throwing a "no gene specified" exception,
1. does not match on any gene ontology term ID, and
1. allows the backend to return all tissue names and cell types for the selected set of tissues.

I think we have two options here:

1. Leave this workaround as-is, add comments explaining the "." and also create a corresponding ticket to update the backend to handle the "no gene specified" case. The workaround will then be removed once the backend is updated.
1. Remove this workaround and update the backend to handle the "no gene specified case".

What are your thoughts, @seve?

MillenniumFalconMechanic:
Hey @atarashansky! Is there a separate ticket for the "cell count" functionality referenced in [scExpression populate cell type names when a user selects a tissue](https://github.com/chanzuckerberg/single-cell-data-portal/issues/2697)? If not, can you update this PR's description with "part 1 of 2" text, or similar?

frontend/src/common/queries/wheresMyGene.ts
MillenniumFalconMechanic: I think my preference would be to go with option 2 (update the backend to handle the "no gene specified" case) but am happy to go with the team's decision!

seve: If we want to get this unblocked asap, option 1 is best while we wait for folks to come back online.  I'd defer to @signechambers1

signechambers1: this is not urgent so we can try for option 2, @atarashansky can you make the backend change or do we need to wait for madison to come back?

atarashansky: @signechambers1 I can try to go for it, but there is some weirdness with how the backend deals with empty queries. That's why I opted for the frontend change. 

Personally, I think we should go with option 1. It's a bit of a hack but it won't require significant changes to the backend. If I were to do this in the backend, I'd do the same thing anyway - check if it's an empty gene list and then add a dummy gene to not break anything.

atarashansky: @signechambers1 
I tried forcing the backend to accept no genes (by setting [this](https://github.com/chanzuckerberg/single-cell-data-portal/blob/ae3315869c4655b9e83db4c32bc435bb9a3dec9c/backend/wmg/data/query.py#L14) line), and it works - the only issue is that it takes forever to load because of how tiledb is handling the query with no genes selected. 

Simplest route forward would be to go with the frontend fix.

signechambers1: ok, we can go with option 1. @atarashansky can you create a backend ticket for madison to take a look at to determine if we should eventually make the change there without negatively impacting performance?

seve (approved):

PR #2730, opened 2022-06-17
chore: reordered dockerfile to improve rebuild speed

atarashansky:
Rebuilding the stack took forever because any changes to the backend code triggers the reinstallation of all the requirements in the `wmg_processing` image. We can fix this by moving the `pip3 install` further up in the `data_cube_creation` Dockerfile. This also means we don't need to push the bulky python dependency layer every time.
### Reviewers
**Functional:** 
@MDunitz


Bento007 (approved):
Nice fix

MDunitz (approved):
Lgtm thanks!

PR #2745, opened 2022-06-21
chore(scExpression): updated tests for WMG to fix failing tests on deployment

atarashansky:
@Bento007 Just need to activate the tests for this PR to confirm they're good.


atarashansky:
@seve Tests look good now.

Issue #1588, opened 2021-10-26
Only send 'Dataset failed to process' Slack notification after last retry


metakuni:
The dataset ingestion step function retries failed tasks a maximum of 2 times. Currently, this can result in 3 Slack notifications for the same error. A Slack notification should only be triggered after the last retry fails.

Example:
* [notification #1](https://czi-sci.slack.com/archives/C0244PQK934/p1653351152568249)
* [notification #2](https://czi-sci.slack.com/archives/C0244PQK934/p1653351494499309)
* [notification #3](https://czi-sci.slack.com/archives/C0244PQK934/p1653351652932509)

metakuni:
Moving to cool down as per [Triaging Rollover Q4 Work](https://docs.google.com/document/d/1QEZfVRVO4TgHgaf3riDEiKRi9YN8MG6HBRlFCq0cSwM/edit#heading=h.4no2xl67bsql)

metakuni:
Removed the `P0` label since this is a non-critical fix.

PR #2746, opened 2022-06-21
Only send 'Dataset failed to process' Slack notification after last retry

nayib-jose-gloria:
- #1588

### Reviewers
**Functional:** 
@metakuni 
**Readability:** 
@Bento007  
---

## Changes
- Moved functions related to slack alerts to upload_failure lambda / HandleErrors step. As a result, slack notification only occurs once, after retries, at the end of processing.

## QA steps (optional)
- Tested in rdev--uploaded dataset with validation error and confirmed that only 1 Slack alert was fired, after failing all retries

## Notes for Reviewer
- NOTE: Because the frontend is polling the db for dataset status while we are uploading, it will display errors to the user before retries are done. This means, with this change, users will see an error message before we are notified on Slack (since it waits for the last retry to fail on any previous step before alerting). This leaves a window where the user could delete the offending dataset before we are alerted about the failure, which would silence the alert. Should we consider alerting only once, but at the first failure? Another solution? Is that edge case ok to have?


metakuni:
> NOTE: Because the frontend is polling the db for dataset status while we are uploading, it will display errors to the user before retries are done. This means, with this change, users will see an error message before we are notified on Slack (since it waits for the last retry to fail on any previous step before alerting). This leaves a window where the user could delete the offending dataset before we are alerted about the failure, which would silence the alert. Should we consider alerting only once, but at the first failure? Another solution? Is that edge case ok to have?

I was thinking about this a bit more last night and this doesn't only affect the portal frontend, but also applies to the curation API. I think it's worthy of a discussion in #single-cell-curation-api to get some consensus on it.

Bento007 (approved):

backend/corpora/dataset_processing/process.py
```diff
@@ -219,12 +216,8 @@ def create_artifact(
...
         )
 
-    except Exception as e:
-        update_db(
-            dataset_id,
-            processing_status={processing_status_type: ConversionStatus.FAILED},
-        )
-        raise e
+    except Exception:
+        raise Exception({processing_status_type: ConversionStatus.FAILED})
```
Bento007, line 220: Won't this raise a generic Exception and not the original exception? We wont get the same error information by calling a generic Exception.

Bento007: you can add the `processing_status` to the exception so it can be used in the exception handler in `main`

```
e.cellxgene_processing_status = {processing_status_type: ConversionStatus.FAILED}
```

This way we can still preserve the original exception. Or you can log the exception here instead and then call your own exception.

Bento007 (changes requested):
If an exception occurs we still want to know what type occurred.

danieljhegeman:
@nayib-jose-gloria what do you mean by
> Monitored dataset status throughout with the dataset curation endpoint, and it also provided updates that indicated progress, but not failure, until all retries were done.

?

danieljhegeman:
> > NOTE: Because the frontend is polling the db for dataset status while we are uploading, it will display errors to the user before retries are done. This means, with this change, users will see an error message before we are notified on Slack (since it waits for the last retry to fail on any previous step before alerting). This leaves a window where the user could delete the offending dataset before we are alerted about the failure, which would silence the alert. Should we consider alerting only once, but at the first failure? Another solution? Is that edge case ok to have?
> 
>  ...since API clients will also have to poll for ingestion status.

@metakuni ??

danieljhegeman:
I think one outstanding question is whether we actually want to retry multiple times.

.happy/terraform/modules/sfn/main.tf
```diff
@@ -17,6 +17,9 @@ resource "aws_sfn_state_machine" "state_machine" {
           "JobDefinition": "${var.job_definition_arn}",
           "JobName": "download-validate",
           "JobQueue": "${var.job_queue_arn}",
+          "RetryStrategy": {
+            "Attempts": ${var.max_attempts}
```
danieljhegeman, line 21: Is this the total number of attempts? Or total number of retries, i.e., `total number of attempts` - 1

backend/corpora/dataset_processing/process.py
```diff
@@ -218,7 +218,8 @@ def create_artifact(
 
     except Exception as e:
         logger.error(e)
-        raise Exception({processing_status_type: ConversionStatus.FAILED})
+        e.args = ({processing_status_type: ConversionStatus.FAILED})
```
danieljhegeman: 👍

backend/corpora/dataset_processing/process.py
```diff
@@ -220,10 +217,8 @@ def create_artifact(
         )
 
     except Exception as e:
-        update_db(
-            dataset_id,
-            processing_status={processing_status_type: ConversionStatus.FAILED},
-        )
+        logger.error(e)
+        e.args = {processing_status_type: ConversionStatus.FAILED}
```
danieljhegeman, line 221: 👍

nayib-jose-gloria:
> @nayib-jose-gloria what do you mean by
> 
> > Monitored dataset status throughout with the dataset curation endpoint, and it also provided updates that indicated progress, but not failure, until all retries were done.
> 
> ?
> 
> Indicated progress, but not failure?

Users hitting the curation API status endpoint still see updates to statuses that indicate progress (UPLOADING, UPLOADED, VALIDATING, SKIPPED, etc.) in real-time. However, failure statuses (FAILED, INVALID) are not saved or shown to users until a step fails on the last retry.

nayib-jose-gloria:
> I think one outstanding question is whether we actually want to retry multiple times.

Yeah, this is a key question. The [RetryStrategy config](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-batch-jobdefinition-retrystrategy.html) actually allows us to limit retries to specific Errors, which could be an interesting approach should we want to only retry on AWS or network failures.

Otherwise, this PR is set up such that we could switch max attempts to 1 painlessly in the future. Unless we have a strong inclination now, I think we should probably make a ticket re: retry strategy and revisit in a separate PR.

Bento007 (approved):
Great job

Issue #2740, opened 2022-06-19
bug(scExpression): adding a gene or a tissue should only query the backend once


atarashansky:
Adding a gene or tissue currently queries the backend four times(!) 

Setup: Local frontend server hooked up to `dev` backend.

This is likely to do with the fact that queries are happening separately in `useExpressionSummary` and `useTermIdLabels`. As these functions are querying with the same request body, we should consolidate them into one request. But this only explains 2/4 queries.

`primary_filter_dimensions` is fetched twice when loading scExpression --> this could be coming from whatever is 2x-ing the `query` fetches.

atarashansky:
Debugging: Added log statements to `useTermIdLabels` and `useExpressionSummary` in `wheresMyGene.ts` and received the following output:
<img width="246" alt="image" src="https://user-images.githubusercontent.com/16548075/174494765-e76453f9-91cf-4442-9f03-c2a5fc91faf6.png">

Every time the frontend state gets updated, these same log statements are output.

atarashansky:
Starting at `2022-06-19T11:15:10.335-07:00`, these [cloudwatch logs](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Fdp$252Fdev$252Fdevstack$252Fbackend/log-events/1c1486f2377c83e1a916dd215c794b2508645119e42fd2f37d924b50caf80f79) show the same result when displaying the heatmap from the `dev` (not local) frontend.

`primary_filter_dimensions` is triggered twice, and `query` four times.

atarashansky:
`useCellTypesByTissueName` and `useGeneExpressionSummariesByTissueName` each call `useTermIdLabels` and `useSummaryExpression` which totals to 4 `query` fetches. Fetches should be memo-ized by the request body.

atarashansky:
Looking into it more, I realized that `useQuery` is already memo-ized. The problem is that the query is getting populated with values asynchronously, triggering multiple calls to `useQuery`.

atarashansky:
Also, the latest snapshot ID in the devstack s3 bucket is `rankit2`. However, initially the snapshot ID logged in the frontend is `1654985300`. After querying the backend, the latest snapshot ID is properly updated to `rankit2`, which triggers `useQuery` to run again. Checking the cellxgene-wmg-dev bucket, the snapshot ID should just always be `rankit2`, so I'm not sure where `1654985300` is coming from.

PR #2756, opened 2022-06-22
fix(scExpression): scExpression api calls are only triggered once now #2740

atarashansky:
- #2740 

### Reviewers
**Functional:** 
@seve 

This PR fixes one of the underlying causes of the bug described in #2740. Some of the filter attributes were initialized as undefined and then subsequently set to empty arrays, triggering the memo-ized `useQuery` function to run twice. Now, these filter attributes will be immediately converted to empty arrays if they are undefined.

The other issue resulting in multiple API calls when loading scExpression is when `latest_snapshot_identifier` is _not a timestamp_ (e.g. `rankit1`). This is a weird bug that @MDunitz and I haven't figured out yet, but seems to be a development edge case since snapshots are almost always timestamped except for testing purposes.


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -618,13 +617,13 @@ function useWMGQueryRequestBody(options = { includeAllFilterOptions: false }) {
 
     return {
       filter: {
-        dataset_ids: datasets,
-        development_stage_ontology_term_ids: developmentStages,
-        disease_ontology_term_ids: diseases,
-        ethnicity_ontology_term_ids: ethnicities,
+        dataset_ids: datasets ?? [],
```
MillenniumFalconMechanic, line 620: I think it would be worthwhile determining the root cause of the `undefined` values here (and the need for the nullish coalescing); the answer will likely be valuable knowledge for the team. @atarashansky, can you take another look at determining where the `undefined` values are coming from?

seve: bump!

atarashansky: Long story short - I haven't been able to figure it out yet.

atarashansky: @MillenniumFalconMechanic I found the root cause. It turns out they were getting initialized to undefined in the reducer. I am setting it now to empty arrays by default and removed the nullish coalescing.

seve (approved):

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -11,6 +11,13 @@ export interface PayloadAction<Payload> {
   payload: Payload;
 }
 
+const EMPTY_FILTERS: State["selectedFilters"] = {
```
MillenniumFalconMechanic, line 14: Does it make sense to reuse the existing `EMPTY_FILTERS` object [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2756/files#diff-cf34339b9c061a8efa652b2ec5e97f93ea9d16bb21cc3a69af0578bd0175b910R539)? You wouldn't then have to remember to keep both in sync. (This isn't an option of course if the existing `EMPTY_FILTERS` is mutated somewhere in the code.)

Issue #2691, opened 2022-06-13
Implement GET /v1/collections/{collection_uuid}


danieljhegeman:
The curator should be able to `GET /v1/collections/{collection_uuid}` for full Collection metadata (including full Dataset data).

Issue #2083, opened 2022-03-09
The description for GET /curation/collections needs more detail


brianraymor:
https://app.swaggerhub.com/apis/danieljhegeman/Curation/1#/default/get_curation_collections

_Fetch all Collections metadata. Authorization required for private collections._

It neither fetches _all collections_ nor _all collections metadata_.  Perhaps something more like:

1. When the `visibility` parameter is unspecified or set to `"PUBLIC"`, a list of all public collections is returned. The Authorization header is not required.  If a collection in the list has been deleted, then it is annotated with `tombstone` set to `True`.
1. When the `visibility` parameter is set to `"PRIVATE"`, a list of all private collections that the user is authorized to access is returned. The Authorization header is required. If a collection in the list is a private revision of a public collection, then it is annotated with `revision_of`.
1. Each collection in the list contains a subset of collection metadata as a preview.
1. To retrieve full collection metadata for a specific collection, `GET /curation/collections/{collection_uuid}` must be used.

**(danieljhegeman note: despite lengthy discussion below, these rules still stand as stated)**

danieljhegeman:
> It neither fetches all collections nor all collections metadata

What collection metadata does it not fetch? Do you mean to refer to the fact that _dataset_ metadata is only a preview?

danieljhegeman:
> Each collection in the list contains a subset of collection metadata as a preview.

what do you mean by subset? Are you intending to contrast that with what is returned for a single collection with GET `/curation/collections/{collection_id}` ?

brianraymor:
> Is this a reference to the fact that the metadata for the associated datasets is different?

Yes. There is a _preview_ of the dataset metadata which is part of the collection metadata.

brianraymor:
Quick assessment:

|                                     | GET collections | GET collection |
|-------------------------------------|:-----------:|:----------:|
| access_type                         |      x      |      x     |
| contact_email                       |      x      |      x     |
| contact_name                        |      x      |      x     |
| created_at                          |      x      |      x     |
| curator_name                        |      x      |      x     |
| datasets/assay                      |      x      |      x     |
| datasets/batch_condition            |             |      x     |
| datasets/cell_count                 |             |      x     |
| datasets/cell_type                  |             |      x     |
| datasets/curator_tag                |      x      |      x     |
| datasets/dataset_assets             |             |      x     |
| datasets/dataset_deployments        |             |      x     |
| datasets/disease                    |      x      |      x     |
[...]

PR #2758, opened 2022-06-22
refactor: reshape_for_curator_api function

nayib-jose-gloria:
- #2691

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 
@danieljhegeman 
---

## Changes
- extracted access_type eval logic from reshape_for_curator_api function, into helper function for logical separation of tasks.
- Updated corresponding API handlers using the call

## Notes for Reviewer
- I did end up using the uuid_provided flag, but I think overall this reads a little simpler 
- Will look into tests, etc. if you like the idea but will otherwise drop it if you don't think it's necessary


Bento007 (commented):
I like it. See if there are other places where `get_access_type` can replace `is_user_owner_or_allowed`.

danieljhegeman:
@nayib-jose-gloria I may have us wait on addressing this due to these [recent discussions](https://github.com/chanzuckerberg/single-cell-data-portal/issues/2083) (at the end)

PR #2761, opened 2022-06-22
feat(curator api): Add endpoint to update curator tags

Bento007:
### Reviewers

---

## Changes
- Add test cases
- move reused regex into corpora/common/utils

## QA steps (optional)
- run unit-tests


backend/corpora/common/utils/regex.py
```diff
@@ -0,0 +1,7 @@
+USERNAME_REGEX = r"(?P<username>[\w\-\|]+)"
+UUID_REGEX = r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
+EXTENSION_REGEX = r"(?P<extension>h5ad)"
+DATASET_ID_REGEX = f"(?P<dataset_uuid>{UUID_REGEX})"
+COLLECTION_ID_REGEX = f"(?P<collection_uuid>{UUID_REGEX})"
+CURATOR_TAG_NAME_REGEX = r"(?P<tag>.*)"
```
danieljhegeman, line 6:
```suggestion
CURATOR_TAG_NAME_REGEX = r"(?P<tag>\.*)"
```

nayib-jose-gloria (approved):

backend/corpora/common/utils/regex.py
danieljhegeman: Need to capture the entire tag but not extend into the extension, delimited by `.`

danieljhegeman: Actually, nvm, I think the way it's being used in `REGEX` below will prevent that from happening 👍

danieljhegeman:
The `curator_tag` needs to _always_ include `.h5ad` or _never_ include it.

I vote for the former.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/datasets/actions.py
```diff
@@ -0,0 +1,31 @@
...
+@dbconnect
+def patch(token_info: dict, collection_uuid: str, body: dict, curator_tag: str = None, dataset_uuid: str = None):
+    db_session = g.db_session
+    dataset = get_dataset_else_error(db_session, dataset_uuid, collection_uuid, curator_tag)
+    get_collection_else_forbidden(
+        db_session, collection_uuid, visibility=CollectionVisibility.PRIVATE, owner=owner_or_allowed(token_info)
+    )
+    tag = body["curator_tag"]
+    if validate_curator_tag(tag):
+        dataset.update(curator_tag=tag)
```
danieljhegeman, line 28: this could throw an error because of the unique constraint on the combination of `collection_id` and `curator_tag`

danieljhegeman:
@Bento007 Actually I am now (strongly) in favor of designating only `<curator_tag>` from `<curator_tag>.h5ad` as the "curator tag entity" on the backend and elsewhere. That way it will be consistent with how we treat the dataset uuid: the extension is _not_ considered a part of the entity.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/datasets/actions.py
Bento007: good catch

backend/corpora/common/utils/regex.py
```diff
@@ -0,0 +1,7 @@
...
+CURATOR_TAG_FULL_REGEX = f"{CURATOR_TAG_NAME_REGEX}.{EXTENSION_REGEX}"
```
Bento007, line 7:
```suggestion
CURATOR_TAG_FULL_REGEX = f"{CURATOR_TAG_NAME_REGEX}\\.{EXTENSION_REGEX}"
```

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -35,5 +35,76 @@ def test__delete_dataset(self):
...
+                        self.assertEqual(expected_status_code, response.status_code)
+                        self.session.expire_all()
+                        if expected_status_code == 204:
+                            self.assertEqual(dataset.curator_tag, new_tag)
+                        else:
+                            self.assertEqual(dataset.curator_tag, starting_tag)
+
+    def test___invalid_curator_tag(self):
+        collection = self.generate_collection(self.session, visibility=CollectionVisibility.PRIVATE.name)
+        test_tags = ["new.cxg", "no_extention", "uuid"]
```
danieljhegeman, line 80: nit: create a list of tuples for this test of `(<tag>, <extension>)` so that we get the semantics of tag vs extension right

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -35,5 +35,76 @@ def test__delete_dataset(self):
...
+                test_url = f"/curation/v1/collections/{collection.id}/datasets"
+                query_string = {"dataset_uuid": dataset.id}
+                headers = self.get_auth_headers()
+                response = self.app.patch(test_url, headers=headers, query_string=query_string, json=test_body)
+                self.assertEqual(400, response.status_code)
+                self.session.expire_all()
+                self.assertIsNone(dataset.curator_tag)
+
+    def test___conflict_curator_tag(self):
+        test_tag = "tag.h5ad"
```
danieljhegeman, line 94: same issue here, perhaps rename as `test_tag_with_extension`

Bento007: I'm not sure what you're asking me to change here.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/datasets/actions.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@dbconnect
+def patch(token_info: dict, collection_uuid: str, body: dict, curator_tag: str = None, dataset_uuid: str = None):
+    db_session = g.db_session
+    dataset = get_dataset_else_error(db_session, dataset_uuid, collection_uuid, curator_tag)
+    get_collection_else_forbidden(
+        db_session, collection_uuid, visibility=CollectionVisibility.PRIVATE, owner=owner_or_allowed(token_info)
+    )
+    tag = body["curator_tag"]
```
danieljhegeman, line 27: Related to the issue of `tag.h5ad` vs `tag`: what if the curator submits `arbitrary_tag.h5ad` as the tag? In order to overwrite/update that Dataset via the local upload, they would need to upload to `<s3_key_prefix>/arbitrary_tag.h5ad.h5ad`.

danieljhegeman: I think we should, in accordance with general "path" rules, prohibit the `.` character from being contained within the tag.

danieljhegeman (changes requested):
See ongoing concerns about the tag

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
danieljhegeman: the name of the variable. `tag.h5ad` is not just a tag, it is a tag + extension.

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -35,5 +35,85 @@ def test__delete_dataset(self):
                     self.assertEqual(expected_status_code, response.status_code)
 
 
+class TestPatchDataset(BaseAuthAPITest):
+    def test__patch_dataset(self):
+        new_tag = "new_tag.h5ad"
```
danieljhegeman, line 40: same thing, this is a tag + extension

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/datasets/actions.py
Bento007: > Related to the issue of `tag.h5ad` vs `tag`: what if the curator submits `arbitrary_tag.h5ad` as the tag? In order to overwrite/update that Dataset via the local upload, they would need to upload to `<s3_key_prefix>/arbitrary_tag.h5ad.h5ad`.

Actually, they would upload to `<s3_key_prefix>/arbitrary_tag.h5ad` to update the dataset with tag `arbitrary_tag.h5ad`. If they wanted to upload to `<s3_key_prefix>/arbitrary_tag.h5ad.h5ad` they could and it would be valid. Not recommended, but I don't see why we need to include the logic to prevent this.

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
Bento007: The whole thing is the tag.

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
Bento007: curator_tag = a name + extension

Bento007: the tag name is never used on its own, or given to the user without the extension. They are single unit of information.

danieljhegeman: Ok, I think that makes things a little less compatible in a logical sense because the extension is _not_ included as part of the uuid when the curator uses the uuid as the identifier for the s3 upload, i.e. `../<uuid>.h5ad`

But I agree it is totally viable to have curator_tag = name + extension. So we can do that. But it means we need to update the regex variable names above to make it clear that the extension is part of the curator tag. There should just be one sub-regex at the end of `REGEX`, and it should be named appropriately as the curator tag regex, and it should _include_ the extension.

backend/corpora/common/utils/regex.py
danieljhegeman, line 6:
```suggestion
CURATOR_TAG_NAME_REGEX = r"(?P<tag>.*\.h5ad)"
```

danieljhegeman: extension should be part of the curator tag regex. 

I'd prefer we not introduce this concept of the "curator tag _name_", i.e. the tag with the extension removed, unless it is absolutely necessary. Because it just adds to confusion around what constitutes the curator tag entity.

Bento007: I'll change the name.

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
Bento007: > Ok, I think that makes things a little less compatible in a logical sense because the extension is _not_ included as part of the uuid when the curator uses the uuid as the identifier for the s3 upload, i.e. `../<uuid>.h5ad`

they do have to append it with h5ad.

backend/corpora/lambdas/api/v1/curation/collections/collection_uuid/datasets/actions.py
```diff
@@ -0,0 +1,39 @@
...
+
+from backend.corpora.api_server.db import dbconnect
+from backend.corpora.common.corpora_orm import CollectionVisibility
+from backend.corpora.common.entities import Dataset
+from backend.corpora.common.utils.http_exceptions import InvalidParametersHTTPException, ConflictException
+from backend.corpora.common.utils.regex import DATASET_ID_REGEX, CURATOR_TAG_PREFIX_REGEX, EXTENSION_REGEX
+from backend.corpora.lambdas.api.v1.authorization import owner_or_allowed
+from backend.corpora.lambdas.api.v1.common import get_dataset_else_error, get_collection_else_forbidden
+
+REGEX = f"^({DATASET_ID_REGEX}|{CURATOR_TAG_PREFIX_REGEX})\\.{EXTENSION_REGEX}$"
```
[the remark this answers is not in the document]
danieljhegeman: Oh, sorry, I also made a mistake. The function would return `False` if the tag was in UUID format, wouldn't it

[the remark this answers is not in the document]
Bento007: Yes, it will return false if the curator tag is in UUID format.

danieljhegeman (approved):
🚀

PR #2762, opened 2022-06-23
Update issue templates

ainfeld:


ainfeld:
Hmm it seems like the lint is still failing even though I updated it to "chore: Update issue templates". Any idea why?

atarashansky:
There was only one commit so it was failing for weird semantic reasons. I just merged main into this branch so that should fix the issue as another commit has been added.

ainfeld:
Thanks Alec, can I merge now?

Issue #2696, opened 2022-06-13
scExpression add cell count to dot plot


signechambers1:
We got user feedback from @BAevermann users of scExpression need to know when there are only a few cells for a given cell type,  in order to interpret the results correctly and determine how strong of a signal the dot plot is reporting. 

We discussed many different implementations (removing cell types with a certain number of cells, creating visual indicators, etc) and landed on surfacing cell counts to the user because we thought this was the most informative and flexible option. It also provides the user with feedback when they select a filter that reduces the number of cells powering the visualization. This ticket covers the work to add cell counts to the dot plot. These cell counts should change if the user updates the filters.

If the cell count is greater than 10,000, UI should show ">10K."

[Figma](https://www.figma.com/file/htICFl7zihfs1z9eIYVwzw/Gene-Search?node-id=3757%3A640373)

atarashansky:
In the Figma, cell types with zero expression for a particular gene have small dots and their number of cells are reported.

In the current implementation, cell types with zero expression display no data and so the number of cells within them are not passed to the frontend.

There are two possibilities:
1) Only display the # of cells for cell types with non-zero expression
2) Make scExpression report information about cell types with zero expression.

What are your thoughts @signechambers1 ?

atarashansky:
I'm sorry, @signechambers1, I am not sure what conversation you are referring to!

PR #2772, opened 2022-06-24
feat(scExpression): added cell count to scExpression

atarashansky:
- #2696

### Reviewers
**Functional:** 
@hthomas-czi 
@signechambers1 
@seve 

image


seve (commented):
TS looks good to me, could you add a python reviewer?

atarashansky:
@MDunitz can you take a look at the backend changes? Turns out that what I had done previously was incorrect as the `joined` dataframe that was being iterated through was the output of `groupby(...).first()`, so it didn't actually capture the total number of cells.

backend/wmg/api/v1.py
```diff
@@ -128,7 +130,7 @@ def build_expression_summary(query_result: DataFrame) -> dict:
     return structured_result
 
 
-def build_dot_plot_matrix(query_result: DataFrame, cell_counts: DataFrame) -> DataFrame:
+def build_dot_plot_matrix(query_result: DataFrame, cell_counts: DataFrame) -> Tuple[DataFrame, DataFrame]:
```
MDunitz, line 133: Can you update this function name and description to match the added functionality?
or even better refactor to  create a new function that calls `build_dot_plot_matrix` and a separate function calling fetching the cell counts

backend/wmg/api/v1.py
```diff
@@ -142,9 +144,12 @@ def build_dot_plot_matrix(query_result: DataFrame, cell_counts: DataFrame) -> Da
     cell_counts_tissue_agg = cell_counts.groupby(["tissue_ontology_term_id"], as_index=True).sum()
     cell_counts_tissue_agg.rename(columns={"n_total_cells": "n_cells_tissue"}, inplace=True)
 
-    return expr_summary_agg.join(
-        cell_counts_cell_type_agg, on=["tissue_ontology_term_id", "cell_type_ontology_term_id"], how="left"
-    ).join(cell_counts_tissue_agg, on=["tissue_ontology_term_id"], how="left")
+    return (
+        expr_summary_agg.join(
```
MDunitz, line 148: can you please add a unit test specifically testing this logic

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -444,27 +444,27 @@ def test__query_total_cell_count_per_cell_type(self, load_snapshot, ontology_ter
...
                         "cell_type_ontology_term_id": "cell_type_ontology_term_id_0",
-                        "total_count": expected_count,
+                        "total_count": expected_count * 64,
                         "depth": 0,
                     },
                     {
                         "cell_type": "cell_type_ontology_term_id_1_label",
                         "cell_type_ontology_term_id": "cell_type_ontology_term_id_1",
-                        "total_count": expected_count,
+                        "total_count": expected_count * 64,
```
MDunitz, line 467: Please add an explanation for the `* 64`

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -186,6 +186,15 @@ def test__query_request_multi_primary_dims_only__returns_200_and_correct_respons
             )
             assert expected_cell_count_per_cell_type == 729
 
+            # there are 729 possible combinations per tissue-cell type given the above filtering criteria,
```
MDunitz, line 189: thank you!

MDunitz (approved):
lgtm

frontend/src/views/WheresMyGene/components/HeatMap/components/XAxisChart/style.ts
```diff
@@ -17,6 +17,7 @@ export const XAxisMask = styled.div`
   width: ${Y_AXIS_CHART_WIDTH_PX}px;
   position: sticky;
   left: 0;
+  opacity: 0;
```
tihuan, line 20: Ah sorry the opacity is needed here for the sticky header behavior when the user adds enough genes and scrolling starts to happen!

To debug, we can make overflow happen sooner with less added genes by increasing the screen size with `cmd` + `+`

mask opacity

tihuan: Currently the `Cell Count` header is not sticky while the values are. I'm assuming the product requirement is wanting both to be sticky when scrolling?

If so, I'm thinking maybe we can create a new DOM element to display the Cell Count column without relying on echarts, and make sure the new element still covers the gene labels when scrolling happens?

PR #2779, opened 2022-06-26
fix: update API response schema for Datasets

danieljhegeman:
Apparently there exists some data corruption in deployed envs. Allow processing status
entity to be null in response for flexibility. Handle key error.

- #TICKET_NUMBER

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 
@nayib-jose-gloria 
---

## Changes
A number of small tweaks need to be made to make sure the API response schema matches up with the db schema and with the data that actually exist in the deployed environments.

### DIFF
--- a/backend/corpora/common/entities/collection.py
+++ b/backend/corpora/common/entities/collection.py
@@ -174,14 +174,16 @@ class Collection(Entity):
         resp_collections = []
         for collection in session.query(cls.table).filter(*filters).all():
             # Add a Collection-level processing status
-            status = ProcessingStatus.SUCCESS
+            status = None
             for dataset in collection.datasets:
                 processing_status = dataset.processing_status
-                if processing_status.processing_status == ProcessingStatus.PENDING:
-                    status = ProcessingStatus.PENDING
-                elif processing_status.processing_status == ProcessingStatus.FAILURE:
-                    status = ProcessingStatus.FAILURE
-                    break
+                if processing_status:
+                    if processing_status.processing_status == ProcessingStatus.PENDING:
+                        status = ProcessingStatus.PENDING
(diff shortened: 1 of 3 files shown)


backend/corpora/common/entities/collection.py
```diff
@@ -174,14 +174,19 @@ def list_collections_curation(
...
-                    status = ProcessingStatus.FAILURE
-                    break
+                if processing_status:
+                    has_statuses = True
+                    if processing_status.processing_status == ProcessingStatus.PENDING:
+                        status = ProcessingStatus.PENDING
+                    elif processing_status.processing_status == ProcessingStatus.FAILURE:
+                        status = ProcessingStatus.FAILURE
+                        break
+            if has_statuses and not status:  # At least one dataset processing status exists, and all were SUCCESS
```
nayib-jose-gloria, line 188: in the situation where there are no dataset processing statuses, collection-level status would be None. Just want to confirm--is that the expected behavior?

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -34,9 +47,21 @@ def reshape_for_curation_api_and_is_allowed(collection, token_info, uuid_provide
     if "datasets" in collection:
         for dataset in collection["datasets"]:
             if "artifacts" in dataset:
```
nayib-jose-gloria, line 49: looks like we're no longer removing the "artifacts" entry from datasets--not seeing it in the OpenAPI spec, should we remove after this block?

nayib-jose-gloria (approved):

danieljhegeman:
> Do any tests need to be added or updated to verify the new behaviors?

@Bento007 going to add a CXG asset to the test data to make sure it gets filtered out.

Issue #2773, opened 2022-06-24
feat(scExpression): queries should only trigger after a user has finished selecting tissues and genes


atarashansky:
Use case: I tried throwing down 100 genes by clicking through the gene selection menu. Every time I clicked a gene, it triggered a query to the backend. The result is that now the frontend is taking ages to process through all those queries and yield the heatmap. I saw there was debouncing happening somewhere in the code so I expected this to not be an issue, but checking the cloudwatch logs I see hundreds of API calls.

The implementation should be to wait for the popover to close after adding genes before sending a query.

atarashansky:
Will put this on the backlog - this will be relevant when the number of users scale up.

PR #2780, opened 2022-06-27
feat(scExpression): queries should only trigger after user closes quickselect menus

atarashansky:
- #2773 

### Reviewers
**UX** 
@hthomas-czi 
@signechambers1 

WMG query will only be called when user closes or escapes the quickselect menu.


atarashansky:
TODO FIX: adding new tissues / genes with popper open clears previous expressions

atarashansky:
Closing for now after discussion with @hthomas-czi. Will keep the branch alive because I believe this issue will be relevant when more users are active on scExpression. We would like to limit API calls as much as possible.

PR #2782, opened 2022-06-27
chore(scExpression): fixed dropdown button behavior

atarashansky:
### Reviewers
**Functional:** 
@seve 

In current scExpression, clicking on the + button when one of the tissue/gene dropdowns are already open will close and open them again. Instead, they should just close.


frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
```diff
@@ -176,13 +176,14 @@ export default function QuickSelect<
...
   const handleClose = (
-    _: React.ChangeEvent<Record<string, never>>,
+    e: React.ChangeEvent<Record<string, never>>,
     reason: AutocompleteCloseReason
   ) => {
     if (reason === "toggleInput") {
       return;
     }
-    setOpen(false);
+    // @ts-ignore: relatedTarget is a valid property as the event will sometimes be a FocusEvent depending on the action taken.
```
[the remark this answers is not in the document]
atarashansky: I am not sure how to do this. The change handler expects a specific type but it's getting a different type (FocusEvent) depending on the action.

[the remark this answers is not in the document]
atarashansky: I had to set it to `any` type because `e.nativeEvent` is typed as generic `Event` when we need it to be a `FocusEvent`. Casting it to `FocusEvent` doesn't work because there's no sufficient overlap in the interfaces.

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
```diff
@@ -176,13 +176,14 @@ export default function QuickSelect<
...
-    _: React.ChangeEvent<Record<string, never>>,
+    e: React.ChangeEvent<Record<string, never>>,
     reason: AutocompleteCloseReason
   ) => {
     if (reason === "toggleInput") {
       return;
     }
-    setOpen(false);
+    const nativeEvent: any = e.nativeEvent;
+    if (nativeEvent.relatedTarget?.id !== `${dataTestId}-id`) setOpen(false);
```
seve, line 186:
```suggestion
    const { nativeEvent } = e;
    if (
      nativeEvent instanceof FocusEvent &&
      nativeEvent.relatedTarget instanceof Element &&
      nativeEvent.relatedTarget?.id !== `${dataTestId}-id`
    )
      setOpen(false);

```

Try this out

atarashansky: Unfortunately, this breaks the behavior when the target clicked is not an Element.

atarashansky: Also breaks when you try closing with the ESC key since it's a KeyboardEvent.

atarashansky:
```
    if (
      nativeEvent instanceof FocusEvent &&
      nativeEvent.relatedTarget?.id !== `${dataTestId}-id` ||
      !(nativeEvent instanceof FocusEvent)
    )
      setOpen(false);
```
this works!

seve (approved):

Issue #2770, opened 2022-06-24
Updating a private dataset revision does not update the dataset


atolopko-czi:
1. Start a collection revision, having a dataset from h5ad A
2. Update the dataset with a new h5ad, A'
3. Update the dataset with a new h5ad, A'', which differs from A'
Displaying the private dataset revision in Explorer shows data from A', not A''.
It seems the UUID assigned to A' and A'' is the same, which would allow cached data for A' to be shown when A'' data should instead be shown.
If the collection is published, the bug persists, because the UUID originally assigned to A' is used for the dataset artifact which is copied over to the published dataset.

Originally reported by Jason Hilton,  missing development_stage data
Steps: 
1. Start revision to published Collection
2. Upload a dataset to replace the existing one. Observed that development_stage was all 'unknown' (i.e., missing data). Other annotations shown as expected: cell_type, sex shown in LSB w/multi-values and assay, disease, ethnicity, organism, tissue shown single-values in info pane.
3. Generated a new h5ad with development_stage included
4. Updated the dataset once more with the new h5ad. Still seeing development_stage 'unknown' (picture) but @baevermann confirmed that he sees the various development_stage values when he goes to the same link (revised Collection)

Canceling & re-starting revision did _not_ produce the problem again.

[Original Slack thread](https://czi-sci.slack.com/archives/C024HCSH9PT/p1655837065495239) reporting the issue (includes private URLs).

atolopko-czi:
The presumptive fix is to have any dataset updates that occur within the same revision cycle fully replace the previous private dataset record with a new dataset record, assigning a new ID.  A unit test should be added for this case, as well, since there doesn't seem to be a test for this scenario.

atolopko-czi:
curious how this ultimately got resolved? separate story/PR?

PR #2796, opened 2022-06-29
fix(2770): Generate a unique s3_uri for cxg files add to the cellxgene bucket.

Bento007:
- #2770 

### Reviewers

**Readability:** @atolopko-czi 

---

## Changes
- new cxg files are upload to the cxg bucket with a unique UUID. This UUID changes whenever ever the dataset is reuploaded.
- The `s3_uri` and the `explorer_url` have been updated to use their own unique UUID.

## QA steps (optional)
1. Create a collection
2. upload a dataset
3. open the explorer and look at the resulting URL path.
   dataset:  https://t2770-explorer.rdev.single-cell.czi.technology/e/391f6c86-950e-49e4-b51c-02e1aaede204.cxg/
4. publish
5. start a revision
    dataset-R: https://t2770-explorer.rdev.single-cell.czi.technology/e/3ee1fa9f-ae40-4662-adad-c1c041a54be1.cxg/
6. update dataset.
    dataset-R':https://t2770-explorer.rdev.single-cell.czi.technology/e/a3053715-359c-4417-bebc-e765ca2f9c68.cxg/
7. update again
   dataset-R'': https://t2770-explorer.rdev.single-cell.czi.technology/e/cc44db47-ec42-484f-9af2-9bb6f78c6689.cxg/
8. publish
    dataset-R'': https://t2770-explorer.rdev.single-cell.czi.technology/e/391f6c86-950e-49e4-b51c-02e1aaede204.cxg/

## Notes for Reviewer
rdev environment https://t2770-frontend.rdev.single-cell.czi.technology

### DIFF
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -564,7 +564,7 @@ def process_cxg(local_filename, dataset_id, cellxgene_bucket):
         metadata = {
             "explorer_url": join(
                 DEPLOYMENT_STAGE_TO_URL[os.environ["DEPLOYMENT_STAGE"]],
-                asset_id + ".cxg",
+                dataset_id + ".cxg",
                 "",
             )
         }
--- a/backend/corpora/dataset_processing/process.py
+++ b/backend/corpora/dataset_processing/process.py
@@ -481,13 +481,11 @@ def make_cxg(local_filename):
     return cxg_output_container
 
 
-def copy_cxg_files_to_cxg_bucket(cxg_dir, object_key, cellxgene_bucket):
+def copy_cxg_files_to_cxg_bucket(cxg_dir, s3_uri):
     """
     Copy cxg files to the cellxgene bucket (under the given object key) for access by the explorer
-    returns the s3_uri where the cxg is stored
     """
     command = ["aws"]
-    s3_uri = f"s3://{cellxgene_bucket}/{object_key}.cxg/"
     if os.getenv("BOTO_ENDPOINT_URL"):
         command.append(f"--endpoint-url={os.getenv('BOTO_ENDPOINT_URL')}")
 
@@ -506,7 +504,6 @@ def copy_cxg_files_to_cxg_bucket(cxg_dir, object_key, cellxgene_bucket):
         command,
         check=True,
     )
-    return s3_uri
 
 
 def convert_file_ignore_exceptions(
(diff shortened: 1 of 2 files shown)


danieljhegeman:
Ah, this was a missed item from the primary key migration, wasn't it? (for the sake of identifying when the bug was introduced)

Because when the id _didn't_ change, prior to the migration, there was no need to worry about the uuid in the `dataset_artifact.s3_uri` column.

backend/corpora/dataset_processing/process.py
```diff
@@ -548,19 +546,18 @@ def get_bucket_prefix(dataset_id):
...
-        s3_uri = copy_cxg_files_to_cxg_bucket(cxg_dir, bucket_prefix, cellxgene_bucket)
-        metadata = {
-            "explorer_url": join(
-                DEPLOYMENT_STAGE_TO_URL[os.environ["DEPLOYMENT_STAGE"]],
-                dataset_id + ".cxg",
-                "",
-            )
-        }
         with db_session_manager() as session:
+            # generate an s3_uri
```
Bento007, line 552: the URI is unique

backend/corpora/dataset_processing/process.py
```diff
@@ -548,19 +546,18 @@ def get_bucket_prefix(dataset_id):
...
+            while True:
+                cxg_id = generate_uuid()
+                bucket_prefix = get_bucket_prefix(cxg_id)
+                s3_uri = f"s3://{cellxgene_bucket}/{bucket_prefix}.cxg/"
+                # check if the s3_uri is in the asset table
+                filters = [DatasetAsset.table.s3_uri == s3_uri]
+                if session.query(DatasetAsset.table).filter(*filters).first():
+                    break
```
atolopko-czi, line 560: Isn't this searching for a _newly generated_ UUID in the db and then breaking out of the infinite loop only if it exists? Is my understanding of the logic check inverted? Sorry if I'm being dense...

Bento007: you're right, I gotta invert it.

atolopko-czi: 😅 I stared at this for a while doubting myself, so glad I caught something in the end!

If this is trying to handle the case was a UUID might be reused, I'm not sure it's worth the defensive code tbh. Probability is squarely on our side here. And if the db is checking uniqueness, we have that defense as well.

Bento007: the db doesn't check for the uniqueness of the s3_uri so it would pass through

danieljhegeman (approved):
Looks good, but needs a minor change in the logging verbiage, and also should change the variable naming in the `get_bucket_prefix` method on ln 539 since it no longer only handles `dataset_id`s.

backend/corpora/dataset_processing/process.py
Bento007: I modified the code to use the asset id rather than generate a new UUID. This will give us the DB uniqueness check.

Issue #2797, opened 2022-06-29
feat( scExpression): precompute relationships between secondary filters in a graph structure


atarashansky:
## Need
Currently, the runtime bottleneck in the query API is building the filter dimensions. For each attribute, you need to find all of its valid values given the other selected filter values ([this line](https://github.com/chanzuckerberg/single-cell-data-portal/blob/56b0998799fb110fee43cdcbe99c999a2639bb91/backend/wmg/api/v1.py#L96)). Each of these operations is a TileDB [query](https://github.com/chanzuckerberg/single-cell-data-portal/blob/56b0998799fb110fee43cdcbe99c999a2639bb91/backend/wmg/api/v1.py#L83).

## Alternative approach
We should precompute a graph structure. Each node is a value from a filter and nodes are connected if they cover overlapping sets of cells.

So, if a user selects a particular filter (like a dataset), all we need to do is to pull out that node and all of its neighbors from the graph. These neighbors are the set of available values from all the other filter attributes. The nodes in the graph will include both primary and secondary filters.

PR #2798, opened 2022-06-29
feat(scExpression): faster filtering with precomputed graph structure encoding filter relationships

atarashansky:
- #2797

Previously, `build_filter_dims_values` runtime ranges from 1-4 seconds. Now it is around 0.02-0.07 seconds.

- added graph creation to the cube creation pipeline. the graph encodes edges between filter labels if they have overlapping sets of cells. it is saved as a hash table pickle file in the snapshot
-  changed `build_filter_dims_values`. Instead of querying `expression_summary`, it is reading the hash table to pull out related filter values

### DIFF
--- a/backend/wmg/data/wmg_cube.py
+++ b/backend/wmg/data/wmg_cube.py
@@ -282,6 +282,8 @@ def _to_dict(a, b):
     """
     convert a flat key array (a) and a value array (b) into a dictionary with values grouped by keys
     """
+    a = np.array(a)
+    b = np.array(b)
     idx = np.argsort(a)
     a = a[idx]
     b = b[idx]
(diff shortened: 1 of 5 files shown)


backend/wmg/data/wmg_cube.py
```diff
@@ -49,6 +50,27 @@ def create_cell_count_cube(corpus_path: str):
             .size()
         )
         df = df.rename(columns={"size": "n_cells"})
+
+        mat = np.tile(df.columns[:-1].values[None, :], (df.shape[0], 1)) + "__" + df.iloc[:, :-1].values
+        nodes = np.unique(mat.astype("str"))
+
+        mapper = pd.Series(index=nodes, data=np.arange(nodes.size))
```
atarashansky, line 57: Initially thought I would be encoding in an adjacency matrix, so I translated strings into indices. But I ended up saving relationships as a hash table, so you don't need this mapper/reverse mapper anymore. Pls remove.

atarashansky:
[This is the rdev stack](https://atar-faster-filters-frontend.rdev.single-cell.czi.technology/scExpression). Things seem to be working.

TODO: Update the unit tests.

MDunitz (changes requested):
Can you please add tests?

PR #2799, opened 2022-06-30
feat(FindMarkerGenes): update cube generation to save sum of squares

atarashansky:
Added calculation of the sum of squared expression values to the cube generation pipeline. This will be important for being able to efficiently calculate the variance of a population given an arbitrary set of filter values (Variance = Mean_of_squares - Mean^2).

### DIFF
--- a/backend/wmg/data/wmg_cube.py
+++ b/backend/wmg/data/wmg_cube.py
@@ -116,18 +116,19 @@ def load_data_into_cube(tdb_group, uri: str):
     n_groups = len(cube_index)
 
     cube_sum = np.zeros((n_groups, n_genes), dtype=np.float32)
+    cube_sq_sum = np.zeros((n_groups, n_genes), dtype=np.float32)
     cube_nnz = np.zeros((n_groups, n_genes), dtype=np.uint64)
     # TODO: These do not appear to be used. Remove?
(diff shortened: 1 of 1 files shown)


backend/wmg/data/wmg_cube.py
```diff
@@ -175,6 +177,7 @@ def build_in_mem_cube(gene_ids, cube_index, other_attrs, cube_sum, cube_nnz):
         dims[2][idx : idx + n_vals] = organism_ontology_term_id
 
         vals["sum"][idx : idx + n_vals] = cube_sum[cube_idx, mask]
+        vals["sumsq"][idx : idx + n_vals] = cube_sq_sum[cube_idx, mask]
```
atolopko-czi, line 180: Does this actually get added to the cube? I figured we'd need to add a TileDB schema attribute [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/1dd0fac3c57c7d2c385171e75282a6c442824697/backend/wmg/data/schemas/cube_schema.py#L40).  Even if this somehow working without that attribute defined, I would add there. 

In the schema, I would comment how/why this attribute is being used (it's not for WMG iirc?).

A unit test should be added/updated to assert that the attribute is being stored.

atolopko-czi (changes requested):
Code looks correct. A few considerations still.

PR #2802, opened 2022-07-01
refactor: wmg pipeline cube creation

MDunitz:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


backend/corpus_asset_pipelines/summary_cubes/cell_count.py
```diff
@@ -0,0 +1,66 @@
...
+                "organism_ontology_term_id",
+            ],
+            as_index=False,
+        )
+        .size()
+    ).rename(columns={"size": "n_cells"})
+    return df
+
+
+def load(corpus_path: str, df: pd.DataFrame) -> str:
```
ebezzi, line 48: `load` is a pretty bad name for a function that writes stuff to disk, IMHO.

backend/corpus_asset_pipelines/summary_cubes/cell_count.py
```diff
@@ -0,0 +1,66 @@
...
+    tiledb.from_pandas(uri, df, mode="append")
+    return uri
+
+
+@log_func_runtime
+def create_cell_count_cube(corpus_path: str):
+    """
+    Create cell count cube and write to disk
+    """
+    obs = extract(corpus_path)
```
ebezzi, line 63: Not a fan of the extract/transform/load naming convention here. I would rather use names that are more relevant to what each operation is effectively doing (for instance, the name `load` is overloaded and if you're not familiar with the ETL pattern, it will be confusing since it's effectively a `write` operation). @atolopko-czi your thoughts?

atolopko-czi: I've personally disliked the "load" term for the same reason (I always think of it as loading data into memory, for whatever reason). But I figured sticking with ETL terminology would help communicate the pattern to anyone learning the code. If most of us dislike the "load" term (or other terms) I don't feel strongly if we use different terms, as long as the terms are consistent across all the (logical) steps in the pipeline.

MDunitz: I completely agree  re `load` but I do think that there is real value is sticking with the etl pattern

backend/corpus_asset_pipelines/summary_cubes/expression_summary/transform.py
```diff
@@ -52,47 +52,48 @@ def reduce_X(tdb_group: str, cube_indices, *accum):
...
+                    cube_nnz,
                 )
 
-        return accum
-
 
 # TODO: this could be further optimize by parallel chunking.  Might help large arrays if compute ends up being a bottleneck. # noqa E501
 @nb.njit(fastmath=True, error_model="numpy", parallel=False, nogil=True)
-def coo_cube_pass1_into(data, row, col, row_groups, sum_into, nnz_into, min_into, max_into):
+def gene_expression_sum_x_cube_dimension(
```
atolopko-czi, line 71: Bless your heart for this rename!

atolopko-czi (approved):
Latest cleanup changes look great! Really happy to see this refactoring make it into the code base. 👏

backend/corpus_asset_pipelines/summary_cubes/expression_summary/transform.py
MDunitz: loll its still not perfect but I figure its waaaaaay better and someone else can build on it in the future

Issue #2774, opened 2022-06-24
Address failing scExpression performance unit test


metakuni:
To deploy portal to prod on 2022-06-24, I had to temporarily skip the failing scExpression performance test `test_secondary_filters_extreme_case`. I ran it 3 times and it's consistently slower than the expected max response time of 150 seconds by over 100 seconds:
* https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/2557528296/attempts/1
* https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/2557528296/attempts/2
* https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/2557528296/attempts/3

Based on the commit history, this looks like a recently introduced test and one that only runs on prod so wasn't sure if the threshold needs tuning or if this actually indicates a performance regression.

MDunitz:
This does represent a performance regression. We were expecting some degradation due to the introduction additional queries to support the secondary filters. @signechambers1 can help determine if the change significantly harms the user experience/if its  worth investigating performance improvements.

metakuni:
Moving to Q2 cool down

atarashansky:
I'm working on a PR (#2798) to address the performance regression. Will keep you updated!

PR #2804, opened 2022-07-01
fix: avoid unnecessary queries

MDunitz:
- #2774 

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes

- modify secondary filters query to avoid queries when secondary filters not set

### DIFF
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -37,7 +37,9 @@ def query():
 
     include_filter_dims = request.get("include_filter_dims", False)
 
-    response_filter_dims_values = build_filter_dims_values(criteria, query, expression_summary) if include_filter_dims else {}
+    response_filter_dims_values = (
+        build_filter_dims_values(criteria, query, expression_summary) if include_filter_dims else {}
+    )
     return jsonify(
         dict(
             snapshot_id=snapshot.snapshot_identifier,
@@ -94,8 +96,8 @@ def build_filter_dims_values(criteria: WmgQueryCriteria, query: WmgQuery, expres
         "ethnicity_ontology_term_id": "",
     }
     for dim in dims:
-        if len(criteria[dim + "s"]) == 0:
-                dims[dim] = expression_summary.groupby(dim).groups.keys()
+        if len(criteria.dict()[dim + "s"]) == 0:
+            dims[dim] = expression_summary.groupby(dim).groups.keys()
         else:
             dims[dim] = find_dim_option_values(criteria, query, dim)
 
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -37,7 +37,7 @@ def query():
 
     include_filter_dims = request.get("include_filter_dims", False)
 
-    response_filter_dims_values = build_filter_dims_values(criteria, query) if include_filter_dims else {}
+    response_filter_dims_values = build_filter_dims_values(criteria, query, expression_summary) if include_filter_dims else {}
     return jsonify(
         dict(
             snapshot_id=snapshot.snapshot_identifier,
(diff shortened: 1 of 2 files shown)


tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -776,6 +777,65 @@ def test__query_request_with_filter_dims__returns_valid_filter_dims(
                 self.assertEqual(dev_stage_terms_eth_2_dev_2, dev_stage_terms_eth_2_no_dev_filter)
                 self.assertNotEqual(eth_stage_terms_eth_2_dev_2, ethnicity_terms_eth_2_no_dev_filter)
 
+            with self.subTest("Additional queries are not performed when the secondary dimensions are not set"):
```
atolopko-czi, line 780:
```suggestion
            with self.subTest("Additional queries are not performed when the secondary dimension filters are not set"):
```

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -776,6 +777,65 @@ def test__query_request_with_filter_dims__returns_valid_filter_dims(
...
+                with patch("backend.wmg.api.v1.find_dim_option_values") as mock_dims:
+                    mock_dims.side_effect = find_dim_option_values
```
atolopko-czi, line 782: nit: I think you can use `patch("...", wraps=find_dim_option_values)` to achieve this pass-through, which might be 0.5% clearer that we're using this mock only to "spy" on the call count. I haven't used this, but interwebs suggest as much.

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -776,6 +777,65 @@ def test__query_request_with_filter_dims__returns_valid_filter_dims(
...
+                        sex_ontology_term_ids=[],
+                        development_stage_ontology_term_ids=["development_stage_ontology_term_id_0"],
+                        ethnicity_ontology_term_ids=["ethnicity_ontology_term_id_0"],
+                    )
+                    two_secondary_filters_request = dict(
+                        filter=two_secondary_filters,
+                        include_filter_dims=True,
+                    )
+                    self.app.post("/wmg/v1/query", json=two_secondary_filters_request)
+                    self.assertEqual(mock_dims.call_count, 2)
```
atolopko-czi, line 837: nit: would break this up into 3 (?) separate subtests so we can see at structural level what behaviors are being tested; not 100% sure the refactoring will be cleaner tho.

backend/wmg/api/v1.py
```diff
@@ -37,7 +37,9 @@ def query():
 
     include_filter_dims = request.get("include_filter_dims", False)
 
-    response_filter_dims_values = build_filter_dims_values(criteria, query) if include_filter_dims else {}
+    response_filter_dims_values = (
+        build_filter_dims_values(criteria, query, expression_summary) if include_filter_dims else {}
```
atarashansky, line 41: elegant!

atolopko-czi (approved):
Looks right. A few nits. 

Also, do we want to wait on merging thus until we have that larger discussion about how filtering should work? I suspect we may end up with a solution where we only have to do a single TileDB query for both the matrix and filter dim values after each user update of a single-dim filter.

atarashansky (approved):
looks great

atarashansky:
> Looks right. A few nits.
> 
> Also, do we want to wait on merging thus until we have that larger discussion about how filtering should work? I suspect we may end up with a solution where we only have to do a single TileDB query for both the matrix and filter dim values after each user update of a single-dim filter.

I think this PR would improve performance for free in the meantime while we sort out the filtering behavior. I would merge

Issue #1735, opened 2021-12-08
Mirror prod env corpus to lower envs on deploy


atolopko-czi:
When staging and dev deployments are performed, we could mirror the prod corpus data to these envs.  For dev, only a "representative" smaller subset of the prod data would be needed.

* Allows for reliable staging and dev env testing when manual testing is targeted to existing prod datasets.
* E2E tests could rely upon on real prod data

This should result in fewer "surprises" when proceeding through dev->staging->prod promotions.

maniarathi:
[This issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1375) could be accomplished by accomplishing this.

metakuni:
Since this requires a non-trivial amount of work and requires some planning first, deferring this to the next cool down.

MDunitz:
@ebezzi and I worked with @edulop91 to update the bucket permissions to allow cross account copying. It is currently possible to copy from 3 buckets in prod (data-wrangling-prod, corpora-data-prod and dunitzm-tmp) while assuming the single-cell-dev role or from a dev ec2 instance that is assuming the bruce-ec2-poweruser role.

atolopko-czi:
[Plan/Design doc](https://docs.google.com/document/d/1rzI2xasAvXnKSSyds9V8z_WzPCeIopiLcnnSQ4Zmbmg/edit)

atolopko-czi:
* Wrote [docs](https://docs.google.com/document/d/1n5cngEIz-Lqk9737zz3makXGTMrEKT5kN4lsofXPRso/edit#heading=h.29q7ugvfyr5j).
* Added step to staging deploy checklist.

PR #2829, opened 2022-07-07
env mirroring

atolopko-czi:
- #1735 

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


backend/scripts/set_src_dest_envs.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+export SRC_ENV=${1:-prod}
+export DEST_ENV=${2:-dev}
+
+if [[ ! "$SRC_ENV" =~ ^(dev|staging|prod)$ ]]; then
```
MDunitz, line 6: given line 16... is there any situation where we'd want the src env to be dev?

atolopko-czi: Only for testing, so basically no.  I thought about just enumerating the possible allowed combinations to make it clearer. Maybe I'll do that.

MDunitz: I think just removing dev as a src option would work?

atolopko-czi: I thought there might be times when we wanted to mirror from staging to dev. E.g., someone finds a bug in staging and we want to verify it's fixed in dev, and that requires having the same data as in staging. The scenario is a bit of a stretch, but I didn't want to rule it out.

Basically, wanted to allow any src/dest pair that wasn't "upward".

backend/scripts/mirror_rds_data.sh
```diff
@@ -0,0 +1,64 @@
...
+DB_DUMP_FILE=`mktemp`
+
+export DEPLOYMENT_STAGE=$SRC_ENV
+if [[ $SRC_ENV == 'staging' ]]; then
+   export AWS_PROFILE=single-cell-dev
+else
+   export AWS_PROFILE=single-cell-${SRC_ENV}
+fi
+# TODO: move kill ssh tunnel to Makefile db/tunnel command
+kill_ssh_tunnel
```
MDunitz, line 33: What do you think of moving the kill_ssh_tunnel command into its own make command (instead of the db/tunnel command)? Then you could call it here and we could also update the backend readme to recommend calling it after  completing any work with the db as a safety measure

backend/scripts/set_src_dest_envs.sh
MDunitz: wait, wouldnt staging be the src and dev the destination then? so we still wouldnt want to dev as the src?

backend/scripts/mirror_s3_data.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -e
+
+SCRIPTS_DIR=`dirname $0`
+. $SCRIPTS_DIR/set_src_dest_envs.sh
+
+echo Mirroring S3 data from $SRC_ENV to $DEST_ENV
+
+# NOTE: Add --delete if you want to clean up orphaned data. This would make the operation more destructive, of course!
+S3_SYNC_CMD="aws s3 sync --no-progress"
```
MDunitz, line 17: If we arent including `--delete` as a default, how can we ensure that datasets removed from prod are also removed from staging/dev? They'll be orphaned/not available but we will still be storing the data which seems a little dangerous to me.

backend/scripts/mirror_rds_data.sh
MDunitz, line 33: What do you think of moving the kill_ssh_tunnel command into its own make command (instead of the db/tunnel command)? Then you could call it here and we could also update the backend readme to recommend calling it after  completing any work with the db as a safety measure

backend/scripts/mirror_s3_data.sh
atolopko-czi: It came down to data preservation/safety concerns, after I realized there were some unexpected data lying around in the buckets. I figured it's better to err on the side of caution in case some put some "extra" data in the dev or staging buckets for testing purposes. It's probably fine to add it in, particularly since these buckets are versioned. I just didn't want to cause additional problems before seeing that this mirroring was overall working.  I could make it a TODO instead of a NOTE? 😄

backend/scripts/mirror_env_data.sh
```diff
@@ -0,0 +1,17 @@
...
+# to a specified destination deployment environment (dev or staging)
+#
+# THIS IS DESTRUCTIVE for the destination env! The source env will
+# never be modified, but the dest env's data will be replaced.
+
+set -e
+
+SCRIPTS_DIR=`dirname $0`
+. $SCRIPTS_DIR/set_src_dest_envs.sh
+# Note: we run RDS mirror first to avoid possibility of referencing S3 data artifacts that are created after RDS dump
```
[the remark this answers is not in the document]
atolopko-czi: My understanding was that the pipeline records the existence of an artifact in the db only after it's been created (in the src env).

We _will_ have a short period whereby the RDS db is ahead of the S3 artifact data in the _dest_ env, but that seems acceptable.

Reasonable?

backend/Makefile
```diff
@@ -46,25 +46,29 @@ db/new_migration_auto:
...
 	PYTHONPATH=.. alembic -c=./config/database.ini revision --autogenerate --message "$(MESSAGE)"
 
+# interactive mode usage: AWS_PROFILE=single-cell-dev DEPLOYMENT_STAGE=dev make db/connect
+# ARGS usage: AWS_PROFILE=single-cell-dev DEPLOYMENT_STAGE=dev make db/connect ARGS="-c \"select * from dataset_artifact where filetype='CXG'\""
 db/connect:
-	# Assuming you've created a tunnel to the DB. Check the docs for information on how to do that
 	$(eval DB_PW = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/${DEPLOYMENT_STAGE}/database --region us-west-2 | jq -r '.SecretString | match(":([^:]*)@").captures[0].string'))
-	# interactive mode usage: AWS_PROFILE=single-cell-dev DEPLOYMENT_STAGE=dev make db/connect
-	# ARGS usage: AWS_PROFILE=single-cell-dev DEPLOYMENT_STAGE=dev make db/connect ARGS="-c \"select * from dataset_artifact where filetype='CXG'\""
+	$(MAKE) db/tunnel/up
```
MDunitz, line 53: Thank you!!

backend/scripts/set_src_dest_envs.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+export SRC_ENV=${1:-prod}
+export DEST_ENV=${2:-dev}
+
+# allow dev->dev for testing
+if [[ "$SRC_ENV" == dev && "$DEST_ENV" == dev ]]; then
+    if [[ -z $NO_PROMPT ]]; then
+        read -n 1 -p "ATTENTION: You are mirroring the dev env to itself. Useful for testing the mirroring operation, but otherwise useless. Proceed? (Y/n) " ANS
```
MDunitz, line 9: loll

MDunitz (approved):
lgtm

PR #2844, opened 2022-07-11
chore: refactor tests to not require the auth sever

Bento007:
- #TICKET_NUMBER

### Reviewers

---

## Changes
- remove the need for a mock Oauth server on most tests
- mock token validation endpoints
- move TOKEN_EXPIRE to tests.unit.backend.corpora.api_server since it is used by multiple modules
- renamed get_auth_token to get_cxguser_token to better describe it's functionality
- using multiprocess.Process for the mock_auth_server to give us better control regarding when it is cleaned up.
- tests that require the Mock Auth server can subclass AuthServerAPITest all other tests that need credentials mocked can continue to use BaseAuthAPITest
- renamed get_auth_header to make_owner_header. This makes the call for an auth header similar to the call to get a none owner or super-curator header.
- Split test__auth_flow into two tests, because the previous test tested too many things.
- TestSession uses its own test server. This keeps the required code to complete the test smaller and more concise.


tests/unit/backend/corpora/api_server/mock_auth.py
```diff
@@ -98,41 +89,24 @@ def start(self):
...
-                if response.status_code == 200:
-                    self.server_okay = True
-                    break
+                response.raise_for_status()
             except Exception:
-                pass
-
-            # wait one second and try again
-            time.sleep(1)
+                time.sleep(0.1)
```
atarashansky, line 102: Just curious - why was this reduced to 0.1 seconds? Also, what happens if the server isn't ready in 0.5 seconds?

tests/unit/backend/corpora/api_server/base_api_test.py
```diff
@@ -48,7 +53,69 @@ def _remove_timestamps(jrb, removing):
...
+    elif token == "super":
+        return {"sub": "super", "email": "fake_user@email.com", "scope": ["write:collections"]}
+    else:
+        raise Exception()
+
+
+def get_cxguser_token(user="owner"):
+    """
+    Generated an auth token for testing.
+    :param app: a WSGI app.
```
atarashansky, line 70: nit: this line should be updated to `user`

tests/unit/backend/corpora/api_server/base_api_test.py
```diff
@@ -48,7 +53,69 @@ def _remove_timestamps(jrb, removing):
...
+    :return:
+    """
+    cxguser = base64.b64encode(
```
atarashansky, line 73: just curious - why are we encoding in utf8-->b64 and then decoding back to utf8?

atarashansky (approved):
LGTM - just one nit, and a couple questions. Feel free to reply at your leisure.

tests/unit/backend/corpora/api_server/mock_auth.py
Bento007: then we will have flakey tests, but I'm confident we won't have that problem. Basically, it either starts almost immediately or doesn't start at all.

tests/unit/backend/corpora/api_server/base_api_test.py
Bento007:
```suggestion
    :param user: the type of use the token will simulate.
```

tests/unit/backend/corpora/api_server/base_api_test.py
Bento007: Since the data portal API uses a different Auth method than the curation portal, we decode the token as part of the authentication process. Right here to be exact https://github.com/chanzuckerberg/single-cell-data-portal/blob/63154b35242213da2d98a596f999a870f7507a0b/backend/corpora/lambdas/api/v1/authentication.py#L225

PR #2851, opened 2022-07-12
chore: Remove python happy and references to it

alexlokshin-czi:


Bento007 (approved):

.github/workflows/push-rdev.yml
```diff
@@ -21,9 +21,6 @@ env:
 jobs:
   build-push-images:
     runs-on: ubuntu-20.04
-    strategy:
-      matrix:
-        component: ['frontend', 'backend', 'upload_failures', 'processing', 'upload_success', 'dataset_submissions']
```
atolopko-czi, line 26: So it `happy push` builds all images now, it seems, since we're removing this explicit list. Should we remove the image names in [push-tests.yml](https://github.com/chanzuckerberg/single-cell-data-portal/blob/21e2e0b0e61737799993d0b1fe8a81d4a169b838/.github/workflows/push-tests.yml) as well?

PR #2853, opened 2022-07-12
fix(workflow): don't run push-tests smoke tests in happy container 

seve:
This speeds up smoke test runs for push-tests from ~20 minutes to <5.

We're also experiencing issues with our frontend docker container running smoke tests, so this will both unblock merges to main and allow for deployments.

We'll be attempting to address containerized smoke test runs in a separate PR.


danieljhegeman (approved):

Bento007 (approved):
Please create a P0 bug to fix running the smoke test in a container.

PR #2860, opened 2022-07-13
feat: replace AWS RDS with TileDB

roaga:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


backend/corpora/common/entities/tiledb_data.py
```diff
@@ -0,0 +1,408 @@
...
+        a6 = tiledb.Attr(name="contact_email", dtype="U1")
+        a7 = tiledb.Attr(name="links", dtype="U1")
+        a8 = tiledb.Attr(name="datasets", dtype="U1")
+        a9 = tiledb.Attr(name="revision_of", dtype="U1")
+        a10 = tiledb.Attr(name="curator_name", dtype="U1")
+        a11 = tiledb.Attr(name="created_at", dtype=np.float32)
+        a12 = tiledb.Attr(name="updated_at", dtype=np.float32)
+        a13 = tiledb.Attr(name="publisher_metadata", dtype="U1")
+
+        schema = tiledb.ArraySchema(domain=dom, sparse=True, attrs=[a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, a12, a13])
```
ebezzi, line 143: you should not use `sparse=True` in this context

backend/corpora/common/entities/tiledb_data.py
```diff
@@ -0,0 +1,408 @@
...
+        """Get all public collections"""
+        with tiledb.open(self.location + "/collections", mode="r") as A:
+            qc = tiledb.QueryCondition(f"visibility == 'PUBLIC'") # TODO: query conditions don't respect overwrites?
+            q = A.query(attr_cond=qc)
+            res = q.df[:].to_dict("records")
+            for i in res:
+                res[i] = Utils.parse_stored_data(res[i], "collections")
+            return res
+
+    def get_published_collections(self, user_id, from_date, to_date):
```
ebezzi, line 219: Not urgent, but this method and the one above should be merged. Maybe put a TODO

backend/corpora/common/entities/tiledb_data.py
roaga: But don't you need a sparse array to handle string (uuid) dimensions?

backend/corpora/common/entities/tiledb_endpts.py
```diff
@@ -0,0 +1,125 @@
...
+
+def list_published_datasets():
+    """/v1/datasets/index GET"""
+    db = TileDBData(location)
+    datasets = db.get_all_datasets()
+    return make_response(jsonify(datasets), 200)
+
+def get_dataset_metadata(explorer_url: str):
+    """/v1/datasets/meta GET"""
+    # TODO
```
ebezzi, line 103: You can probably skip this for now, it is only used by the explorer

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,167 @@
...
+    def setUp(self):
+        self.metadata = {
+            "name": "test_collection",
+            "description": "This is a test collection.",
+            "owner": "Rohan Agarwal",
+            "contact_name": "Emanuele Bezzi",
+            "contact_email": "ebezzi@chanzuckerberg.com",
+            "links": ["https://nature.com/fake-article", "https://science.com/test-paper"],
+        } # TODO: fill in rest of required metadata
+        self.location = "/Users/ragarwal/code/single-cell-data-portal/tests/unit/backend/fixtures/test_tiledb/metadata"
```
ebezzi, line 15: Replace with a relative path

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,167 @@
...
+        TileDBData.init_db(self.location)
+        self.db = TileDBData(self.location)
+
+    def tearDown(self):
+        TileDBData.destroy_db(self.location)
+
+    def createCollection(self):
```
ebezzi, line 22: replace with snake case (`create_collection`)

backend/corpora/common/entities/tiledb_data.py
```diff
@@ -0,0 +1,416 @@
...
+
+    @staticmethod
+    def pack_input_data(data: dict, array: str) -> dict:
+        for a in Utils.attrs_to_parse[array]:
+            if a in data:
+                data[a] = str(data[a])
+        return data
+
+
+class TileDBData():
```
ebezzi, line 113:
```suggestion
class TileDBData:
```

backend/corpora/common/entities/tiledb_data.py
```diff
@@ -0,0 +1,416 @@
...
+        ]
+    }
+  
+    @staticmethod
+    def new_id():
+        return uuid.uuid4().hex
+
+    # TODO: figure out actual list and dict support in TileDB
+    @staticmethod
+    def parse_stored_data(data: dict, array: str) -> dict:
```
ebezzi, line 99: Why is this method required?

backend/corpora/common/entities/tiledb_data.py
ebezzi: I don't think so?

backend/corpora/common/entities/tiledb_data.py
roaga: I don't know how to use dict/object attributes in TileDB, so I turn them into a string for storage and parse it when needed.

backend/corpora/common/entities/tiledb_data.py
roaga: Docs may be wrong but I'm going off this: https://docs.tiledb.com/main/how-to/arrays/creating-arrays/creating-dimensions

backend/corpora/common/entities/tiledb_data.py
ebezzi: And why do you need to `eval` them?

roaga: To convert the string into whatever data type it should be.

backend/corpora/common/upload.py
```diff
@@ -64,44 +72,39 @@ def upload(
...
-            dataset.reprocess()
+            db.bulk_edit_dataset(dataset_id, Utils.empty_dataset)
 
     else:
         # Add new dataset
-        dataset = Dataset.create(db_session, collection=collection, curator_tag=curator_tag)
+        dataset_id = db.add_dataset(collection_id, Utils.empty_dataset)
 
-    dataset.update(processing_status=dataset.new_processing_status())
+    db.edit_dataset(dataset_id, "processing_status", new_processing_status())
```
ebezzi, line 105: Mostly a nitpick, but I would keep the methods that change the dataset status separate from `edit_dataset`. In fact, it would probably be a good idea to add a check to `edit_dataset` so that only a dataset that belongs to a private collection can be edited.

backend/corpora/dataset_processing/process.py
```diff
@@ -691,16 +668,19 @@ def main():
             logger.error(f"Step function configuration error: Unexpected STEP_NAME '{step_name}'")
 
     except ProcessingCancelled:
-        cancel_dataset(dataset_id)
+        # cancel_dataset(dataset_id)
```
ebezzi, line 671: Is the cancellation deprecated? If not, please leave a TODO since we'll have to keep this feature.

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,273 @@
...
+        id = self.create_collection()
+        coll = self.db.get_collection(id)
+        self.assertEqual(coll['name'], self.metadata['name'])
+        self.assertDictEqual(coll['publisher_metadata'], self.metadata['publisher_metadata'])
+
+    def test_edit_collection(self):
+        # Create collection
+        id = self.create_collection()
+        # Edit collection name
+        self.db.edit_collection(id, "name", "edited_name")
```
ebezzi, line 132: I would move the assertion right next to this line.

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,273 @@
...
+        # Create collection
+        id = self.create_collection()
+        # Add dataset to collection
+        did1 = self.db.add_dataset(id, self.dataset_metadata)
+        did2 = self.db.add_dataset(id, self.dataset_metadata)
+        # Get datasets and delete one
+        self.db.delete_dataset(id, did1)
+        # Get dataset/collection
+        datasets = self.db.get_datasets(id)
+        self.assertEqual(datasets[0]['id'][0].decode("utf-8"), did2)
```
ebezzi, line 174: I would also assert that there are two datasets before the deletion and 1 after.

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,273 @@
...
+        datasets = self.db.get_datasets(id)
+        dataset_ids = self.db.get_attribute(id, "datasets")
+        self.assertEqual(datasets[0]['cell_count'], 12)
+        self.assertEqual(dataset_ids[0], dataset_id)
+
+    def test_delete_dataset_from_collection(self):
+        # Create collection
+        id = self.create_collection()
+        # Add dataset to collection
+        did1 = self.db.add_dataset(id, self.dataset_metadata)
```
ebezzi, line 168: Let's use full variable names here: `dataset_id_to_delete_1` or something like that

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,273 @@
...
+        self.assertEqual(dataset['cell_count'], 12)
+
+    def test_create_revision_on_collection(self):
+        # Create a collection
+        id = self.create_collection()
+        # Create the revision
+        rev_id = self.db.create_revision(id)
+        # Get the revision
+        rev = self.db.get_collection(rev_id)
+        self.assertEqual(rev['name'], "test_collection")
```
ebezzi, line 211: Add more assertions:
1. the revised collection is private
2. the revised collection id is != from the original collection id

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,273 @@
...
+        # Create a collection
+        id = self.create_collection()
+        # Create a revision
+        rev_id = self.db.create_revision(id)
+        self.db.edit_collection(rev_id, "name", "edited_name")
+        # Publish the revision
+        self.db.publish_revision(rev_id)
+        # Get the new collection
+        coll = self.db.get_collection(id)
+        self.assertEqual(coll['name'], "edited_name")
```
ebezzi, line 234: Add more assertions:
1. The published collection is public
2. The published collection id is the same as the original collection

tests/unit/backend/test_tiledb.py
```diff
@@ -0,0 +1,273 @@
...
+        _ = self.db.add_dataset(id1, self.dataset_metadata)
+        id2 = self.create_collection()
+        d2 = self.db.add_dataset(id2, self.dataset_metadata)
+        self.db.publish_collection(id2)
+        datasets = self.db.get_published_datasets()
+        self.assertEqual(datasets[0]['id'], d2)
+
+    def test_check_collection_access(self):
+        id = self.create_collection()
+        self.db.edit_collection(id, "visibility", "PRIVATE")
```
ebezzi, line 265: This is a bit borderline. `edit_collection` should only be able to modify metadata, not an attribute that should only be managed internally. Maybe create a `_edit_collection` method that works on each attribute and expose an `edit_collection` that works on metadata only.

backend/corpora/dataset_processing/process.py
```diff
@@ -426,13 +427,13 @@ def _get_batch_condition():
         "ethnicity": _get_term_pairs("ethnicity"),
         "development_stage": _get_term_pairs("development_stage"),
         "cell_count": adata.shape[0],
-        "mean_genes_per_cell": numerator / denominator,
+        # "mean_genes_per_cell": numerator / denominator,
```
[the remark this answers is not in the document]
roaga: It's not in the OpenAPI schema definition, so I haven't added this attribute yet, but really I don't know what to trust as the true schema.

backend/corpora/dataset_processing/process.py
roaga: I'm thinking that it should be deprecated. Tt looks like all it does is just delete the assets, and I thought we wanted to avoid deleting assets to allow going back in history.

PR #2863, opened 2022-07-14
feat: Single-Cell Team needs help moving one service to Fargate

alexlokshin-czi:
This migrates rdev, dev, staging and prod to FARGATE profile.


.happy/terraform/modules/deletion/variables.tf
```diff
@@ -8,6 +8,12 @@ variable task_role_arn {
   description = "ARN for the role assumed by tasks"
 }
 
+variable execution_role {
+  type        = string
+  description = "Execution role to use for fargate tasks - required for fargate services!"
+  default     = ""
```
jakeyheath, line 14: Should we make this a required parameter? Otherwise, should we make an assert statement that if they choose to use fargate this variable needs to be set? I feel like an empty string might lead to unexpected bugs at apply time.

.happy/terraform/modules/migration/variables.tf
```diff
@@ -8,6 +8,12 @@ variable task_role_arn {
   description = "ARN for the role assumed by tasks"
 }
 
+variable execution_role {
+  type        = string
+  description = "Execution role to use for fargate tasks - required for fargate services!"
+  default     = ""
```
jakeyheath, line 14: see note above

jakeyheath (approved):

ebezzi (approved):

Bento007:
is this ready for review yet?

PR #2868, opened 2022-07-14
fix: smoke tests on deployment

nayib-jose-gloria:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- mirror push test workflow smoke testing while containerized smoke tests are broken

## QA steps (optional)
- smoke tests


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:21.10
```
seve, line 1: Just a heads up that this Dockerfile is for the backend and not the frontend

PR #2883, opened 2022-07-15
fix: reimplement all perf tests in prod

MDunitz:
- #2774 

### Reviewers
**Functional:** 
- @metakuni 
**Readability:** 

---

## Changes

- modify perf tests to run in prod and log run times

## QA steps (optional)
- ran against prod locally. Tests ran within desired time limits

### DIFF
--- a/tests/performance/test_wmg_apis.py
+++ b/tests/performance/test_wmg_apis.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import timeit
 import unittest
 import os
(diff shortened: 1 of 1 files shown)


tests/performance/test_wmg_apis.py
```diff
@@ -11,6 +12,8 @@
 
 # Note that these tests share fixtures and general test paths with the wmg api functional tests
 
+logger = logging.getLogger(__name__)
+
 
 @unittest.skipIf(os.getenv("DEPLOYMENT_STAGE") != "prod", "this test should only run in prod")
```
metakuni, line 18: Is this only running in prod due to potential timing differences caused by differences between the staging and prod infrastructure?

The problem with this is that we don't catch performance regressions until it's too late, i.e. prod deployment day. If the staging infra != prod infra, perhaps we should make them identical and also run this test in staging?

metakuni (approved):
One question posted re: running this test in prod only, otherwise 👍

tests/performance/test_wmg_apis.py
MDunitz: Yea there is a difference in infrastructure and (prior to the env mirror work) the cube in dev/staging was much smaller so the perf would be way better in those envs/not actually useful for determining much about how it would perform in prod

Issue #2878, opened 2022-07-15
scExpression padding between genes and cell type moves


signechambers1:
When you only have one gene there is a lot of padding between gene and cell type label, then when you add additional genes, all the genes move to the left, closer to the cell type label. 

Desired behavior: genes do not move and when you add more genes, genes are added to the right hand side.

atolopko-czi:
@signechambers1 still relevant?

signechambers1:
Still an issue but I'll close, not a high priority.

atarashansky:
Fixed by #3811.

PR #2887, opened 2022-07-16
feat(scExpression): prevent gene columns from compressing/expanding when adding/removing genes

atarashansky:
- #2878 

It seemed impossible to get echarts to stop autoscaling the X axis when adding/removing genes without resorting to incredibly janky tactics. The solution presented in this PR is to give each gene its own echarts instance. Adding new echarts divs (positioned with`relative`, using `flex` layout along the rows) would create new columns with fixed widths. Because each echarts instance is now only 1 column wide, I figured I could remove all the debouncing logic implemented previously, since the charts are no longer getting update dynamically - rather, they are mounting/unmounting. 

@tihuan One drawback with the current implementation is that if we have:
```
Intestine heatmap
------------------
Stomach heatmap
------------------
etc
```
and we remove "Intestine", Stomach gets unmounted and remounted. Do you know why this is happening?


atarashansky:
Currently explodes when removing a tissue or gene that is not the last one (index-wise)

atarashansky:
All tissues unmount and remount when removing a prior tissue -- same with genes. Dot spacing is messed up too.

tihuan:
Thanks so much for the PR, Alec! 

I found some minor regressions in the [Cell Count PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2772/files#r924731676), so maybe we can discuss that first before we continue with this PR, since potentially a new solution for Cell Count could affect this PR too!

For this PR, a few things I'm thinking are:

1. Creating an echart instance per gene could be too heavy on memory consumption, since it adds both an echart object and a new canvas element

3. Existing solution seems to have UI regression of gene labels wrapping when enough genes are added, like so:

    Screen Shot 2022-07-18 at 2 45 17 PM

4. It seems like there's performance regression that now adding genes the screen freezes more frequently. Probably need profiling to see if it's caused by having too many echarts instance or something. Since probably every echart calculation blocks the main thread sequentially, which blocks the browser from updating the screen and handling user events?

Given that the product requirement is for fixing paddings, maybe we can see if we could find another way to make it happen without resulting in significant refactoring of the system?

PR #2899, opened 2022-07-20
feat: get ensembl id from request with only gene name

sienacizdziel:
### Reviewers
**Functional:** 
@MDunitz 
**Readability:** 

---

## Changes
- add functionality to convert gene names to gene ensembl IDs in the gene info api endpoint
- make calls to the gene info endpoint possible with only gene name or only gene ID or both
- update unit tests to include changes

## Notes for Reviewer
Because cxg files do not contain a column for gene IDs, implement functionality from code & ontology files from the single-cell-curation repo to search for a gene ensembl ID given a gene name (see [zenhub issue here](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/322)). This would allow for calls to the API using either gene name or gene ID (or both) to successfully return gene information


backend/gene_info/api/v1.py
```diff
@@ -5,11 +5,20 @@
     NotFoundHTTPException,
     ForbiddenHTTPException,
 )
+from backend.gene_info.api.ensembl_ids import GeneChecker
 
 
-def gene_info(geneID: string):
+def gene_info(gene: string = "", geneID: string = ""):
```
MDunitz, line 11: I would use None as the default and then check for that.
Also since it will ignore the gene name if the gene id is given its probably worth adding a note in case of a mismatch. And adding an explicit test case to show thats the expected behavior

tests/unit/backend/gene_info/test_v1.py
```diff
@@ -121,3 +144,14 @@ def test_parse_xml_tree_missing_values(self):
...
+
+    def test_gene_checker(self):
+        """
+        GeneChecker successfully creates a dictionary of gene names to gene IDs,
+        checks validity of gene names, and returns ensembl ID for a gene name.
+        """
+        gene_checker = ensembl_ids.GeneChecker()
+        self.assertTrue(gene_checker.gene_dict)
+        self.assertTrue(gene_checker.is_valid_label("APOE"))
+        self.assertFalse(gene_checker.is_valid_label("not a gene"))
```
MDunitz, line 156: great test!

backend/config/gene-info-api.yml
```diff
@@ -17,12 +17,13 @@ servers:
 paths:
   /v1/gene_info:
     get:
-      summary: Given a gene ensembl ID, returns that gene's summary information, full name, synonyms, and NCBI URL. 
+      summary: Given a gene ensembl ID and/or gene name, returns that gene's summary information, full name, synonyms, and NCBI URL.
```
MDunitz, line 20: I would specify that if the gene id is given the gene name will be ignored

backend/corpora/api_server/app.py
```diff
@@ -47,6 +48,10 @@ def add_api(base_path, spec_file):
     add_api(base_path="/wmg", spec_file="wmg-api.yml")
     add_api(base_path="/gene_info", spec_file="gene-info-api.yml")
 
+    # initialize gene checker to go ahead and create a dictionary of all
```
MDunitz, line 51: can you also add a comment about why this is being done here (instead of on request) so future engineers arent confused

tests/unit/backend/gene_info/test_v1.py
```diff
@@ -49,20 +49,43 @@ def test_api_calls(self, mock_load_search_result, mock_get):
...
+
+        # parameters contain only gene ID
         res = self.app.get("/gene_info/v1/gene_info?geneID=ensembl1")
         self.assertTrue(mock_fetch_gene_uid.called)
         self.assertTrue(mock_fetch_gene_info_tree.called)
         self.assertTrue(mock_parse_gene_info_tree.called)
         self.assertEqual(res.status_code, requests.codes.ok)
         self.assertEqual(json.loads(res.data), self.final_gene_info_result)
 
+        # parameters contain both gene ID and gene name
```
MDunitz, line 72: re my comment above you could probably just check that mock_get_id wasnt called.
Also I think you'll want to either reset the mocks or pull this into a separate test to avoid incorrect results. I'd recommend creating a separate test

backend/gene_info/api/ensembl_ids.py
```diff
@@ -0,0 +1,73 @@
...
+        :rtype str
+        :return A gene ENSEMBL id
+        """
+
+        # in processing, if there are duplicate gene labels,
+        # a gene might be gene_label + "_" + gene_id
+        if "_" in gene_label:
+            gene_label = gene_label.split("_")[1]
+
+        if not self.is_valid_label(gene_label):
```
MDunitz, line 66: you could probably rewrite this to be 
`if not gene_label in GeneChecker.gene_dict:` and remove the is_valid_label function for simplicity

backend/gene_info/api/ensembl_ids.py
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    gene_dict = {}
+
+    def __init__(self):
+
+        if not GeneChecker.gene_dict:
+            for file in self.GENE_FILES.keys():
+                with gzip.open(self.GENE_FILES[file], "rt") as genes:
+
```
MDunitz, line 30: nit can you just remove some of these extra blank lines?

MDunitz (changes requested):
see inline comments

MDunitz (approved):
lgtm! Nice work

PR #2903, opened 2022-07-20
test(wmg): Fix cell type name count

tihuan:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
1. The cell type name comparison got thrown off a bit with the introduction of cell count because we were grabbing all `text` elements within the y-axis `g` element. Halving the count fixes it!
2. Update to only select one tissue instead of 3 tissues to limit flakiness -- it feels that sometimes the test code will grab cell type counts from different tissues, thus causing false negative
3. Extract reused constants

NOTE:
1. This bug only got discovered in dev smoke test because we are NOT running scExpression smoke test in PRs due to related API being unavailable in the container. Not sure if this is still true today? If so, we'll need to run scExpression tests locally ourselves! 

Thank you!

### DIFF
--- a/frontend/tests/features/wheresMyGene.test.ts
+++ b/frontend/tests/features/wheresMyGene.test.ts
@@ -10,6 +10,8 @@ import { getTestID, getText } from "../utils/selectors";
 
 const GENE_LABELS_ID = "gene-labels";
 const CELL_TYPE_LABELS_ID = "cell-type-labels";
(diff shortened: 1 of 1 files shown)


atarashansky (approved):
Oof, I was unaware WMG smoke tests are not being run on push. Will run locally in the future.

Thanks for fixing this! LGTM

PR #2909, opened 2022-07-20
test(smoketest): Add logs

tihuan:
It seems like adding Playwright play-by-play logs would help with debugging, so proposing to add it for all GHA smoke tests here!

---

## Changes
- add
- remove
- modify


seve:
I'm in favor of having verbose logs off by default, debugging should be done locally

tihuan:
Actually just ran into a failed smoke test in GHA that I can't reproduce locally 😆 Having logs would have been useful

https://github.com/chanzuckerberg/single-cell-data-portal/runs/7438739738?check_suite_focus=true

Screen Shot 2022-07-20 at 3 23 42 PM

seve:
I think the main problem with verbose logs over the past few weeks is that non-FE devs are having trouble even finding what's failing when the logs are so verbose.

I'm having difficulty finding an example, I believe it's in explorer, but sometimes the test results will be sandwiched between the debug logging.

I think this is worth bringing up as a discussion point during an eng meeting.


On the other hand, my points might all be moot if I succeed in implementing a testing summary for our smoke tests on the data portal

tihuan:
Yeah I think having the test result summary would help engineers get the big picture at a glance, but unless the summary also captures exactly which step fails, it can't replace Playwright logs?

Also might be worth looking into replacing Jest with Playwright's own test runner. Jess from GenEpi team did it last quarter because it offers more features and better integration with Playwright. Now I wonder if they even have better test summary too 😆

Added to eng discussion for today! Thanks Seve!

ebezzi (approved):

PR #2915, opened 2022-07-21
test(smoketest): Collection revision prevent race condition

tihuan:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
This PR prevents the race condition when selecting a published collection to test for revision

The solution is just to:

1. Make sure we have at least 4 published collections to select from
2. Randomly pick one collection to start a revision

Before this change, we always pick the first available collection to start the revision, which resulted in a race condition and flaky test failure

CAVEAT: The code assumes we will have at least 4 published collections to work with, but ideally 10 published collections would be safer!


frontend/tests/features/collection/revision.test.ts
```diff
@@ -142,26 +142,45 @@ describeIfDeployed("Collection Revision", () => {
...
 async function startRevision(): Promise<string> {
-  // (thuang): NOTE: the `*` captures the "collection-row" selector
-  const COLLECTION_ROW_SELECTOR_START = `*${getTestID(
-    COLLECTION_ROW_ID
-  )} >> text="Start Revision"`;
-
   await goToPage(TEST_URL + ROUTES.MY_COLLECTIONS);
-  await page.waitForLoadState();
+  // (thuang): Wait for collections to load to prevent race condition
+  await page.waitForLoadState("networkidle");
```
tihuan, line 147: If we don't wait for network idle, Playwright will execute the rest of the code with different amounts of collections loaded on the page each time, which introduces unpredictability and flakiness

frontend/tests/features/collection/revision.test.ts
```diff
@@ -142,26 +142,45 @@ describeIfDeployed("Collection Revision", () => {
...
+
+  const MIN_USABLE_COLLECTION_COUNT = 4;
```
tihuan, line 149: This is arbitrary, originally I was thinking about 5, but generally at least 50% of the seeded published collections need to be available. So if we seed 10 published collections, around 4 to 5 should be available for revision (AKA don't have a revision in progress already)

seve (approved):

seve:
Need to update tests!

seve:
@MillenniumFalconMechanic Do you know if any of the links surrounding the breadcrumbs on the explorer will have to be changed in order to support any of these changes? I know a link is passed via [the config here](https://github.com/chanzuckerberg/single-cell-explorer/blob/1b10f5ed62715ee3ecce1a352765c3a069f4db5d/client/src/globals.ts#L31-L47), but I'm not sure what is actually expected/sent

Issue #2840, opened 2022-07-11
Uploads of datasets via links not working when dataset is still processing


danieljhegeman:
Steps to reproduce:

1) Upload a Dataset to a Collection 
2) While the Dataset is still processing, upload another datafile to the same Dataset id (with `id=<uuid>` in the data of the `PUT` request).
3) Server returns a 405 presumably due to [this](https://github.com/chanzuckerberg/single-cell-data-portal/blob/9af989384aaf660182036996655f817186f5978d/backend/corpora/lambdas/api/v1/collection_id/upload.py#L77)

We need to add the checks (to see if a Dataset with the matching curator tag or id is processing) that are in the lambda used by local file uploads to the code here for link uploads.

PR #2969, opened 2022-07-26
fix(upload-link): Add better error message

Bento007:
- #2840

### Reviewers

---

## Changes
- modify the error message returned when submitting a dataset while the dataset is still processing.

## QA steps (optional)
- Upload a dataset using a link. Then upload to the same dataset again using and upload-link. The response will be a 405 with the new error message.


nayib-jose-gloria (approved):

danieljhegeman (approved):
🚀

nayib-jose-gloria (approved):

danieljhegeman (commented):
Shouldn't this tell them to _either_ cancel via DELETE endpoint _or_ wait until it has finished?

backend/corpora/lambdas/api/v1/collection_id/upload.py
```diff
@@ -75,6 +75,10 @@ def upload_from_link(collection_id: str, token_info: dict, url: str, dataset_id:
     except NonExistentCollectionException:
         raise ForbiddenHTTPException()
     except InvalidProcessingStateException:
-        raise MethodNotAllowedException()
+        raise MethodNotAllowedException(
+            detail="Submission failed. A dataset cannot be updated while a previous update for the same dataset is in "
+            "progress. Please cancel the current submission with "
+            f"DELETE /v1/collections/collection/dataset/{dataset_id}"
```
nayib-jose-gloria, line 81: actually, should we specify the mechanism for canceling? Curators have more user-friendly notebooks they could use instead if they'd like (to more easily set headers and whatnot), or the UI could be available. What do you think of just saying please cancel the current submission?

danieljhegeman (approved):
LGTM

PR #2985, opened 2022-07-27
fix: Make explorer_url and is_primary_data nullable in dataset response

danieljhegeman:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -587,6 +587,7 @@ components:
         all non-canonical (SECONDARY), or contain a mixture (BOTH).
       type: string
       enum: [PRIMARY, SECONDARY, BOTH]
+      nullable: true
     links:
       type: array
       items:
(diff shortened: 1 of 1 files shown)


backend/config/curation-api.yml
```diff
@@ -686,6 +687,7 @@ components:
     explorer_url:
       type: string
       description: The url at which a given Dataset may be explored using the cellxgene visualization tool
+      nullable: true
```
danieljhegeman, line 690: fixing this _before_ we see it cause an error

backend/config/curation-api.yml
```diff
@@ -587,6 +587,7 @@ components:
         all non-canonical (SECONDARY), or contain a mixture (BOTH).
       type: string
       enum: [PRIMARY, SECONDARY, BOTH]
+      nullable: true
```
danieljhegeman, line 590: this was the culprit

Bento007 (approved):

PR #2988, opened 2022-07-28
fix(logging): Log 500 error exceptions

Bento007:
- #TICKET_NUMBER

### Reviewers

**Readability:** 

---

## Changes
- add log message when auth0 management credentials are refreshed
- log the exceptions resulting from an internal server error.


backend/corpora/api_server/app.py
```diff
@@ -164,5 +167,11 @@ def handle_corpora_error(exception):
     )
 
 
+@app.errorhandler(InternalServerError)
+def handle_internal_server_error(exception):
+    app.logger.exception("InternalServerError", exc_info=exception.original_exception)
+    return FlaskApi.get_response(problem(500, "Internal Server Error"))
```
danieljhegeman, line 173: From my reading of the docs, this will replace the original response object with a new response object... right? Are we worried about losing the message detail from the error thrown by the code (which this middleware is catching)?

danieljhegeman (approved):

backend/corpora/api_server/app.py
Bento007: Correct, we hide the details of the server error from the user and log it on your side.

PR #2990, opened 2022-07-28
chore: add cellxgene schema 3.0.0 associated onotology files

nayib-jose-gloria:
- #2994 

### Reviewers
**Functional:** 
@MDunitz 
**Readability:** 

---

## Changes
- update ontology files, as processed by owl file versions specified by cellxgene schema 3.0.0 (to merge into schema 3.0.0 feature branch for now)

## QA steps (optional)
- pipeline tests

## Notes for Reviewer
- @brianraymor [suggested](https://github.com/chanzuckerberg/single-cell-curation/pull/298#issuecomment-1198506771) that we should not be copying ontology files. Should we consider using git submodules here (which would still require a manual step of running git submodule updates to pick up changes) or another solution to reduce the chance of incongruencies across repos?


brianraymor:
@nayib-jose-gloria - the other discussion has been around (over time) ensuring that all special processing of ontology files is centralized in single-cell-curation, since other features have dependencies on properties of ontologies.

ebezzi (approved):

Issue #2980, opened 2022-07-27
POST /v1/auth/token intermittently returns an unexpected 500 status code


brianraymor:
See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1658945805445059). 

The [swagger API](https://api.cellxgene.cziscience.com/curation/ui/#/authentication/backend.corpora.lambdas.api.v1.curation.auth.token.post) does not document a 500 response, but a 500 does get returned on a regular if intermittent basis:

`HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: https://api.cellxgene.cziscience.com/curation/v1/auth/token`

brianraymor:
On Dev, I'm also seeing:

`HTTPError: 502 Server Error: Bad Gateway for url: https://api.cellxgene.dev.single-cell.czi.technology/curation/v1/auth/token`

PR #2999, opened 2022-07-28
fix: Add retry logic to auth0 calls

Bento007:
- #2980

### Reviewers

**Readability:** @ebezzi 

---

## Changes
- add retry logic for calls to auth0


backend/corpora/common/authorizer.py
```diff
@@ -2,13 +2,20 @@
...
+from requests.packages.urllib3.util import Retry
 from jose import jwt
 from jose.exceptions import ExpiredSignatureError, JWTError, JWTClaimsError
 
 from .corpora_config import CorporaAuthConfig
 from backend.corpora.common.utils.http_exceptions import UnauthorizedError
 
 
+auth0_session_with_retry = requests.Session()
+retry_config = Retry(total=3, backoff_factor=1, status_forcelist=[429, 500, 502, 503, 504])
```
danieljhegeman, line 15: why are we specifying `status_forcelist` twice? Both here and in the `retry_config`

danieljhegeman (commented):
One question: it seems like we're specifying the status forcelist twice when we (probably?) don't need to be; could lead to mistakes when adding/removing statuses later

backend/corpora/common/authorizer.py
Bento007: I can consolidate

backend/corpora/common/auth0_manager.py
```diff
@@ -12,12 +15,20 @@ class Auth0ManagementSession:
 
     _session = None
     _domain = None
+    config = CorporaAuthConfig()
```
danieljhegeman, line 18: 👍

danieljhegeman (approved):
Looks good, but random q: do we have any hard evidence that the Auth0 requests were timing out? Or is this just our best guess at what was causing transient failures, and at worst it's simply precautionary?

Bento007:
This is our best guess. I have a separate PR to log 500 errors from the server.

PR #3000, opened 2022-07-28
fix(functional tests): Add retries to overcome transient errors in revisions test

Bento007:
- #TICKET_NUMBER

### Reviewers

**Readability:** @ebezzi 

---

## Changes
- add retry logic in backend functional tests to account for transient error.
- documentation for the method here https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html


tests/functional/backend/common.py
```diff
@@ -28,6 +30,9 @@ def setUpClass(cls):
         super().setUpClass()
         cls.deployment_stage = os.environ["DEPLOYMENT_STAGE"]
         cls.config = CorporaAuthConfig()
+        cls.session = requests.Session()
+        retry_config = Retry(total=5, backoff_factor=2, status_forcelist=[500, 502, 503, 504])
```
danieljhegeman, line 34: Does this mean the retries will only occur for the specified status codes?

Weren't these tests failing due to timeout, and not 500s? I may be lacking context.

danieljhegeman:
Updating branch to incorporate unit test fixes

tests/functional/backend/common.py
Bento007: We were getting 502s and 500s at various times. The 503 and 504 errors are added as extra precautions because of CloudFront.

danieljhegeman (approved):
🔁 🚀

tests/functional/backend/common.py
danieljhegeman: Ok I just want to clarify something even though I approved already:

The `status_forcelist` will force a retry _continuously_ on those listed status codes, up to some large number of retries? And then all other failures, regardless of whether their status code is in the `status_forcelist`, will be retried up to `5` times?

I guess I'm asking if the `status_forcelist` causes the `5` retries to only apply to responses that have those specific status codes...

Bento007: status_forcelist will cause the session to only retry on the errors in the list, up to 5 times.

PR #3003, opened 2022-07-29
fix: Handle Dropbox 200 for head req for nonexistent resources

danieljhegeman:
- unit tests updated
- Dropbox now returns 200 in response to a HEAD request for a resource
that does not exist.

https://czi-sci.slack.com/archives/C023Q1APASK/p1659053375076559

### Reviewers
**Functional:** 
@atolopko-czi @ebezzi 
**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/tests/unit/backend/corpora/api_server/test_v1_collection_upload_link.py
+++ b/tests/unit/backend/corpora/api_server/test_v1_collection_upload_link.py
@@ -67,7 +67,9 @@ class TestCollectionPostUploadLink(BaseAuthAPITest):
             test_url = furl(path=path)
             response = self.app.post(test_url.url, headers=headers, data=json.dumps(body))
             self.assertEqual(400, response.status_code)
-            self.assertEqual("The URL provided causes an error with Dropbox.", json.loads(response.data)["detail"])
+            self.assertEqual(
+                "'content-disposition' not present in the header.", json.loads(response.data)["detail"][-48:]
+            )
 
     @patch(
         "backend.corpora.common.utils.dl_sources.url.DropBoxURL.file_info", return_value={"size": 1, "name": "file.txt"}
--- a/tests/unit/backend/corpora/common/utils/test_dropbox.py
+++ b/tests/unit/backend/corpora/common/utils/test_dropbox.py
@@ -65,18 +65,18 @@ class TestDropbox(unittest.TestCase):
         postive_test.file_info()
 
     def test__get_file_info__neg(self):
-        with self.subTest("404"):
-            negative_test = from_url("https://www.dropbox.com/s/12345678901234/test.h5ad?dl=1")
-            self.assertRaises(HTTPError, negative_test.file_info)
-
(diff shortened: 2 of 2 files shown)


tests/unit/backend/corpora/common/utils/test_dropbox.py
```diff
@@ -65,18 +65,18 @@ def test__get_file_info__OK(self):
         postive_test.file_info()
 
     def test__get_file_info__neg(self):
-        with self.subTest("404"):
+        with self.subTest("Missing Headers"):
             negative_test = from_url("https://www.dropbox.com/s/12345678901234/test.h5ad?dl=1")
```
danieljhegeman, line 69: swap fabricated "missing headers" test for a fabricated 404 (below) in case Dropbox switches back to returning 404.

tests/unit/backend/corpora/api_server/test_v1_collection_upload_link.py
```diff
@@ -67,7 +67,9 @@ def test__bad_link__400(self):
             test_url = furl(path=path)
             response = self.app.post(test_url.url, headers=headers, data=json.dumps(body))
             self.assertEqual(400, response.status_code)
-            self.assertEqual("The URL provided causes an error with Dropbox.", json.loads(response.data)["detail"])
+            self.assertEqual(
+                "'content-disposition' not present in the header.", json.loads(response.data)["detail"][-48:]
+            )
```
danieljhegeman, line 72: this would fail if Dropbox switches back but oh well, we'll cross that bridge when we come to it

atolopko-czi: nit: slightly more robust to use `assertIn()` than `["detail"][-48:]`

atolopko-czi: this test is no longer asserting a bad link, but rather a missing header. do want to _add_ the header to maintain the original intent of the test?

tests/unit/backend/corpora/common/utils/test_dropbox.py
```diff
@@ -65,18 +65,18 @@ def test__get_file_info__OK(self):
...
                 response = requests.Response()
-                response.status_code = 200
+                response.status_code = 404
                 response.headers = {}
                 return response
 
             with mock.patch("requests.head", return_value=make_response()):
                 negative_test = from_url("https://www.dropbox.com/s/12345678901234/test.h5ad?dl=1")
-                self.assertRaises(MissingHeaderException, negative_test.file_info)
+                self.assertRaises(HTTPError, negative_test.file_info)
```
atolopko-czi, line 82: I don't understand the purpose of these tests.  We're testing mocked responses, afaict, so this is just a test of from_url behavior?  this seems disconnected from testing anything dropbox-specific.

ebezzi (approved):

Issue #2978, opened 2022-07-27
GET /v1/collections/{collection_uuid} fails and returns a 500 status


brianraymor:
Reported by @jahilton on [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1658940138740329)

metakuni:
Assigning to @danieljhegeman for triage

brianraymor:
The following public collections are failing:

```

'6e8c5415-302c-492a-a5f9-f29c57ff18fb : 500'
'c9706a92-0e5f-46c1-96d8-20e42467f287 : 500'
'51544e44-293b-4c2b-8c26-560678423380 : 500'
'558385a4-b7b7-4eca-af0c-9e54d010e8dc : 500'
'9132fae8-bdfe-480f-9e45-45bc77f320b3 : 500'
'ae1420fe-6630-46ed-8b3d-cc6056a66467 : 500'
'9b02383a-9358-4f0f-9795-a891ec523bcc : 500'
'24d42e5e-ce6d-45ff-a66b-a3b3b715deaf : 500'
'60358420-6055-411d-ba4f-e8ac80682a2e : 500'
'7edef704-f63a-462c-8636-4bc86a9472bd : 500'
'cdfb9ead-cb58-4a53-879d-5e4ed5329e73 : 500'
'9c8808ce-1138-4dbe-818c-171cff10e650 : 500'
'0b9d8a04-bb9d-44da-aa27-705bb65b54eb : 500'
'a238e9fa-2bdf-41df-8522-69046f99baff : 500'
'db468083-041c-41ca-8f6f-bf991a070adf : 500'
'b52eb423-5d0d-4645-b217-e1c6d38b2e72 : 500'
'5e469121-c203-4775-962d-dcf2e5d6a472 : 500'
```

brianraymor:
Testing the PR in Dev:

https://api.cellxgene.dev.single-cell.czi.technology/curation/v1/collections/6e8c5415-302c-492a-a5f9-f29c57ff18fb

```
{
  "detail": "'male' is not of type 'object'\n\nFailed validating 'type' in schema['allOf'][1]['properties']['datasets']['items']['allOf'][1]['properties']['sex']['items']:\n    {'properties': {'label': {'type': 'string'},\n                    'ontology_term_id': {'type': 'string'}},\n     'type': 'object'}\n\nOn instance['datasets'][2]['sex'][0]:\n    'male'",
[...]

PR #3008, opened 2022-07-29
fix: track active vs tombstoned datasets correctly for API response

danieljhegeman:
- #2978 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- modify algorithm for determining which datasets should be returned. Previous logic had a mistake because it was mutating the list while iterating over it, allowing possibility for datasets to sneak through without being reshaped if tombstoned datasets exist within the collection. cc @brianraymor

### DIFF
--- a/backend/corpora/lambdas/api/v1/curation/collections/common.py
+++ b/backend/corpora/lambdas/api/v1/curation/collections/common.py
@@ -46,7 +46,7 @@ def reshape_for_curation_api_and_is_allowed(collection, token_info, id_provided=
 
     if datasets := collection.get("datasets"):
         active_datasets = []
-        for i, dataset in enumerate(datasets):
+        for dataset in datasets:
             if dataset.get("tombstone"):
                 continue  # Remove tombstoned Datasets
             active_datasets.append(dataset)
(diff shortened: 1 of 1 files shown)


Bento007 (changes requested):
Add a test to verify this works as expected

danieljhegeman:
> Add a test to verify this works as expected

To recreate the exact scenario that was occurring would be extremely tedious, because it was only occurring in certain scenarios with certain arrangements of datasets with tombstones interspersed. 

The difference is in accidentally _skipping_ items in a list before (because I `pop`'d items out of the list); now I'm just doing a normal `for / in` loop.

That is to say, it's not a new feature. It's basically fixing a typo-equivalent bug.

Bento007:
If we had proper test coverage then this would have shown up in the unit tests. It's an edge case we didn't account for in our tests and should be covered to prevent regression. This will make future changes to that code area more stable.

danieljhegeman:
> If we had proper test coverage then this would have shown up in the unit tests. It's an edge case we didn't account for in our tests and should be covered to prevent regression. This will make future changes to that code area more stable.

Ok I'll produce something.

But bear in mind that edge cases are infinite. Testing is never perfect. There are many edge cases equivalent in degree to this one that we're not testing right now, only because we never screwed up the code in a way to reveal them in the first place.

danieljhegeman:
Given the nature of the problem, the extremely narrow picture we'd be testing for, and the fact that we need to unblock curators (the problem isn't silent; it is known immediately if it is occurring because of the API schema check), we should just merge this now and come back and decide on how to add a test for this later, which may require a reworking of the shared test db data.

Bento007 (approved):
Great job investigating and fixing.

Issue #2659, opened 2022-06-07
Add endpoint to verify the s3 upload name.


Bento007:
the goal is to save a user from wasting their time uploading a dataset that will be rejected. The endpoint will look at the path the user is uploading to and check if they have permission do upload there, and if a new dataset will be created or upated.

danieljhegeman:
>The endpoint will look at the path the user is uploading to and check if they have permission do upload there

This is handled by retrieving the `UploadKeyPrefix` from the s3-upload-credentials response.

> and if a new dataset will be created or upated.

consider: this could be done with some pretty simple client-side logic using `GET /collections/{collection_uuid}`.

danieljhegeman:
I suppose this endpoint would provide the logic for distinguishing between uuid and tag, which would be nice. 

Either way, whether we put this logic on the client side or make an endpoint for it, it points to the utility of modular, tested client-side code (client side code still needed to hit this potential endpoint and then interpret / log the results) for this API.

I'm leaning toward (eventually) taking the notebook modules and distributing them via pypi. I think that would be more wieldy than require that curators import from src in the curation repo. Also better than a CLI, which would be a lot more work as well. Thoughts, @Bento007 ?

PR #3019, opened 2022-08-01
feat(curation API): GET Dataset

Bento007:
- #2659

### Reviewers

---

## Changes
- Move Delete dataset. This is to keep the directory consistently organized with the other API handlers.
- Move and rename post_s3_credentials to .datasets.upload_s3.post
- Adding GET /datasets endpoint for curation API. This allows users to retrieve only the datasets from a collection or a single dataset from a collection.
- Add logging of 500 errors from the flask app. This will help us debug exception on the server side.
- Splitting up datasets parsing code from reshape_collection_for_curation_api so it can be reused in the GET datasets endpoint
- change POST to GET for s3_upload_credentials since nothing is being stored in the database.
-


backend/config/curation-api.yml
```diff
@@ -356,7 +356,37 @@ paths:
           $ref: "#/components/responses/401"
         "404":
           $ref: "#/components/responses/404"
-
+  /v1/checks/submission_key:
```
danieljhegeman, line 359: I think I like `/v1/verify/` more than `/v1/checks/`

What do you think?

backend/config/curation-api.yml
```diff
@@ -356,7 +356,37 @@ paths:
...
+    put:
+      summary: Verify the result of submitting a dataset
```
danieljhegeman, line 361: > Verify

☝️ 👍

danieljhegeman:
```suggestion
      summary: Verify the result of submitting a dataset: 'create' vs 'update'
```

backend/config/curation-api.yml
```diff
@@ -356,7 +356,37 @@ paths:
...
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  result:
+                    type: string
+                    enum: ["UPDATE", "NEW", "FAIL"]
```
danieljhegeman, line 384: when would it `FAIL`? 🤔

danieljhegeman: Oh, if it doesn't have `.h5ad`...

backend/corpora/common/submission.py
```diff
@@ -0,0 +1,115 @@
...
+    COLLECTION_ID_REGEX,
+    DATASET_ID_REGEX,
+    CURATOR_TAG_PREFIX_REGEX,
+    EXTENSION_REGEX,
+)
+from backend.corpora.common.upload import upload
+
+logger = logging.getLogger(__name__)
+
+REGEX = f"^{USERNAME_REGEX}/{COLLECTION_ID_REGEX}/({DATASET_ID_REGEX}|{CURATOR_TAG_PREFIX_REGEX})\\.{EXTENSION_REGEX}$"
```
danieljhegeman, line 21: I think we should put this in a common file, have a single source, given that it is already [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/9af989384aaf660182036996655f817186f5978d/backend/corpora/dataset_submissions/app.py#L22)

backend/config/curation-api.yml
danieljhegeman: Ah, or if it's to the wrong user prefix, or `/super/` when they don't have permission... right...

danieljhegeman (changes requested):
I appreciate the work you have put in to modularize / share code between this endpoint and the actual submission & processing portions of code!

High-level comment: my original understanding of this ticket, and what I think it still should be, is to validate only the _identifier_ + _collection_id_ to determine whether an upload will produce a new dataset or update an existing one. I agree that it technically would be useful to validate the entire s3 upload key, but recall that we are already providing the s3 key prefix to them programmatically as part of the response from `POST /s3-upload-credentials` which they have to do before uploading. So the only thing they need to verify is the collection_id + identifier combo.

That would simplify things substantially and would make the design of the API a bit more parsimonious in my opinion.

backend/corpora/common/utils/corpora_constants.py
```diff
@@ -14,5 +14,6 @@ class CorporaConstants(object):
         "default_embedding",
         "X_approximate_distribution",
     ]
-    SUPER_CURATOR_NAME = "super|curator"
+    SUPER_CURATOR_NAME = "super-curator"
     SUPER_CURATOR_SCOPE = "write:collections"
+    SUPER_CURATER_SUBMISSON_PATH = "super"
```
nayib-jose-gloria, line 19: typo: CURATER -> CURATOR

backend/config/curation-api.yml
Bento007: I considered using verify. However, I am avoiding the use of verbs in the URL path as is the REST way.

backend/config/curation-api.yml
Bento007:
```suggestion
      summary: Quickly check the result of submitting a dataset.
```

backend/config/curation-api.yml
Bento007: I skipped checking if the user is the owner or a super. This check is better handled by our S3 credentials and bucket permissions. I assume if the user knows to use super in the title then they must already know if they have it or not. We only check that the use in the key is the owner of the collection. If the user us super then we don't check if they are the owner. 

Also notice that this endpoint does not require Auth. I'm still trying to decide if this is the best course. I figured I'd leave it up to the reviewers and PM.

danieljhegeman: I don't think the user should be passing in the entire s3 key. They should only pass in the collection_id and the identifier, whether tag or dataset_id.

backend/config/curation-api.yml
danieljhegeman:
```suggestion
      summary: Determine whether a Dataset submission under the given `collection_id` and identifier (`curator_tag` or `dataset_id`) will produce a new Dataset or update an existing Dataset.
```

This, assuming that we change the query to take a `collection_id` and a `curator_tag` or `dataset_id`.

danieljhegeman:
@Bento007 let me know your thinking on my general comment above about configuring this to take a `collection_id` + `dataset_id`/`curator_tag` rather than a `s3_object_key`.

nayib-jose-gloria:
From my understanding, this endpoint is effectively meant to be a dry run for dataset uploads, and its main value is in quickly confirming that their inputs to an upload function will result in the output they expect, or in pointing out unexpected behavior quicker than a real upload would so that they can correct faster. 

Per the curation API spec, users are expected to upload either via the upload-link endpoint (which has inputs: path param collection_id, body param identifier, and body param link to file) or through a boto3 upload function directly to our s3 bucket locally. If we are asking users to use a python library for local uploads anyway, I think it makes sense to instead point them to our own curation package + notebooks for local uploads. Our provided local upload function inputs very closely match the upload-links endpoint inputs (collection_id, identifier, and filepath). Neither asks for the full submission key. 

To that end, I think it makes sense to reshape this endpoint to match the input pattern of collection_id, identifier, and filepath (albeit filepath/link might only be useful for verifying whether h5ad is appended at the end) and be presented as a dry run for upload by either mechanism. That way, there is no confusion stemming from different expected inputs from the dry-run verification vs. a real upload. If we do go that route, we should also rename the endpoint path to reflect that its a dataset upload dry-run (maybe v1/checks/upload or v1/dry-runs/upload or something similar)

Open to disagreement with this, of course. let me know if I'm misrepresenting the user needs/use case

backend/config/curation-api.yml
```diff
@@ -356,7 +356,46 @@ paths:
           $ref: "#/components/responses/401"
         "404":
           $ref: "#/components/responses/404"
-
+  /v1/collections/{collection_id}/datasets/s3-upload-credentials/submission_key:
+    put:
```
danieljhegeman, line 360: I don't think this should be a `PUT`. I think it should be a `GET`.

This call does not modify any resource. It requests information about a resource.

Bento007: Since there is a request body it needed to be a `PUT`, but a `GET` does make more sense. The parameters will need to query parameters in this case. Not a big problem, but the tag or dataset_id will need to be made URL friendly.

backend/config/curation-api.yml
```diff
@@ -272,6 +272,36 @@ paths:
...
+      description: >-
+        Return one or more datasets from a collection. If a dataset ID or a curator tag is provided as a query parameter
+        then only that dataset will be returned.
+      operationId: backend.corpora.lambdas.api.v1.curation.collections.collection_id.datasets.actions.get
+      tags:
+        - collection
+      parameters:
+        - $ref: "#/components/parameters/path_collection_id"
+        - $ref: "#/components/parameters/query_curator_tag"
+        - $ref: "#/components/parameters/query_dataset_id"
```
danieljhegeman, line 286: Would now be an okay time to move to putting the identifier in the url path?
```
/v1/collections/{collection_id}/datasets/[rest/of/path/is/tag/or/id]
```

It is the most conventional approach for REST

danieljhegeman: Although obviously it would apply to the other endpoints as well... so perhaps better put in a different PR

backend/config/curation-api.yml
```diff
@@ -394,7 +424,7 @@ paths:
 
       security:
         - curatorAccess: []
-      operationId: backend.corpora.lambdas.api.v1.curation.collections.collection_id.dataset.post_s3_credentials
+      operationId: backend.corpora.lambdas.api.v1.curation.collections.collection_id.datasets.upload_s3.post
```
danieljhegeman, line 427: 👍

danieljhegeman: btw this endpoint should really be a GET.

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -45,31 +47,38 @@ def reshape_for_curation_api_and_is_allowed(collection, token_info, id_provided=
...
+    for ontology_element in DATASET_ONTOLOGY_ELEMENTS:
+        if dataset_ontology_element := dataset.get(ontology_element):
+            if not isinstance(dataset_ontology_element, list):
+                # Package in array
+                dataset[ontology_element] = [dataset_ontology_element]
+        else:
+            dataset[ontology_element] = []
+    return dataset
+
+
```
danieljhegeman, line 81: 👍

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -114,5 +114,53 @@ def test___conflict_curator_tag(self):
...
+    def test_get_multiple_datasets_in_a_collection_ignore_tombstones_200(self):
+        collection = self.generate_collection(self.session, visibility=CollectionVisibility.PRIVATE.name)
+        self.generate_dataset(self.session, collection=collection)
+        self.generate_dataset(self.session, collection=collection, tombstone=True)
+        test_url = f"/curation/v1/collections/{collection.id}/datasets"
+        response = self.app.get(test_url)
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(1, len(response.json["datasets"]))
+
+    def test_get_fake_dataset_404(self):
```
danieljhegeman, line 148:
```suggestion
    def test_get_nonexistent_dataset_404(self):
```
"fake" implies "stub" to me, esp in context of tests

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -114,5 +114,53 @@ def test___conflict_curator_tag(self):
...
+        test_url = f"/curation/v1/collections/{collection.id}/datasets"
+
+        test_query_strings = [{"dataset_id": dataset.id}, {"curator_tag": dataset.curator_tag}]
+        for query_string in test_query_strings:
+            with self.subTest(query_string):
+                response = self.app.get(test_url, query_string=query_string)
+                self.assertEqual(200, response.status_code)
+                self.assertEqual(dataset.id, response.json["id"])
+
+    def test_get_datasets_fake_collection(self):
```
danieljhegeman:
```suggestion
    def test_get_datasets_nonexistent_collection(self):
```

danieljhegeman:
@Bento007 I thought we agreed that we would turn this into a `dataset`, not `datasets`, endpoint (https://czi-sci.slack.com/archives/C02UGUV5R3J/p1659643024820679) ("Regarding adding the GET _dataset_ endpoint..."). 

I don't think a `/datasets` endpoint is useful because that is essentially what `GET /collections/{collection_id}` is already doing.

A `GET /collection/{collection_id}/datasets/[tag/or/id]` endpoint would work for the desired (pre-s3 upload check) purpose, and help fulfill the API along REST conventions (i.e., distinct resource-oriented), because the curator would just hit this endpoint with the tag they intend to use to see if the dataset exists already or not.

backend/config/curation-api.yml
danieljhegeman: I made the above comments thinking the endpoint was `/dataset` and not `/datasets`. See my comment on the main thread of this PR.

backend/config/curation-api.yml
```diff
@@ -272,6 +272,28 @@ paths:
           $ref: "#/components/responses/403"
         "404":
           $ref: "#/components/responses/404"
+    get:
+      summary: Return a datasets from a collection
```
danieljhegeman, line 276:
```suggestion
      summary: Fetch a dataset from a collection
```

backend/config/curation-api.yml
```diff
@@ -272,6 +272,28 @@ paths:
...
+      description: >-
+        Return a from a collection. If a dataset ID or a curator tag is provided as a query parameter
```
danieljhegeman, line 278:
```suggestion
        Fetch a Dataset from a collection. If a `dataset_id` or a `curator_tag` is provided as a query parameter
```

backend/config/curation-api.yml
```diff
@@ -272,6 +272,28 @@ paths:
...
+        the associated dataset metadata is returned.
```
danieljhegeman, line 279:
```suggestion
        the associated Dataset metadata is returned.
```

danieljhegeman (approved):

PR #3025, opened 2022-08-02
fix(deployment): Run tests in parallel during upgrade

Bento007:
- #TICKET_NUMBER

### Reviewers
**Functional:** @ebezzi 

**Readability:** @seve 

---

## Changes
- modify the upgrade Github actions to run the tests in parallel after deployment has finished.


danieljhegeman:
These three test steps all run against the same deployed environment. Previously we didn't have to worry about making sure tests didn't interfere with each other. Do we know that the tests use complete separate data surfaces, and won't conflict with each other? I'm not sure they were designed with that in mind...

But I wouldn't be opposed to trying this and going with it if it seems to produce ✅ runs regularly. I like speed-ups.

ebezzi (approved):

Bento007:
@danieljhegeman if they aren't then they should be. I know the functional tests are independent. I think the smoke test are independent from what I've read in the code. The performance test should be.

Issue #2982, opened 2022-07-27
GET /v1/collections/{collection_id} does not return processing_status for the collection


brianraymor:
The [Swagger API](https://api.cellxgene.cziscience.com/curation/ui/#/collection/backend.corpora.lambdas.api.v1.curation.collections.collection_id.actions.get) documents `processing_status` but it's not present in this [response](/https://api.cellxgene.cziscience.com/curation/v1/collections/968834a0-1895-40df-8720-666029b3bbac) for example.

PR #3026, opened 2022-08-03
fix(curation): GET collection_id returns the processing_status

Bento007:
- #2982

### Reviewers

---

## Changes
- add function `add_collection_level_processing_status` to move the dataset processing status up to the collection level. This was removed from `list_collections_curation` to make it reusable.
- Move `add_collection_level_processing_status` and `list_collections_curation`  out of the Collection class. The entity class and all of its hellspawn are a horrible plague on our code base that should have never been created.


danieljhegeman (changes requested):
@Bento007 Add a simple test to make sure it's coming through on the `GET /collections/{collection_id}` endpoint?

EDIT: nvm, I see that's being addressed with the one-line change in the existing test. 🚀

danieljhegeman (approved):
Good refactor of existing code to use for this endpoint

PR #3027, opened 2022-08-03
fix(curation_api): timestamps are JaSON friendly

Bento007:
- #TICKET_NUMBER

### Reviewers

---

## Changes
- modify how date-time stamps are returned from the APIs. They are now returned in an ISO human-readable way.
- update curation api spec to expect string for time values
- Use curationJSONEncoder in the curation api


danieljhegeman (approved):
I thought I already reviewed this a day ago. Looks good! 🚀

Issue #3010, opened 2022-07-29
PATCH /collections/{collection_uuid} descriptions and examples do not conform to the schema


brianraymor:
**August 3**

RE links: capturing design discussion from [curation-api](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1659561250170139?thread_ts=1659476056.513829&cid=C02UGUV5R3J).

**MVP [now with BAD REQUEST}**
1. Links are in the request. Completely replaces the existing links in the collection. This must be clearly documented. If Links are [], then return BAD REQUEST.
2. Links are not in the request. The existing links in the collection are not changed.

--- 

See the [Swagger API](https://api.cellxgene.cziscience.com/curation/ui/#/collection/backend.corpora.lambdas.api.v1.curation.collections.collection_id.actions.patch). See also [editorial: update_collection.ipynb instructions are misleading](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/3009). 

For the record, I prefer an implementation that allows curators to only pass the fields that need to be changed since this is a PATCH of an existing collection that has already contains the required fields.

The description states:

_Update a private Collection or a Revision. **If any Collection metadata fields are included in the body of the request, these fields SHALL be updated for the specified Collection**._

This is reinforced by the examples:

1. Update a description

```
{
  "description": "New Collection description"
}
```

2. Update a name

```
{
  "name": "New Collection name"
}

```

But the schema indicates that the following fields are required:

```
collection_form_metadata {
contact_email*	string email of contact person for the collection

contact_name*	string name of the primary person of contact for the collection

description*	string description of the collection

links	links[...]

name*	string name of the collection
}
```

brianraymor:
@danieljhegeman - I connected the PR. Is this issue assigned to you?

danieljhegeman:
Ah, I see that my 'mention' of this issue in the PR is not the same as actually _connecting_ the issue to the PR. Noted. 📝

brianraymor:
@danieljhegeman - Please ensure that the swagger api is also updated.

brianraymor:
Failed testing because existing links are deleted during a partial update:

In the current implementation:
1. A collection has links.
1. Only its description is updated with PATCH.
1. A collection no longer has links, but does have a new description.

PR #3028, opened 2022-08-03
fix: keep links when updating Collections via Curation API

danieljhegeman:
- add optional parameter to shared update_collection function
- unit test coverage for empty payload to `PATCH /collections/{collection_id}`

- #3010 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
cc @brianraymor 
For `PATCH /collections/{collection_id}`:

- Links management:
  - reject request that has empty array `[]` for `links` -- `400`
  - replace all existing links with all new links if links are provided
  - keep all existing links if new links are _not_ provided
- Response object is the _full_ schema for [collection_form_metadata](https://github.com/chanzuckerberg/single-cell-data-portal/blob/2e5ddb0244b254fd83c8c840a90e69232b169277/backend/config/curation-api.yml#L181-L192)
- Partial data updates are accepted without modifying fields which are not provided
- update/add unit tests
- [update Swagger API description](https://github.com/chanzuckerberg/single-cell-data-portal/blob/2e5ddb0244b254fd83c8c840a90e69232b169277/backend/config/curation-api.yml#L153-L158)


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -264,7 +264,7 @@ def delete_collection(collection_id: str, token_info: dict):
 
 
 @dbconnect
-def update_collection(collection_id: str, body: dict, token_info: dict):
+def update_collection(collection_id: str, body: dict, token_info: dict, keep_links: bool = False):
```
Bento007, line 267: Since this is used directly by the portal API we shouldn't put parameters that are not in the portal API. Break this function into its own common function that is used by both curation and portal.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -408,17 +411,39 @@ def test__update_collection__OK(self):
...
-            f"/curation/v1/collections/{collection_id}",
-            data=json.dumps(metadata),
-            headers=self.make_owner_header(),
-        )
-        self.assertEqual(200, response.status_code)
-        response = self.app.get(f"curation/v1/collections/{collection_id}")
-        self.assertEqual(response.json["name"], metadata["name"])
+    @patch("backend.corpora.common.entities.Collection.get_doi")
+    # Patch Collection.get_doi to ensure the additional collection.update call is made in the update_collection method
+    def test__update_collection_partial_data__OK(self, get_doi: Mock):
```
Bento007, line 416: Also add a test where we try to update some links. I have a feeling there is some unexpected behavior around that.

Bento007 (changes requested):

backend/corpora/lambdas/api/v1/curation/collections/collection_id/actions.py
```diff
@@ -36,4 +40,12 @@ def get(collection_id: str, token_info: dict):
...
 def patch(collection_id: str, body: dict, token_info: dict):
-    return update_collection(collection_id, body, token_info)
+    if "links" in body:
+        if not body["links"]:
+            raise InvalidParametersHTTPException(detail="If provided, the 'links' array may not be empty")
+        keep_links = False  # links have been provided; replace all old links
+    else:
+        keep_links = True  # links have NOT been provided; keep existing links
+    collection = update_collection_common(collection_id, body, token_info, keep_links=keep_links)
+    columns_to_return = ("name", "description", "contact_name", "contact_email", "links")
```
danieljhegeman, line 50: @Bento007 just out of curiosity, is there a way to dynamically inspect the Response fields of the API definition here? Tried looking up an answer to this and didn't have much luck.

Not necessarily but could be nice, would be a bit cleaner and more maintainable.

Bento007: Not an easy way that I know of. You'd need to dig deep into the connexions code, or parse the swagger yourself. I think parsing the swagger would be the easy option, but I wouldn't recommend it. Personally I'd like to move away from connexion, and use something like fast-api, or [flask + marshmallow](https://flask-marshmallow.readthedocs.io/en/latest/).

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -408,17 +411,83 @@ def test__update_collection__OK(self):
...
         response = self.app.get(f"curation/v1/collections/{collection_id}")
-        self.assertEqual(response.json["name"], metadata["name"])
+        self.assertEqual(response.json["name"], new_name)
+        self.assertEqual(response.json["description"], description)
+        self.assertEqual(response.json["contact_name"], contact_name)
+        self.assertEqual(response.json["contact_email"], contact_email)
+        self.assertEqual(response.json["links"], links)
+
+    @patch("backend.corpora.common.entities.Collection.get_doi")
+    # Patch Collection.get_doi to ensure the additional collection.update call is made in the update_collection method
```
Bento007, line 451: We test this in the last test. Do we need to test it again in this test?

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -408,17 +411,83 @@ def test__update_collection__OK(self):
...
+        name = "partial updates test collection"
+
+        links_configurations = (
+            ("With links already in place; new links replace old", links, new_links, 200, new_links),
+            ("With no links in place; new links get added", None, new_links, 200, new_links),
+            ("With links in place, but empty request; links persist", links, None, 200, links),
+            ("With links in place, empty array passed; BAD REQUEST 400", links, [], 400, links),
+        )
+
+        for links_config in links_configurations:
```
Bento007, line 471: unpack the link_config here rather than bellow.

Bento007 (changes requested):

backend/corpora/lambdas/api/v1/curation/collections/collection_id/actions.py
Bento007: There is even a [flask+pydantic](https://pypi.org/project/flask-pydantic-spec/)

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: I think it's better to have it whenever links are being tested, for the reason given in the comment: ensures that all possible `update` calls (where data can be mutated, esp links) are made.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: Yikes, great point. will change. 🤦‍♂️

Bento007 (approved):

Issue #2694, opened 2022-06-13
scExpression Reduce padding between cell type names and dot plot


signechambers1:
1. Go to https://cellxgene.cziscience.com/scExpression
2. Add Lymph node as a tissue
3. Add some genes

Notice the cell type names are preemptively cut off - there is more room for them to show before hitting the dot plot. this is not true for all cell types (CD141 positive myeloid dendritic cell shows up correctly but others do not - see screenshots)

![Screen Shot 2022-06-13 at 12.34.20 PM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/7ca566ad-d579-417e-991c-e153ea0283d8)
![Screen Shot 2022-06-13 at 12.36.35 PM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/51f13ce5-7820-4f53-9d48-de55aa25fbdc)

ashin-czi:
Just curious while I'm working on this, would it be preferred if the whole cell name was displayed?

signechambers1:
@ashin-czi i think we want to maximize dot plot space while holding enough room for the cell type name - some of the cell type names are very long so we haven't found it feasible to display the entire cell type name in all circumstances.

ashin-czi:
Hi all, I've made some changes but wanted to do a quick run by you guys before I commit anything. Here are some examples:
![image.png](https://images.zenhubusercontent.com/62dee801e47e5625ea12d27f/e0a687d2-5853-402b-b0c6-5b85d16fb509)
![image.png](https://images.zenhubusercontent.com/62dee801e47e5625ea12d27f/96ae55ff-c4d4-4185-bc85-96df27efb3ff)

You can see that the text cut off is much more uniform now. I've played around with some tissues and it seems to look good, but wanted to see if this was along the lines of what was being asked for this ticket.

atarashansky:
This looks great! @signechambers1 can correct me if I'm wrong, but this seems to address the problem. Feel free to go ahead and open a PR.

signechambers1:
looks great to me! thanks andrew!

PR #3038, opened 2022-08-04
fix: uniform truncation of cell type names

ashin-czi:
- No longer relying on echarts truncate logic for cell type names
- Manually truncating cell type names based on width of text

Echarts truncate logic for axis labels caused uneven trimming of text depending on the content. This fix truncates text based on width in pixels instead of character count.

- #2694

### Reviewers
**Functional:** @seve @atarashansky 

**Readability:** @seve @atarashansky 

---

## Changes
- add
    - Function that takes in text and returns the truncated text trimmed to the given width in pixels

- remove
    - Removed echarts truncate line for axis labels: `["overflow" as string]: "truncate"`

- modify


frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -268,14 +297,18 @@ export function createYAxisOptions({
 
             const displayDepth = Math.min(depth, MAX_DEPTH);
 
-            const paddedName = " ".repeat(displayDepth * 8) + name;
+            const paddedName = getFixedWidth(
+              " ".repeat(displayDepth * 8) + name,
+              200
```
atarashansky, line 302: How is this number determined? If it's related to `Y_AXIS_CHART_WIDTH_PX` (at the top of this file), could you make it scale with that variable?

atarashansky (changes requested):

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
ashin-czi: The 200 number was just what fit the best for the current wrapper/container width. But you're right, it would be better to have it scale to that constant in case we want to make the Y-axis wider

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -224,6 +224,35 @@ interface CreateYAxisOptionsProps {
...
+  text: string,
+  maxLength: number,
+  font = "12px sans-serif"
+): string {
+  CTX!.font = font;
+  for (let i = 0; i < text.length; i++) {
+    const substring = text.substring(0, i);
+    const textMetrics = CTX!.measureText(substring);
+    if (textMetrics!.width >= maxLength) {
+      return substring + "...";
```
seve: @signechambers1 If we're creating a custom truncation solution for cell type names. Should we follow the explorer and use middle truncation?

ashin-czi:
Now scaling with different widths: 

image
image
image

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
signechambers1: yes! that would be great for consistency

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -224,6 +224,36 @@ interface CreateYAxisOptionsProps {
...
+  maxLength: number,
+  font = "12px sans-serif"
+): string {
+  CTX!.font = font;
+  const ellipsisWidth = CTX!.measureText("...").width;
+  for (let i = 0; i < text.length; i++) {
+    const substring = text.substring(0, i + 1);
+    const textMetrics = CTX!.measureText(substring);
+    if (textMetrics.width >= maxLength - ellipsisWidth) {
+      return text.substring(0, i) + "...";
```
atarashansky, line 251: Can you update to do middle truncation (i.e. ellipses in the center of the word)?

atarashansky (changes requested):

ashin-czi:
Just pushed some changes for the middle truncation logic, see below for an example of how it will look: 
image

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -224,6 +224,81 @@ interface CreateYAxisOptionsProps {
...
+ */
+const CTX = document.createElement("canvas").getContext("2d");
+
+/**
+ * Formats and truncates the cell type name to a given width
+ *
+ * @param name The text to truncate
+ * @param maxWidth The max width in pixels the string should be
+ * @param font The font family and font size as a string. Ex. "12px sans-serif"
+ * @returns The string fixed to a certain pixel width
```
atarashansky, line 238: nit: add a comment about `displayDepth`

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -268,14 +343,20 @@ export function createYAxisOptions({
 
             const displayDepth = Math.min(depth, MAX_DEPTH);
 
-            const paddedName = " ".repeat(displayDepth * 8) + name;
+            const paddedName = formatCellLabel(
+              name,
+              Y_AXIS_CHART_WIDTH_PX - 75, // scale based on y-axis width
+              "bold 12px sans-serif", // prevents selected style from overlapping count
```
atarashansky, line 349: nit: Could we avoid hardcoding this by using `SELECTED_STYLE` defined in line 152? It doesn't specify the font size, but maybe it should. I leave this up to you.

atarashansky (approved):
A couple nits, but looks good! Approving as is. Thanks, Andrew!

PR #3041, opened 2022-08-05
fix: API - reinstate the (default) commit when DOI is removed

danieljhegeman:
- #TICKET_NUMBER

### Reviewers
**Functional:** 
@ebezzi 
**Readability:** 

---

## Changes
- Remove `commit=False`, which was added by the previous commit to `main`, so that the old doi removal is committed if the new doi is not found, regardless of whether or not the rest of the updates in the body end up throwing errors.


danieljhegeman:
The Curation API usage of the `update_collection_common` function is broken, but I wanted to reinstate the original function of this method -- this was the only change made -- to how it was working for the Corpora API previously.

I had thought it made sense to not immediately commit, but that's not necessarily the case, and I should have erred on the side of not changing the functionality of anything Corpora API-related.

PR #3043, opened 2022-08-05
fix(smoke tests): Create Test Collections for Smoke Tests if missing

nayib-jose-gloria:
- #3053 

### Reviewers
**Functional:** 
@metakuni 
**Readability:** 

---

## Changes
- created script that will check if the test account has the min needed amount of published collections for smoke testing (4). If less than the min, will create a max of 10 collections (amount recommended by [Timmy](https://github.com/chanzuckerberg/single-cell-data-portal/pull/2915) to reduce flakiness for the associated test), upload a small example dataset via dropbox link for each, and publish. 

## QA steps (optional)
- Ran script locally for both cases where collections are created and collections are sufficient

## Notes for Reviewer
- Creating collections adds about 10-15 min to dev/stage smoke test runs (largely from waiting for the test dataset to upload, which is required to publish). Will only occur after mirroring or if the tests account's collections have been deleted in some other scenario. 
- API doesn't seem to support filtering collections by user ID. I am using contact name as a proxy and manually filtering after receiving the results, but very open to better ideas about how to check for existing public collections owned by a particular account via the backend API if possible
- Any error handling we should add?
- Could I be doing the parallelization better without overcomplicating? 
- I'm assuming we can handle a max of 10 simultaneous sets of calls, but any concern about the request load? Should I check the dataset status less frequently (10 sec is the current interval for functional tests, but its only running 1 thread)?


scripts/smoke_tests/requirements.txt
```diff
@@ -0,0 +1,3 @@
+requests
+tenacity
+boto3
```
maniarathi, line 3: Nit: please alphabetize python dependencies. It improves readability.

scripts/smoke_tests/setup.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def get_collection_count(self):
+        res = self.session.get(f"{self.api}/curation/v1/collections?visiblity=PUBLIC", headers=self.headers)
+        res.raise_for_status()
+        data = json.loads(res.content)
+        num_collections = 0
+        for collection in data['collections']:
+            if collection["contact_name"] == "Smoke Test User":
+                num_collections += 1
+            if num_collections >= MIN_TEST_COLLECTIONS:
```
maniarathi, line 32: num_collections will never be > than MIN_TEST_COLLECTIONS right? Because you're incrementing by 1 and as soon as you hit MIN_TEST_COLLECTIONS, then it returns so the max it can ever be is MIN_TEST_COLLECTIONS. 

If my logic is correct, I would simplify to `if num_collections == MIN_TEST_COLLECTIONS:`

scripts/smoke_tests/setup.py
```diff
@@ -0,0 +1,109 @@
...
+import json
+import time
+import threading
+
+import requests
+
+from tests.functional.backend.common import BaseFunctionalTestCase
+
+# min amount of published collections expected by smoke tests
+MIN_TEST_COLLECTIONS = 4
```
maniarathi, line 12: Question: why have a MIN_TEST_COLLECTIONS instead of always have 10?

scripts/smoke_tests/setup.py
```diff
@@ -0,0 +1,109 @@
...
+# amount advised by Timmy Huang to reduce flakiness
```
maniarathi, line 13: Do you know why 10 helps reduce flakiness? If so, maybe add a little more insight here?

maniarathi (commented):
You'll have to pardon me -- I am going through a handful of PRs and since this one showed up in my inbox, I took a look (especially because I was interested in learning more about smoke tests) and had a few comments! They are non-blocking, though I would very much appreciate if you alphabetized the dependencies :)

scripts/smoke_tests/setup.py
nayib-jose-gloria: Yeah, I think I'll change this. Initially, I was thinking about this distinction as a time-saving procedure--as in, if we already have >= the minimum amount needed to pass the test, it might not be worthwhile to add to the deployment time to get to the Max every time (vs. with less than the min, we already have to take the time to trigger at least 1 create/upload/publish workflow, so we may as well kick off enough in parallel to reach the max). But if we're already asserting confidence that 10 will significantly reduce flakiness over 4, it probably saves time long-term to make sure we always have at least 10.

ebezzi (approved):

scripts/smoke_tests/setup.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def publish_collection(self, collection_id):
+        body = {"data_submission_policy_version": "1.0"}
+        res = self.session.post(
+            f"{self.api}/dp/v1/collections/{collection_id}/publish", headers=self.headers, data=json.dumps(body)
+        )
+        res.raise_for_status()
+
+    # override
+    def upload_and_wait(self, collection_id, dropbox_url):
```
ebezzi, line 63: Out of curiosity, why do we need to override?

nayib-jose-gloria: The parent class includes a cleanup 'DELETE {api_url}/v1/datasets' after the dataset is finished uploading

Bento007 (approved):
Ideally, all of this would be written directly into the smoke test, but this is a great solution in the interim.

How hard would it be to create an additional script that runs at the end of the smoke test to clean up excess test collections? As it stands now, the smoke-tests become flakey after a while due to a build-up of test old test collections.

scripts/smoke_tests/setup.py
ebezzi: Makes sense. Maybe we can add a `delete=True` parameter and pass False here? Only if it's not too complicated though.

nayib-jose-gloria:
> Ideally, all of this would be written directly into the smoke test, but this is a great solution in the interim.
> 
> How hard would it be to create an additional script that runs at the end of the smoke test to clean up excess test collections? As it stands now, the smoke-tests become flakey after a while due to a build-up of test old test collections.

Not that hard, though if we're going to be mirroring prod into dev/staging regularly and resetting the test collections through that, would it still be an issue?

Issue #2810, opened 2022-03-02
Should a public collection include a reference to its revision for authorized users?


brianraymor:
The private revisions of public collections will include a `revision_of` attribute that is the UUID of their public collection so a curator can map from a revision to the original collection. And perhaps this should be named `revision_of_collection_id`. 

Should there be a symmetrical attribute in the public collection such as `revised_in_collection_id` that is the UUID of the private revision?

There is a security issue because this attribute MUST only be returned in responses to authorized users which is a different model from how public collections will be accessed.

Another proposal from @jahilton is:
> boolean if a revision is in progress or not?

danieljhegeman:
I plan on adding a 'view' to the db, constructed by the equivalent of 
```create view project_revising_in as select * from project as p left join (select id as revising_in, revision_of as id from project) t using(id);```

danieljhegeman:
I rebased my code changes on [3019](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3019)

brianraymor:
@Bento007 @danieljhegeman - let's discuss [https://github.com/chanzuckerberg/single-cell-data-portal/pull/3045#discussion_r941701575] here.

Per the top-level summary comment:

**There are still security concerns. revising_in MUST only be present in the response if the curator is authorized to access the revision. Otherwise, the portal would be leaking the bearer token.**

Bento007:
I don’t think we need both the None response and the Not authorized response. One or the other would be sufficient. We should keep the None response and remove the Not Authorized response and eliminated the need for an extra code path.

brianraymor:
Clarifying:

So - instead of filtering out the `revisioning_in` if I'm not authored to access the revision in question, then value is set to "None"? 

Example:

Brian creates a revision.
Brian is returned the UUID.

Jason is a super curator.
Jason is returned the UUID because he's authorized.

Trent is a curator or consumer.
Trent is returned "None" because he's not authorized.

Bento007:
@brianraymor yes. That is how I see it.

brianraymor:
I have no objections to this alternative. It matches the _house style_. It will still need careful testing similar to an implementation that would _filter out_ the field for unauthorized users to ensure that information is not leaking.

danieljhegeman:
@Bento007 @brianraymor that is not quite how the current code works.

At present, here is the scheme:

[...]

PR #3045, opened 2022-08-06
feat: add `revising_in` attribute

danieljhegeman:
- #2810

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- Add `revising_in` attribute to responses to `GET /collections` and `GET /collections/{collection_uuid}`. 
  - `None` if collection is private or revision does not exist
  - `"NOT AUTHORIZED"` if collection is public and user is not authorized
  - `"<revision_id>"` if collection is public, user is authorized, and revision exists

## QA steps (optional)
- in dev, `revising_in` field for a collection:
  - with auth
    - no revision: returned `None`
    - after revision created: returned revision id
  - without auth
    - returned `"NOT AUTHORIZED"` regardless of whether revision existed


backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -14,6 +14,7 @@
     Base,
     ProcessingStatus,
 )
+from ......common.entities import Collection
```
Bento007, line 17: use explicit path name.

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -221,3 +207,47 @@ class EntityColumns:
         DbDatasetArtifact: dataset_asset_cols,
         DbDatasetProcessingStatus: dataset_processing_status_cols,
     }
+
+
+def list_collections_curation(
```
Bento007, line 212: Why the move? Any actual changes?

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -143,6 +143,25 @@ def test__get_collections_with_auth__OK(self):
         res_auth = self.app.get("/curation/v1/collections", headers=self.make_owner_header())
         self.assertEqual(200, res_auth.status_code)
         self.assertEqual(6, len(res_auth.json["collections"]))
+        with self.subTest("The 'revising_in' attribute is \"NOT AUTHORIZED\" for unauthorized public collections"):
+            for c in res_auth.json["collections"]:
+                if c["id"] in (
+                    "test_collection_id_public_for_revision_one",
+                    "test_collection_id_public_for_revision_two",
+                ):
+                    self.assertEqual("NOT AUTHORIZED", c["revising_in"])
```
Bento007, line 152: couldn't we just leave this field out if they are not authorized? Do they need to know that there is a revision?

backend/corpora/lambdas/api/v1/curation/collections/common.py
danieljhegeman: Shoot. This would have been an accident. Stupid merge conflicts. Let me see if I can revert this portion so that the commit history is accurate.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: It's set to `NOT_AUTHORIZED` for all public collections for which you are not allowed to view, regardless of whether there actually is an existing revision.  cc @brianraymor for comment

brianraymor: If the design or requirements are unclear, I'd prefer to discuss in the issue and not the PR to capture outcomes. I'm going to respond in https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2810.

Bento007:
I don’t think we need both the None response and the Not authorized response. One or the other would be sufficient. We should keep the None response and remove the Not Authorized response and eliminated the need for an extra code path.

backend/corpora/lambdas/api/v1/curation/collections/common.py
danieljhegeman: Oh, ok, even though this got moved, it still retains you as the author. So we're good. No actual changes.

backend/config/curation-api.yml
```diff
@@ -560,6 +560,9 @@ components:
         revision_of:
           type: string
           nullable: true
+        revising_in:
```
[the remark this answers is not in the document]
danieljhegeman: done!

Bento007 (approved):

Issue #2944, opened 2022-07-25
During collection publication, the direct personal identifiers information is clipped


brianraymor:
1. Publish a collection
2. When the "Are you sure?" modal appears, select direct personal identifiers.

Its text is clipped at the top. 

![Screen Shot 2022-07-25 at 10.48.10 AM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/f82450c5-bb41-4077-b7a6-07a2e4b03985)

PR #3060, opened 2022-08-08
fix(collection): 2944 Publish Collection PII cutoff

tihuan:
- #2944 

### Reviewers
**Functional:** 
@seve 

**Readability:** 
@hthomas-czi 
---

## Changes
Update PII popover height to be 50vh + scroll overflow, so it works with different zoom levels and viewport heights

## QA steps (optional)
1. Invoke the publish collection modal and hover over the PII target
2. Play with different viewport heights and zoom levels, the content should no longer be cutoff

demo

### DIFF
--- a/frontend/src/components/Collections/components/PublishCollection/components/Policy/components/DirectPersonalIdentifiers/index.tsx
+++ b/frontend/src/components/Collections/components/PublishCollection/components/Policy/components/DirectPersonalIdentifiers/index.tsx
@@ -4,7 +4,7 @@ import {
   PopoverPosition,
 } from "@blueprintjs/core";
 import { IconNames } from "@blueprintjs/icons";
-import { FC } from "react";
+import { FC, useEffect, useRef } from "react";
 import { GRAY } from "src/components/common/theme";
 import {
   BulletWrapper,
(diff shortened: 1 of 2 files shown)


seve (approved):

seve:
Could we possibly force the scroll bar to show, since it's not intuitive that it's scrollable?

seve:
https://stackoverflow.com/a/7855592

tihuan:
Thanks Seve! Ooh great question! Scrollbar is pretty interesting because different browsers have different scrollbar styles, and thus sometimes product/design don’t want scrollbar to show up by default, which also matches browser’s default behavior

@hthomas-czi do you have any preference in this case?

Thanks all!

hthomas-czi:
@tihuan @seve 

Forcing the scrollbar to appear makes sense to me.

hthomas-czi (approved):

tihuan:
Hey @seve @hthomas-czi ! I did a little more research on the scrollbar behavior, turned out that it's a MacOS thing to autohide scrollbars and Mac users can customize their OS in settings here:

Screen Shot 2022-08-10 at 2 07 40 PM

Interestingly Firefox solves this problem by temporarily showing the scrollbar and then fading away, so users get the hint that the content is scrollable. I felt that's a pretty good middle ground between MacOS user's own OS preference vs. our web app affordance, so I applied the same trick in our code to do the same for all browsers. What do you think?

demo

Issue #2972, opened 2022-07-26
dataset preview schema in GET /v1/collections does not match the response


brianraymor:
See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1658874554748979?thread_ts=1658868624.213879&cid=C02UGUV5R3J).

The dataset preview schema in [GET /v1/collections](https://api.cellxgene.cziscience.com/curation/ui/#/collection/backend.corpora.lambdas.api.v1.curation.collections.actions.get) documents that a subset of the full dataset metadata is returned in the response; however, the response actually includes the full dataset metadata with empty values for fields not included in the preview (cell_type, development_stage, ethnicity, sex, etc.) which may lead developers to believe that metadata is missing:

```
     {
        "assay": [
          {
            "label": "CEL-seq2",
            "ontology_term_id": "EFO:0010010"
          }
        ],
        "cell_type": [],
        "curator_tag": null,
        "development_stage": [],
        "disease": [
          {
            "label": "normal",
            "ontology_term_id": "PATO:0000461"
          }
        ],
        "ethnicity": [],
        ...
        "organism": [
          {
            "label": "Homo sapiens",
            "ontology_term_id": "NCBITaxon:9606"
          }
        ],
        "sex": [],
        "tissue": [
          {
            "label": "pancreas",
            "ontology_term_id": "UBERON:0001264"
          }
```

This is not a case where null values are appearing per [@atolopko-czi 's earlier guidance](https://czi-sci.slack.com/archives/C023Q1APASK/p1654867613679049). These are fields that were never intended to be included in the response. They must be removed.

PR #3063, opened 2022-08-08
fix(GET Collections): Return correct dataset preview

Bento007:
- #2972 

### Reviewers

**Readability:** 

---

## Changes
- Add boolean to create the dataset preview when listing collections. This makes the response match the expected format
- Add tests that verify the correct fields are returned


backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -28,15 +28,17 @@
...
     Reshape Collection data for the Curation API response. Remove tombstoned Datasets.
     :param collection: the Collection being returned in the API response
     :param token_info: user access token
     :param id_provided: bool - whether or not the collection uuid was provided by the user, for access purposes
+    :param preview: boool - whether the dataset is in preview form or not.
     :return: whether or not the Collection should be included in the response per ownership/access rules
     """
-    owner = collection["owner"]
+
+    owner = collection.pop("owner")  # Don't actually want to return 'owner' in response
```
danieljhegeman, line 41: thank you 🙈

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -72,13 +72,14 @@ def reshape_dataset_for_curation_api(dataset: dict) -> dict:
...
-        else:
-            dataset[ontology_element] = []
+    if not preview:
+        for ontology_element in DATASET_ONTOLOGY_ELEMENTS:
+            if dataset_ontology_element := dataset.get(ontology_element):
+                if not isinstance(dataset_ontology_element, list):
+                    # Package in array
+                    dataset[ontology_element] = [dataset_ontology_element]
+            else:
+                dataset[ontology_element] = []
```
danieljhegeman, line 82: This isn't going to work. There are [four elements returned in the dataset preview](https://github.com/chanzuckerberg/single-cell-data-portal/blob/5fa765b0a030a04ba28f9901e1e946bc85234bf5/backend/config/curation-api.yml#L678-L685) which need to have the reshaping logic applied to them.

Best to add a `DATASET_PREVIEW_ONTOLOGY_ELEMENTS` constant list and use it accordingly

danieljhegeman (changes requested):
See comment. Need to apply reshape logic to the 4 ontology elements that are returned as part of dataset preview.

danieljhegeman (approved):
🚀

Issue #2793, opened 2022-06-29
Add query param flag for retrieving collections by curator name and visibility in  `GET /collections` index endpoint


danieljhegeman:
**August 8**: There was agreement to remove this issue from the curation API MVP. It's not required for the 3.0.0 migration and can be considered in parallel with the symmetrical [curation UX request.](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2638).

---

**Update**: See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1659040871197119). `access_type` MUST also be removed from responses and the Swagger API. 

---

see https://github.com/chanzuckerberg/single-cell-data-portal/issues/2083#issuecomment-1163794647

Add a query param `MyCollections` flag (or similar) that causes `GET /collections` to return only Collections which are owned by the curator making the call.

@brianraymor @jahilton should visibility be specified? Or should the usage of this flag default to returning all of a curator's Collections, both public and private? _\*\*edit\*\*: see comment below_

danieljhegeman:
@brianraymor I'm seeing a comment on the other issue from you said 

> Use GET /collections?MyCollections which is familiar to the curators from the UX - which returns nothing without an access token and is specifically for curators since all other consumers only have "READ" access.

I'm assuming this means then that we should include both public _and_ private Collections in the response (since it's mimicking the UI endpoint that curators are familiar with, which displays both).

brianraymor:
> I'm assuming this means then that we should include both public and private Collections in the response (since it's mimicking the UI endpoint that curators are familiar with, which displays both).

That is correct. 

To be more precise, it should return all public collections and all authorized private collections.

danieljhegeman:
> To be more precise, it should return all public collections and all authorized private collections.

Wait, this is different than what is returned when visiting `My Collections` on the UX: the UX shows public collections but _only the public collections which belong to the user_ (and of course it shows the private collections as well).

brianraymor:
Doh. Writing too fast.

Yes - all collections private or public that I'm authorized to see. Lead curators see everything. Other curators see only their collections.

danieljhegeman:
I'm still unclear on whether that means a non-super curator should get back...

 _all **public** Collections_
 
 OR

  _just their own **public** Collections_.

brianraymor:
[...]

PR #3064, opened 2022-08-09
fix(GET Collections): Add filter by owner

Bento007:
- #2793 

### Reviewers

**Readability:** 

---

## Changes
- Move list_collections_curation, because it's only used in one place.
- Add query parameter `owned` to get collections to allow filtering collections by the owner. The access token is used to determine the owner.
- Add a test case to only retrieve the collections owned by a user.

## Notes for Reviewer
- The test leaves something to be desired. Since the owner is not in the collection data returned there is no definitive way to show the collection belongs to the user. Instead, I retrieve all the collections owned by a user and check that they have WRITE permission. In the same test, we retrieve all of the collections owned by the super curator, which should be 0 since the super-curator doesn't own any collection because none were created by the.

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -64,6 +64,14 @@ paths:
         - collection
       parameters:
         - $ref: "#/components/parameters/query_visibility"
+        - name: owned
+          description: >-
+            Only show collections owned by the curator.
+          in: query
+          required: false
+          schema:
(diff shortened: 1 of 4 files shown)


danieljhegeman (approved):
This looks good. @brianraymor how do we feel about `owner` as the boolean query param flag?

Issue #3047, opened 2022-08-07
Improve error messages for 403's


danieljhegeman:
Attempts to update or delete public collections return a standard 403 message with no useful detail. Let's add verbiage along the lines of "Cannot delete a public collection via Curator API" and "Updating a public collection is not allow; you must create a revision."

I suggest simply adding an `error_detail` parameter to [this function](https://github.com/chanzuckerberg/single-cell-data-portal/blob/9af989384aaf660182036996655f817186f5978d/backend/corpora/lambdas/api/v1/common.py#L14), where it can then be passed as `detail=<error detail>` to `ForbiddenHTTPException` constructor.

danieljhegeman:
I think this _may_ be ameliorated by my updates to the notebooks per [my comments here](https://github.com/chanzuckerberg/single-cell-data-portal/issues/3042).

danieljhegeman:
In fact, yes: you should see the error [we return here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/5fa765b0a030a04ba28f9901e1e946bc85234bf5/backend/corpora/lambdas/api/v1/curation/collections/collection_id/actions.py#L34) now with the latest notebooks on my development branch in the curation repo

danieljhegeman:
Made edits to give a good message for 403 in response to `PATCH /collections/{collection_id}`

PR #3065, opened 2022-08-09
fix: add detail to error when attempting to update public collection

danieljhegeman:
- #3047 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add logic to indicate when a update has been attempted on a collection that the curator must first create a revision for before updating
- add meaningful message to 403 error detail


backend/corpora/lambdas/api/v1/curation/collections/collection_id/actions.py
```diff
@@ -60,7 +61,14 @@ def patch(collection_id: str, body: dict, token_info: dict) -> Response:
             raise InvalidParametersHTTPException(detail="If provided, the 'links' array may not be empty")
         keep_links = False  # links have been provided; replace all old links
 
-    collection, errors = get_collection_and_verify_body(db_session, collection_id, body, token_info)
+    try:
+        collection, errors = get_collection_and_verify_body(db_session, collection_id, body, token_info)
+    except ForbiddenHTTPException as err:
+        if Collection.get_collection(
+            db_session, collection_id, CollectionVisibility.PUBLIC, owner=owner_or_allowed(token_info)
```
Bento007, line 68: A 405 would be more appropriate. It is what we use in other areas where a user tries to modify a public collection.

Bento007 (changes requested):

backend/corpora/lambdas/api/v1/curation/collections/collection_id/actions.py
danieljhegeman: Ah, that's a good point.

danieljhegeman: fixed

Bento007 (approved):

PR #3078, opened 2022-08-10
chore(curation API): Adding a test where a collection has no datasets

Bento007:
- #3079

### Reviewers

**Readability:** 

---

## Changes
- remove extra variables from add_collection_level_processing_status
- Adding a test where a collection has no datasets


backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -230,20 +230,17 @@ def list_collections_curation(
...
-                status = ProcessingStatus.PENDING
-            elif processing_status.processing_status == ProcessingStatus.FAILURE:
-                status = ProcessingStatus.FAILURE
+        if processing_status := dataset.processing_status:
+            status = processing_status.processing_status
+            if status == ProcessingStatus.PENDING:
+                return_status = status
+            elif status == ProcessingStatus.FAILURE:
+                return_status = status
                 break
```
danieljhegeman, line 245:
```suggestion
                return status
```

danieljhegeman: Prevents accidents in the case where some code doing who-knows-what later gets inserted between the break and the final `return`.

danieljhegeman (approved):
@Bento007 👍 would just ask that you incorporate the nit suggestion but otherwise ✅

Issue #2997, opened 2022-07-28
Deprecate `access_type` in responses


brianraymor:
See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1659040871197119). `access_type` MUST be removed from all responses and the Swagger API.

brianraymor:
@Bento007 - tested for all roles and checked the swagger. Ready for production.

PR #3080, opened 2022-08-10
fix: remove access_type from curation API

Bento007:
- #2997 
- #2793 

### Reviewers

**Readability:** @ebezzi 

---

## Changes
- remove access_type from the curation_api
- modify how we determine what filters to apply when listing collections. This is to accommodate the changes needed to support the removal of the access_type. It also reduces the number of collections returned by database to only the ones the user can view.
- [Fix the error message when curator is used by not a super curator](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3080/commits/00c8c16e4469575fce062d59aba56be07148fc74)
-
## QA steps (optional)
- updated tests to use access_type.
- update tests to handle new errors return when using the curator parameter

## Notes for Reviewer
blocked by #3089


backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -50,14 +50,9 @@ def reshape_for_curation_api_and_is_allowed(
     """
 
     owner = collection.pop("owner")  # Don't actually want to return 'owner' in response
-    if is_user_owner_or_allowed(token_info, owner):
-        collection["access_type"] = "WRITE"
-    elif not id_provided and collection["visibility"] == CollectionVisibility.PRIVATE:
+    if not id_provided and collection["visibility"] == CollectionVisibility.PRIVATE:
```
danieljhegeman, line 53: this breaks [Super Curator #3](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3089#issuecomment-1213348180)

danieljhegeman: You should do as you see fit, but I think that the right way to resolve the issue here with the refactoring you're introducing in #2793 is to basically do away with the "is_allowed" portion of this method (this method was always trying to do two things which was a bad design choice on my part), move the auth logic outside of the reshaping logic (i.e., auth logic occurs in body of `...collections.actions.get` handler method, with just a single `is_super_curator(token_info)` call), and then this collection reshaping method gets called only if the user is authorized for the collection 🏄‍♂️ 🤙

danieljhegeman (dismissed):
We're now running into the problem with my design choice from a while back for the reshape method 😅 

See comment about breaking Super Curator #3

backend/corpora/lambdas/api/v1/curation/collections/actions.py
```diff
@@ -18,23 +18,28 @@ def get(visibility: str, token_info: dict, curator: str = None):
...
-    elif visibility:
+    if visibility:
         filters.append(DbCollection.visibility == getattr(CollectionVisibility, visibility))
+        if visibility == CollectionVisibility.PRIVATE.name:
+            if not token_info:
+                raise ForbiddenHTTPException(detail="Not authorized to query for PRIVATE collection.")
+            else:
+                result = owner_or_allowed(token_info)
+                if result:
+                    filters.append(DbCollection.owner == result)
```
nayib-jose-gloria, line 29: to accommodate Super Curator #3 outlined [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3089#issuecomment-1213348180), I think the owner filter should only be appended if the user is not a super curator. Maybe turn this else block into an `elif not is_super_curator(token_info)`

nayib-jose-gloria: nvm, owner_or_allowed already returns None for superusers. Maybe could use a rename, but that's out of scope for this ticket

backend/corpora/lambdas/api/v1/curation/collections/actions.py
```diff
@@ -18,23 +18,28 @@ def get(visibility: str, token_info: dict, curator: str = None):
...
+    if curator:
+        if not is_super_curator(token_info):
+            raise ForbiddenHTTPException(detail="Not authorized to use the curator query parameter.")
+        else:
+            filters.append(DbCollection.curator_name == curator)
 
     db_session = g.db_session
     resp_collections = []
     for collection in db_session.query(DbCollection).filter(*filters).all():
         resp_collection = collection.to_dict_keep(EntityColumns.columns_for_collections)
```
nayib-jose-gloria, line 40: suggestion, non-blocking: I think for a 'reshape_for_curation_api' function, it makes sense to refactor all the recurring reshaping logic (to_dict_keep, add_collection_level_processing_status) into that call. It taking in the collection object from the DB and returning a resp_collection (rather than being in-place) also seems like a better pattern to me.

backend/corpora/lambdas/api/v1/curation/collections/actions.py
nayib-jose-gloria: I'll leave this open as a non-blocking suggestion in case you do think of a better name, because it isn't clear just reading a call to owner_or_allowed what it actually returns. but admittedly, I can't come up with anything particularly great off the top of my head

nayib-jose-gloria (approved):
The change itself looks good! I left a few non-blocking suggestions for readability / code clean-up, but I recognize those issues didn't originate with this fix. We can also revisit later, or you can let me know if you disagree

backend/corpora/lambdas/api/v1/curation/collections/actions.py
Bento007: Great idea.

backend/corpora/lambdas/api/v1/curation/collections/actions.py
Bento007: I change the name to owner_filter and added a comment.

PR #3088, opened 2022-08-11
feat(landing): use real stats for hero numbers

seve:
### Reviewers
**Functional:**  @tihuan 

**Readability:** @sethfeingold @ambrosejcarr 

---

Shows real time data for number of datasets, cell count, and number of cell types

https://user-images.githubusercontent.com/8716829/184206701-de58e5cc-5147-4492-8fa5-c54c3e73babe.mov

### DIFF
--- a/frontend/src/common/queries/filter.ts
+++ b/frontend/src/common/queries/filter.ts
@@ -220,6 +220,45 @@ export function useFetchDatasetRows(): FetchCategoriesRows<DatasetRow> {
   };
 }
 
+interface CorpusSummary {
+  cellCount: number;
+  cellTypesCount: number;
+  datasets: number;
+  isError: boolean;
+  isLoading: boolean;
+}
+
+export function useCorpusSummary(): CorpusSummary {
+  const datasetRows = useFetchDatasetRows();
+  const summary = datasetRows.rows.reduce(
+    (acc, row) => {
+      const newCellTypes = row.cell_type.reduce((acc1, cellType) => {
+        if (acc.cellTypes.indexOf(cellType.ontology_term_id) === -1) {
+          acc1.push(cellType.ontology_term_id);
+        }
+        return acc1;
+      }, new Array<string>());
+      return {
+        ...acc,
+        cellCount: acc.cellCount + (row.cell_count || 0),
+        cellTypes: acc.cellTypes.concat(newCellTypes),
+        datasets: acc.datasets + 1,
+      };
+    },
+    {
+      cellCount: 0,
+      cellTypes: new Array<string>(),
+      datasets: 0,
+      isError: datasetRows.isError,
+      isLoading: datasetRows.isLoading,
+    }
+  );
+
+  const { cellTypes, ...ret } = summary;
+
+  return { ...ret, cellTypesCount: cellTypes.length };
+}
+
 /**
  * Cache-enabled hook for fetching public, non-tombstoned, datasets returning only filterable and sortable fields.
  * @returns Array of datasets - possible cached from previous request - containing filterable and sortable dataset
--- a/frontend/src/common/queries/filter.ts
+++ b/frontend/src/common/queries/filter.ts
@@ -238,9 +238,15 @@ export function useCorpusSummary(): CorpusSummary {
         }
         return acc1;
       }, new Array<string>());
+
+      const isPrimaryDataset = row.is_primary_data === IS_PRIMARY_DATA.PRIMARY;
+
+      const newCellCount =
+        acc.cellCount + (isPrimaryDataset ? row.cell_count ?? 0 : 0);
+
       return {
         ...acc,
-        cellCount: acc.cellCount + (row.cell_count || 0),
+        cellCount: newCellCount,
         cellTypes: acc.cellTypes.concat(newCellTypes),
         datasets: acc.datasets + 1,
       };
--- a/frontend/src/common/queries/filter.ts
+++ b/frontend/src/common/queries/filter.ts
@@ -237,23 +237,23 @@ export function useCorpusSummary(): CorpusSummary {
           acc1.push(cellType.ontology_term_id);
         }
         return acc1;
-      }, new Array<string>());
+      }, [] as string[]);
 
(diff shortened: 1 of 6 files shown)


seve:
> It looks like we're using "all cells" for the count. Would it be possible to count only primary cells? I think that's a more reasonable number, and will reflect what users will ultimately see via computational APIs.

good point, I'll try and see if I can make that change

seve:
@ambrosejcarr Only counting datasets with `is_primary_data === PRIMARY` 
image

ambrosejcarr:
Getting closer. This reports a lower cell count than our analytics (27,186,039). This might under-estimate the cell count since there are datasets that contain both primary and secondary cells? Might we need to calculate at the observation level?

seve:
> Getting closer. This reports a lower cell count than our [analytics](https://czi-sci.slack.com/archives/C02BHFB6DJ8/p1659961170168979) (27,186,039) - as of Monday. This might under-estimate the cell count since there are datasets that contain both primary and secondary cells? Might we need to calculate at the observation level?

This is might be due the variance between dev and prod, I'm going to test this against prod and see if the number is more accurate.

seve:
@ambrosejcarr Looks like this won't be an easy attempt due to some security settings we have on the prod API. I can give it another shot on Monday

seve:
> This might under-estimate the cell count since there are datasets that contain both primary and secondary cells?

I have the option of selecting datasets that have `is_primary_data === BOTH` that may be what the analytics is reading

ambrosejcarr:
If is_primary_data === BOTH returns a value that matches analytics, then
that will work.

frontend/src/common/queries/filter.ts
```diff
@@ -220,6 +220,51 @@ export function useFetchDatasetRows(): FetchCategoriesRows<DatasetRow> {
...
+export function useCorpusSummary(): CorpusSummary {
+  const datasetRows = useFetchDatasetRows();
+  const summary = datasetRows.rows.reduce(
+    (acc, row) => {
+      const newCellTypes = row.cell_type.reduce((acc1, cellType) => {
+        if (acc.cellTypes.indexOf(cellType.ontology_term_id) === -1) {
+          acc1.push(cellType.ontology_term_id);
+        }
+        return acc1;
+      }, new Array<string>());
```
tihuan, line 240: let's avoid using `new Array()` 🙏 

https://coderwall.com/p/h4xm0w/why-never-use-new-array-in-javascript
```suggestion
      }, [] as string[]);
```

frontend/src/common/queries/filter.ts
```diff
@@ -220,6 +220,51 @@ export function useFetchDatasetRows(): FetchCategoriesRows<DatasetRow> {
...
+      return {
+        ...acc,
+        cellCount: newCellCount,
+        cellTypes: acc.cellTypes.concat(newCellTypes),
+        datasets: acc.datasets + 1,
+      };
+    },
+    {
+      cellCount: 0,
+      cellTypes: new Array<string>(),
```
tihuan, line 256:
```suggestion
      cellTypes: [] as string[],
```

frontend/src/common/queries/filter.ts
```diff
@@ -220,6 +220,51 @@ export function useFetchDatasetRows(): FetchCategoriesRows<DatasetRow> {
...
+        if (acc.cellTypes.indexOf(cellType.ontology_term_id) === -1) {
+          acc1.push(cellType.ontology_term_id);
+        }
+        return acc1;
+      }, new Array<string>());
+
+      const isPrimaryDataset = row.is_primary_data === IS_PRIMARY_DATA.PRIMARY;
+
+      const newCellCount =
+        acc.cellCount + (isPrimaryDataset ? row.cell_count ?? 0 : 0);
```
tihuan, line 245: somehow the double 0 threw me off. Maybe the following?
```suggestion
        acc.cellCount + ((isPrimaryDataset && row.cell_count) || 0);
```

frontend/src/common/queries/filter.ts
```diff
@@ -220,6 +220,51 @@ export function useFetchDatasetRows(): FetchCategoriesRows<DatasetRow> {
...
+
+      return {
+        ...acc,
+        cellCount: newCellCount,
+        cellTypes: acc.cellTypes.concat(newCellTypes),
```
tihuan, line 250: discussion: any special reason to use `.concat()` over `...`?
```suggestion
        cellTypes: [...acc.cellTypes, ...newCellTypes],
```

The article seems to say only use `concat` when the array is large:
https://www.educative.io/answers/spread-operator-vs-arrayprototypeconcat-in-javascript

frontend/src/views/Landing/index.tsx
```diff
@@ -25,6 +27,76 @@ import SingleCellDataIconActive from "./icons/single-cell-data-active";
 import SingleCellDataIconInactive from "./icons/single-cell-data-inactive";
 import styles from "./index.module.scss";
 
+const DEFAULT_NUM_CELLS = 33000000;
```
tihuan, line 30:
```suggestion
const DEFAULT_NUM_CELLS = 33_000_000;
```

frontend/src/views/Landing/index.tsx
```diff
@@ -25,6 +27,76 @@ import SingleCellDataIconActive from "./icons/single-cell-data-active";
...
+const COUNT_UP_DURATION = 1.5;
+
+const HeroStatsCounter = ({
+  className,
+}: {
+  className: string;
+}): JSX.Element => {
+  const [hasUpdated, setHasUpdated] = useState(false);
+  const corpusSummary = useCorpusSummary();
+  const formatNumber = useCallback((n: number): string => {
```
tihuan, line 43: since this function doesn't depend on any component specific variable, we could just extract it as a function outside of the component?

E.g.,

```tsx
const HeroStatsCounter: () => {
   ...
  // use formatNumber
}

function formatNumber(n: number) {
   ...
}
```

tihuan (commented):
Looks really good! Thanks for using the real numbers, Seve! Just some tiny comments PTAL 🙏

Issue #3092, opened 2022-08-11
The processing_status for an invalid dataset returns PENDING and not FAILURE


brianraymor:
When an invalid dataset is uploaded and fails validation, its `processing_status` is reported as `PENDING` and not `FAILURE`.

Bento007:
Was this uploaded through a link or uploaded directly to s3.

Bento007:
Was this uploaded through a link or uploaded directly to s3.

brianraymor:
It was uploaded directly.

Bento007:
can you send me a link to the dataset you're using?

brianraymor:
I just used one of the old small(er) schema 1.0.0 datasets in that dropbox folder. This will immediately fail validation.

https://www.dropbox.com/s/jgunwv8amblvjk1/7MB.h5ad?dl=0

Here's a [private collection with the invalid dataset](https://cellxgene.dev.single-cell.czi.technology/collections/0ff1eb4f-d818-4eaf-9ef4-fd9ca87805a3).

When I GET /collection and check the dataset status for the tag:

'tag: badcurator.h5ad PENDING'

brianraymor:
This failed on [re-test](https://cellxgene.dev.single-cell.czi.technology/collections/632c301c-0a3a-4009-ab1f-ff949a75ca7b).

brianraymor:
@Bento007 - please do not move bugs to _Ready for Prod_ until either @jahilton  or I have re-tested and verified. This is the second re-test for this case. Testing was blocked by the x_normalization issue earlier.

That said, it looks correct:

The collection `processing_status` is "'FAILURE'.

The individual dataset `processing_status` are:

'tag: badcurator.h5ad FAILURE'
'tag: goodcurator.h5ad SUCCESS'

PR #3098, opened 2022-08-12
fix(submission pipe): add upload_failure directory to root of docker …

Bento007:
…image.

- #3092

### Reviewers
**Functional:** @MDunitz  

---

## Changes
- add the upload_failure directory to the root of the docker image. The upload_failure handler was failing because it couldn't find `app`. This was because `app` was put in the wrong place during docker image location.
- remove extra requirements from the submission handlers. This will shrink the size of the docker image, speed up build time, and reduce the cost of lambda execution.


danieljhegeman (approved):
Ok, good catch on this one. Thanks Trent

MDunitz (approved):
lgtm

PR #3107, opened 2022-08-16
fix(upload-failure): Fixing the build for upload-failure lambda

Bento007:
- #3098

### Reviewers
**Functional:** @atarashansky 

**Readability:** 

---

## Changes
- add make recipe to test upload-success. This should have been created but never was. This was used to verify the upload failure works the same way upload success works.
- remove the upload file and role into the `upload_failure.app`. The file is small enough to include in the app and fixes an import error.
- remove extra volumes from docker-compose. Not needed
- 

## QA steps (optional)
- build upload-failure locally and invoke using the make command.
- build upload-success locally and invoke using the make command.


atarashansky (approved):
lgtm

Issue #1705, opened 2021-12-02
Upgrade `next` and `node-sass` to the latest version


tihuan:
Context:
We have a vuln related to `node-sass`: https://github.com/chanzuckerberg/single-cell-data-portal/pull/1248/files

Solution:
1. Upgrade `next` first at least to latest v12, since they have fixed their `node-sass` deps [here](https://github.com/vercel/next.js/pull/28710)
1. Now we should be able to upgrade `node-sass`!
2. `next` 11.1.3 upgrade alert: https://github.com/chanzuckerberg/single-cell-data-portal/security/dependabot/frontend/package-lock.json/next/open
3. https://github.com/chanzuckerberg/single-cell-data-portal/pull/1980

seve:
Just a heads-up, if this takes too many cycles. This trim-newlines vulnerability doesn't actually affect us.  If the benefits of upgrading `next` aren't P0, I would recommend dropping the priority of this ticket.

cc: @maniarathi

maniarathi:
@tihuan Do you think it makes sense to drop to P1 then given Seve's note?

tihuan:
Thanks so much for the headsup, @seve ! 

@maniarathi : Yeah I think dropping to P1 sounds great, and we can just upgrade `node-sass` when we upgrade `next` in a few months once they port over Styled Components parser to SWC 🙆‍♂️ 

Github Issue: https://github.com/vercel/next.js/issues/30802

Thanks both!

maniarathi:
Can do this sometime in Q2 2022.

tihuan:
Since `next` finally released 11.1.3, which is a patch version bump, I'll send out a PR to upgrade it after Perf! Thank you!

https://github.com/chanzuckerberg/single-cell-data-portal/security/dependabot/frontend/package-lock.json/next/open

tihuan:
Turned out that Next only included the patch in v12:

https://github.com/vercel/next.js/issues/34139#issuecomment-1034290706

tihuan:
I think we've waited long enough for Next 12 to be stable, so maybe I can help upgrade to 12 this week? @maniarathi thank you!

PR #3119, opened 2022-08-17
chore(fe): upgrade to nextjs12 blueprint4 node16

tihuan:
- #1705

### Reviewers
**Functional:** 
@seve @hthomas-czi @frano-m 

**Readability:** 
@atarashansky 

---

## Changes
1. Upgrade Next.js from `11` to `12`
2. Upgrade Blueprint.js from `3` to `4`. [Changelog](https://github.com/palantir/blueprint/wiki/Blueprint-4.0)
3. Pin node from `14` to `16`. UPDATE: Pinning to `16.14.2` to avoid GH Action lint error exit code 243. [[Source]](https://stackoverflow.com/a/71892226/3120863)
4. Remove `styled-components` and use `emotion` completely (so we can use Next's fast SWC compiler and only have one css-in-js dependency)

Notable changes:
1. Next12 doesn't use Babel anymore and defaults to use its new SWC rust based compiler, which is supposedly 17x faster [migration doc](https://nextjs.org/docs/upgrading#upgrading-from-11-to-12). So in order for `jest` test runner to work with TS files, we need to update jest's config files to use `next/jest` to compile TS files before jest consumes them

3. Blueprint3 to Blueprint4 upgrade was a pain. It includes removing `node-sass` as a direct dependency (which is now officially deprecated in favor of `sass`(dart-sass)) AND also install `@vgrid/sass-inline-svg` to process Blueprint SVG files. On top of that, because we define a custom theme via sass, I had to follow [custom sass processing setup steps](https://github.com/palantir/blueprint/issues/4759#issuecomment-1112218581) to overcome Blueprint's very limited support of theming. Currently the Blueprint team hasn't come up with a solution for official theming support, as shown in [this comment](https://github.com/palantir/blueprint/issues/4032#issuecomment-1087723314). Given that `@vgrid/sass-inline-svg` actually still depends on `node-sass`, we might need to move away from Blueprint completely sooner than later to avoid indirectly depending on a deprecated package for too long

3. Migration from `styled-components` to `emotion` includes a small change of ~removing the usage of `import { css } from 'styled-components'`, because `@emotion/css` outputs a class string instead of a styling function. So I just rewrote those instances with a styling function instead! See `frontend/src/components/Collections/components/Dataset/common/style.ts` for example~

    UPDATE: Thanks to Alec, we just had to change the import from `{ css } from 'styled'` to `{ css } from '@emotion/react'`

## QA steps (optional)
RDEV:
https://thuang-next-12-frontend.rdev.single-cell.czi.technology/

Local:
1. Make sure your node is v16
2. Run `rm -rf node_modules && npm i` to install all new dependencies
3. Run the app via `make local-sync` OR if your local app connects to dev/staging BE, then just `npm run dev`
5. App should work and all the styles should look the same!

## Notes for Reviewer
Tagging all the people that work on FE for visibility!


seve (changes requested):
we should also bump https://github.com/chanzuckerberg/single-cell-data-portal/blob/c85d9a4770d1418342a8c65cfb2429b75654bc82/frontend/Dockerfile#L2

tihuan:
> 

Oh thank you! Done ✅

frano-m:
Hey @tihuan woah, epic and brave :star2:!
Looks like there are few minor and possibly unwanted changes from upgrading blueprint.
- At a glance the N+ tag colour is different. It looks like Figma has moved on since we originally implemented N+ tag and so the PR doesn't match the current design or current implementation. If the current Figma is the desired design, we should update all N+ tags for consistency (the differences at a glance are border radius, colour, font size and padding).
- The close icon `x` on the selected terms should be white. The classname `bp3-tag-remove` inherited its colour, however in updated classname `bp4-tag-remove` the colour is now specified.
- The filter menu search input placeholder "Search" is a slightly lighter colour than the original implementation.
- BP MenuItem with prop icon appears to have wrapped an additional element around `bpx-icon` with classname `bp4-menu-item-icon`. This had added unwanted styles like margin-right and height. Each menu item is now 2px taller and 4px wider as a result.

I'd love to see this on rdev to test the new collections form. And that's it for now! Nice work!!

atarashansky:
For the record, I think I got `import { css } from "@emotion/react"` to work in place of `import { css } from "styled-components"`

tihuan:
Ooh I didn't think of using `css` from `@emotion/react` instead of `@emotion/css`! I'll use that now. Thanks, Alec!

.gitignore
```diff
@@ -141,3 +141,6 @@ dmypy.json
 .envrc
 .env.ecr
 *.sqlc
+
+# Typescript
+tsconfig.tsbuildinfo
```
tihuan, line 146: a quick research shows that `tsbuildinfo` is not worth checking in 

https://github.com/vercel/next.js/discussions/30815

frontend/.nvmrc
```diff
@@ -1 +1 @@
-v15
+v16.14.2
```
tihuan, line 1: Pinning to  avoid GH Action lint error exit code 243
See: https://stackoverflow.com/a/71892226/3120863

frontend/Dockerfile
```diff
@@ -1,26 +1,32 @@
...
+# Pinning to 16.14.2 to avoid GH Action lint error exit code 243
+# See: https://stackoverflow.com/a/71892226/3120863
+FROM node:16.14.2
 
 # install dependencies first, in a different location for easier app bind mounting for local development
 # due to default /opt permissions we have to create the dir with root and change perms
 RUN mkdir /opt/node_app && chown node:node /opt/node_app
 WORKDIR /opt/node_app
+
+RUN apt-get update && apt-get install -y make wget \
```
tihuan, line 11: copied from GenEpi's Dockerfile 😆

frontend/Dockerfile
```diff
@@ -1,26 +1,32 @@
...
 # the official node image provides an unprivileged user as a security best practice
 # but we have to manually enable it. We put it here so npm installs dependencies as the same
 # user who runs the app.
 # https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user
-
-
-# -- TODO, we should try turning this back on later.
-# USER node
-COPY package.json package-lock.json* ./
+USER node
```
tihuan, line 26: we can enable this now, because we switch back to root before running `ENTRYPOINT` below, so the blast radius is a little smaller

frontend/Makefile
```diff
@@ -16,8 +16,7 @@ init:
 
 .PHONY: lint
 lint:
-	npm run prettier-check
-	npm run lint
+	npm run prettier-check & npm run lint
```
tihuan, line 19: run both commands in parallel for speed

frontend/next.config.js
```diff
@@ -87,4 +90,27 @@ module.exports = {
...
+        path.join(__dirname, "blueprint-icons/icons"), // Replaced path to icons
+        {
+          // minimal "uri" encoding is smaller than base64
+          encodingFormat: "uri",
+          // run through SVGO first
+          optimize: true,
+        }
+      ),
+    },
+  },
```
tihuan, line 115: I removed `swcMinify: true`, because it breaks the build

See: https://github.com/vercel/next.js/issues/38436#issuecomment-1220155395

frontend/src/components/common/Grid/components/ActionButton/index.tsx
```diff
@@ -12,12 +14,7 @@ export default function ActionButton({
   isAnchorButton = false,
   ...props /* Spread props to allow for data-test-id and button specific attributes e.g. "href", "target", or "disabled". */
 }: Props): JSX.Element {
-  return (
-    <StyledActionButton
-      as={isAnchorButton ? AnchorButton : undefined}
```
tihuan, line 17: @emotion styled's `as` seems to only work with native DOM elements

E.g.,
```tsx
const button = styled.button`
`
```

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -312,11 +316,11 @@ const DocContent = styled.div`
   }
 
   & h1,
-  h2,
-  h3,
-  h4,
-  h5,
-  h6 {
+  & h2,
```
Comment (tihuan) [inline frontend/src/pages/docs/[...slug].tsx:319]: @seve I got a stylelint error saying &h1 needs to go after h2, h3, h4, etc., but I assume that you wanted `&` for all header tags?

frontend/src/views/WheresMyGene/components/HeatMap/style.ts
```diff
@@ -40,12 +40,12 @@ export const YAxisWrapper = styled.div`
   top: 0;
   left: 0;
   z-index: 1;
-  // Somehow Firefox requires this to scroll
+  /* Somehow Firefox requires this to scroll */
```
tihuan, line 43: `//` comment style doesn't work in emotion

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -229,7 +229,10 @@ interface CreateYAxisOptionsProps {
 /**
  * Used to calculate text pixel widths. Should be only created once.
  */
-const CTX = document.createElement("canvas").getContext("2d");
+const CTX =
```
tihuan, line 232: nextjs 12 somehow invokes this file on the server side, which doesn't have access to `document`, so we need to check `typeof document` now

frontend/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     "target": "esnext",
     "types": ["jest-playwright-preset", "expect-playwright", "jest"],
     "forceConsistentCasingInFileNames": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "incremental": true
```
tihuan, line 23: this is enabled by nextjs, which speeds up TS compilation

https://www.typescriptlang.org/tsconfig#incremental

tihuan:
I'm going through the UI one more time after Fran's comment to see if I can catch more issues 😆

tihuan:
> Hey @tihuan woah, epic and brave 🌟! Looks like there are few minor and possibly unwanted changes from upgrading blueprint.
> 
> * At a glance the N+ tag colour is different. It looks like Figma has moved on since we originally implemented N+ tag and so the PR doesn't match the current design or current implementation. If the current Figma is the desired design, we should update all N+ tags for consistency (the differences at a glance are border radius, colour, font size and padding).

Wow I didn't even notice the border radius changed from 2px to 1px 🤯 , how did you spot that? Too cool!  

It seems to me that the font size and padding of N+ tag are the same as before?

And since the border radius and color change are both part of the Blueprint4 design, I wonder if it's okay to keep them as is!

@hthomas-czi PTAL 🙏 

demo

> * The close icon `x` on the selected terms should be white. The classname `bp3-tag-remove` inherited its colour, however in updated classname `bp4-tag-remove` the colour is now specified.
> * The filter menu search input placeholder "Search" is a slightly lighter colour than the original implementation.
> * BP MenuItem with prop icon appears to have wrapped an additional element around `bpx-icon` with classname `bp4-menu-item-icon`. This had added unwanted styles like margin-right and height. Each menu item is now 2px taller and 4px wider as a result.

Thank you so much for catching this 🙏 Updated to match the current design!

> I'd love to see this on rdev to test the new collections form. And that's it for now! Nice work!!

The rdev is now available! https://thuang-next-12-frontend.rdev.single-cell.czi.technology
And I've compared this branch to `dev`, and the diff seems to be just the gray color as intended in [Blueprint 4 design](https://github.com/palantir/blueprint/wiki/Blueprint-4.0) for better a11y color contrast, so I'm thinking maybe we should keep that? Let's see what @hthomas-czi thinks!

demo

Thanks again for the wonderful review, Fran! I really appreciate your insights here!

tihuan:
The PR should be ready for everyone to review now! Thanks again!

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -333,13 +337,13 @@ const DocContent = styled.div`
...
 
-const StyledImage = styled(Image)``;
+const StyledImage = styled(Image)();
 
 const ImageContainer = styled.div`
   width: 100%;
   margin: 24px 0;
 
-  > div {
+  > span {
```
Comment (tihuan) [inline frontend/src/pages/docs/[...slug].tsx:346]: this change is needed since Next12 changed the Image wrapper from `div` to `span`:

https://nextjs.org/docs/upgrading#nextimage-changed-wrapping-element

Comment (tihuan) [reply frontend/src/pages/docs/[...slug].tsx:346]: Images look good now! Thanks so much for catching this, @seve 🙏 

Screen Shot 2022-08-24 at 10 41 49 AM

frontend/src/pages/_document.tsx
```diff
@@ -1,53 +1,8 @@
...
-  Main,
-  NextScript,
-} from "next/document";
-import { ServerStyleSheet } from "styled-components";
+import RawDocument, { Head, Html, Main, NextScript } from "next/document";
 
 const OG_PAGE_TITLE = "Cellxgene Data Portal";
 
 export default class Document extends RawDocument {
-  static async getInitialProps(
```
atarashansky, line 14: Just curious, why is this function going away?

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -275,7 +279,7 @@ const StyledLeftNav = styled.div`
   ::-webkit-scrollbar-thumb {
     background-clip: padding-box;
     border-right: 4px #f8f8f8 solid;
-    background: grey;
+    background-color: grey;
```
Comment (atarashansky) [inline frontend/src/pages/docs/[...slug].tsx:282]: gray vs grey? are both recognized?

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
atarashansky: weird...

atarashansky (commented):
Readability is good, I have no suggestions other than a couple small questions. I'll wait for @seve to approve.

seve (approved):

frontend/src/pages/docs/[...slug].tsx
Comment (tihuan) [reply frontend/src/pages/docs/[...slug].tsx:282]: ooh never thought about that 💡 !

Just confirmed that they are both recognized!

https://www.w3schools.com/colors/colors_names.asp

Screen Shot 2022-08-24 at 11 07 21 AM

frontend/src/pages/_document.tsx
tihuan: Ah sorry for missing to document this change!

`styled-components` needed the extra config to server side render the stylesheets, but `@emotion` handles it automatically, so less code for us to maintain!

tihuan:
Thanks so much, @seve and @atarashansky !

I'll also wait for @hthomas-czi and @frano-m for approval before merging 🚀

tihuan:
Harley approved in [Slack](https://czi-sci.slack.com/archives/C023Q1APASK/p1661364876590849?thread_ts=1661284800.833469&cid=C023Q1APASK)!

Issue #304, opened 2020-05-19
Update Corpora database to reflect finalized schema


maniarathi:
Perform a database migration to such that any necessary changes to reflect parity with the finalized Corpora schema.

Tasks
- [ ] Update high level architecture document to reflect new changes
- [ ] Update database schema
- [ ] Perform database migration

brianraymor:
Closed during triage with Arathi. No longer required based on completing the schema design.

PR #3125, opened 2022-08-18
feat: add donor_id to the backend

ebezzi:
Adds `donor_id` to the supported fields.

Also contains the following preparatory changes for the schema 3.0.0 transition:
1. Migrate requirements to `schema_version = 3.0.0`
1. Add cellxgene-schema version 3 (for now, using a gzipped release)
1. Remove `X_normalization` and rename `ethnicity` - those are partial changes but are required for testing uploads with 3.0.0 compliant datasets. The full changes will be done in separate PRs.

TODO:
1. Curation API


backend/corpora/dataset_processing/process.py
```diff
@@ -416,23 +416,26 @@ def _get_batch_condition():
...
+
     metadata = {
         "name": adata.uns["title"],
         "organism": _get_term_pairs("organism"),
         "tissue": _get_term_pairs("tissue"),
         "assay": _get_term_pairs("assay"),
         "disease": _get_term_pairs("disease"),
         "sex": _get_term_pairs("sex"),
-        "ethnicity": _get_term_pairs("ethnicity"),
+        "ethnicity": _get_term_pairs("self_reported_ethnicity"), # TODO
```
ebezzi, line 429: Let's keep the field named `ethnicity` for now until we're ready for #304.

nayib-jose-gloria (commented):
There are dataset (and dataset preview) schemas in the curation API yaml, do we need to update those and any associated endpoints as part of this PR?

ebezzi:
> There are dataset (and dataset preview) schemas in the curation API yaml, do we need to update those and any associated endpoints as part of this PR?

We definitely should. No easy way to test it using rdev though

backend/corpora/common/corpora_orm.py
```diff
@@ -323,6 +323,7 @@ class DbDataset(Base, AuditMixin, TimestampMixin):
     development_stage = Column(JSONB)
     cell_type = Column(JSONB)
     cell_count = Column(Integer)
+    donor_id = Column(ARRAY(String), nullable=True)
```
nayib-jose-gloria, line 326: question: It makes sense to set nullable=True so we can do the migration with the existing corpus. Should we add a TODO/ticket to update this after the migration to remove nullable=True, since it's a required field?

tests/unit/backend/corpora/api_server/test_v1_collection.py
```diff
@@ -193,6 +193,7 @@ def test__get_collection_id__ok(self):
                     "x_normalization": "test_x_normalization",
                     "x_approximate_distribution": "NORMAL",
                     "batch_condition": ["batchA", "batchB"],
+                    "donor_id": ["donor_1", "donor_2"]
```
nayib-jose-gloria, line 196: typo: missing comma

tests/unit/backend/utils.py
```diff
@@ -75,6 +75,7 @@ def get(cls, **kwargs):
             x_normalization="normal",
             x_approximate_distribution=XApproximateDistribution.NORMAL.name,
             batch_condition=np.array(["batchA", "batchB"], dtype="object"),
+            donor_id=["donor_1", "donor_2"]
```
nayib-jose-gloria, line 78: typo" missing comma after list

tests/unit/backend/utils.py
```diff
@@ -75,6 +75,7 @@ def get(cls, **kwargs):
...
             schema_version="2.0.0",
```
nayib-jose-gloria, line 79: todo: update schema version

nayib-jose-gloria (commented):
small comments + typos, but generally this looks good! Let's add the curation API updates as well

nayib-jose-gloria:
> > There are dataset (and dataset preview) schemas in the curation API yaml, do we need to update those and any associated endpoints as part of this PR?
> 
> We definitely should. No easy way to test it using rdev though

Good point. I think its safe to rely on unit tests for now since its just a response shape change, but lets think about it and assess whether the unit tests seem sufficient

backend/corpora/common/corpora_orm.py
ebezzi: I think `nullable=True` is the default and I wanted to be explicit here. I don't think that enforcing non-nullability is important since this will already be handled by the pipeline.

backend/corpora/common/entities/dataset.py
```diff
@@ -169,16 +169,29 @@ def get_assets(self):
         """
         return self.artifacts
 
+    @staticmethod
+    def parse_version(version):
```
maniarathi, line 173: There's a nice Python package that does this instead of us writing our own custom code: https://pypi.org/project/semver/. I know it's a simple change, but it helps reduce the general surface area of this file, especially since this function is a general utility function, not something specific to the object that this class represents.

ebezzi: I don't know how I feel about adding another dependency for a single usage. Also, I just realized that versions are lexicographically sorted so we should be able to just do `< "2.0.0"` and it should work.

maniarathi (commented):
General comments:

- I would actually add a PR description to this PR and make it very descriptive. Not only does this PR add the donor_id, but it appears to also disallow further processing of H5AD files that have been created using the 2.x.x schema which is a pretty major change, even if the code change itself is small.
- It appears to add support for a function `transform_sex_for_schema_2_0_0` in a situation where the `schema_version` doesn't exist, which seems like quite a departure from current assumptions which is that a `schema_version` MUST exist.

ebezzi:
> General comments:
> 
> * I would actually add a PR description to this PR and make it very descriptive. Not only does this PR add the donor_id, but it appears to also disallow further processing of H5AD files that have been created using the 2.x.x schema which is a pretty major change, even if the code change itself is small.
> * It appears to add support for a function `transform_sex_for_schema_2_0_0` in a situation where the `schema_version` doesn't exist, which seems like quite a departure from current assumptions which is that a `schema_version` MUST exist.

1. The portal won't support both 2.0.0 and 3.0.0 by requirement. This PR will be merged to a feature branch, which will then be released when we're ready for the transition. 
2. The function was already in place because of some old datasets that were not migrated, if I recall correctly (possibly the ones that don't have a UUID as an id). I remember having to introduce it during the previous 1.0.0 -> 2.0.0 transition. We could verify if those can be fixed/migrated, but for now I don't think it's safe to skip that check.

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -151,6 +151,7 @@ class EntityColumns:
         "disease",
         "organism",
         "tombstone",
+        "donor_id",
```
nayib-jose-gloria, line 154: per brian and jason's [specifications](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/308), we don't want the donor_id as part of the dataset preview (but we do want it as part of the full dataset metadata response)

PR #3126, opened 2022-08-18
feat: cube details logging

MDunitz:
### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- Remove excess log setups
- Log the dataset, cell, and gene count for integrated corpus
- calculate the gene, cell and dataset counts from the generated cubes.
- add caching to is_dataset_already_loaded and removed caching from get_all_datasets. get_all_datasets value should not be cached since the input value never changes. This will cause us to never get the update list of unique datasets as we continue to add datasets to the cube.
- add tiledb query to get gene_count from the expressions cube
- add tiledb query to get the cell_count from the cell_count cube
- add tests for the load_data_and_create_cube stats
- rename cube_ to expression_summary_


maniarathi (commented):
Are there any unit tests that need to get updated to ensure that the return values of all the functions that have changed in this PR are tested and return the expected value?

Bento007 (approved):

atolopko-czi (approved):
This seems okay so approving as is, but a few considerations and nits:
* Instead of returning counts from various methods, the counts could be ascertained in a black-box fashion by querying the generated cubes; this would report what actually gets written to disk (I still have some distrust of TileDB). Code changes would be simpler as well.
* It could be worth logging the counts from the intermediate "integrated corpus" in addition  the generated snapshot counts. There is some filtering that goes on, but I don't remember the details; either way it could raise a red flag the pipeline starts misbehaving.
* "The cube contains # cells" is more accurate as "The snapshot was generated from # cells". (There are 2 cubes and they don't contain cells, per se, but rather aggregate stats for these cells.)
* No test has been added or updated; one high-level test to check the generated message might be reasonable

tests/unit/wmg_processing/test_cube_pipeline.py
```diff
@@ -1,10 +1,41 @@
...
+
+class TestCubePipe(unittest.TestCase):
+    @mock.patch("backend.wmg.data.cube_pipeline.notify_slack")
+    def test_exception_handle_catches_errors(self, mock_notify_slack: Mock):
+        with tempfile.TemporaryDirectory() as tempdir:
+            last_wd = os.getcwd()
+            os.chdir(tempdir)
+            with self.assertLogs(logger, logging.ERROR) as logs:
+                try:
+                    main()
```
atolopko-czi, line 29: curious, what error is generated?

tests/unit/wmg_processing/test_load_corpus.py
```diff
@@ -32,25 +32,26 @@ def fixture_file_path(relative_filename):
         parent_dir_of_curr_file = "/".join(os.path.dirname(__file__).split("/")[:-1])
         return os.path.abspath(os.path.join(parent_dir_of_curr_file, relative_filename))
 
+    @staticmethod
+    def create_anndata_test_object(path, dataset_name, num_genes=1000, num_cells=5000):
```
atolopko-czi, line 36: I suppose this could also live in `tests.unit.backend.wmg.fixtures.test_anndata_object`

tests/unit/wmg_processing/test_load_corpus.py
```diff
@@ -121,40 +122,62 @@ def test_expression_matrix_matches_global_index(self):
...
+                self.assertTrue(
+                    expected_obs_df.development_stage_ontology_term_id.equals(
+                        actual_obs_df.development_stage_ontology_term_id
+                    )
+                )
+                self.assertTrue(expected_obs_df.obs_idx.equals(actual_obs_df.obs_idx))
+                self.assertTrue(expected_obs_df.dataset_id.equals(actual_obs_df.dataset_id))
+                self.assertTrue(expected_obs_df.dataset_local_cell_id.equals(actual_obs_df.dataset_local_cell_id))
+
+                # check vars
```
atolopko-czi, line 166:
```suggestion
                # check var
```

tests/unit/wmg_processing/test_load_corpus.py
```diff
@@ -121,40 +122,62 @@ def test_expression_matrix_matches_global_index(self):
...
-        with tiledb.open(self.fixture_file_path("fixtures/small-corpus/X"), "r") as x:
-            expected_x_df = x.df[:]
-
-        self.assertTrue(expected_x_df.equals(actual_x_df))
+    @patch("backend.corpus_asset_pipelines.integrated_corpus.transform.GENE_EXPRESSION_COUNT_MIN_THRESHOLD", 1)
+    @patch("backend.corpus_asset_pipelines.integrated_corpus.job.tiledb.consolidate", new=Mock())  # Slow
+    @patch("backend.corpus_asset_pipelines.integrated_corpus.job.tiledb.vacuum", new=Mock())  # Slow
+    @patch("backend.wmg.data.cube_pipeline.upload_artifacts_to_s3", new=Mock())  # Don't upload the cube.
+    @patch("backend.wmg.data.cube_pipeline.extract.get_dataset_s3_uris", new=Mock(return_value={}))
+    def test_corpus_creation_works_as_expected(self):
```
atolopko-czi, line 130: technically, this is testing _snapshot_ creation; the corpus is the input, as a test fixture; rename?

atolopko-czi (approved):
LGTM! Thanks for all the extra refactoring and unit test work.  A few minor nits you can review/address before merging.

tests/unit/wmg_processing/test_cube_pipeline.py
Bento007: `No such file or directory: 'datasets'`

I'm going to mock raise an error, so it is consistent in it's error.

Issue #3116, opened 2022-08-17
Forbidden actions should return the same status code


brianraymor:
See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660600181278149).

For example:

1. Revise a revision -> 403 (Forbidden) - **not documented in the swagger API.**
2. Create second revision while a first revision is in progress -> 409 (Conflict)

It would be simpler for users if 403 (Forbidden) was returned for both cases because the portal *policy* does not allow either action. 

>    The 403 (Forbidden) status code indicates that the server understood
   the request but refuses to authorize it.

PR #3131, opened 2022-08-19
fix(curation api): fix error returned when a revision already exists

Bento007:
- #3116 

### Reviewers

---

## Changes
- modify the error message returned when a revision already exists.


ebezzi (approved):
I think 409 actually made more sense, but I guess this has been discussed with the product team.

Issue #3118, opened 2022-08-17
PATCH /collections/{collection_uuid} MUST fail when required fields have null values


brianraymor:
bad_collection_form_metadata = {
    "name": "",  # required
    "description": "",  # required
    "contact_name": "",  # required
    "contact_email": "",  # required
}

If this is passed to `update_collection` in the notebook, then all the required fields are missing in the collection:

![Screen Shot 2022-08-17 at 12.58.09 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/0da94981-0d19-407a-a8f8-11714b577212)

brianraymor:
@Bento007 - are we also ensuring that the email value is valid if present?

brianraymor:
@Bento007 = Re-tested with "" values. Fails as expected with 400.

PR #3132, opened 2022-08-19
fix(curation api): If fields are included they cannot be blank for required fields

Bento007:
- #3118

### Reviewers

---

## Changes

- modify the code that verifies the collection body. Required fields cannot be blank, but they can be missing from the response body.

### DIFF
--- a/backend/corpora/lambdas/api/v1/collection.py
+++ b/backend/corpora/lambdas/api/v1/collection.py
@@ -200,22 +200,22 @@ def verify_collection_links(body: dict, errors: list) -> None:
 
 
 def verify_collection_body(body: dict, errors: list) -> None:
-    def _check_if_blank(key) -> bool:
+    def check_if_blank(key) -> bool:
         if key in body.keys():
             if not body[key]:
                 errors.append({"name": key, "reason": "Cannot be blank."})
             else:
                 return body[key]
 
-    contact_email = _check_if_blank("contact_email")
+    contact_email = check_if_blank("contact_email")
     if contact_email:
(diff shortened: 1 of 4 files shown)


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -199,19 +199,23 @@ def _error_message(i: int, _url: str) -> dict:
...
-    if not body.get("description") and not allow_none:  # Check if description is None or 0 length
-        errors.append({"name": "description", "reason": "Cannot be blank."})
-
-    if not body.get("name") and not allow_none:  # Check if name is None or 0 length
-        errors.append({"name": "name", "reason": "Cannot be blank."})
-
-    if not body.get("contact_name") and not allow_none:  # Check if contact_name is None or 0 length
-        errors.append({"name": "contact_name", "reason": "Cannot be blank."})
+def verify_collection_body(body: dict, errors: list) -> None:
+    def _check_if_blank(key) -> bool:
```
danieljhegeman, line 203: I thought the leading `_` was a convention to indicate the method should be treated as private. I don't think this applies to methods defined in function scope. 🤔

danieljhegeman (approved):
see nit comment on helper method name 

I am approving though I want to flag [this issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/3134) concerning unicode points 0 thru 31. 

Approving because this would not be a regression in any way (consistent with current UI behavior), but we do still need to improve the validation.

PR #3135, opened 2022-08-22
feat: remove X_normalization

ebezzi:
### Reviewers
**Functional:** 
@nayib-jose-gloria

### DIFF
--- a/frontend/doc-site/032__Contribute and Publish Data.mdx
+++ b/frontend/doc-site/032__Contribute and Publish Data.mdx
@@ -39,7 +39,6 @@ Datasets need the following information added to a single h5ad (AnnData 0.7) for
 - **Dataset-level metadata in uns**:
   - schema_version: `2.0.0`
   - title: title of the individual dataset visualization
-  - X_normalization: method used to normalize the data stored in X and any scaling or logging; 'none' if data in X is raw
   - optional: batch_condition: list of obs fields that define that “batches” that a normalization or integration algorithm should be aware of
 - **Data in .X and raw.X**:
   - raw counts are required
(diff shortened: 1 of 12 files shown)


frontend/doc-site/032__Contribute and Publish Data.mdx
```diff
@@ -39,7 +39,6 @@ Datasets need the following information added to a single h5ad (AnnData 0.7) for
 - **Dataset-level metadata in uns**:
   - schema_version: `2.0.0`
```
nayib-jose-gloria, line 40: todo: update schema version while we're here. Might be worthwhile to do a quick check over any docs that refer to 2.0.0 and update, especially now that we're deprecating a field from 2.0.0

nayib-jose-gloria: actually, maybe we should have a separate documentation ticket--noticing a couple of other spots that could use an update. Maybe this should just link out to the single-cell-curation schema docs? so as to not always have to update two different spots.

either way, I don't want to block this specific ticket with this issue

nayib-jose-gloria (commented):
Feel free to merge after lint checks pass. I did leave 1 comment about updating outdated documentation but I think we should make a separate ticket for that since there's more than 1 spot to update, so as to not block this ticket

frontend/doc-site/032__Contribute and Publish Data.mdx
ebezzi: I created a ticket to audit the whole docs: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/3156

Issue #3105, opened 2022-08-15
datasets.artifacts field temporarily appears during upload?


brianraymor:
1. Upload a dataset using either local file or links
2. GET ../collection{uuid} at the start of the upload
3. There is an `artifacts` field reported for the new dataset. This disappears before processing is finished.

```

{'artifacts': [],
   'assay': [],
   'cell_count': None,
   'cell_type': [],
   'curator_tag': 'fromlink.h5ad',
   'development_stage': [],
   'disease': [],
   'ethnicity': [],
   'explorer_url': None,
   'id': '0e32c878-eed3-4557-b616-91139d4c7478',
   'is_primary_data': None,
   'mean_genes_per_cell': 0.0,
   'name': '',
   'organism': [],
   'processing_status': 'PENDING',
   'revised_at': None,
   'revision': 0,
   'schema_version': None,
   'sex': [],
   'tissue': [],
   'tombstone': False,
   'x_approximate_distribution': None,
   'x_normalization': None}],

```

brianraymor:
@jahilton also noticed this case.

PR #3137, opened 2022-08-22
fix(curation api): remove artifact from GET collection_id response

Bento007:
- #3105 

### Reviewers

---

## Changes
- modify how the collection id response is built to remove artifacts from the response even if it's an empty list.

### DIFF
--- a/backend/corpora/lambdas/api/v1/curation/collections/common.py
+++ b/backend/corpora/lambdas/api/v1/curation/collections/common.py
@@ -96,12 +96,11 @@ def reshape_datasets_for_curation_api(datasets: typing.List[dict], preview=False
 
 
 def reshape_dataset_for_curation_api(dataset: dict, preview=False) -> dict:
-    if artifacts := dataset.get("artifacts"):
+    if artifacts := dataset.pop("artifacts", []):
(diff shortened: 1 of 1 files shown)


backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -96,12 +96,11 @@ def reshape_datasets_for_curation_api(datasets: typing.List[dict], preview=False
 
 
 def reshape_dataset_for_curation_api(dataset: dict, preview=False) -> dict:
-    if artifacts := dataset.get("artifacts"):
+    if artifacts := dataset.pop("artifacts", []):
```
ebezzi, line 99: nit: since this is an operation with a side effect, I think It would be better to split this into two separate statements: the pop and the if. Just an opinion though.

ebezzi (approved):

backend/corpora/lambdas/api/v1/curation/collections/common.py
Bento007: with the walrus operator (`:=`) we are doing both of those operations in one.

Issue #3102, opened 2022-08-12
Remove`.h5ad` suffix requirement for identifiers


danieljhegeman:
See conversation [here](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660162834937169)

The curator should not be required to, but _may_, include the `.h5ad` suffix when communicating either the curator tag OR the dataset id to the API. The backend must handle the determination accordingly _for all endpoints_. This includes the lambda logic that parses the s3 key.

Updated tag and id handling rules: 
- a tag may not match a uuid. (cc @Bento007 for approval)
- A curator tag may not be an empty string.
- Curators may not assign their own ids. They _must_ assign a curator tag when uploading to s3, but are not required to use it thereafter; they may reference it by the dataset id which they can retrieve via GET /collections/{collection_id} and filtering for the new dataset, whether manually or by searching with the curator tag.

brianraymor:
I would avoid special-casing .h5ad, because it increases the test matrix.

Bento007:
no h5ad at all please. We only need to add the h5ad suffix when they download the file, but even then they can modify the name themselves.

danieljhegeman:
Ok.

I think we should make the change by allowing files whose url includes `s3://cellxgene-dataset(env?-)submissions` to skip [this check](https://github.com/chanzuckerberg/single-cell-data-portal/blob/9af989384aaf660182036996655f817186f5978d/backend/corpora/common/upload.py#L63)

brianraymor:
See a [recent case](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660779700824279?thread_ts=1660778105.668779&cid=C02UGUV5R3J) where Jason tripped over the .h5ad requirement.

brianraymor:
Per triage with @jahilton and @Bento007 - there was agreement that this refinement was required to simplify the API.

jahilton:
Special attn to one of the cases I hit, in case it isn't covered in the '.h5ad' work.
I believe it was rooted in uuid-formatted curator_tags, which should be blocked according to the notebook documentation. There was no error, file reports uploading but then nothing happens on the portal.

brianraymor:
> There was no error, file reports uploading but then nothing happens on the portal.

I captured this case [above](https://github.com/chanzuckerberg/single-cell-data-portal/issues/3102#issuecomment-1218676449).

brianraymor:
@Bento007 - you'll also update the notebooks?

[...]

Issue #3096, opened 2022-08-12
Malformed tags are producing generic errors


danieljhegeman:
When a curator_tag query parameter has a suffix other than `.h5ad`, the error returned is a generic 400 ("One or more parameters is invalid.").

In such an event, the error message should contain detail about the suffix since this logic will need to be done in a function that does some forgiving parsing of curator tags and dataset ids. See [this issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2631) and [this conversation thread](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660162834937169).

brianraymor:
This is no longer relevant since the .h5ad suffix requirement is being deprecated per curator request.

PR #3151, opened 2022-08-22
fix(curation api):Remove h5ad suffix requirements

Bento007:
- #3096
- #3102 

### Reviewers

**Readability:** 

---

## Changes
- remove .h5ad suffix from uploads and curator tags
- check for embedded UUID in curator tag


backend/corpora/common/utils/regex.py
```diff
@@ -1,6 +1,24 @@
...
+    """
+    Verify the correct curator tag format is obeyed.
+
+    :param curator_tag: the tag name to validate.
+    :return: True if CURATOR_TAG_PREFIX_REGEX is matched.
+    """
+    regex = f"^({DATASET_ID_REGEX}|{CURATOR_TAG_REGEX})$"
+    matched = re.match(regex, curator_tag)
+    if matched and (tag := matched.groupdict().get("tag")):
+        if not re.search(ID_REGEX, tag):
```
ebezzi, line 20: I think you can only check and raise for the positive case and skip the else.

Bento007: `re.search` returns None if no matches are found. `None` is `false` in python.

backend/corpora/common/utils/regex.py
```diff
@@ -1,7 +1,26 @@
...
+            return
+        else:
+            ValueError("Curator tag cannot contain the same shape as a UUID.")
```
ebezzi, line 25:
```suggestion
            raise ValueError("Curator tag cannot contain the same shape as a UUID.")
```

ebezzi (approved):

Issue #3134, opened 2022-08-20
Non-printable characters are (wrongly) being accepted for Collection names


danieljhegeman:
_I could've sworn I made an issue for this already a month or so ago, so don't be surprised if you find a duplicate (the one which I can't seem to find right now)_

The [validation of Collection metadata](https://github.com/chanzuckerberg/single-cell-data-portal/blob/5fa765b0a030a04ba28f9901e1e946bc85234bf5/backend/corpora/lambdas/api/v1/collection.py#L202-L216) is not adequate. It currently permits [characters which do not yield a physical character representation aka grapheme](https://www.google.com/search?q=non-printable+ascii+characters&oq=non-printable+&aqs=chrome.2.69i57j0i512l9.3035j0j7&sourceid=chrome&ie=UTF-8) (anything with a decimal code 1 thru 31; I take issue with the linked post claiming that line feeds and vertical tabs are "non-printable" because they can be/are printed--they just can't be used for hyperlink text).

This is a problem because it would produce scenarios we don't want on the FE, among them being nonexistent Collection links on the Collections page(s) in the Data Portal (I've already demonstrated this can be done), which means no way to click into a Collection.

danieljhegeman:
can use `re.sub(r'[\x00-\x1f]', " ",` to replace each such character with a single space (though perhaps not on the trailing end, in which case just remove? I don't know if we want to worry about that). At the very least use it to remove the characters.

brianraymor:
> edit: on second thought, simply reject an attempt to create/update a Collection with any of those characters present in the name.

Assuming that we went with the second thought. I'm seeing a 400 when I attempt to include these characters during either creation of the private collection or update of collection information:

FAILED

```
BAD REQUEST

{
  "detail": [
    {
      "name": "description",
      "reason": "Invalid characters detected."
    }
  ],
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}
```

@Bento007 - This is now ready for production.

danieljhegeman:
This was intended to be applied _only_ to the Collection `name` field because we need hyperlinks to be functioning on the UI; the `name` is used for the hyperlink.

It should *NOT* be applied to `description`.

[...]

PR #3154, opened 2022-08-22
fix(curation api):Check the collection body for none printable characters

Bento007:
- #3134 

### Reviewers

---

## Changes
- add check for in collection request body for none printable characters.


backend/corpora/lambdas/api/v1/collection.py
```diff
@@ -200,22 +200,24 @@ def _error_message(i: int, _url: str) -> dict:
 
 
 def verify_collection_body(body: dict, errors: list) -> None:
-    def check_if_blank(key) -> bool:
+    def check(key) -> bool:
         if key in body.keys():
             if not body[key]:
                 errors.append({"name": key, "reason": "Cannot be blank."})
+            elif re.search(r"[\x00-\x1f]", body[key]):
```
nayib-jose-gloria, line 207: I've seen some places include "\x7f" (DEL command) as a non-printable ascii character as well. Mind looking into it briefly to see if we should include in this regex? My instinct is yes

Bento007: I added a few more none printable characters to the regex

ebezzi (approved):

Issue #3162, opened 2022-08-23
get_collection failing with internal server error


brianraymor:
While testing processing_status for uploads, get_collection failed:

FAILED

Internal Server Error

{
  "detail": "None is not one of ['COUNT', 'NORMAL']\n\nFailed validating 'enum' in schema['allOf'][0]['properties']['datasets']['items']['allOf'][1]['properties']['x_approximate_distribution']:\n    {'description': 'Cellxgene runs a heuristic to detect the approximate '\n                    'distribution of the data in X so that it can '\n                    'accurately calculate statistical properties of the '\n                    'data. This field enables the curator to override this '\n                    'heuristic and specify the data distribution '\n                    'explicitly.',\n     'enum': ['COUNT', 'NORMAL'],\n     'type': 'string'}\n\nOn instance['datasets'][0]['x_approximate_distribution']:\n    None",
  "status": 500,
  "title": "Response body does not conform to specification",
  "type": "about:blank"
}

---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
Input In [13], in <cell line: 3>()
      1 from src.collection import get_collection
----> 3 collection = get_collection(new_collection_id)
      4 display(collection['processing_status'])
      5 display(collection)

File ~/Notebooks/curation_api/python/src/collection.py:72, in get_collection(collection_id)
     70 except requests.HTTPError as e:
     71     failure(logger, e)
---> 72     raise e
     73 return res.json()

File ~/Notebooks/curation_api/python/src/collection.py:69, in get_collection(collection_id)
     67 try:
     68     res = requests.get(url, headers=headers)
---> 69     res.raise_for_status()
     70 except requests.HTTPError as e:
     71     failure(logger, e)

File ~/.pyenv/versions/3.9.11/envs/env/lib/python3.9/site-packages/requests/models.py:960, in Response.raise_for_status(self)
    957     http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
    959 if http_error_msg:
--> 960     raise HTTPError(http_error_msg, response=self)

HTTPError: 500 Server Error: Internal Server Error for url: https://api.cellxgene.dev.single-cell.czi.technology/curation/v1/collections/8b5a0c61-a226-4bb7-9f1e-70ab3792966a

Issue #2983, opened 2022-07-27
Swagger API should correctly sort fields in schemas


brianraymor:
**This was reported very early in the design process and has not been addressed.** 

The schema should match the ordering in the JSON response. 

It's more difficult than it needs to be for users to understand the responses and reviewers to validate the responses. For example:

![Screen Shot 2022-07-27 at 12.51.49 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/85c6eae3-cffb-4697-b9e9-79643618a571)

brianraymor:
This is not done. For example, `datasets` are not sorted in properly:

![Screen Shot 2022-08-16 at 3.11.26 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/0b4fab86-c329-4746-923b-bc629b54d520)


And X_normalization and X_approximate_distribution are not capitalized in the JSON response or sorted in properly:

![Screen Shot 2022-08-16 at 3.16.14 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/9854a87a-21a5-4bbc-898c-8a757eb3917a)

brianraymor:
@Bento007 - I still see datasets sorted at the bottom in the dev swagger?

brianraymor:
LGTM @Bento007 !

PR #3163, opened 2022-08-23
fix: x_approximate_distribution can be None

Bento007:
- #2983 
- #3162

### Reviewers
**Functional:** @atarashansky 

---

## Changes
- modify allow x_approximate_distribution to be None. If the field is not provided by the user, we want to return it as None in the API request.

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -734,7 +734,7 @@ components:
               type: boolean
             x_approximate_distribution:
               "$ref": "#/components/schemas/distribution"
-            X_normalization:
+            x_normalization:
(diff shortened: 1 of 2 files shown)


backend/config/curation-api.yml
```diff
@@ -780,6 +780,7 @@ components:
         - COUNT
         - NORMAL
       type: string
+      nullable: true
```
atarashansky, line 783: I could only find `x_approximate_distribution`, not `distribution`, in `DbDataset` (am I looking in the right spot?)

atarashansky (approved):

Issue #3084, opened 2022-08-11
feat: Add extra dimension to integrated corpus, summary cubes, and snapshot of scExpression


pablo-gar:
To address the cube creation step of the [eng design ](simplify tissue dropdown) to simplify tissue dropdown of scExpression.

We need to:

In the dimensions of the integrated corpus and summary cubes:

- Rename the following of the tiledb objects:
   - Dimension: tissue_ontology_term_id --> tissue_original_ontology_term_id
   - Attr: tissue--> tissue_original 
- Add the following:
   - Dimension: tissue_ontology_term_id (for now a copy, but in the future high-level definition of tissue)
   - Attr: tissue (for now a copy, but in the future high-level definition of tissue)
- Modify the cube creation to account for added dimension and attribute
- Modify the snapshot and its querying to account for added dimension and attribute

metakuni:
Oops, I just noticed that the PR merged for this issue was merged to a feature branch, not main! Moving this back to `In Progress` until the feature branch is merged to main.

PR #3171, opened 2022-08-24
feat: Add extra dimension to integrated corpus, summary cubes, and snapshot of scExpression

pablo-gar:
Addresses #3084 

### Reviewers
**Functional:** 

@ebezzi (?) Let me know if I should choose someone else

**Readability:** 

@atarashansky 

---

## Changes
- renames integrated corpus and wmg cube **dimension** `tissue_ontology_term_id` to `tissue_original_ontology_term_id`
- renames integrated corpus and wmg cube **attribute** `tissue` to `tissue_original`
- the steps above results in an added dimension and attribute to integrated corpus and cubes 
- `tissue_ontology_term_id` and `tissue` will contain the high-level tissues. For now, only children of `lung` are mapped to `lung`, everything else is untouched, another PR will be submitted for mapping all tissues
- modify querying to snapshot to handle extra dimension and attribute and send correct information to frontend
- update unit tests of querying to account for extra dimension and attribute 

## QA

- Integrated corpus and cubes were manually created with small sets of data AND all data in prod. TileDB objects were inspected manually and everything looks good:
   - Proper addition of dimension `tissue_original_ontology_term_id`
   - Proper addition of attribute `tissue_original`
   - Children of lung were properly mapped to lung in`tissue_ontology_term_id` and `tissue`

- FE is working as tested in rdev -- only lung and none of its children are shown.

### DIFF
--- a/backend/corpus_asset_pipelines/integrated_corpus/transform.py
+++ b/backend/corpus_asset_pipelines/integrated_corpus/transform.py
@@ -5,6 +5,7 @@ import time
 import anndata
 import numpy
 import numpy as np
+from pandas import DataFrame
 import scanpy
 import tiledb
 from scipy import sparse
@@ -35,6 +36,32 @@ def apply_pre_concatenation_filters(
         anndata_object.obs["assay_ontology_term_id"].isin(included_assay_ontology_ids), :
     ].copy()
     return anndata_object
+
+
+def create_high_level_tissue(anndata_object: anndata.AnnData):
+    anndata_object.obs["tissue_original"] = anndata_object.obs["tissue"]
+    anndata_object.obs["tissue_original_ontology_term_id"] = anndata_object.obs["tissue_ontology_term_id"]
+    anndata_object.obs = get_high_level_tissue(anndata_object.obs)
+
+
+# TODO finalize this function
+def get_high_level_tissue(obs: DataFrame):
+
+    for i in range(len(obs)):
+
+        if "lung" in obs["tissue"][i]:
+            if "UBERON:0002048" not in obs["tissue_ontology_term_id"].cat.categories:
+                obs["tissue_ontology_term_id"].cat.add_categories("UBERON:0002048", inplace=True)
+                obs["tissue"].cat.add_categories("lung", inplace=True)
+
+            obs["tissue_ontology_term_id"][i] = "UBERON:0002048"
+            obs["tissue"][i] = "lung"
+            continue
+
+        obs["tissue_ontology_term_id"][i] = obs["tissue_original_ontology_term_id"][i]
+        obs["tissue"][i] = obs["tissue_original"][i]
+
+    return obs
 
 
 def transform_dataset_raw_counts_to_rankit(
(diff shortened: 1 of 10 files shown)


backend/corpus_asset_pipelines/integrated_corpus/transform.py
```diff
@@ -37,6 +38,32 @@ def apply_pre_concatenation_filters(
...
 
 
+def create_high_level_tissue(anndata_object: anndata.AnnData):
+    anndata_object.obs["tissue_original"] = anndata_object.obs["tissue"]
+    anndata_object.obs["tissue_original_ontology_term_id"] = anndata_object.obs["tissue_ontology_term_id"]
+    anndata_object.obs = get_high_level_tissue(anndata_object.obs)
+
+
+# TODO finalize this function
+def get_high_level_tissue(obs: DataFrame):
```
atarashansky, line 48: I think we should finalize this function in this PR, as I would prefer to avoid hardcoding the tissue rollup (even if it is temporary).

backend/corpus_asset_pipelines/integrated_corpus/transform.py
```diff
@@ -37,6 +38,32 @@ def apply_pre_concatenation_filters(
...
+
+    for i in range(len(obs)):
+
+        if "lung" in obs["tissue"][i]:
```
atarashansky, line 52: If the file that will eventually be used to map tissues to high-level tissues is of the following format:
`Row: (original tissue, high level tissue)`
then I would create a temporary file mapping original lung tissues to lung and use that here.

atarashansky (changes requested):

backend/corpus_asset_pipelines/integrated_corpus/transform.py
pablo-gar: I was planing to keep them separate for better and encapsulated review. 

That work is done in the branch pablo-gar/wmg-tissue-roll-up-tissueMapper

[](https://github.com/chanzuckerberg/single-cell-data-portal/blob/pablo-gar/wmg-tissue-roll-up-tissueMapper/backend/wmg/data/tissue_mapper.py)

[](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d708135988ed5da2614a66a881d41550ce506583/backend/corpus_asset_pipelines/integrated_corpus/transform.py#L45-L74)

backend/corpus_asset_pipelines/integrated_corpus/transform.py
pablo-gar: This is a placeholder will be replaced, see comment above

backend/wmg/data/query.py
```diff
@@ -7,13 +7,14 @@
 
 from backend.wmg.data.snapshot import WmgSnapshot
 
-EMPTY_DIM_VALUES = ""
-
+# ASK madison if there's a better way to do this
+INDEXED_DIMENSIONS_TO_IGNORE = ["tissue_original_ontology_term_ids"]
```
ebezzi, line 11: Looks like this isn't used anywhere?

pablo-gar: Oh yeah Madison suggested something else, will remove

Issue #3122, opened 2022-08-18
PUT ../upload-link must support HTTPS scheme for S3 URL(s)


brianraymor:
See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660838077260049?thread_ts=1660778105.668779&cid=C02UGUV5R3J). 

I've added a related note to [Update and complete all documentation of APIs in the OpenAPI spec](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2161) about user confusion over which links are supported.

Bento007:
I still need to update the opnepapi spec to document what URLs are supported

brianraymor:
@jahilton to re-test his case per the thread in the top-level comment.

brianraymor:
@jahilton reported some _odd UX behavior_ on [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1661374966291379) that needs review.

brianraymor:
Also regarding the swagger:

> Upload a Dataset. The Dataset will be uploaded from the provided link. MUST include a curator_tag in the request body; if an existing Dataset has this tag, the existing Dataset SHALL be replaced, otherwise a new Dataset will be added. MAY include the id of an existing Dataset, in which case the existing Dataset SHALL be replaced. Only presigned AWS S3 URLs and Dropbox shared file links are supported.

May I suggest a minor rewrite to move related information closer?

> Upload a Dataset. The Dataset will be uploaded from the provided link which must either be a presigned AWS S3 URLor Dropbox shared file. MUST include a curator_tag in the request body; if an existing Dataset has this tag, the existing Dataset SHALL be replaced, otherwise a new Dataset will be added. MAY include the id of an existing Dataset, in which case the existing Dataset SHALL be replaced.

Bento007:
I'll add this to the swagger docs update PR I'm working on.

brianraymor:
@jahilton's [seal of approval ](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1661466887273089?thread_ts=1661374966.291379&cid=C02UGUV5R3J)- _Then we're all good here_. Ready for production.

PR #3176, opened 2022-08-24
fix: document upload links that are suported

Bento007:
- #3122

### Reviewers

**Readability:** 

---

## Changes

- modify upload link documentation to include the types of links we currently support.

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -73,10 +73,6 @@ paths:
           schema:
             type: string
           examples:
-            current_curator:
-              summary: Return collections owned by current curator.
-              value:
-                name: "self"
             specified_curator:
               summary: Return collections owned by the specified curator
               value:
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -429,9 +429,8 @@ paths:
 
         To upload the files to S3, use the Python boto3 package. Once a file is successfully uploaded, it wil be
         processed and added to the Collection (specified in the S3 key path) with no further user action required.
-        Include an email address in the S3 file upload metadata to receive processing and error notifications.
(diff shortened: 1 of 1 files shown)


backend/config/curation-api.yml
```diff
@@ -355,6 +355,8 @@ paths:
         request body; if an existing Dataset has this tag, the existing Dataset SHALL be replaced, otherwise a new
         Dataset will be added. MAY include the **id** of an existing Dataset, in which case the existing Dataset
         SHALL be replaced.
+
+        Only presigned AWS S3 URLs and Dropbox shared file links are supported.
```
nayib-jose-gloria, line 359: question, non-blocking: do you think its worth noting here for users that they can get AWS S3 URLs using the curation API via the datasets/{dataset_id}/assets endpoint?

nayib-jose-gloria (commented):
let me know if you'd rather I create a separate issue for this, but I did spot another line I think we need to update in the swagger:

The endpoint reference [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/b5c19c6a5af8942188a05d171f66f27ac9ba7ce1/backend/config/curation-api.yml#L433) isn't part of the API anymore, right? processing_status should be monitored from the GET collections/{collections_id} endpoint right?

backend/config/curation-api.yml
Bento007: At the moment if they were to download one of those datasets and try uploading it, it would fail. We add information to the data that our processing pipeline doesn't expect. I don't think encouraging them to do that would be wise.

nayib-jose-gloria: ah right right, makes sense

nayib-jose-gloria (approved):

PR #3200, opened 2022-08-29
style: remove unused regex and consolidate curator tag logic

danieljhegeman:
- #3102 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- remove unused regex
- enforce non-empty curator tag in backend validator code
- do not prohibit uuid shape _within_ a curator tag; only prohibit uuid _as_ tag.

### DIFF
--- a/backend/corpora/common/utils/regex.py
+++ b/backend/corpora/common/utils/regex.py
@@ -4,23 +4,20 @@ USERNAME_REGEX = r"(?P<username>[\w\-\|]+)"
 ID_REGEX = r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
 DATASET_ID_REGEX = f"(?P<dataset_id>{ID_REGEX})"
 COLLECTION_ID_REGEX = f"(?P<collection_id>{ID_REGEX})"
-CURATOR_TAG_PREFIX_REGEX = r"(?P<tag_prefix>.*)"
 CONTROL_CHARS = r"[\x00-\x1f\x7f-\xa0]"
 CURATOR_TAG_REGEX = r"(?P<tag>.*)"
 
 
 def validate_curator_tag(curator_tag: str) -> bool:
     """
-    Verify the correct curator tag format is obeyed.
+    Verify the correct curator tag format is obeyed (i.e., it is not a UUID)
 
     :param curator_tag: the tag name to validate.
-    :return: True if CURATOR_TAG_PREFIX_REGEX is matched.
+    :return: True if CURATOR_TAG_REGEX is matched.
     """
     regex = f"^({DATASET_ID_REGEX}|{CURATOR_TAG_REGEX})$"
     matched = re.match(regex, curator_tag)
-    if matched and (tag := matched.groupdict().get("tag")):
-        if not re.search(ID_REGEX, tag):
-            return
-        else:
-            raise ValueError("Curator tag cannot contain the same shape as a UUID.")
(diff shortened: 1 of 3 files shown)


backend/corpora/common/utils/regex.py
```diff
@@ -4,23 +4,23 @@
 ID_REGEX = r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
 DATASET_ID_REGEX = f"(?P<dataset_id>{ID_REGEX})"
 COLLECTION_ID_REGEX = f"(?P<collection_id>{ID_REGEX})"
-CURATOR_TAG_PREFIX_REGEX = r"(?P<tag_prefix>.*)"
 CONTROL_CHARS = r"[\x00-\x1f\x7f-\xa0]"
-CURATOR_TAG_REGEX = r"(?P<tag>.*)"
+CURATOR_TAG_REGEX = r"(?P<tag>.+)"
```
danieljhegeman, line 8: no empty tags

backend/corpora/common/utils/regex.py
```diff
@@ -4,23 +4,23 @@
 ID_REGEX = r"[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
 DATASET_ID_REGEX = f"(?P<dataset_id>{ID_REGEX})"
 COLLECTION_ID_REGEX = f"(?P<collection_id>{ID_REGEX})"
-CURATOR_TAG_PREFIX_REGEX = r"(?P<tag_prefix>.*)"
```
danieljhegeman, line 7: unused

backend/corpora/common/utils/regex.py
```diff
@@ -4,23 +4,23 @@
...
+    Verify the correct curator tag format is obeyed (i.e., it is not a UUID)
 
     :param curator_tag: the tag name to validate.
-    :return: True if CURATOR_TAG_PREFIX_REGEX is matched.
+    :return: True if CURATOR_TAG_REGEX is matched.
     """
     regex = f"^({DATASET_ID_REGEX}|{CURATOR_TAG_REGEX})$"
     matched = re.match(regex, curator_tag)
-    if matched and (tag := matched.groupdict().get("tag")):
-        if not re.search(ID_REGEX, tag):
```
danieljhegeman, line 22: removing this because we are only in the business of making sure that we know when to treat an identifier as a uuid, and that is when the identifier takes the precise form of a uuid. Otherwise, if it is not exactly the form of a uuid, we treat it as a curator tag.

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -88,7 +88,7 @@ def _test(_tag, _dataset):
             self.assertIsNone(_dataset.curator_tag)
 
         dataset = self.generate_dataset(self.session, collection=collection)
-        tests = [dataset.id, "prefix" + dataset.id, dataset.id + "suffix", "prefix" + dataset.id + "suffix"]
+        tests = [dataset.id, ""]
```
danieljhegeman, line 91: (following on from comments above) `"prefix" + dataset.id` is therefore a valid tag

backend/corpora/common/utils/regex.py
Bento007: This requirement was explicitly in the ticket when I implemented it. I updated the ticket to reflect the actual requirements.

Issue #2438, opened 2022-04-27
All ingestion failures MUST be reported to #single-cell-ops


brianraymor:
Per @jahilton in [#single-cell-data-wrangling](https://czi-sci.slack.com/archives/C024HCSH9PT/p1651072467672089), curators were blocked when CZI failed to investigate and contact them per the message in the portal UX:

![Screen Shot 2022-04-27 at 12.10.54 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/ea568bba-d2b1-4b1a-8970-ad24dd0fc205)

Currently, only validation failures are reported to #single-cell-ops, but as noted by [Madison's comment ](https://docs.google.com/document/d/1NuqQuFvTOV4PQpi1RsbZhRDRPV6O4YV0vj-l5Tj2xGg/edit?disco=AAAATnreg2c) in the ingestion pipeline design document, ALL failures **MUST** be reported. Otherwise, _on calls_ will have no visibility into failures to investigate.

See also:
- https://docs.google.com/document/d/16xPhBQb1nOaAAMEfpr9sq4nz18DM0GuAIgUFQN6Zr20/edit?disco=AAAAcOrDTMM

metakuni:
[Example](https://czi-sci.slack.com/archives/C024HCSH9PT/p1661534310715979) of ingestion failure that didn't get reported to #single-cell-ops:
```
jahilton
 Today at 10:18
@ebezzi
 
@kuni
 Dataset failing in https://cellxgene.cziscience.com/collections/a3d5316d-f953-46dd-95b8-a4a11e22e175
h5ad & rds generated appears successful, but no cxg
```

metakuni:
Side note: We might want to remove 'and follow-up in an email' from the error message displayed by the FE since we don't actually do this.

cc: @brianraymor

nayib-jose-gloria:
Currently, we set processing_status to SUCCESS so long as the h5ad is validated and uploaded successfully, even if then there are subsequent conversion failures for CXG or RDS. Do we want to maintain that status quo? It will change how I approach the solution.

brianraymor:
@jahilton @danieljhegeman @Bento007 ...

For the curation API, I would prefer that `processing_status` returns FAILURE with a note that the problem must be addressed by CZI OPS. In general, conversions SHOULD NOT fail under ordinary circumstances. The validator should catch cases that would blow up conversions. Testing should catch any cases related to internal dependency updates.

Issue #2289, opened 2022-04-07
Dataset ingestion failure notifications should @sc-oncall-eng in #single-cell-ops


metakuni:
When dataset ingestion failures occur, the Slack notification message posted to #single-cell-ops should include `@sc-oncall-eng` so that the on-call engineers are explicitly pinged.

See:
* https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/corpora/dataset_processing/slack.py#L10

brianraymor:
@metakuni - how does this differ from [All ingestion failures MUST be reported to #single-cell-ops](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/2438)?

metakuni:
> @metakuni - how does this differ from All ingestion failures MUST be reported to #single-cell-ops?

This issue was to specifically to add the `@sc-oncall-eng` in the Slack notification so that the on-call engineer would be explicitly pinged on Slack. It's a trivial one-line change.

PR #3205, opened 2022-08-29
feat: fail overall dataset processing status if any file conversion step fails

nayib-jose-gloria:
- #2438
- #2289 

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- refactor 'convert_file_ignore_exceptions' to 'convert_file' and raise ConversionFailed exceptions
- in main process function, catch and handle ConversionFailed exceptions in the existing HandleErrors pipeline. 
- overall dataset processing_status will now be FAILURE if any conversion step fails. By extension, this will also raise the FAILURE status at the collection-level processing_status (which looks at dataset-level processing_status to determine value)
- Update HandleErrors step function to check which step error stemmed from, for new artifact clean-up behavior. cxg artifacts will only be deleted on cxg failures. h5ad + rds artifacts will only be deleted on download-validate failures. I propose that seurat failures will not yield artifact clean-up because there is a standalone seurat-process step that accounts for replacing existing rds artifacts. This will not mark a change in behavior, as we currently are not going through the HandleErrors step with artifact clean-up on seurat failures.
- Append an @sc-oncall-eng to slack ops error messages
- updated docstrings
- update tests

## QA steps (optional)
- went through upload scenarios in rdev: validation failure, cxg failure, upload success. Did not have example dataset available to test a seurat failure


backend/corpora/dataset_processing/process.py
```diff
@@ -92,23 +92,34 @@
     processing_status = ProcessingStatus.PENDING
     upload_status = UploadStatus.UPLOADED
     upload_progress = 1.0
-    validation_status = ValidationStatus
+    validation_status = ValidationStatus,VALID
```
Bento007, line 95:
```suggestion
    validation_status = ValidationStatus.VALID
```

backend/corpora/upload_failures/app.py
```diff
@@ -11,14 +12,19 @@
...
-    delete_many_from_s3(cellxgene_bucket, object_key)
     update_dataset_processing_status_to_failed(dataset_id, event["error"]["Cause"])
 
+    error_step_name = get_error_step_name(event)
+    object_key = os.path.join(os.environ.get("REMOTE_DEV_PREFIX", ""), dataset_id).strip("/")
+    # clean up artifacts depending on error
+    if error_step_name == "download-validate":
+        delete_many_from_s3(os.environ["ARTIFACT_BUCKET"], object_key)
+    if error_step_name == "cxg":
+        cellxgene_bucket = f"hosted-cellxgene-{os.environ['DEPLOYMENT_STAGE']}"
```
ebezzi, line 23: Not introduced by your PR, but I'd recommend an if/else here.

ebezzi (approved):

PR #3209, opened 2022-08-30
feat: Update cxg pipeline to create row- and column-optimized tileDB gene expression arrays and dictionary-encoded obs arrays

atarashansky:
- #3210 

 - The current layout of our gene expression tileDB array (X) has been optimized to compromise between fast row slicing and column slicing. However, column slicing for fetching genes is quite slow (~2 seconds for 10 genes in the 1.46m dataset). This PR updates data portal to generate CXGs with two 1D X* arrays, one for row-slicing (`Xr`) and one for column-slicing (`Xc`). The new CXG parameters were loosely optimized in a Jupyter notebook using the following approach. I kept all other parameters fixed, varied capacity by powers of 2, and picked the capacity that gave the fastest slicing for 10 random indices. Using the new capacity, I did the same for picking the number of tiles. With the new schema, column slicing (i.e. row slicing `Xc`) 10 random genes takes ~80ms. Note: CXGs will only have two 1D arrays if the original 2D expression array was sparse.
 - I also updated the obs array to store categorical data as integer codes. This offers two benefits to explorer: (1) integer arrays take up less memory and are faster to load (2) we can avoid having to cast untyped arrays of strings into pandas Categoricals as we already have the codes (for high-cardinality categories this is the computational bottleneck in the explorer API call). The list of categories in the schema type hints (ObsArray.meta['cxg_schema']) is ordered to correspond to the integer codes (e.g. for categories ["A","B","C"], 0 in the codes array corresponds to "A", 1 to "B", and so on).
 - The remastering pipeline has been adapted from its previous version to `v2` for handling the evolution of `X`, `obs`, and `cxg_group_metadata`.


atarashansky:
- [ ] Update categoricals to be dictionary encoded as tileDB arrays. Store JSON mapping in `meta` for that array.

scripts/cxg_admin_scripts/reprocess_datafile.py
```diff
@@ -85,6 +85,52 @@ def cxg_remaster(ctx):
...
+                dataset = Dataset.get(session, dataset_id=record.id)
+                artifacts = [a.s3_uri for a in dataset.artifacts if a.filetype == DatasetArtifactFileType.CXG]
+                if len(artifacts) > 0:
+                    cxg = artifacts[0]
+
+                    p = urlparse(cxg)
+                    bucket = p.hostname
+                    dataset_id = p.path.strip("/").strip(".cxg")
+
+                    if dataset_id == "2e5273bd-aa36-4478-8f6f-62fa0abcea43":
```
atolopko-czi, line 110: why?

scripts/cxg_admin.py
```diff
@@ -267,6 +267,11 @@ def cxg_remaster(ctx):
     """Cxg remaster"""
     reprocess_datafile.cxg_remaster(ctx)
 
+@cli.command()
+@click.pass_context
+def cxg_remaster_v2(ctx):
```
atolopko-czi, line 272: what's the diff between this cmd and cxg_remaster_v2.py script?

backend/corpora/dataset_processing/remaster_cxg_v2.py
```diff
@@ -0,0 +1,215 @@
...
+        {
+            "py.init_buffer_bytes": 4 * 1024**3,
+            "sm.tile_cache_size": 1 * 1024**3,
+        }
+    )
+
+    st = time.time()
+    with tiledb.scope_ctx(ctx):
+        with tiledb.open(f"{cxg}/{source_array}", "r") as old_X:
+            in_sparse = old_X.schema.sparse
```
atolopko-czi, line 116: `is_sparse`

backend/corpora/dataset_processing/remaster_cxg_v2.py
```diff
@@ -0,0 +1,215 @@
...
+                target_array,
+                compression,
+                X_extent,
+                old_X.schema,
+                cell_order,
+                tile_order,
+                capacity,
+            )
+            logger.info("created, starting to read...")
+            with tiledb.open(f"{cxg}/{target_array}", "w") as new_X_r:
```
atolopko-czi, line 131: Why not call the array `Xr`? Overloading `X` is potentially confusing and `Xr` makes it clear its intended for row-wise access. `X` would be exclusively for dense arrays.

backend/corpora/dataset_processing/remaster_cxg_v2.py
```diff
@@ -0,0 +1,215 @@
...
+                with tiledb.open(f"{cxg}/{target_array}c", "w") as new_X_c:
+                    i, chunk = 0, 200_000
+                    while i < old_X.shape[0]:
+                        logger.info(f"Chunk {i}")
+                        dat = old_X[i : i + chunk]
+                        logger.info("Got dat")
```
atolopko-czi, line 137: make `logger.debug` or rm?

scripts/cxg_remaster_v2.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+
+pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))  # noqa
+sys.path.insert(0, pkg_root)  # noqa
+
+from backend.corpora.dataset_processing.remaster_cxg_v2 import compute
+
+"""
+Re-master an existing "X" tiledb array in a CXG
+Creates two new arrays (Xr and Xc) with the same data but row- and column-optimized schemas
```
atolopko-czi, line 13: "X and Xc", right? Tho `Xr and Xc` seems better.

backend/corpora/dataset_processing/process.py
```diff
@@ -207,6 +207,10 @@ def main():
         elif step_name == "cxg_remaster":
             from backend.corpora.dataset_processing.remaster_cxg import process
 
+            process(dataset_id, os.environ["CELLXGENE_BUCKET"], dry_run=False)
+        elif step_name == "cxg_remaster_v2":
+            from backend.corpora.dataset_processing.remaster_cxg_v2 import process
```
atolopko-czi, line 212: do we still need to maintain `remaster_cxg(_v1)`, above?

scripts/cxg_remaster_v2.py
atarashansky: I forgot to update this docstring. I changed it to "X" and "Xc" because it makes backwards compatibility in explorer less verbose.

backend/corpora/dataset_processing/remaster_cxg_v2.py
atolopko-czi: Other naming possibilities:
X_{by_,}cell, X_{by_,}feature
X_{by_,}obs, X_{by_,}var

backend/corpora/dataset_processing/remaster_cxg_v2.py
atarashansky: This was adapted from `backend/corpora/dataset_processing/remaster_cxg.py` where they used `in_sparse` and `out_sparse`. Here, `out_sparse` always equals True so I deleted that variable.

backend/corpora/dataset_processing/remaster_cxg_v2.py
atarashansky: Overloading the `X` array with the 1D row array made it easy to achieve backwards compatibility in explorer (e.g. if the array is dense or has not been remastered yet)

backend/corpora/dataset_processing/process.py
atarashansky: probably not! i can remove

backend/corpora/dataset_processing/remaster_cxg_v2.py
atarashansky: In explorer, I added logic to appropriately handle cases where `X` is 1D and sparse, 2D and dense, or 2D and sparse (not remastered yet).

scripts/cxg_admin.py
atarashansky: I'll need to consult with @ebezzi about this!

scripts/cxg_admin_scripts/reprocess_datafile.py
atarashansky: I'll ask @ebezzi !

scripts/cxg_admin.py
atarashansky: I will run the remastering pipeline from a Jupyter notebook, so this stuff will likely not be used.

.happy/terraform/modules/sfn/main.tf
```diff
@@ -211,21 +211,21 @@ resource "aws_sfn_state_machine" "state_machine_seurat" {
...
-resource "aws_sfn_state_machine" "state_machine_cxg_remaster" {
-  name     = "dp-${var.deployment_stage}-${var.custom_stack_name}-cxg-remaster-sfn"
+resource "aws_sfn_state_machine" "state_machine_cxg_remaster_v2" {
+  name     = "dp-${var.deployment_stage}-${var.custom_stack_name}-cxg-remaster-v2-sfn"
   role_arn = var.role_arn
 
   definition = <<EOF
 {
-  "StartAt": "CxgRemaster",
+  "StartAt": "CxgRemasterV2",
```
nayib-jose-gloria, line 220: code clarity suggestion: if we're not keeping the previous CxgRemaster step, is there any reason to name this CxgRemasterV2? What do you think of keeping the CxgRemaster naming convention and just changing the implementation? It adds an extra layer of maintenance to version it, and some confusion since we're not really versioning it such that we could still easily use or route to 'v1'

backend/corpora/common/utils/cxg_constants.py
```diff
@@ -1,4 +1,4 @@
 class CxgConstants(object):
     # The CXG container version number.  Must be a semver string (major.minor.patch)
     # DO NOT UPDATE THIS WITHOUT ALSO UPDATING CXG SPECIFICATION.
-    CXG_VERSION = "0.2.0"
+    CXG_VERSION = "0.3.0"
```
nayib-jose-gloria, line 4: question: is this meant to coincide with [this cellxgene schema](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md)? If so, we're actually on 2.0.0 currently and all 3.0.0 changes are in a feature branch. Not sure where the disconnect originated with using 0.2 instead of 2.0, but please let me know if this is meant to refer to something else

backend/corpora/dataset_processing/remaster_cxg_v2.py
```diff
@@ -0,0 +1,311 @@
...
+        pass
+
+
+def increment_version(cxg):
+    """
+    Increments the version number of the CXG
+    :param cxg: the path to the CXG
+    """
+    with tiledb.open(f"{cxg}/cxg_group_metadata", "w") as X:
+        X.meta["cxg_version"] = "0.3.0"
```
nayib-jose-gloria, line 125: nit: should use the CXG_VERSION constant here rather than hard-coding the version check

backend/corpora/common/utils/cxg_constants.py
atarashansky: I'm not 100% sure, but I think there's the schema version and the CXG version, which are different.

atarashansky:
Thanks, @nayib-jose-gloria ! I addressed your comments in the latest commits.

nayib-jose-gloria:
> Thanks, @nayib-jose-gloria ! I addressed your comments in the latest commits.

thanks! It looks like there's a few spots (happy/terraform step config, tests, scripts) still using the v2 notation

backend/corpora/common/utils/cxg_constants.py
nayib-jose-gloria: Minor, and not related to this PR but I did notice that you updated the cxg spec doc but I think we should also update the current version heading at the top from 0.2.0 -> 0.3.0 (+ the date) for clarity: https://github.com/chanzuckerberg/single-cell-explorer/blob/main/dev_docs/cxg.md#cxg-data-format-specification

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -27,48 +32,151 @@ def process(dataset_id: str, cellxgene_bucket: str, prefix=None, dry_run=True):
...
+        )
+        with tiledb.open(f"{cxg}/new_obs", "w") as new_X:
+            new_X[:] = new_data
+            new_X.meta["cxg_schema"] = json.dumps(schema)
+        tiledb.consolidate(f"{cxg}/new_obs")
+
+
+def evolve_X(**kwargs):
+    """
+    Computes the evolved cxg from the sparse source_array and saves it to the target_array destination
```
nayib-jose-gloria, line 179: nit: mind adding the return type to the docstring + explain it briefly in the description?

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -131,103 +227,59 @@ def compute(**kwargs):
...
-            if 1.0 - nnz / size >= sparse_threshold / 100.0:
-                return "sparse"
-
-            return "dense"
-
-
-def create_new_X(cxg, target_array, kind, compression, X_extent, old_schema, cell_order, tile_order, capacity):
-    is_sparse = kind == "sparse"
-    logger.info(f"create_new_X: {is_sparse}")
+def create_new_X(cxg, target_array, compression, X_extent, old_schema, cell_order, tile_order, capacity):
```
nayib-jose-gloria, line 250: suggestion: this didn't originate with your PR, so this is non-blocking but it would be helpful to add docstrings to the functions in this file that are missing them

tests/unit/processing_container/test_process.py
```diff
@@ -94,7 +96,7 @@ def test_main(self, mock_rematser_cxg_process, mock_download_from_source_uri):
         with EnvironmentSetup(test_environment):
             main()
 
-        test_environment["STEP_NAME"] = "cxg_remaster"
+        test_environment["STEP_NAME"] = "cxg_remaster_v2"
         with EnvironmentSetup(test_environment):
             main()
             mock_rematser_cxg_process.assert_called()
```
nayib-jose-gloria, line 102: nit: not your typo, but we should fix while we're here (rematser -> remaster)

tests/unit/processing_container/test_process.py
```diff
@@ -64,7 +64,9 @@ def clean_generated_files(cls):
                 os.remove(f)
 
     @patch("backend.corpora.dataset_processing.process_download_validate.download_from_source_uri")
-    @patch("backend.corpora.dataset_processing.remaster_cxg.process")  # TODO: provide test data to properly test this.
+    @patch(
+        "backend.corpora.dataset_processing.remaster_cxg.process"
+    )  # TODO: provide test data to properly test this.
     def test_main(self, mock_rematser_cxg_process, mock_download_from_source_uri):
```
nayib-jose-gloria, line 70: nit: not your typo, but we should fix while we're here (rematser -> remaster)

tests/unit/processing_container/test_process.py
```diff
@@ -94,7 +96,7 @@ def test_main(self, mock_rematser_cxg_process, mock_download_from_source_uri):
         with EnvironmentSetup(test_environment):
             main()
 
-        test_environment["STEP_NAME"] = "cxg_remaster"
+        test_environment["STEP_NAME"] = "cxg_remaster_v2"
```
nayib-jose-gloria, line 99: update to just cxg_remaster

.happy/terraform/modules/sfn/main.tf
nayib-jose-gloria: still need to revert this back to cxg_remaster

backend/corpora/common/utils/cxg_constants.py
atolopko-czi: > I'm not 100% sure, but I think there's the schema version and the CXG version, which are different.
correct, they're different

atarashansky:
> > Thanks, @nayib-jose-gloria ! I addressed your comments in the latest commits.
> 
> thanks! It looks like there's a few spots (happy/terraform step config, tests, scripts) still using the v2 notation

Whoops, thanks for catching!

backend/corpora/dataset_processing/remaster_cxg.py
```diff
@@ -27,48 +32,151 @@ def process(dataset_id: str, cellxgene_bucket: str, prefix=None, dry_run=True):
...
+    download_command = ["aws", "s3", "sync", obs_path, f"{local_path}/old_obs", "--quiet"]
+    # Let errors fail the pipeline
+    subprocess.run(download_command, check=True)
 
-    if not dry_run:
-        upload_command = ["aws", "s3", "sync", "--delete", f"{local_path}/X_new", path]
-        subprocess.run(upload_command, check=True)
+    logger.info(f"Downloading {meta_path} to {local_path}/cxg_group_metadata")
+    download_command = ["aws", "s3", "sync", meta_path, f"{local_path}/cxg_group_metadata", "--quiet"]
+    # Let errors fail the pipeline
```
nayib-jose-gloria, line 52: currently, on errors, we route all failures to a HandleErrors lambda which will update the processing status to FAILED. In this case, we don't want to do that. Let's wrap a 'try-catch' around remaster_cxg.process in the main function in process.py, in which we just log the error w/o re-raising (or just `pass` if logging isn't needed for the use case) so that errors don't get routed to HandleErrors

nayib-jose-gloria (approved):
looks good! Feel free to merge once tests pass. I'm not the biggest expert for reviewing the actual remastering logic but looks good from my pov and testing will catch remaining issues, if any

Issue #3208, opened 2022-08-30
Allow reference to revision Datasets via `original_id`


danieljhegeman:
See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1661819222477019). 

When a revision is created, not only does the revision Collection have a different `id` than the published Collection, but each Dataset in the revision has a different `id` than each published Dataset (with revision Datasets tracking their published counterparts through `dataset.original_id`.

Revision Datasets should be specifiable by both their `id` and their `original_id`. Otherwise, it is much more difficult for users to make meaningful use of the dataset id through the revision iteration process.

Nothing should change for the handling of `curator_tag`. Tags are copied from published Dataset to revision Dataset as-is. If the tag gets updated, which would only ever happen by direct action taken by the curator, then it's on the curator to use the new tag to specify that revision Dataset going forward during that revision rather than the original/old tag for that Dataset.

## Requirements
- Update `GET collection/{collection_id}` to return `revision_of` for each dataset.
- -Update `GET collection/{collection_id}/datasets/` to return `revision_of` for each dataset.
- `revison_of` in a dataset is the dataset_id of the published dataset. This field is only populated when retrieving a collection by is ID or a dataset by its identifier. Otherwise, the field is None.
- `revision_of` for a dataset is *not* returned when listing collections from `GET collections/` 
_**\*\*\*SEE FIRST COMMENT BELOW\*\*\***_

danieljhegeman:
Ok, so, the one problem with actually _allowing_ this reference-by-proxy thing as described above is that it opens Pandora's box when it comes to how all revision entities might be treated, aka "why not allow a curator to refer to a revision Collection with it's public id when making edits?". It would be sort of unclear where we would draw the line.

I am leaning toward adding `original_id` to the `dataset` schema that is returned by `GET /collections/{collection_id}`. From there, curators can programmatically grab the revision `id` by looking for the dataset with `original_id` == <published dataset id>.

With the convenience methods, it's quite trivial:
```
published_dataset_id = <some uuid>
collection = get_collection(<collection_id>)
for dataset in collection["datasets"]:
    if dataset["original_id"] == published_dataset_id:
        revision_dataset_id = dataset["id"]
        ...
```

Thoughts, @Bento007 and @brianraymor and @jahilton ?

jahilton:
>why not allow a curator to refer to a revision Collection with it's public id when making edits

[...]

PR #3215, opened 2022-08-31
feat: Curation API - return revision_of attribute for Datasets

danieljhegeman:
This commit also streamlines logic for prior code additions that are
also involved in conditionally adding attributes to the Dataset response
object during reshaping

- #3208 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -721,6 +721,10 @@ components:
             revision:
               nullable: true
               type: integer
+            revision_of:
+              description: the id of the published version of this Dataset
+              nullable: true
+              type: string
             schema_version:
               nullable: true
               type: string
--- a/backend/corpora/lambdas/api/v1/curation/collections/common.py
+++ b/backend/corpora/lambdas/api/v1/curation/collections/common.py
@@ -105,10 +105,12 @@ def reshape_dataset_for_curation_api(dataset: dict, preview=False) -> dict:
         else:
             dataset[ontology_element] = []
 
-    if value := dataset.pop("name", None):
-        dataset["title"] = value
-    if value := dataset.pop("is_primary_data", None):
-        dataset["is_primary_data"] = is_primary_data_mapping.get(value)
+    if not preview:  # Add these fields only to full (and not preview) Dataset metadata response
+        dataset["revision_of"] = dataset.pop("original_id", None)
+        dataset["title"] = dataset.pop("name", None)
+        if value := dataset.pop("is_primary_data", None):
+            dataset["is_primary_data"] = is_primary_data_mapping.get(value, [])
+
     return dataset
 
 
@@ -160,6 +162,7 @@ class EntityColumns:
 
     dataset_cols = [
         *dataset_preview_cols,
+        "original_id",
         "name",
         "revision",
         "revised_at",
--- a/tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
+++ b/tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
@@ -366,6 +366,7 @@ class TestGetCollectionID(BaseAuthAPITest):
                 "processing_status": "PENDING",
                 "revised_at": None,
                 "revision": 0,
+                "revision_of": None,
                 "schema_version": "2.0.0",
                 "sex": [
                     {"label": "test_sex", "ontology_term_id": "test_obo"},


backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -105,10 +105,12 @@ def reshape_dataset_for_curation_api(dataset: dict, preview=False) -> dict:
...
-    if value := dataset.pop("name", None):
-        dataset["title"] = value
-    if value := dataset.pop("is_primary_data", None):
-        dataset["is_primary_data"] = is_primary_data_mapping.get(value)
+    if not preview:  # Add these fields only to full (and not preview) Dataset metadata response
+        dataset["revision_of"] = dataset.pop("original_id", None)
+        dataset["title"] = dataset.pop("name", None)
+        if value := dataset.pop("is_primary_data", None):
+            dataset["is_primary_data"] = is_primary_data_mapping.get(value, [])
+
```
danieljhegeman, line 113: @Bento007 pls approve of the restructuring of this entire portion. Tests are passing but I want to make sure I'm not unintentionally screwing this up.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -144,24 +144,42 @@ def test__get_collections_with_auth__OK(self):
         self.assertEqual(200, res_auth.status_code)
         self.assertEqual(6, len(res_auth.json["collections"]))
         with self.subTest("The 'revising_in' attribute is None for unauthorized public collections"):
+            conditions_tested = 0
```
Bento007, line 147: rather than doing this condition thing, make each condition a subtest.

Bento007: Also this test is too long, please break it up into small tests.

backend/corpora/lambdas/api/v1/curation/collections/common.py
```diff
@@ -105,10 +105,12 @@ def reshape_dataset_for_curation_api(dataset: dict, preview=False) -> dict:
         else:
             dataset[ontology_element] = []
 
-    if value := dataset.pop("name", None):
-        dataset["title"] = value
-    if value := dataset.pop("is_primary_data", None):
-        dataset["is_primary_data"] = is_primary_data_mapping.get(value)
+    if not preview:  # Add these fields only to full (and not preview) Dataset metadata response
+        dataset["revision_of"] = dataset.pop("original_id", None)
```
Bento007, line 109: Why only in the preview?

backend/corpora/lambdas/api/v1/curation/collections/common.py
Bento007: This means that if `name` or `origin_id` are not populated in the database, we will return `null` for those fields.

Bento007 (changes requested):

backend/corpora/lambdas/api/v1/curation/collections/common.py
danieljhegeman: you mean why _not_ in the preview...?

None of these attributes are supposed to be returned on the preview response object.

backend/corpora/lambdas/api/v1/curation/collections/common.py
danieljhegeman: correct, which is what we're supposed to do

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: >rather than doing this condition thing, make each condition a subtest.

That would be a lot of code duplication. Also I'm not seeing how we would do this without the `conditions_tested` flag. You need that to ensure that you actually run the `assert` that is doing the testing since it is nested, necessariyl, into a conditional.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -144,24 +144,42 @@ def test__get_collections_with_auth__OK(self):
...
                     self.assertIsNone(c["revising_in"])
+            self.assertEqual(3, conditions_tested)
         with self.subTest("The 'revising_in' attribute is equal to the id of the revision Collection"):
+            conditions_tested = 0
             for c in res_auth.json["collections"]:
                 if c["id"] == "test_collection_id":
+                    conditions_tested += 1
                     self.assertEqual("test_collection_id_revision", c["revising_in"])
+            self.assertTrue(1, conditions_tested)
+        with self.subTest(""):
```
Bento007, line 174: did you forget to add a description?

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
```diff
@@ -144,24 +144,42 @@ def test__get_collections_with_auth__OK(self):
         self.assertEqual(200, res_auth.status_code)
         self.assertEqual(6, len(res_auth.json["collections"]))
```
Bento007, line 145: These two checks above are not needed since we are testing that in the test above.

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
Bento007: I see why you're using `conditions_tested`, but it's still an odd pattern. I also don't like seeing the test iterating through the whole list every time, but this is not intended to be performant code so I'll let that go.  Breaking it up would not be much code duplication, but I don't see a compelling reason to do that at this time.

Carry on

backend/corpora/lambdas/api/v1/curation/collections/common.py
Bento007: We include revision_of in the collection portion. If a response returns information about the collection `revision_of`, it makes sense to also return information about the dataset `revision_of`

tests/unit/backend/corpora/api_server/curator/collection/test_curation_collections.py
danieljhegeman: 😆 maybe

backend/config/curation-api.yml
```diff
@@ -646,12 +646,13 @@ components:
...
+            Processing status for Datasets in a Collection
 
-            - **"FAILURE":** one or more Datasets ran into an error during processing. The CELLxGENE team will reach out to help you.
-            - **"PENDING":** one or more Datasets are still being processed.
-            - **"SUCCESS":** all Datasets have been processed successfully.
-            - **null:** the Collection has no Datasets.
+            * `"FAILURE"`: one or more Datasets ran into an error during processing. The CELLxGENE team will reach out to help you.
+            * `"PENDING"`: one or more Datasets are still being processed.
+            * `"SUCCESS"`: all Datasets have been processed successfully.
+            * `null`: the Collection has no Datasets.
```
danieljhegeman, line 655: just a stylistic change to improve accuracy/readability

Bento007: double check that it works. I kept it as it was because it was having trouble rendering, because of how it parses the `:`

Bento007 (approved):

Issue #3242, opened 2022-09-07
Single 400 schema differs from multiple 400 schema


danieljhegeman:
[Schema for a single 400 error](https://github.com/chanzuckerberg/single-cell-data-portal/blob/fc0b18d8a19998cbd9e97e717c4594c3341659e4/backend/config/curation-api.yml#L854-L863)

- detail
- title
- type

[Schema for a multiple 400 error](https://github.com/chanzuckerberg/single-cell-data-portal/blob/fc0b18d8a19998cbd9e97e717c4594c3341659e4/backend/config/curation-api.yml#L976-L992)

- name
- reason
- value

I would think these should be identical... @Bento007 you committed both of these sections. Is there context I am missing on these?

Bento007:
Change it to 
```
{ 
"details": str,
"title": str,
"type": str,
"invalid-param": {
 "name": str,
 "reason": str,
 "value": str
```

Bento007:
Trying to follow the spirit of this https://www.rfc-editor.org/rfc/rfc7807 the best we can.

danieljhegeman:
I'm not clear on whether your answer is intended to apply to one or both of '400' and '400 multiple'.

brianraymor:
Does this mean that we expect to fill in the [template](https://www.rfc-editor.org/rfc/rfc7807#section-3.1)? 

 >  This document defines a "problem detail" as a way to carry machine-
   readable details of errors in a HTTP response to avoid the need to
   define new error response formats for HTTP APIs.

This RFC is focused on a specific problem/solution.

Bento007:
We will be doing the latter @brianraymor.

Bento007:
@danieljhegeman  If we can consolidate it into a single error that'd be great.

Bento007:
keep in mind that the UI also uses these error messages. We might need to update the frontend to match that new error shape, or reshape the errors for the curator API specifically.

Bento007:
This does not seem like a P0. What is the motivation for this issue?

danieljhegeman:
@Bento007 we have been treating all issues with documentation--and especially error messages--as p0

PR #3255, opened 2022-09-08
fix(error): return 400 parameter error in invalid_parameter field

Bento007:
- #3242

### Reviewers

---

## Changes
- update openapi to include invalid_parameters in 400 responses. This is an optional field and is only populated when multiple parameter errors are found
- update endpoints that returned multiple error messages to match the expected error message format.
- update tests to check for the correct error message shape.


backend/corpora/lambdas/api/v1/curation/collections/actions.py
```diff
@@ -40,3 +41,12 @@ def get(visibility: str, token_info: dict, curator: str = None):
         resp_collections.append(resp_collection)
 
     return jsonify(resp_collections)
+
+
+def post(body: dict, user: str):
+    try:
+        return create_collection(body, user)
+    except InvalidParametersHTTPException as ex:
+        ex.ext = dict(invalid_parameters=ex.detail)
```
danieljhegeman, line 50: I'm a little confused. The `ex` instance of an `InvalidParametersHTTPException` is already going to have a `.detail` attribute on it, no? So we're putting the `errors` array on both `ex.detail` and `ex.ext`??

danieljhegeman (commented):
See my question regarding the assignment to the `.detail` vs `.ext` attributes on the `InvalidParametersHTTPException` instance

backend/corpora/lambdas/api/v1/curation/collections/actions.py
Bento007: We are putting  `InvalidParametersHTTPException._detail` in `ex.detail`. See the next line down.

danieljhegeman: Shouldn't the next line down be `ex._detail`?

Or is this `_detail` attribute a _class_ attribute/detail?

Bento007: it's a class attribute, which is why is why put the `_` prefix. I'll make it`__` since it's a private class attribute.

Bento007: actually, the convention is still to a single under-score prefix for private class variables. but I updated the variable name to make it more obvious what its purpose is./

danieljhegeman (approved):

Issue #3153, opened 2022-08-22
Deprecate curator_tag


brianraymor:
@jahilton [remarked earlier](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1652979065526479?thread_ts=1652977958.567249&cid=C02UGUV5R3J):

> I anticipate using the dataset id primarily, rather than the curator tag
So an elegant way to avoid a curator_tag would be ideal

and more [recently](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660838317051469):

> How would I go about uploading a new dataset without the curator tag and just have it auto-assigned as the dataset id?

danieljhegeman:
Woa. I did not see this issue before.

I am totally opposed to this without further discussion/explanation.

danieljhegeman:
>How would I go about uploading a new dataset without the curator tag and just have it auto-assigned as the dataset id?

If he doesn't want to use the id, then he can just upload to an arbitrary tag `1` or `2` or `redfish` or `blue` or... and then forget the tag (and he can retrieve the dataset id and use that going forward).

**There is no reason to deprecate the `curator_tag`, and there are many reasons not to.**

brianraymor:
> I am totally opposed to this without further discussion/explanation.

Which is why this is a _future_ - we're arm wresting in this issue.

danieljhegeman:
I see so a _future_ is a discussion item for later; it does not necessarily entail that it is _to be done in the future_?

danieljhegeman:
Having issues like this baking in the background without any context as to the level of certainty and/or reasoning for the proposed work isn't totally ideal IMO. It seems like it would lead to confusion if someone peruses 'future' work; they might make a design decision thinking that this was actually work that was desired.

But perhaps this is a tried-and-true tracking convention that I am just not used to.

brianraymor:
@jahilton, @danieljhegeman, @Bento007, @brianraymor met and agreed to deprecate `curator_tag` and its dependencies prior to release. 

@Bento007 - can you add your related technical proposal?

Bento007:
Curator tags are presently required to upload a dataset directly to S3. This is to give curators a handle that they can use to query the status of their datasets before an actual `dataset_id` has been generated. We will eliminate this need by adding an endpoint to generate a `dataset_id`

## POST /collections/{collection_id}/datasets
[...]

PR #3259, opened 2022-09-12
feat(curation api): dataset path parameter.

Bento007:
- #3153 

### Reviewers

**Readability:** 

---

## Changes
- pass dataset id as a path parameter
- move


backend/config/curation-api.yml
```diff
@@ -288,13 +288,41 @@ paths:
           $ref: "#/components/responses/401"
         "403":
           $ref: "#/components/responses/403"
-
   /v1/collections/{collection_id}/datasets:
+    post:
+      summary: Create a Dataset ID to be used in a future upload.
+      description: |
+        Create an new empty Dataset and return the `dataset_id`. The `dataset_id` is used when uploading a Dataset file 
+        to S3 to identify what Dataset the file belongs to.
```
danieljhegeman, line 296: Replace this with the verbiage that was in this section when it was further below and was introduced by the curator tag removal PR 👇

backend/config/curation-api.yml
```diff
@@ -347,34 +375,6 @@ paths:
           $ref: "#/components/responses/400"
         "404":
           $ref: "#/components/responses/404"
-    post:
-      summary: Create an empty Dataset and return the `id` to be used in a future upload.
-      description: |
-        Create a new empty Dataset and return the `id`. The `id` is used when uploading a datafile 
-        to S3 to identify what Dataset the file belongs to.
```
danieljhegeman, line 354: this 👆

backend/config/curation-api.yml
```diff
@@ -427,7 +427,7 @@ paths:
         "413":
           $ref: "#/components/responses/413"
 
-  /v1/collections/{collection_id}/datasets/s3-upload-credentials:
+  /v1/collections/{collection_id}/s3-upload-credentials:
```
danieljhegeman, line 430: hm... I don't think I agree. You're getting upload credentials that are specific  to a _Collection_, and not to a _Dataset_.

danieljhegeman: Although I guess we aren't specifying a dataset in the path. Hm... ok I suppose this is alright.

backend/config/curation-api.yml
danieljhegeman: actually not just the `description`. Everything needs to be copied over, including the `dataset_id` -> `id` change.

backend/config/curation-api.yml
```diff
@@ -347,34 +375,6 @@ paths:
...
-                type: object
-                properties:
-                  id:
-                    $ref: "#/components/schemas/dataset_id"
-        "401":
-          $ref: "#/components/responses/401"
-        "403":
-          $ref: "#/components/responses/403"
-        "405":
-          $ref: "#/components/responses/405"
```
danieljhegeman, line 377: reinstate this, but further up

backend/config/curation-api.yml
```diff
@@ -1023,11 +1023,11 @@ components:
...
       schema:
         $ref: "#/components/schemas/collection_id"
-    query_dataset_id:
+    path_dataset_id:
       description: The uuid of the Dataset.
-      in: query
+      in: path
       name: dataset_id
-      required: false
+      required: true
```
danieljhegeman, line 1030: Since everything is a uuid now, I think we should consider using the [pattern validator](https://swagger.io/docs/specification/data-models/data-types/#pattern), for both collection_id and dataset_id

tests/unit/backend/corpora/api_server/curator/collection/test_dataset.py
```diff
@@ -6,56 +6,45 @@
...
+                collection = self.generate_collection(self.session, visibility=CollectionVisibility.PRIVATE.name)
+                dataset = self.generate_dataset(
+                    self.session,
+                    collection=collection,
+                    processing_status=processing_status,
+                )
+                test_url = f"/curation/v1/collections/{collection.id}/datasets/{dataset.id}"
+                headers = auth() if callable(auth) else auth
+                response = self.app.delete(test_url, headers=headers)
+                self.assertEqual(expected_status_code, response.status_code)
```
danieljhegeman, line 26: 👍

backend/config/curation-api.yml
Bento007: connexion doesn't support this, so there will be 0 enforcement unless we implement it. Also, this value is never generated by the user, so there is no reason or them to know the format. Lastly, if we validate the shape of every id, then we'll have to update a bunch of tests.

danieljhegeman: Hm, I don't agree. They could think a given dataset simply doesn't exist, when in reality they entered the value wrong or accidentally truncated it with a keystroke when they weren't paying attention. It'd be useful for them to know that the id they entered isn't actually a uuid in that situation. Admittedly it's something we could handle in server code (a more detailed error message if the id they provide isn't actually a uuid). But yes, I agree it would incur extra work in the form of modifying all test id's that we use. But we don't have to do this right now, so I'm okay with ignoring it.

I didn't know connexion doesn't support the built-in `pattern` keyword though 😢

danieljhegeman (approved):
👍

Issue #3272, opened 2022-09-13
feat: Indicate to a higher precision the number of cells on Gene Expression's cell count column


pablo-gar:
## Need

Currently the cell counts cap at 10K and anything greater than that will be shown as ">10K".

The feature needs to show higher precision

<img width="305" alt="image" src="https://user-images.githubusercontent.com/29237819/190012691-31ba152f-7345-4d69-957f-9b0be68129de.png">

## Approach

Show exact amount in thousands:

15940 --> 15K
95600 --> 95K
125930 --> 125K

PR #3273, opened 2022-09-13
chore: Formatted cell counts with greater precision

atarashansky:
- #3272

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -390,7 +390,11 @@ export function createYAxisOptions({
             const { total_count } = deserializeCellTypeMetadata(
               value as CellTypeMetadata
             );
-            return total_count > 10000 ? ">10k" : `${total_count}`;
+            let formattedCount = `${total_count}`;
+            if (total_count > 10000) {
+              formattedCount = `${Math.round(total_count / 1000)}k`;
+            }
+            return formattedCount;
```
seve, line 397: const formattedString = Intl.NumberFormat("en-US", {
      maximumFractionDigits: 1,
      notation: "compact",
    }).format(n);
    return `${formattedString}${formattedString !== n.toString() ? "+" : ""}`;
    
    for when we hit 1M of a single cell type 🚀, will also give us  `x.xK`

seve (changes requested):

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
atarashansky: Oh right, will fix!

atolopko-czi (approved):
LGTM, with minor Q's:
* Test update needed?
* Is it ok if values [1000..10000) are abbreviated with `k` suffix?

seve (approved):

atarashansky:
> LGTM, with minor Q's:
> 
> * Test update needed?
I couldn't find tests for the cell count in the tooltip. I'm not sure if they exist.
> * Is it ok if values [1000..10000) are abbreviated with `k` suffix?
The point of the suffix initially was so that the counts do not occupy too much horizontal space. Any number below 10,000 is okay to display without abbreviation.

Issue #3277, opened 2022-09-14
s3-upload-credentials endpoint needs improved description


danieljhegeman:
For someone _not_ using the notebooks, the current description lacks sufficient clarity and specificity for performing a local upload.

PR #3278, opened 2022-09-14
docs: improve description for s3 upload protocol

danieljhegeman:
This commit also changes the example code for using bearer tokens to be prefaced with "Bearer" instead of "bearer"

- #3277 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- improve description for s3-upload-credentials endpoint to allow a user to comfortably perform a local datafile upload without having to reference code or instructions in our notebooks.
- change `bearer` to `Bearer`

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -469,7 +469,7 @@ paths:
 
         Once a file is successfully uploaded, it wil be processed and added to the Collection that you specified in your
         initial request to this endpoint, with no further user action required. Use
-        `GET /curation/collections/{collection_id}/datasets/` with the `id` query parameter to retrieve metadata for the
+        `GET /curation/collections/{collection_id}/datasets/{dataset_id}` to retrieve metadata for the
         Dataset in question in order to check on its processing status.
 
         A Dataset uploaded directly to S3 in this manner will appear in the Collection only once the Dataset has started
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -61,12 +61,12 @@ paths:
 
         When the visibility parameter is set to PRIVATE, a list of all private Collections or private revisions of 
         public Collections that the user is authorized to access is returned. An access token MUST be set in the request
-        header as: `Authentication: bearer <access_token>`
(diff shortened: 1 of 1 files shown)


backend/config/curation-api.yml
```diff
@@ -61,12 +61,12 @@ paths:
 
         When the visibility parameter is set to PRIVATE, a list of all private Collections or private revisions of 
         public Collections that the user is authorized to access is returned. An access token MUST be set in the request
-        header as: `Authentication: bearer <access_token>`
+        header as: `Authentication: Bearer <access_token>`
```
danieljhegeman, line 64: @Bento007 instances I'm seeing online all use `Bearer`... not that it really matters...? It's what I've been using in my testing.

danieljhegeman: although I recognize that we're just splicing all of that out anyways... unsure if it could have functional implications at some point though

Bento007: I don't think it will have any functional implications in our stack, but using `Bearer` could save us from weird bugs in the future if we change our stack.

backend/config/curation-api.yml
```diff
@@ -453,15 +453,29 @@ paths:
...
 
-        A Dataset uploaded directly to S3 will appear in the Collection only once the Dataset has started
+        * the object or key: `<UploadKeyPrefix>/<dataset_id>` (note: you must include the joining slash `/`)
+
+        * the bucket: `<Bucket>`
+
+
+        Once a file is successfully uploaded, it wil be processed and added to the Collection that you specified in your
+        initial request to this endpoint, with no further user action required. Use
+        `GET /curation/collections/{collection_id}/datasets/` with the `id` query parameter to retrieve metadata for the
```
Bento007, line 472: This endpoint no longer exists.

Bento007, line 472:
```suggestion
        `GET /curation/collections/{collection_id}/datasets/{dataset_id}` to retrieve metadata for the
```

Bento007 (approved):

PR #3280, opened 2022-09-14
feat: added cell type ancestors. #2933.

MillenniumFalconMechanic:
- #2933

### Reviewers
**Functional:** 
@pablo-gar

---

## Changes
- Added `cell_type_ancestors` to `datasets/index`.


backend/corpora/common/utils/tissue_ontology_mapping.py
```diff
@@ -1,5 +1,5 @@
 # Generated by
-# https://github.com/chanzuckerberg/single-cell-curation/blob/main/notebooks/compute_tissue_ancestor_mapping.ipynb
+# https://github.com/chanzuckerberg/single-cell-curation/blob/main/notebooks/compute_tissue_and_cell_type_mappings.ipynb
```
MillenniumFalconMechanic, line 2: I have consolidated tissue and cell type notebooks; corresponding PR will be ready shortly.

tests/unit/backend/corpora/api_server/test_v1_dataset.py
```diff
@@ -212,6 +212,34 @@ def test__enrich_tissue_with_ancestors_missing_key_ok(self):
...
+                "CL:0000988",
+                "CL:0000738",
+                "CL:0000548",
+                "CL:0000219",
+                "CL:0000003",
+                "CL:0002242",
+            ],
+        )
+
+    def test__enrich_cell_type_with_ancestors_empty_key_ok(self):
```
pablo-gar, line 233: curious, will we ever have this case? And if so, shouldn't it be an error since no datasets should contain empty cell types

pablo-gar, line 233: curious, will we ever have this case? And if so, shouldn't it be an error since no datasets should contain empty cell types, or empty metadata?

pablo-gar (approved):
LGTM

tests/unit/backend/corpora/api_server/test_v1_dataset.py
MillenniumFalconMechanic: Hi @ebezzi, what are your thoughts on this?

MillenniumFalconMechanic: Hi @pablo-gar, good question! I am not super familiar with the error handling conventions on the BE so it’s hard for me to answer. I am following a pattern used in the original development stage effort and I am not sure of the reasoning behind it, perhaps it was to protect against bad data in lower environments? If we do want to guard against an empty dataset (or empty cell type etc) case, then we would have to make sure it's a truly unexpected case and make any necessary adjustments in the calling code.

Issue #2124, opened 2022-03-11
The original dataset should be saved as a mitigation for potential data corruption during processing


brianraymor:
**August 29 Update** - Team should review @atolopko-czi 's guidance in the [Curation API specification](https://docs.google.com/document/d/1OpBpkST75rRqTAXhPKxwYEVA2ddfU4HvjbU7BYoNi6M/edit?disco=AAAAVH1yrU8). 

**August 10 Update** - See [**curation-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1660155815198679?thread_ts=1660086715.773679&cid=C02UGUV5R3J):
1. @ebezzi to review caching the original dataset as part of this epic per @atolopko-czi 's [recommendation](https://docs.google.com/document/d/1OpBpkST75rRqTAXhPKxwYEVA2ddfU4HvjbU7BYoNi6M/edit) as a backup in cases of data corruption. 
1. There is no Q3 work for the curation API at this time. If `1` is implemented, it can be exposed later in the API if needed.

### Context

See [**wrangling**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1646270820321339) for reference. If the technical requirement is implemented, then there are potential product uses. 

When a dataset is submitted to the cellxgene portal, human-readable labels are applied for gene and ontology terms. If a dataset is later downloaded to be revised by a curator, these terms must be deleted before re-submitting the dataset to the portal; otherwise, validation fails. 

One mitigation is to add a label remover to `cellxgene-schema` or to overwrite the labels if present. (This was included in 3.0.0).

Another approach is to allow curators to download the original dataset if available. Once the curation API is implemented, all submitted datasets are stored during ingestion in their original format.

#### Updating ontology versions

The pinned ontology versions in the cellxgene schema are updated. Many datasets in the corpus must be updated to reflect the change, including:

1. Addressing obsolete terms
1. Addressing changes to human readable names (re-labeling)
1. Updating cases where a less specific CL term was used until a new version of the CL ontology was supported

It would be more efficient if a curator could simply download the original dataset to make such changes.

#### (Temporary) differences in submission and download file formats

[...]

PR #3294, opened 2022-09-16
feat: Save the original h5ad as an artifact in the dataset.

Bento007:
- #2124

### Reviewers

---

## Changes
- Add saving of the original h5ad file. The original h5ad is named raw.h5ad in the dataset_artifact.filename field. This adds a second H5AD artifact to the dataset. This artifact will be filtered out from responses to users.
- Add filenames to CorporaConstants
- Upgrade sqlalchemy version so all services are using the same version.
- Storing the original uploaded file
- Removing the original upload files when converting ORM object to dict so it doesn't show up in API responses.


backend/corpora/dataset_processing/process_download_validate.py
```diff
@@ -54,6 +54,9 @@ def process(dataset_id: str, dropbox_url: str, artifact_bucket: str):
 
     # Upload the labeled dataset to the artifact bucket
     bucket_prefix = get_bucket_prefix(dataset_id)
+    create_artifact(
```
nayib-jose-gloria, line 57: nit: mind adding a comment (or adding to the comment right above) to explain that we're saving both the unlabeled and labeled h5ads?

nayib-jose-gloria (approved):
one non-blocking nit to clarify a comment, but looks good!

tests/unit/backend/corpora/api_server/test_v1_dataset.py
```diff
@@ -315,7 +316,13 @@ def test__get_dataset_assets(self):
             user_submitted=True,
             s3_uri="s3://mock-bucket/mock-key.h5ad",
         )
-        dataset = self.generate_dataset(self.session, id="test_dataset", artifacts=[artifact_0, artifact_1])
+        artifact_2 = dict(
+            filename=CorporaConstants.ORIGINAL_H5AD_ARTIFACT_FILENAME,
+            filetype=DatasetArtifactFileType.H5AD,
+            user_submitted=True,
+            s3_uri="s3://mock-bucket/mock-key.h5ad",
```
nayib-jose-gloria, line 323: for clarity on which artifact should be found in assertions below, we should rename this s3_uri to something else (right now its matching artifact_1)

nayib-jose-gloria (commented):
One test set-up change requested, otherwise looks good!

nayib-jose-gloria (approved):
ty!

atolopko-czi:
@Bento007 Realizing this is already merged, but would it worth be splitting `datasetartifactfiletype` psql enum value `H5AD` into `H5AD_RAW` and `H5AD_LABELED`?

Issue #3297, opened 2022-09-17
chore: Add automatic validation of roll up tissue functionality in Gene Expression cubes


pablo-gar:
# Need
Steps to validate that the extra dimension as part of the tissue roll up happen correctly

# Approach

- Compare semi-manual aggregation of lung children against the extra dimension for lung for the following:
    - Cell counts
    - Sum of gene expression across all cell types

Those two comparisons must be equal for validation to pass

pablo-gar:
Duplicate of #3112

Issue #3112, opened 2022-08-16
feat: scExpression cube validation checks tissue aggregation


pablo-gar:
## Need
- To update validation of gene expression and cell count cubes to ensure that tissues can be aggregated correctly
- Validation should include:
    - Aggregating cubes by `tissue` should yield the same results as subsetting by corresponding `original_tissue`:
       - Same number of rows
       - Same cell counts
       - Same sum of rankit values
    - It is sufficient to check one `tissue` but desirable to check all of them

pablo-gar:
Done in https://github.com/chanzuckerberg/single-cell-data-portal/pull/3302

PR #3302, opened 2022-09-19
chore: Add automatic validation of roll up tissue functionality in Gene Expression cubes

pablo-gar:
- #3297 
- #3112  

### Reviewers

**Functional:** 
@ebezzi 

**Readability:** 
@atarashansky 

---

## Changes
**add Automatic validation of cubes for tissue roll up**

It compares metrics between querying the cube by "Lung" on the rolled-up tissues VS all children of Lung on the original tissues.

It compares that:

- The number of cells is equal
- The sum of expression per cell types is equal
    - Ensures cell types are the same
    - Ensures  expression is the same 

## QA steps (optional)

(DONE) Test that validation passes with test data (141 datasets from prod)


backend/wmg/data/validation/validation.py
```diff
@@ -115,6 +120,67 @@ def validate_cube_species(self):
...
+        with tiledb.open(self.cell_count_path, "r") as cell_count_cube:
+            human_ontology_id = fixtures.validation_species_ontologies["human"]
+            all_lung_tissues = fixtures.validation_all_lung_tissues
+            lung_high_level_tissue = fixtures.validation_lung_high_level
+
+            original_tissues_cell_count = cell_count_cube.df[:, all_lung_tissues, human_ontology_id].n_cells.sum()
+            rollup_tissues_cell_count = cell_count_cube.df[lung_high_level_tissue, :, human_ontology_id].n_cells.sum()
+
+            if original_tissues_cell_count != rollup_tissues_cell_count:
+                logger.error(
```
ebezzi, line 144: Question: is this validator meant to be run manually by users or as part of the pipeline? I recall the latter being the original idea but I'm not sure we ended up implementing it. Regardless, I think it would be a good idea to make sure these methods either return a Boolean (valid/invalid) or a list of errors, kinda like the cellxgene-schema validator.

ebezzi (approved):
One comment but probably for the future. Approving 👍

backend/wmg/data/validation/validation.py
pablo-gar: It runs automatically as part of the pipeline

https://github.com/chanzuckerberg/single-cell-data-portal/blob/9a8d94c20debc40cbbcdd63e79c775e6e8543fa1/backend/corpus_asset_pipelines/summary_cubes/__init__.py#L14-L30

PR #3325, opened 2022-09-20
fix: proceed with processing if status is INITIALIZED

danieljhegeman:
This commit ensures that INITIALIZED Datasets -- those created using the Curation API -- will proceed with processing rather than erroneously throwing an InvalidProcessingStateException

- #TICKET_NUMBER

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/corpora/common/upload.py
+++ b/backend/corpora/common/upload.py
@@ -79,6 +79,7 @@ def upload(
         if dataset.processing_status.processing_status not in [
             ProcessingStatus.SUCCESS,
             ProcessingStatus.FAILURE,
+            ProcessingStatus.INITIALIZED,
         ]:
             raise InvalidProcessingStateException(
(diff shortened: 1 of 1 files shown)


backend/corpora/common/upload.py
```diff
@@ -79,6 +79,7 @@ def upload(
         if dataset.processing_status.processing_status not in [
             ProcessingStatus.SUCCESS,
             ProcessingStatus.FAILURE,
+            ProcessingStatus.INITIALIZED,
```
metakuni, line 82: Will there not be corresponding test(s) that need to be updated?

danieljhegeman: no because we don't have tests for this

danieljhegeman: getting lambdas working in localstack has proven difficult. Andrew dove into this and failed, I think. 😬

metakuni (approved):

Issue #3212, opened 2022-08-30
Authorization headers are not consistently documented in the swagger API


brianraymor:
Is there a [formal way](https://swagger.io/docs/specification/authentication/) to document security schemes in swagger? 

For example>

POST ../token - Returns a bearer access token t**hat must be passed as a parameter to requests that require authorization** such as creating a new Collection.

but when I review one example:

POST ../collections - Create a new Collection

There is no reference for how the access token is passed in the request? 

It only seems to be documented in the description for:

GET ../collections - 
When the visibility parameter is unspecified or set to PUBLIC, a list of all public collections is returned. **The Authorization header is not required.** If a collection in the list has been deleted, then it is annotated with tombstone set to True.

When the visibility parameter is set to PRIVATE, a list of all private collections that the user is authorized to access is returned. **The Authorization header is required**. If a collection in the list is a private revision of a public collection, then it is

Bento007:
We only documented it when it is an optional parameter that affects the response. We can add more documentation to the `/v1/auth/token: POST` endpoint with instructions on how it should be used.

The openapi spec does inform the user when authentication is required. What other details would you like?

brianraymor:
I was curious about whether there is *formal* markdown that can be added as a security schema to all requests that MAY require an authorization header.

Bento007:
I'll add some markdown to endpoint that required auth. Openapi doesn't do it for us. It only gives us the little padlock next to the endpoint.

brianraymor:
For example, I was wondering about - https://github.com/OAI/OpenAPI-Specification/issues/14#issuecomment-297457320

Bento007:
We have that defined in the openapi spec. https://github.com/chanzuckerberg/single-cell-data-portal/blob/6404a803a1650ed414c585a5f3e39e3725445262/backend/config/curation-api.yml#L60-L62

brianraymor:
GET /v1/collections/{uuid}

**Optional: provide your access token in the request header as Authentication: Bearer <access_token>.**

Is it optional or is it required in specific scenarios? 

---

PATCH /v1/collections/{uuid}

**Optional: provide your access token in the request header as Authentication: Bearer <access_token>.**

Isn't it always required?

---

POST /v1/collections/{uuid}/datasets

No reference to authentication token?

---

GET /v1/collections/{uuid]/datasets/{uuid}

No reference to authentication token?

---


GET .../assets

No reference to authentication token?

---

[...]

PR #3332, opened 2022-09-21
fix(curation-api): auth headers

Bento007:
- #3212

### Reviewers

**Readability:** 

---

## Changes
- adding additional details about optional auth headers
- change optional to required for updating collection metadata.


backend/config/curation-api.yml
```diff
@@ -154,6 +154,8 @@ paths:
       description: >
         Fetch full metadata for the Collection and its Datasets.
 
+        If the collection owner's Authentication token is provided, then revision related fields will also be returned.
```
brianraymor, line 157: 1. Any user may request a public collection w/o authorization.
2. A curator may request a **private** collection that always requires authorization.

brianraymor, line 157: I don't think we should use _collection owner's_ because we're discussing authorized users - for example, a super curator may not be the curator who created a collection, but they're authorized to see the details.

Bento007:
```suggestion
        If an Authentication token is provided and the curator is allowed, then revision-related fields will also be returned.
```

backend/config/curation-api.yml
```diff
@@ -323,8 +327,9 @@ paths:
     delete:
       summary: Delete a Dataset or cancel processing for a Dataset in a private Collection or Revision.
       description: >-
-        Delete a Dataset or cancel processing for a Dataset. The **collection_id** MUST reference either a private Collection OR a 
-        Revision. MUST indicate the Dataset by providing the **dataset_id** parameter.
+        Delete or cancel processing for a Dataset with `dataset_id` in `collection_id`. 
+        The **collection_id** MUST reference either a private Collection OR a
```
danieljhegeman, line 331:
```suggestion
        The `collection_id` MUST reference either a private Collection OR a 
```

danieljhegeman: I did a quick check - this is the only such instance in the api file of accidentally using asterisks for a variable instead of backticks 👍

backend/config/curation-api.yml
```diff
@@ -154,6 +154,8 @@ paths:
       description: >
         Fetch full metadata for the Collection and its Datasets.
 
+        If an Authentication token is provided and the curator is allowed, then revision-related fields will also be returned.
```
danieljhegeman, line 157: > revision-related fields

This is slightly confusing. Was this specifically requested by Brian or Jason?

Bento007: yes, they wanted to know what happens if an auth token is provided.

danieljhegeman: Doesn't this comprise solely either the `revision_of` or `revising_in` fields? The current phrasing sort of suggests that it returns fields for the revision collection.

danieljhegeman:
```suggestion
        An Authentication token must be provided in order to retrieve the id of an associated Revision for this Collection, if such a Revision currently exists.
```

danieljhegeman: or something like that

danieljhegeman (approved):
🚀

Issue #3299, opened 2022-09-19
Reorganize dataset upload link endpoint


danieljhegeman:
Since we've deprecated the curator tag: 

`PUT /v1/collections/{collection_id}/datasets/upload-link` with query params `id` and `link` should become `PUT /v1/collections/{collection_id}/datasets/{dataset_id}` with query param `link`.

PR #3337, opened 2022-09-22
fix: move 'PUT /datasets/upload-link' to 'PUT /datasets/{dataset_id}'

danieljhegeman:
- #3299 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add
- remove
- modify


backend/corpora/lambdas/api/v1/curation/collections/collection_id/datasets/actions.py
```diff
@@ -44,3 +45,13 @@ def post(token_info: dict, collection_id: str):
         db_session, collection=collection, processing_status={"processing_status": ProcessingStatus.INITIALIZED}
     )
     return make_response(jsonify({"dataset_id": dataset.id}), 201)
+
+
+def put(collection_id: str, dataset_id: str, body: dict, token_info: dict):
```
Bento007, line 50: thank you for moving this

Bento007 (approved):
So now you have to create an empty dataset before you can upload from a link or s3? It's more verbose but consistent.

Bento007 (changes requested):
We need to rework the description of PUT upload-link

danieljhegeman:
> We need to rework the description of PUT upload-link

@Bento007 done

danieljhegeman:
> So now you have to create an empty dataset before you can upload from a link or s3? It's more verbose but consistent.

Yeah so the reason for this was to avoid the annoyance of manually creating a `ProcessingStatus` of `INITIALIZED`; the testing code infra is not there for it

Bento007 (approved):

Issue #2031, opened 2022-03-03
Update slack dataset processing notification wrt primary key migration


danieljhegeman:
- Update https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/corpora/dataset_processing/slack.py#L36

danieljhegeman:
moving to F&F

brianraymor:
@danieljhegeman - is this assigned to you?

danieljhegeman:
> @danieljhegeman - is this assigned to you?

No it is not. @Bento007 ?

metakuni:
@danieljhegeman will add more details

brianraymor:
@metakuni @ebezzi - no further details were added per Kuni's request above. Can you advise on this ?

ebezzi:
I have no idea what this is about. @danieljhegeman please add more details or close it if no longer relevant.

metakuni:
@danieljhegeman , could you provide more details? The link in the description is also broken.

PR #3339, opened 2022-09-23
fix: remove '/private' from Collection url in Slack notification

danieljhegeman:
Alright, here it is. After countless hours of untangling our pipeline and re-deriving various fundamentals of binary-encoded data theory, this is what I've come up with for this previously-thought-to-be-unsolvable ticket. Even [Emanuele was stumped](https://github.com/chanzuckerberg/single-cell-data-portal/issues/2031#issuecomment-1224093938) by this one.

- #2031 

### Reviewers
**Functional:** 
@Bento007 @seve @metakuni 
**Readability:** 

---

## Changes
- add
- remove
- modify


Bento007 (approved):
nice

Issue #3342, opened 2022-09-26
Expired Key results in uninformative/unhelpful UNAUTHORIZED error message


danieljhegeman:
https://czi-sci.slack.com/archives/C02UGUV5R3J/p1664207979936129

We need to return a message that indicates that the user needs to regenerate their key.

danieljhegeman:
Edit: this is actually a problem with how the response is handled in the client side code in the notebook.

PR #3344, opened 2022-09-26
Tsmith/3342 expired key

Bento007:
- #3342

### Reviewers

---

## Changes
- add error message when the curator API key has expired
- encapsulated JWT-related code, so error messages are returned consistently.

### DIFF
--- a/backend/corpora/common/authorizer.py
+++ b/backend/corpora/common/authorizer.py
@@ -48,7 +48,10 @@ def assert_authorized_token(token: str, audience: str = None) -> dict:
                 token, public_key, algorithms=algorithms, audience=use_audience, issuer=issuer, options=options
             )
         except ExpiredSignatureError:
-            raise
+            raise UnauthorizedError(
+                detail="Your authentication credentials have expired. Please provide updated "
+                "credentials before trying again."
+            )
         except JWTClaimsError:
             raise UnauthorizedError(detail="Incorrect claims, please check the audience and issuer.")
         except Exception:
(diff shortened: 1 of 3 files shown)


danieljhegeman:
@Bento007 by your eye, is there or is there not an issue with the message being provided by the BE for this scenario?

Bento007:
The current error message is a generic 401. We can be a little more detailed in this case.

danieljhegeman (changes requested):
while we're improving the access token acquisition flow, can you please rename [this function](https://github.com/chanzuckerberg/single-cell-data-portal/blob/807a866b6a57b97ccd2ea0fad74ebb0aa79a8f7e/backend/corpora/lambdas/api/v1/authentication.py#L199)

It should not have `apikey` in it. It deals with access tokens and not keys.

backend/corpora/lambdas/api/v1/authentication.py
```diff
@@ -190,18 +189,18 @@ def check_token(token: dict) -> dict:
...
     return payload
 
 
-def curation_apikey_info_func(token):
+def curation_access_token_func(token):
     return assert_authorized_token(token, CorporaAuthConfig().curation_audience)
 
 
-def curation_apikey_info_func_lenient(token: str) -> dict:
+def curation_access_token_func_lenient(token: str) -> dict:
```
danieljhegeman, line 203: thank you for getting this one as well

danieljhegeman (approved):

PR #3382, opened 2022-10-03
fix: cloudfront response logging.

Bento007:
### Reviewers

**Readability:** @ebezzi 

---

## Changes
- fix how we are logging the cloudfront invalidation response.
- Catch and log all exceptions and raise internal error on API.

### DIFF
--- a/backend/corpora/api_server/app.py
+++ b/backend/corpora/api_server/app.py
@@ -3,8 +3,6 @@ import os
 import time
 from urllib.parse import urlparse
 
-from werkzeug.exceptions import InternalServerError
-
 import connexion
 
 from connexion import FlaskApi, ProblemException, problem
@@ -174,7 +172,7 @@ def handle_corpora_error(exception):
     )
 
 
-@app.errorhandler(InternalServerError)
+@app.errorhandler(Exception)
 def handle_internal_server_error(exception):
     app.logger.exception("InternalServerError", exc_info=exception.original_exception)
     return FlaskApi.get_response(problem(500, "Internal Server Error"))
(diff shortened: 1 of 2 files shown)


backend/corpora/common/utils/cloudfront.py
```diff
@@ -34,7 +34,7 @@ def _create_invalidation(distribution: str, paths: List[str]):
             "CallerReference": invalidation_id,
         },
     )
-    logging.info(**response.json, status_code=response.status_code)
+    logging.info(dict(body=response.json, status_code=response.status_code))
```
danieljhegeman, line 37: this was the issue?

backend/corpora/api_server/app.py
```diff
@@ -174,7 +172,7 @@ def handle_corpora_error(exception):
     )
 
 
-@app.errorhandler(InternalServerError)
+@app.errorhandler(Exception)
```
danieljhegeman, line 175: Isn't this also going to be triggered by exceptions that produce responses for the client, such as 400 responses, and which are not server errors? 🤔

danieljhegeman (commented):
see q about catching `Exception`

backend/corpora/api_server/app.py
Bento007: The 400 responses for invalid parameters are thrown by connexion which throws a ProblemException. We have a handler for ProblemExceptions above.

backend/corpora/common/utils/cloudfront.py
Bento007: Yes

backend/corpora/api_server/app.py
Bento007: We would be failing a bunch of unit tests if this wasn't working properly.

danieljhegeman: Ok. As long as we're sure we aren't masking any non-500 errors with this.

danieljhegeman (approved):

Issue #3243, opened 2022-09-07
feat(wmg-download): allow for download of multiple tissues from wmg


seve:
Story:
 When a user has multiple tissues displayed on wmg and attempts to download an image, one heatmap image per tissue should be downloaded

Needs:
- [ ] One image per tissue
- [ ] Identical layout
- [ ] Possibly compress the images into single zip file (@signechambers1, the folders generated from decompressing will softly avoid any duplicates. An example of this behavior is downloading multiple images from Google photos)

PR #3383, opened 2022-10-03
feat(wmg-download): enable downloading of multiple tissues

seve:
- #3243 

### Reviewers
**Functional:** @tihuan 

---

## Changes
- add
   - Add ability to download multiple tissues into a zip  
- modify
   - Changed download method

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,7 +23,7 @@
     "d3-scale-chromatic": "^3.0.0",
     "downloadjs": "^1.4.7",
     "echarts": "^5.2.2",
-    "html-to-image": "github:seve/html-to-image#seve/add-cloned-tag",
+    "html-to-image": "^1.10.8",
     "jschardet": "^2.3.0",
     "lodash": "^4.17.20",
     "ml-hclust": "^3.1.0",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -17,6 +17,7 @@
     "@material-ui/lab": "latest",
     "@types/d3-scale-chromatic": "^3.0.0",
     "@vgrid/sass-inline-svg": "^1.0.1",
+    "client-zip": "^2.2.2",
     "clipboard-copy": "^4.0.1",
     "czifui": "^5.0.0",
     "d3-scale-chromatic": "^3.0.0",
--- a/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
@@ -107,8 +107,6 @@ export default function SaveImage({
       link.remove();
     } catch (error) {
       console.error(error);
-      // this could probably be handled better
-      return { input: "", name: "" };
     }
   }, [fileType, selectedCellTypes, selectedTissues, selectedGenes]);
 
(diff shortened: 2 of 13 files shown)


frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
```diff
@@ -1,74 +1,112 @@
...
+          return {
+            input: isPNG
+              ? base64URLToArrayBuffer(imageURL)
+              : decodeURIComponent(imageURL.split(",")[1]),
+            name: `${tissue}.${fileType}`,
+          };
+        })
       );
+      heatmapNode.classList.remove("CLONED");
+      const { downloadZip } = await import("client-zip");
```
atarashansky, line 90: Why importing here as opposed to top level?

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
```diff
@@ -1,74 +1,112 @@
...
+      const blob = await downloadZip(images).blob();
+      const link = document.createElement("a");
+      link.href = URL.createObjectURL(blob);
+      link.download = `CELLxGENE_gene_expression.zip`;
```
atarashansky, line 94: If only one tissue, I think we should download the image itself instead of zip file.

atarashansky (changes requested):

frontend/package.json
```diff
@@ -17,6 +17,7 @@
     "@material-ui/lab": "latest",
     "@types/d3-scale-chromatic": "^3.0.0",
     "@vgrid/sass-inline-svg": "^1.0.1",
+    "client-zip": "^2.2.2",
```
tihuan, line 20: `github:seve/html-to-image#seve/add-cloned-tag` do we still need to point to your fork?

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
```diff
@@ -1,74 +1,112 @@
...
+      link.click();
+      link.remove();
     } catch (error) {
       console.error(error);
+      // this could probably be handled better
+      return { input: "", name: "" };
```
tihuan, line 111: at first glance, I can't tell why we need to return anything in the catch block, given that `handleDownload` returns `void` when success?

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
seve: downloadZip will crash if SSR

frontend/src/views/WheresMyGene/components/GeneSearchBar/index.tsx
```diff
@@ -121,13 +119,13 @@ export default function GeneSearchBar({
...
-              selectedTissues={selectedTissues}
-              selectedGenes={selectedGenes}
-            />
-          </>
+          <SaveImage
+            {...{
+              selectedCellTypes,
+              selectedGenes,
+              selectedTissues,
+            }}
```
tihuan, line 127: spreading object disables prop type checking, so I think we should do explicit prop passing!

here and throughout
```tsx
<SaveImage
  selectedCellTypes={selectedCellTypes}
  ...
/>
```

atarashansky (approved):

tihuan (approved):
Looks really good! Just some small nits otherwise LGTM 🚀 Thanks for the feature, @seve !! So excited for launching this feature soon 🎉

frontend/package.json
seve: nice catch 💀

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
seve: you're right, this is an artifact from an older implementation

PR #3386, opened 2022-10-04
fix: catch all exception handler

Bento007:
- #TICKET_NUMBER

### Reviewers

**Readability:** 

---

## Changes
- modify the catch all exception handler to log the error and return the correct response.

## QA steps (optional)
- added a new test case to verify the exception is handle correctly


danieljhegeman (approved):
Clever testing 👍

PR #3389, opened 2022-10-04
fix(fe-infra): add 'wasm-unsafe-eval' to CSP to allow for wasm

seve:
https://github.com/Touffy/client-zip#webassembly-and-content-security-policy

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -12,7 +12,7 @@ const PLAUSIBLE_URL = "https://plausible.io";
 
 const TWITTER_URL = "https://cdn.syndication.twimg.com platform.twitter.com";
 
-const SCRIPT_SRC = ["'self'", PLAUSIBLE_URL, TWITTER_URL];
+const SCRIPT_SRC = ["'self'", "'wasm-unsafe-eval'", PLAUSIBLE_URL, TWITTER_URL];
 
 const defaultSecureHeaders = {
   contentSecurityPolicy: {


tihuan:
Reading through https://github.com/WebAssembly/content-security-policy/blob/main/proposals/CSP.md, it sounds like adding `'wasm-unsafe-eval'` is the minimum we need to allow WASM execution. It sucks that it doesn't give us finer controls on what WASM modules we allow to execute (from which origins, etc.), but that would need to happen later when wasm has its own CSP directive!

tihuan (approved):

tihuan:
CC @jakeyheath for visibility! Not sure if you see any immediate red flags by enabling `'wasm-unsafe-eval'` 🙏 Thank you!

seve:
> . It sucks that it doesn't give us finer controls on what WASM modules we allow to execute (from which origins, etc.)

Maybe we should look into adding `strict-dynamic` and nonces?

tihuan:
Yeah if @jakeyheath sees anything really bad with this directive, we could also use https://github.com/Stuk/jszip instead, which doesn't seem to use WASM!

jakeyheath:
Whoaaaa WASM, how cool!

jakeyheath:
I am actually unsure about this directive, I have never seen it before.

tihuan:
Yeah today is my first day seeing it too 😆 

`strict-dynamic` is supported by our targeted browsers, but if we're caching SSR'd pages for SEO, then nonce won't work!

Something to discuss if we get there!

Issue #3432, opened 2022-10-13
bug(documentation nit): Curation API OpenAPI Spec refers to 'Authentication' bearer token request header, rather than 'Authorization'


nayib-jose-gloria:
## Describe the bug
Curation API OpenAPI Spec, in the description of multiple endpoints, instructs the user to provide a request header labeled 'Authentication' for bearer token scheme auth. However, the standard HTTP request header for passing bearer tokens is 'Authorization', and that is the header that will be checked for bearer token auth. If a user follows the instructions as written to use the API, they will get a 401 Unauthorized for bearer token security gated endpoints even if they had a proper, authorized bearer token

## Expected behavior
Docs replace instances of 'Authentication' request header with 'Authorization'

danieljhegeman:
I don't think this impacts any of the actual code... so I will be making the fix in the Swagger doc only

PR #3446, opened 2022-10-14
docs: Curation API - "Authentication" header should be "Authorization"

danieljhegeman:
- #3432 

### Reviewers
**Functional:** 
@nayib-jose-gloria @Bento007 
**Readability:** 

---

## Changes
- modify doc reference for header for including Bearer token in Curation API

### DIFF
--- a/backend/config/curation-api.yml
+++ b/backend/config/curation-api.yml
@@ -24,7 +24,7 @@ paths:
       tags:
         - authentication
       description: |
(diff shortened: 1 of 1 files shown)


Bento007 (changes requested):
The header should be `Authorization` per https://www.rfc-editor.org/rfc/rfc6750#section-2.1

Bento007 (approved):

Issue #3267, opened 2022-09-13
List of Datasets in the right-side bar of Gene Expression doesn't update as intended upon filtering


pablo-gar:
When cells are filtered in Gene Expression in way that the datasets contributing data are not longer there, then the list of "Source Data" doesn't update accordingly

To reproduce:

1. Go to Gene Expression 
2. Add tissue: Lung
3. Add gene: MALAT1
4. Filter Dataset: "Tabula Sapiens  - All Cells"
5. Source data should only contain Tabula Sapiens data but it still list all other collections

<img width="922" alt="image" src="https://user-images.githubusercontent.com/29237819/189971964-48daff60-2220-43a6-affa-c502229fa42a.png">

PR #3447, opened 2022-10-17
fix: Source data updates properly when filtering (3267)

atarashansky:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3267

`useFilterDimensions` returns the filters available for selection in the filter panel. Selecting a dataset does not limit the dataset filters available for selection, because users might wish to select multiple datasets. 

To achieve the desired behavior (source data reflects selected datasets), we look at `StateContext.selectedFilters` to see which datasets are selected, and use those to filter the datasets output by `useFilterDimensions`.


ashin-czi (approved):
Approved, LGTM!

Issue #3394, opened 2022-10-05
Change all python import to be absolute.


Bento007:
We have a mix of relative and absolute imports in our python code. In order to keep consistency across our modules, change all imports to be explicit.

- Update all explorer code to have explicit python imports
- update all portal code to have explicit python imports

Bento007:
Increasing priority since it will speed up refactoring the data portal backend if we do it sooner.

Bento007:
Doing this work will make refactoring the backend code easier.

danieljhegeman:
@Bento007 should this include not using `__init__.py` as a way to globally import resources, in keeping with making all imports absolute (and explicit) in individual files?

Or is that less of a concern

Bento007:
That can be done in a separate pass. The concerns should be handled separately.

PR #3450, opened 2022-10-17
fix: Remove relative imports

Bento007:
- #3394 

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- modify all imports to be absolute imports.


danieljhegeman (approved):
My search tells me you got all instances of `from .[...]`

Issue #2963, opened 2022-07-26
Refactor Backend Code


Bento007:
Remove the Corpora Directory and organize the code to be consistent.

## Backend APIs are scattered in the backend code base
We should consolidate to a single directory for all APIs that are hosted in the same container. Here is my suggestion:
- move backend/corpora/api_server up to backend/api_server
- The portal API lives in backend/corpora/lambda/api. It's not a lambda. Move it to backend/portal
- Move open_api specs from backend/config into backend/api_server

## Group submission related code
move all submission-related code to _backend_. These are the related directories:
- _backend/dataset_processing_
- _backend/dataset_submissions_ 
- _backend/upload_failures_
- _backend/upload_success_, 

## backend/corpora/common is used by other libraries
- move _backend/corpora/common_ to _backend/common_ 

**Update the test directory to match the above changes**

atolopko-czi:
https://docs.google.com/document/d/1H3IdC8frpvJsmm6Z3XTO-fQd3XNkIPvNvcvp8j380OE/edit#

PR #3455, opened 2022-10-18
refactor: backend

Bento007:
- #2963

### Reviewers
**Functional:** @danieljhegeman @ashin-czi 

**Readability:** @ebezzi 

---

## Changes
- move backend/corpora/api_server -> backend/api_server
- move backend/corpora/lambda/api -> backend/portal
- move backend/corpora/dataset_processing -> backend/dataset_pipeline/processing 
- move backend/corpora/dataset_submissions ->  backend/dataset_pipeline/submissions
- move backend/corpora/upload_failures -> backend/dataset_pipeline/upload_failures
- move backend/corpora/upload_success -> backend/dataset_pipeline/upload_success
- move backend/corpora/common -> backend/common
- move backend/corpora/lambda/api/v1/curation -> backend/curation/api/v1/curation
- move backend/wmg/cube_pipeline.py -> backend/wmg/pipeline/cube_pipeline.py
- move backend/corpus_dataset_pipeline -> backend/wmg/pipeline
- move backend/ontology_files -> backend/common/ontology_files
- rename corpora-api.yml -> portal-api.yml
- Update docker files
- Update docs
- Update python import paths
- fix test_crossref_provider tests

**Update the test directory to match the above changes**
## QA steps (optional)
 Ran in happy path and check API works and processing a dataset works.
## Notes for Reviewer
The easiest way to review is to pull the repo and look at the new files structure and see if the files are where you'd expect them.

Tip: Many directories will be left empty after `git pull`, and removing them will reduce confusion (remove `echo` when confident):
```
# First remove any __pycache__ dirs
find backend -type d -name '__pycache__' -exec echo rm -r {} \;
# Next, remove empty leaf directories; repeat the command until all are gone. 
# Note: `find` reports errors for rm'd dirs, but it seems to work anyway
find backend -type d -empty -exec echo rmdir {} \;
```


atolopko-czi (changes requested):
Now that we can see the re-org "in the flesh", I have a few more considerations:
* should `config/*-api.yml` files go under `{portal,gene_info/wmg}/api/`, respectively?
* should `curation` be a sibling of `portal/api/v1`? Doesn't it need to be versioned independently of the (private) portal api?
* should we merge `backend/corpus_asset_pipelines` into `wmg/`? The intermediate corpus building and tightly tied to the wmg cube building pipeline steps, so it seems better to keep them together.  We can move apart if/when the corpus building becomes a producer for multiple downstream consumers.
* move `ontology_files` under `common`?
* `dataset_{submissions,processing}`, `upload_{success,failures}` are steps of the same dataset processing pipeline, so should we house them under a `backend/dataset_pipeline/` dir?

Bento007:
Great suggestions. I actually started moving the API stuff around in a separate PR, but it can all be rolled into this one.

Bento007:
> * should `config/*-api.yml` files go under `{portal,gene_info/wmg}/api/`, respectively?

I have no strong opinions either way.

> * should `curation` be a sibling of `portal/api/v1`? Doesn't it need to be versioned independently of the (private) portal API?

I moved this to backend

> * should we merge `backend/corpus_asset_pipelines` into `wmg/`? The intermediate corpus building and tightly tied to the wmg cube building pipeline steps, so it seems better to keep them together.  We can move apart if/when the corpus building becomes a producer for multiple downstream consumers.

If we do this then we should we also move the dataset processing pipelines closer to the portal and curator apis?

> * move `ontology_files` under `common`?

I have not strong opinions about this.

> * `dataset_{submissions,processing}`, `upload_{success,failures}` are steps of the same dataset processing pipeline, so should we house them under a `backend/dataset_pipeline/` dir?

I think we should go shallow in the directory paths first and then add more depth if it's needed.

atolopko-czi (approved):
This is looking SO much nicer!  👏 

Could the openapi spec yml files go under their respective `api` dirs? I see they are currently siblings to that dir. Any reason not to push then down further?

Bento007:
I moved them closer to the handlers under the API path

Issue #3451, opened 2022-10-17
There MUST be an error message for datasets that exceed UPLOAD_SIZE_LIMIT_BYTES


brianraymor:
See [#single-cell-three](https://czi-sci.slack.com/archives/C03PXFMM7ST/p1666031935254049).

In the curation UX, there was no requirement for an error message, because the dropbox choose includes:

```
    // Optional. A limit on the size of each file that may be selected, in bytes.
    // If specified, the user will only be able to select files with size
    // less than or equal to this limit.
    // For the purposes of this option, folders have size zero.
    sizeLimit: 1024, // or any positive number
```

For the API, there should be an error message surfaced to inform curators that the dataset has exceed size maximums (currently 30 GB).

danieljhegeman:
The wrinkle here is that both the local-upload-to-s3 and the usage of `PUT /datasets/{dataset_id}` (providing link for downloading the datafile) do not provide an easy means of informing the user directly.

I think the "correct" way to do this would be to surface the error in the UI, although that does break the current UI flow, since as you mentioned @brianraymor the UI prevents even _selecting_ files that would exceed the size limit.

danieljhegeman:
@brianraymor does our `cellxgene-schema` validator tool inform the user that a datafile exceeds the 30 gb limit? Perhaps we can consider that sufficient for now? Theoretically everyone should be validating prior to uploading, though perhaps in the case of a migration this isn't happening...

brianraymor:
Alas, it would be inconsistent. In all other cases, the CLI fails validation locally and on the portal for the same reasons. If the curator neglected local validation (which happens) and uploaded a dataset that exceeds the maximum size, then the operation would silently fail because portal validation would never execute.

And wouldn't it be even simpler to ask the curator to:

```
braymor@CZIMACOS3974 Notebooks % ls -lah zero.h5ad
-rw-r--r--@ 1 braymor  staff   252M Jul 14 09:50 zero.h5ad
```

danieljhegeman:
>and the usage of PUT /datasets/{dataset_id} (providing link for downloading the datafile) do not provide an easy means of informing the user directly
This statement of mine is incorrect. This endpoint _does_ warn of a file being too large, because it checks the remote file size metadata before returning a response to the client.

danieljhegeman:
>Alas, it would be inconsistent. 

[...]

PR #3459, opened 2022-10-19
fix: update validation status and message when files exceed size limit

danieljhegeman:
- #3451 

### Reviewers
**Functional:** 
@Bento007 
**Readability:** 
@nayib-jose-gloria 
---

## Changes
Populate the `dataset_processing_status.validation_status` and `dataset_processing_status.validation_message` when a file is encountered that is too large for processing.

### DIFF
--- a/backend/corpora/common/upload.py
+++ b/backend/corpora/common/upload.py
@@ -63,7 +63,7 @@ def upload(
     # Check if file size is within max size limit
     max_file_size_gb = CorporaConfig().upload_max_file_size_gb * GB
     if file_size is not None and file_size > max_file_size_gb:
-        error_message = f"The file exceeds the maximum allowed size of {max_file_size_gb} GB"
+        error_message = f"The file exceeds the maximum allowed size of {max_file_size_gb} bytes"
         if dataset:
             dataset.update(
                 processing_status={
--- a/backend/corpora/common/upload.py
+++ b/backend/corpora/common/upload.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm import Session
 from backend.corpora.common.corpora_config import CorporaConfig
 import os
 
-from backend.corpora.common.corpora_orm import CollectionVisibility, ProcessingStatus
+from backend.corpora.common.corpora_orm import CollectionVisibility, ProcessingStatus, ValidationStatus
 from backend.corpora.common.entities import Collection, Dataset
 from backend.corpora.common.utils.authorization_checks import owner_or_allowed
 from backend.corpora.common.utils.exceptions import (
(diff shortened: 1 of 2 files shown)


backend/corpora/common/upload.py
```diff
@@ -52,9 +52,26 @@ def upload(
...
+        dataset = Dataset.get(db_session, dataset_id, collection_id=collection_id)
+        if not dataset:
+            raise NonExistentDatasetException(f"Dataset {dataset_id} does not exist")
+    else:
+        dataset = None
+
+    # Check if file size is within max size limit
     max_file_size_gb = CorporaConfig().upload_max_file_size_gb * GB
     if file_size is not None and file_size > max_file_size_gb:
-        raise MaxFileSizeExceededException(f"{url} exceeds the maximum allowed file size of {max_file_size_gb} Gb")
```
danieljhegeman, line 57: `url` is not needed, and I don't want to have to think through the implications of saving Collection / Dataset id's here (probably not an issue?) _as well as_ exposing the curator's google auth id (possibly an issue?)

danieljhegeman:
@jahilton @brianraymor this fix will ensure that the curator can see that validation was invalid AND a descriptive message that it was due to file size if they retrieve the Collection using the API. 

It does _not_ include any sort of change to indicate the file size error on the UX, which would be more involved. I am assuming this is not needed, anyways.

backend/corpora/common/upload.py
```diff
@@ -52,9 +52,26 @@ def upload(
...
+        error_message = f"The file exceeds the maximum allowed size of {max_file_size_gb} bytes"
```
Bento007, line 66:
```suggestion
        error_message = f"The file exceeds the maximum allowed size of {max_file_size_gb} Gb"
```

tests/unit/backend/corpora/dataset_processing/test_common.py
```diff
@@ -206,3 +209,22 @@ def converter(_file):
...
+                collection.id,
+                "download_url_placeholder",
+                max_file_size + 1,  # Purport to exceed the max file size
+                "test_user",
+                dataset_id=dataset.id,
+            )
+        self.assertEqual(dataset.processing_status.validation_status, ValidationStatus.INVALID)
+        self.assertEqual(
+            dataset.processing_status.validation_message,
+            f"The file exceeds the maximum allowed size of {max_file_size} bytes",
```
Bento007, line 229:
```suggestion
            f"The file exceeds the maximum allowed size of {max_file_size} Gb",
```

Bento007 (changes requested):
The size is in Gigabytes not bytes.

backend/corpora/common/upload.py
danieljhegeman: sorry this had to be fixed above, because the GB multiplier is already being applied, thus making it bytes

Bento007 (approved):

brianraymor:
> It does not include any sort of change to indicate the file size error on the UI for that Collection, which would be more involved. I am assuming this is not needed, anyways.

I think that's a good compromise. What is the UX experience? Silence? Validation failure with no details?

Issue #3461, opened 2022-10-19
bug: null values must be stringified before generating cxg_schema type hints


atarashansky:
[Relevant slack thread](https://czi-sci.slack.com/archives/C02450L5YAX/p1666127016416749)

To reproduce:
[Published Dataset](https://cellxgene.cziscience.com/e/232f6a5a-a04c-4758-a6e8-88ab2e3a6e69.cxg/) - doublet_status is in left sidebar drop-downs
[Revision Dataset](https://cellxgene.cziscience.com/e/25d14de9-23f5-44e3-bf14-11366dd407bc.cxg/) (no updates to that field) - doublet_status in the dataset drawer and not in the left sidebar

The difference is the following: In Published dataset, nan is a string and treated as a label. In Revision dataset, nan is a NaN (null value) and ignored. This is an issue in the processing pipeline. Categorical data should be converted to strings first.

PR #3462, opened 2022-10-19
fix: Cast categorical values to strings before generating cxg_schema type hints.

atarashansky:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3461


atolopko-czi (commented):
We should probably have a unit test to ensure this code path is covered.

atolopko-czi (approved):
LGTM

PR #3463, opened 2022-10-19
feat: redesign persistence layer (interface+mock)

ebezzi:
- #TICKET_NUMBER

### Reviewers
**Functional:** 
@atolopko-czi 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/layers/common/entities.py
+++ b/backend/layers/common/entities.py
@@ -36,6 +36,12 @@ class Dataset:
 
 
 @dataclass
+class DatasetVersion:
+    version_id: str
+    dataset: Dataset
+
+
+@dataclass
 class CollectionMetadata:
     name: str
     description: str
@@ -49,8 +55,8 @@ class CollectionMetadata:
 class Collection:
     id: str
     metadata: CollectionMetadata
-    publisher_metadata: dict # TODO: use a dataclass
-    datasets: List[Dataset]
+    publisher_metadata: dict  # TODO: use a dataclass
+    datasets: List[DatasetVersion]
 
 
 @dataclass
@@ -58,8 +64,3 @@ class CollectionVersion:
     version_id: str
     collection: Collection
     published_at: datetime
-
-@dataclass
-class DatasetVersion:
-    version_id: str
-    dataset: Dataset
--- a/backend/layers/common/entities.py
+++ b/backend/layers/common/entities.py
@@ -29,7 +29,7 @@ class DatasetMetadata:
 
 @dataclass
 class DatasetVersion:
-    id: str
+    dataset_id: str
     version_id: str
     processing_status: Optional[DatasetStatus]
     metadata: DatasetMetadata
--- a/backend/layers/common/entities.py
+++ b/backend/layers/common/entities.py
@@ -1,0 +1,65 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import List
+
+
+@dataclass
+class DatasetStatus:
+    status: str  # TODO: use an enum
+
+
+@dataclass
+class DatasetArtifact:
+    type: str
+    uri: str
+
+
+@dataclass
+class DatasetMetadata:
+    organism: str
+    tissue: str
+    assay: str
+    disease: str
+    sex: str
+    self_reported_ethnicity: str
+    development_stage: str
+    cell_type: str
+    cell_count: int
+
+
+@dataclass
+class Dataset:
+    id: str
+    status: DatasetStatus
+    metadata: DatasetMetadata
+    artifacts: List[DatasetArtifact]
+
+
+@dataclass
+class CollectionMetadata:
+    name: str
+    description: str
+    owner: str
+    contact_name: str
+    contact_email: str
+    links: List[dict]  # TODO: use a dataclass
+
+
+@dataclass
+class Collection:
+    id: str
+    metadata: CollectionMetadata
+    publisher_metadata: dict # TODO: use a dataclass
+    datasets: List[Dataset]
+
+
+@dataclass
+class CollectionVersion:
+    version_id: str
+    collection: Collection
+    published_at: datetime
+
+@dataclass
+class DatasetVersion:
+    version_id: str
+    dataset: Dataset
--- a/backend/layers/common/entities.py
+++ b/backend/layers/common/entities.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from datetime import datetime
-from typing import List
+from typing import List, Optional
 
 
 @dataclass
@@ -28,17 +28,12 @@ class DatasetMetadata:
 
 
 @dataclass
-class Dataset:
+class DatasetVersion:
     id: str
-    status: DatasetStatus
+    version_id: str
+    processing_status: Optional[DatasetStatus]
     metadata: DatasetMetadata
     artifacts: List[DatasetArtifact]
-
-
-@dataclass
-class DatasetVersion:
-    version_id: str
-    dataset: Dataset
 
 
 @dataclass
@@ -52,15 +47,11 @@ class CollectionMetadata:
 
 
 @dataclass
-class Collection:
-    id: str
-    metadata: CollectionMetadata
-    publisher_metadata: dict  # TODO: use a dataclass
-    datasets: List[DatasetVersion]
-
-
-@dataclass
 class CollectionVersion:
+    collection_id: str
     version_id: str
-    collection: Collection
-    published_at: datetime
+    owner: str
+    metadata: CollectionMetadata
+    publisher_metadata: Optional[dict]  # TODO: use a dataclass
+    datasets: List[DatasetVersion] 
+    published_at: Optional[datetime]
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -1,0 +1,62 @@
+from typing import List
+from backend.corpora.common.entities.dataset import Dataset
+from backend.layers.common.entities import Collection, CollectionMetadata, CollectionVersion, DatasetAsset, DatasetMetadata, DatasetStatus
+
+
+class DatabaseProviderInterface:
+    def create_collection(self, collection_metadata: CollectionMetadata):
+        pass
+
+    def get_collection(self, collection_id: str) -> Collection:
+        pass
+
+    def get_all_collections(self) -> List[Collection]:  # TODO: add filters if needed
+        pass
+
+    def delete_collection(self, collection_id: str) -> None:
+        pass
+
+    def update_collection_metadata(self, collection_id: str, collection_metadata: CollectionMetadata) -> None:
+        pass
+
+    def update_collection_publisher_metadata(self, collection_id: str, publisher_metadata: dict) -> None:
+        pass
+
+    def create_collection_version(self, collection_id: str) -> str:
+        pass
+
+    def delete_collection_version(self, collection_id: str, version_id: str) -> None:
+        pass
+
+    def get_collection_version(self, collection_id: str, version_id: str) -> CollectionVersion:
+        pass
+
+    def publish_collection(self, collection_id: str) -> None:
+        pass
+
+    def get_dataset(self, dataset_id: str) -> Dataset:
+        pass
+
+    def get_all_datasets(self) -> List[Dataset]:  # TODO: add filters if needed
+        pass
+
+    def get_dataset_assets(self, dataset_id: str) -> List[DatasetAsset]:
+        pass
+
+    def create_dataset(self, name: str, dataset_metadata: DatasetMetadata) -> None:
+        pass
+
+    def create_dataset_asset(self, dataset_id: str, asset: DatasetAsset) -> None:
+        pass
+
+    def update_dataset_status(self, dataset_id: str, status: DatasetStatus) -> None:
+        pass
+
+    def add_dataset_to_collection_version(self, version_id: str, dataset_id: str) -> None:
+        pass
+
+    def delete_dataset_from_collection_version(self, version_id: str, dataset_id: str) -> None:
+        pass
+
+    def replace_dataset_in_collection_version(self, version_id: str, dataset_id: str) -> None:
+        pass
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -8,7 +8,7 @@ class DatabaseProviderInterface:
         
     def create_canonical_collection(self, owner: str, collection_metadata: CollectionMetadata) -> CollectionVersion:
         """
-        Initializes a canonical collection, generating a canonical collection_id and a new version_id.
+        Creates a new canonical collection, generating a canonical collection_id and a new version_id.
         Returns the newly created CollectionVersion
         """
         pass
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -64,4 +64,5 @@ class DatabaseProviderInterface:
     def delete_dataset_from_collection_version(self, collection_version_id: str, dataset_version_id: str) -> None:
         pass
 
-    def replace_dataset_in_collection_version(
+    def replace_dataset_in_collection_version(self, collection_version_id: str, old_dataset_version_id: str, dataset_metadata: DatasetMetadata) -> None:
+        pass
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -109,7 +109,7 @@ class DatabaseProviderInterface:
         """
         pass
 
-    def initialize_canonical_dataset(self, collection_version_id: str, dataset_metadata: DatasetMetadata) -> DatasetVersion:
+    def create_canonical_dataset(self, collection_version_id: str, dataset_metadata: DatasetMetadata) -> DatasetVersion:
         """
         Initializes a canonical dataset, generating a dataset_id and a dataset_version_id.
         Returns the newly created DatasetVersion.
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -109,7 +109,7 @@ class DatabaseProviderInterface:
         """
         pass
 
-    def inizialize_canonical_dataset(self, collection_version_id: str, dataset_metadata: DatasetMetadata) -> DatasetVersion:
+    def initialize_canonical_dataset(self, collection_version_id: str, dataset_metadata: DatasetMetadata) -> DatasetVersion:
         """
         Initializes a canonical dataset, generating a dataset_id and a dataset_version_id.
         Returns the newly created DatasetVersion.
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -1,46 +1,53 @@
 from typing import List
 from backend.corpora.common.entities.dataset import Dataset
-from backend.layers.common.entities import Collection, CollectionMetadata, CollectionVersion, DatasetAsset, DatasetMetadata, DatasetStatus
+from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetArtifact, DatasetMetadata, DatasetStatus
 
 
 class DatabaseProviderInterface:
-    def create_collection(self, collection_metadata: CollectionMetadata):
+        
+    def create_collection(self, owner: str, collection_metadata: CollectionMetadata):
         pass
 
-    def get_collection(self, collection_id: str) -> Collection:
+    def get_collection(self, collection_id: str) -> CollectionVersion:
         pass
 
-    def get_all_collections(self) -> List[Collection]:  # TODO: add filters if needed
+    def get_all_collections(self) -> Iterable[CollectionVersion]:  # TODO: add filters if needed
         pass
 
     def delete_collection(self, collection_id: str) -> None:
         pass
 
-    def update_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
+    def save_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
         pass
 
-    def update_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
+    def save_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
         pass
 
     def create_collection_version(self, collection_id: str) -> str:
         pass
 
-    def delete_collection_version(self, collection_id: str, version_id: str) -> None:
+    def delete_collection_version(self, version_id: str) -> None:
         pass
 
-    def get_collection_version(self, collection_id: str, version_id: str) -> CollectionVersion:
+    def get_collection_version(self, version_id: str) -> CollectionVersion:
         pass
 
     def publish_collection(self, collection_id: str, version_id: str) -> None:
         pass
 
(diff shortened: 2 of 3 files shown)


backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+    name: str
+    description: str
+    owner: str
+    contact_name: str
+    contact_email: str
+    links: List[dict]  # TODO: use a dataclass
+
+
+@dataclass
+class Collection:
```
atolopko-czi, line 49: Did we decide that we could add a `latest_version_id` to Collection?  Or are we okay finding latest version via max/sort/iteration?

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
+from typing import List
+from backend.corpora.common.entities.dataset import Dataset
+from backend.layers.common.entities import Collection, CollectionMetadata, CollectionVersion, DatasetAsset, DatasetMetadata, DatasetStatus
+
+
+class DatabaseProviderInterface:
+    def create_collection(self, collection_metadata: CollectionMetadata):
+        pass
+
+    def get_collection(self, collection_id: str) -> Collection:
```
atolopko-czi, line 10: wondering if we'll ever need to get the collection itself.  maybe this can be `get_latest_collection_version(self, collection_id: str)`, from which one can retrieve the canonical parent Collection.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
...
+    def update_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
+        pass
+
+    def update_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
+        pass
+
+    def create_collection_version(self, collection_id: str) -> str:
+        pass
+
+    def delete_collection_version(self, collection_id: str, version_id: str) -> None:
```
atolopko-czi, line 28: physical or logical deletion? is it even needed? logically, I think we just need to tombstone canonical Collections, rather than any particular version.  Versions will just be superseded by new versions.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
...
+        pass
+
+    def get_collection_version(self, collection_id: str, version_id: str) -> CollectionVersion:
```
atolopko-czi, line 31: if we end up adding `get_latest_collection_version`, we may not need this.  i.e., will we ever need to get a collection version that isn't the latest one? (not sure, just thinking out loud).   If anything, we might need a `get_all_collection_versions(self, collection_id:str)`, if we have business logic that needs to work with more than just the latest version.

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+    cell_type: str
+    cell_count: int
+
+
+@dataclass
+class Dataset:
+    id: str
+    status: DatasetStatus
+    metadata: DatasetMetadata
+    artifacts: List[DatasetArtifact]
```
atolopko-czi, line 35: These last 3 should be in the `DatasetVersion`, right?

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import List
+
+
```
atolopko-czi, line 5: I think we should we add `created_at` timestamps to all entities

atolopko-czi (changes requested):
Added a number comments that shows that we may not be working off the same understanding of how versioning will work. Most of my comments return my thinking to the original [schema proposal](https://docs.google.com/document/d/1phr7vTz_iqd98yLOldHienwsl403pq89I-KGUlAlF2U/edit#heading=h.aoib77ca9ck0), which seemed like it would work. Let's discuss further before I continue reviewing.

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+@dataclass
+class DatasetStatus:
+    status: str  # TODO: use an enum
```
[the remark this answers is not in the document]
ebezzi: Could just be an enum

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@dataclass
+class DatasetArtifact:
+    type: str
```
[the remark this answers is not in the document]
ebezzi: Definitely

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+    uri: str
+
+
+@dataclass
+class DatasetMetadata:
+    organism: str
```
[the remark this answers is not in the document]
ebezzi: I wasn't even aware that `Optional` existed - good call! 
Using `dict`s is a good idea, or we could take a step further and create a dataclass for ontology terms.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
...
+    def create_collection(self, collection_metadata: CollectionMetadata):
+        pass
+
+    def get_collection(self, collection_id: str) -> Collection:
+        pass
+
+    def get_all_collections(self) -> List[Collection]:  # TODO: add filters if needed
+        pass
+
+    def delete_collection(self, collection_id: str) -> None:
```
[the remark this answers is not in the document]
ebezzi: It is the only option.

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+    contact_email: str
+    links: List[dict]  # TODO: use a dataclass
+
+
+@dataclass
+class Collection:
+    id: str
+    metadata: CollectionMetadata
+    publisher_metadata: dict # TODO: use a dataclass
+    datasets: List[Dataset]
```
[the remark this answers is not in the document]
ebezzi: The latter is definitely true (in fact, that's how the mock is implemented - I need to set up pylance to attempt some static typing and catch those errors !).

For the former, it is mostly a coding design choice - those 4 fields are just grouped under a `Collection` class for ease of use. When you create a version, you want to clone those fields verbatim and copy them to the new version (since initially a version/revision is identical to the published). Grouping them allows for an easier copy. Note that this `Collection` doesn't have anything to do with the `Collections` table we designed, which will only be a mapping between ids.

backend/layers/common/entities.py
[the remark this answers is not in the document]
ebezzi: We probably do. I was initially thinking to leave tombstones entirely within the persistence layer (e.g., `get_collection` could raise an exception in case the selected collection is tombstoned), but it might make more sense to leave that logic up to the business layer. Any preference?

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
+from typing import List
+from backend.corpora.common.entities.dataset import Dataset
+from backend.layers.common.entities import Collection, CollectionMetadata, CollectionVersion, DatasetAsset, DatasetMetadata, DatasetStatus
+
+
+class DatabaseProviderInterface:
```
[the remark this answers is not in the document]
ebezzi: I was not aware of `abc`, will take a look and refactor if it's cleaner.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_all_collections(self) -> List[Collection]:  # TODO: add filters if needed
+        pass
+
+    def delete_collection(self, collection_id: str) -> None:
+        pass
+
+    def update_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
+        pass
+
+    def update_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
```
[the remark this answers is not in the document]
ebezzi: I think my rationale here was that in some cases (e.g. when a DOI isn't changed during a collection edit), the latter will not need to be called. We could achieve the same result by using nullable parameters though.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_collection(self, collection_id: str) -> Collection:
+        pass
+
+    def get_all_collections(self) -> List[Collection]:  # TODO: add filters if needed
+        pass
+
+    def delete_collection(self, collection_id: str) -> None:
+        pass
+
+    def update_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
```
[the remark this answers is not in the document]
ebezzi: That's right but I think immutability should be entirely owned by the persistence layer implementation, therefore `update_collection_metadata` should be implemented via append only but having `update_` in the name isn't bad. Logically creating a new version is still an update ;)

backend/layers/common/entities.py
ebezzi: See my comment above - this class isn't tied to the `Collection` table. I can see the confusion though so we should probably rename this to something better. Unfortunately, the only name that comes to mind right now is `CollectionData`...

backend/layers/persistence/persistence.py
ebezzi: I was thinking that some of the curation API endpoints would only want collection information and not version, so I created this method. I agree though that this layer should probably be very explicit and always return "versioned" objects.

backend/layers/persistence/persistence.py
ebezzi: We do have the option to discard a revision, hence this method. The implementation might not be straightforward though.

backend/layers/persistence/persistence.py
ebezzi: I thought it might be safer not to rely on `get_all_collection_versions`, but to have the API pass the version_id it's currently working on. This can avoid some race conditions. We might also not follow this pattern, and in that case, we won't need this method.

backend/layers/common/entities.py
ebezzi: See my comment about `Collection` (also applies here 😄 )

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import List
+
+
+@dataclass
```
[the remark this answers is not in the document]
ebezzi: Seems like a great library. +1 for using it

backend/layers/common/entities.py
```diff
@@ -0,0 +1,65 @@
...
+class Dataset:
+    id: str
+    status: DatasetStatus
+    metadata: DatasetMetadata
+    artifacts: List[DatasetArtifact]
+
+
+@dataclass
```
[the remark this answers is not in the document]
ebezzi: Won't work. collections depend on datasets so they need to be below.

backend/layers/common/entities.py
```diff
@@ -0,0 +1,57 @@
...
+    sex: str
+    self_reported_ethnicity: str
+    development_stage: str
+    cell_type: str
+    cell_count: int
+
+
+@dataclass
+class DatasetVersion:
+    id: str
```
atolopko-czi, line 32:
```suggestion
    dataset_id: str
```

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,75 @@
+from typing import List
+from backend.corpora.common.entities.dataset import Dataset
+from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetArtifact, DatasetMetadata, DatasetStatus
+
+
+class DatabaseProviderInterface:
+        
+    def create_collection(self, owner: str, collection_metadata: CollectionMetadata):
```
atolopko-czi, line 8: maybe it should return `CollectionVersion`? similarly for all other other `create*` methods.

ebezzi: For the persistence layer, I was thinking to return the id of the newly created entity. That said, if the full entity class can be returned without an additional database call or merge logic, we should take a step further and return it.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,75 @@
...
+    def get_collection(self, collection_id: str) -> CollectionVersion:
+        pass
+
+    def get_all_collections(self) -> Iterable[CollectionVersion]:  # TODO: add filters if needed
+        pass
+
+    def delete_collection(self, collection_id: str) -> None:
+        pass
+
+    def save_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
```
atolopko-czi, line 20: this is to be used for saving repeated edits to a collection version while it remains unpublished, right?

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,75 @@
...
+        pass
+
+    def save_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
+        pass
+
+    def create_collection_version(self, collection_id: str) -> str:
```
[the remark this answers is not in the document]
ebezzi: This would be "start_revision", but since `start` already implies business logic, I went ahead with `create`.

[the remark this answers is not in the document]
atolopko-czi: `add` could work

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        pass
+
+
+    def get_collection_version(self, version_id: str) -> CollectionVersion:
+        """
+        Retrieves a specific collection version by id
+        """
+        pass
+
+    def get_all_version_for_collection(self, collection_id: str) -> List[CollectionVersion]:
```
atolopko-czi, line 23:
```suggestion
    def get_all_versions_for_collection(self, collection_id: str) -> List[CollectionVersion]:
```

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        pass
+
+    def save_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
+        """
+        Saves publisher metadata for a collection version
+        """
+        pass
+
+    def add_collection_version(self, collection_id: str) -> CollectionVersion:
```
atolopko-czi, line 54: we don't expect to pass an initial CollectionMetadata?

ebezzi: The idea is to get it from the existing version. That said, we can add it as a parameter and let the business layer take care of that.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        pass
+
+    def save_collection_metadata(self, version_id: str, collection_metadata: CollectionMetadata) -> None:
+        """
+        Saves collection metadata for a collection version
+        """
+        pass
+
+    def save_collection_publisher_metadata(self, version_id: str, publisher_metadata: dict) -> None:
```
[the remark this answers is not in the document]
ebezzi: There are two different metadata: "standard" and "publisher". I would rather keep the two methods separate.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        pass
+
+    def get_all_versions_for_collection(self, collection_id: str) -> List[CollectionVersion]:
+        """
+        Retrieves all versions for a specific collections, without filtering
+        """
+        pass
+
+    def get_all_collections(self) -> Iterable[CollectionVersion]:
```
atolopko-czi, line 29: would rename:
```suggestion
    def get_all_collections_latest_version(self) -> Iterable[CollectionVersion]:
```
I see `TODO: add a filtering object`.  This will allow client to get just published collections, right?

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
+from datetime import datetime
+from typing import List
+from backend.corpora.common.entities.dataset import Dataset
+from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetArtifact, DatasetMetadata, DatasetStatus, DatasetVersion
+
+
+class DatabaseProviderInterface:
+        
+    def initialize_canonical_collection(self, owner: str, collection_metadata: CollectionMetadata) -> CollectionVersion:
```
atolopko-czi, line 9:
```suggestion
    def create_canonical_collection(self, owner: str, collection_metadata: CollectionMetadata) -> CollectionVersion:
```

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        Initializes a canonical collection, generating a canonical collection_id and a new version_id.
```
atolopko-czi, line 11:
```suggestion
        Creates a new canonical collection, generating a canonical collection_id and a new version_id.
```

backend/layers/persistence/persistence.py
ebezzi: This method needs to be more generic than just `latest_version`. For instance, we also need to retrieve all the private collections (aka, all the collection versions that haven't been previously published) owned by someone, or by all users (in case of a super curator), hence the necessity of a filtering object. Your method still makes sense and can be added as a secondary method that relies on this one.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        pass
+
+    def get_all_collections(self) -> Iterable[CollectionVersion]:
+        """
+        Retrieves all collections.
+        # TODO: add a filtering object.
+        """
+        pass
+
+    def delete_canonical_collection(self, collection_id: str) -> None:
```
atolopko-czi, line 36: Perhaps tombstoning should remain a business layer concern. It seems asymmetric for the persistence layer to support a _logical_ deletion, where the data still exists physically, but then cannot later be found via the persistence layer. 

(It's also not clear to me yet how the physical model will support logical deletion. Presumably the "collection mapping" table would have a tombstoned flag?)

backend/layers/persistence/persistence.py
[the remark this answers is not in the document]
atolopko-czi: OK, let's go with that for now.  I can see we're only providing save methods for the _parts_ of the entity that can be modified.

backend/layers/persistence/persistence.py
atolopko-czi: Yeah, I don't see doing a copy-on-create as being useful. The business layer can decide what new data constitutes the new version (taking it from the previous version).

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        pass
+
+    def add_collection_version(self, collection_id: str) -> CollectionVersion:
+        """
+        Adds a collection version to an existing canonical collection.
+        Returns the new version.
+        """
+        pass
+
+    def delete_collection_version(self, version_id: str) -> None:
```
atolopko-czi, line 61: This _physically_ deletes the collection version, right? (for "cancel revision").  Presumably, it's up to the business layer to ensure that the version is unpublished, right?

backend/layers/persistence/persistence.py
atolopko-czi: But it should only ever return a single collection_version for any given collection_id, right? If so, the method doesn't capture these semantics.

Could the return type be `Dict[CollectionId, CollectionVersion]`?  (we'd have to define `CollectionId`)

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        pass
+
+    def get_all_versions_for_dataset(self, dataset_id: str) -> List[DatasetVersion]:
+        """
+        Returns all dataset versions for a canonical dataset_id
+        """
+        pass
+
+    def get_all_datasets(self) -> Iterable[DatasetVersion]:  # TODO: add filters if needed
```
atolopko-czi, line 99: Similar to collections, is this returning the _all_ DatasetVerions or one DatasetVersion per canonical dataset id?  If the latter, it's always the latest version per canonical dataset, I assume.

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        pass
+
+    def get_dataset_artifacts(self, dataset_version_id: str) -> List[DatasetArtifact]:
+        """
+        Returns all the artifacts for a specific dataset version
+        """
+        pass
+
+    def initialize_canonical_dataset(self, collection_version_id: str, dataset_metadata: DatasetMetadata) -> DatasetVersion:
```
atolopko-czi, line 112:
```suggestion
    def create_canonical_dataset(self, collection_version_id: str, dataset_metadata: DatasetMetadata) -> DatasetVersion:
```
to (now) be consistent with collection

backend/layers/common/entities.py
```diff
@@ -0,0 +1,57 @@
...
+    processing_status: Optional[DatasetStatus]
+    metadata: DatasetMetadata
+    artifacts: List[DatasetArtifact]
+
+
+@dataclass
+class CollectionMetadata:
+    name: str
+    description: str
+    owner: str
```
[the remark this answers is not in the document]
ebezzi: Nope, thanks for the catch!

tests/unit/backend/layers/persistence/persistence_mock.py
```diff
@@ -0,0 +1,176 @@
...
+    def create_dataset_artifact(self, version_id: str, artifact: DatasetArtifact) -> None:
+        version = self.datasets_versions[version_id]
+        version.artifacts.append(artifact)
+
+    def update_dataset_processing_status(self, version_id: str, status: DatasetStatus) -> None:
+        dataset_version = self.datasets_versions[version_id]
+        dataset_version.processing_status = status
+
+    def add_dataset_to_collection_version(self, version_id: str, dataset_id: str) -> None:
+        # Not needed for now - create_dataset does this
```
atolopko-czi, line 153: Also, why would a _canonical_ dataset need to be mapped to a collection version? We have `add_dataset_to_collection_version_mapping(self, collection_version_id: str, dataset_version_id: str)` which is associating a dataset _version_ to a collection _version_. Isn't that all we'll need?

atolopko-czi: moving this comment to the interface...

atolopko-czi: in fact, this is not even an interface method. :)  maybe just remove it here?

backend/layers/persistence/persistence.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        Removes a mapping between a collection version and a dataset version
+        """
+        pass
+
+    def replace_dataset_in_collection_version(self, collection_version_id: str, old_dataset_version_id: str, dataset_metadata: DatasetMetadata) -> None:
+        """
+        Replaces an existing mapping between a collection version and a dataset version
+        """
+        pass
```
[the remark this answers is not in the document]
ebezzi: I am generally in favor of more, smaller methods. We can always defer this choice to the implementation phase.

tests/unit/backend/layers/persistence/persistence_mock.py
ebezzi: Yep, this is now outdated. The mock hasn't kept up with the interface, since we're making a lot of changes, I will update it when we're settled on the interface.

ebezzi:
Incorporated in: https://github.com/chanzuckerberg/single-cell-data-portal/pull/3485

PR #3468, opened 2022-10-19
[DO NOT MERGE - testing GHA] test(push-test): Update push-test.yml

tihuan:
- #TICKET_NUMBER

Restore E2E tests to hit local containers again instead of dev BE. But building containers is so slow though. Not sure if we should do this along with better Docker layer caching?

context: https://github.com/chanzuckerberg/single-cell-data-portal/pull/2853#pullrequestreview-1039667667

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -62,6 +62,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "16.14.2"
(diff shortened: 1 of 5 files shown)


tihuan:
closing this for now, since we don't have plan to switch back yet

Issue #3412, opened 2022-10-07
Investigate and resolve failing 'Invalid DOIs' smoke tests


nayib-jose-gloria:
Address flakey 'Invalid DOIs' tests. Determine if they are providing value as tests. If not, remove them. If so, investigate and update them to fix failures, or otherwise reduce their flakiness or replace them with less flakey test(s) that still provide the expected coverage. 

Example failures (different reported issues): 
- https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/3201753174/jobs/5230126887 
- https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/3276702818/jobs/5395088931

danieljhegeman:
I think [this](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/3372) is the same issue, right?

nayib-jose-gloria:
> I think [this](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/3372) is the same issue, right?

yeah, that's a dupe. I'd like to keep this one since its in the context of the improve testing confidence epic

PR #3477, opened 2022-10-21
fix(Invalid DOI e2e tests): e2e test specifies which response it is waiting for before evaluating test assertions

nayib-jose-gloria:

- #3412

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- Made default # of workers for local e2e test runs 1, to more closely match CI pipeline post-deployment test environment when testing locally
- Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)

## QA steps (optional)
- Locally ran e2e tests against dev env

## Notes for Reviewer
- On true failures, invalid DOI tests will now timeout waiting for a 400 response (3 min) rather than fail quickly on evaluation like before. This is a trade-off, however, I find it preferable to quicker tests with a larger chance for false negatives.

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -86,8 +86,8 @@ const config: PlaywrightTestConfig = {
     viewport: VIEWPORT,
   },
 
-  /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  /* Opt out of parallel tests. */
+  workers: 1,
 
   /* Run your local dev server before starting the tests */
   // webServer: {
--- a/frontend/tests/features/collection/collection.test.ts
+++ b/frontend/tests/features/collection/collection.test.ts
@@ -1,6 +1,7 @@
 import { expect, Page, test } from "@playwright/test";
 import { ROUTES } from "src/common/constants/routes";
 import { Collection } from "src/common/entities";
+import { API_URL } from "src/configs/configs";
 import { sortByCellCountDescending } from "src/components/Collection/components/CollectionDatasetsGrid/components/DatasetsGrid/common/util";
 import { INVALID_DOI_ERROR_MESSAGE } from "src/components/CreateCollectionModal/components/Content/common/constants";
 import { BLUEPRINT_SAFE_TYPE_OPTIONS, TEST_URL } from "tests/common/constants";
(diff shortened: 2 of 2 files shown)


ashin-czi (approved):

danieljhegeman:
>Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)

@nayib-jose-gloria  This is what was causing 200s for invalid DOIs?

nayib-jose-gloria:
> > Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)
> 
> @nayib-jose-gloria This is what was causing 200s for invalid DOIs?



> > Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)
> 
> @nayib-jose-gloria This is what was causing 200s for invalid DOIs?



> > Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)
> 
> @nayib-jose-gloria This is what was causing 200s for invalid DOIs?



> 



> > Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)
> 
> @nayib-jose-gloria This is what was causing 200s for invalid DOIs?



> > Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)
> 
> @nayib-jose-gloria This is what was causing 200s for invalid DOIs?



> > Updated Invalid DOI tests to wait for specific response (URL, status code) they want to evaluate against, they were previously evaluating after the first response event on the page, which is not necessarily always the response they are actually testing (led to flakiness)
> 
> @nayib-jose-gloria This is what was causing 200s for invalid DOIs?

I believe so. These tests are now passing against the dev env when I run them from my machine

tihuan:
Wow this is such a great catch! Thanks so much for fixing this, @nayib-jose-gloria 💡 🚀 👏

PR #3485, opened 2022-10-24
feat: portal redesign

ebezzi:


backend/layers/business/business.py
```diff
@@ -0,0 +1,190 @@
+from ast import Str
+from dataclasses import dataclass
+from typing import Iterable, Optional
+from backend.corpora.common.providers.crossref_provider import CrossrefProvider
+
+from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetArtifact, DatasetStatus, DatasetVersion
```
Bento007, line 6: Will everything be nested under layers?

backend/layers/business/business.py
```diff
@@ -0,0 +1,190 @@
...
+from backend.layers.persistence.persistence import DatabaseProviderInterface
+from backend.layers.thirdparty.crossref_provider import CrossrefProviderInterface
+from backend.layers.thirdparty.step_function_provider import StepFunctionProviderInterface
+
+
+@dataclass
+class CollectionQueryFilter:
+    from_date: Optional[str]
```
Bento007, line 14: We don't use the from_date or to_date filter any more

backend/layers/business/business.py
```diff
@@ -0,0 +1,190 @@
...
+
+@dataclass
+class UserInfo:
+    user_id: str
+    token: str
+
+
+class BusinessLogicInterface:
+
+    # Get_collections
```
Bento007, line 28: this should be a docstring under get_collections

```python
def get_collections():
    """ 
        Replaces get_collections_list and get_collections_index
        Accepts a CollectionFilter class (or kwargs) with:
        Date bounds
        Visibility
        Ownership
        List of fields to be returned
        Accepts an UserInfo object
        Returns a list of dictionaries that only includes the selected fields for each collection
        It should NOT add any information that is required by the current API for compatibility reasons (e.g. access_write). These will be delegated the upper layer
        It should NOT do any operation that is required by Curation API assumptions (e.g. remove None values)
"""
```

Bento007: reading these comments and how mentions the soon to be deprecaited functions, it reads more like notes for future implementation not actual documentation. Is that how I should be interpretting it?

backend/layers/business/business.py
```diff
@@ -0,0 +1,190 @@
...
+from typing import Iterable, Optional
+from backend.corpora.common.providers.crossref_provider import CrossrefProvider
+
+from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetArtifact, DatasetStatus, DatasetVersion
+from backend.layers.persistence.persistence import DatabaseProviderInterface
+from backend.layers.thirdparty.crossref_provider import CrossrefProviderInterface
+from backend.layers.thirdparty.step_function_provider import StepFunctionProviderInterface
+
+
+@dataclass
```
Bento007, line 12: I asume these datatypes will need to be reference by adjacent layers, should they be in a common location?

backend/layers/business/business.py
```diff
@@ -0,0 +1,190 @@
...
+
+    database_provider: DatabaseProviderInterface
+    crossref_provider: CrossrefProviderInterface
+    step_function_provider: StepFunctionProviderInterface
+
+    def __init__(
+        self, 
+        database_provider: DatabaseProviderInterface, 
+        crossref_provider: CrossrefProviderInterface,
+        step_function_provider: StepFunctionProviderInterface,
```
Bento007, line 185: is there anything to initialize in `BusinessLogicInterface`, else remove `super().__init__()`

backend/layers/business/business.py
```diff
@@ -0,0 +1,190 @@
...
+    ) -> None:
+        self.crossref_provider = crossref_provider
+        self.database_provider = database_provider
+        self.step_function_provider = step_function_provider
+        super().__init__()
```
Bento007, line 190: why are these class variables?

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,802 @@
...
+import copy
+from unittest.mock import Mock
+from backend.corpora.common.providers.crossref_provider import CrossrefException, CrossrefProvider
+from backend.layers.business.business import BusinessLogic, CollectionQueryFilter, UserInfo
+from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetMetadata, DatasetStatus, Link
+from backend.layers.persistence.persistence import DatabaseProviderInterface
+from backend.layers.thirdparty.crossref_provider import CrossrefProviderInterface
+from backend.layers.thirdparty.step_function_provider import StepFunctionProviderInterface
+from tests.unit.backend.layers.persistence.persistence_mock import DatabaseProviderMock
+
```
Bento007, line 12: add `@unittest.skip()` to test suites you want to skip

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,802 @@
...
+class BaseBusinessLogicTestCase(unittest.TestCase):
+
+    sample_collection_metadata = CollectionMetadata(
```
Bento007, line 15: don't use class level variables in unittests. Move these the setup or in
```
@classmethod
def setUpClass(cls)
```

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,802 @@
...
+        # By default does nothing. Can be mocked by single test cases.
+        self.crossref_provider = CrossrefProviderInterface()
+        self.step_function_provider = StepFunctionProviderInterface()
+
+        self.business_logic = BusinessLogic(
+            database_provider=self.database_provider, 
+            crossref_provider=self.crossref_provider,
+            step_function_provider=self.step_function_provider,
+        )
+        return super().setUp()
```
Bento007, line 55: Remove. You don't need to call super or return it.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,802 @@
...
+        self.database_provider.finalize_collection_version(version.collection_id, version.version_id, published_at)
+        return version
+        
+
+
+class TestCreateCollection(BaseBusinessLogicTestCase):
+
+    def create_collection_ok(self):
+        """
+        A collection can be created using `create_collection`
```
Bento007, line 98: I appreciate the descriptions

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,802 @@
...
+        """
+        cv = self.business_logic.create_collection(self.sample_collection_metadata, self.user_info)
```
Bento007, line 100: what is `cv`?

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,802 @@
...
+        cv_from_database = self.database_provider.get_collection_version(cv.version_id)
```
Bento007, line 101: start tests with `test_`. This helps separate tests from helper code.

Bento007, line 101: in tests where you pass use `NotImplemented`

Bento007 (changes requested):
You're missing some init files in your python modules

backend/layers/business/business.py
ebezzi: We can either put them all in the common folder or keep them closer to the layer which is most logically connected to them. I'm ok with both solutions.

backend/layers/business/business.py
ebezzi: I will get rid of it, thanks. I think @atolopko-czi suggested to use [ABC](https://docs.python.org/3/library/abc.html) which would make interfaces look neater.

backend/layers/business/business.py
ebezzi: What do you think they should be?

tests/unit/backend/layers/business/test_business.py
ebezzi: Sounds good, this was recommended by VSCode, but I see it isn't needed.

tests/unit/backend/layers/business/test_business.py
ebezzi: `collection version`, but admittedly, it's a poor naming choice. I will rename it.

tests/unit/backend/layers/business/test_business.py
ebezzi: 👍

tests/unit/backend/layers/business/test_business.py
ebezzi: This is a bit tricky because these variables are used as default parameters in the subsequent methods. Moving them below isn't gonna allow that. Any ideas?

backend/layers/business/business.py
Bento007: These
```    
database_provider: DatabaseProviderInterface
crossref_provider: CrossrefProviderInterface
step_function_provider: StepFunctionProviderInterface
```

Should be variables set in the `__init__`

tests/unit/backend/layers/business/test_business.py
Bento007: move them to the global context or you can wrap them in another class called Fixtures.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,778 @@
...
+        self.assertEqual(fetched_version.metadata, version.metadata)
+
+class TestGetAllCollections(BaseBusinessLogicTestCase):
+
+    def test_setUp(self) -> None:
+        self.collection1 = self.initialize_private_collection(owner="test_user_1")
+        self.collection2 = self.initialize_private_collection(owner="test_user_2")
+        self.collection3 = self.initialize_public_collection(published_at=datetime(2022, 1, 1))
+        self.collection4 = self.initialize_public_collection(published_at=datetime(2021, 1, 1))
+        self.collection5 = self.initialize_public_collection(published_at=datetime(2020, 1, 1))
```
atolopko-czi, line 221: Nit: I _much_ prefer inlining all test state setup code in the test itself and avoid having a multi-purpose test state.  This state is serving multiple purposes (owner filtering, published filtering, etc), and it will become harder to maintain a correct, shared test state as the tests grow.

backend/layers/business/business.py
```diff
@@ -0,0 +1,186 @@
...
+    # Replaces the current update_collection
+    # Accepts:
+    # Collection_id
+    # UserInfo
+    # A dataclass with the body to be updated
+    # Should validate the body
+    # Should handle DOI updates (re-use the existing logic with minimal refactors)
+    # Can either return nothing or the metadata of the updated collection
+
+    # TODO: body should be a dataclass?
```
atolopko-czi, line 86: can't it be `CollectionMetadata`? If not using that because it also includes `links` then maybe `links` should be treated as `publisher_metadata`, i.e. moved up a level into CollectionVersion and updated via a separate method.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,778 @@
...
+        self.crossref_provider = CrossrefProviderInterface()
+        self.step_function_provider = StepFunctionProviderInterface()
+
+        self.business_logic = BusinessLogic(
+            database_provider=self.database_provider, 
+            crossref_provider=self.crossref_provider,
+            step_function_provider=self.step_function_provider,
+        )
+
+    def initialize_empty_private_collection(self, owner: str = test_user_name, metadata = sample_collection_metadata) -> CollectionVersion:
```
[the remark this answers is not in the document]
atolopko-czi: could move these to a fixtures module as well

atolopko-czi (commented):
Partial review.  More to come.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,778 @@
...
+        self.assertEqual(2, len(version_from_db.datasets))
+
+        # Get collection retrieves the new version (with one datasets)
+        version = self.business_logic.get_collection(public_collection.collection_id, self.user_info)
+        self.assertEqual(version.version_id, new_version.version_id)
+        self.assertEqual(1, len(version.datasets))
+        self.assertEqual(version.datasets[0].version_id, dataset_version_to_keep)
+
+
+    def test_publish_version_with_added_dataset_ok(self):
```
atolopko-czi, line 672: All of the `test_publish_version_with_*` tests are good to have, as a black box test. In practice, we're no longer doing all of the sophisticated copying and deleting of records that occurs in the current system. So from a white box testing perspective, having all of these tests doesn't add much, right? A "publish" operation is just a simple collection version attribute and collection mapping update, right?

I'm not suggesting we should remove them, but rather trying to ensure my understanding is correct.

backend/layers/business/business.py
ebezzi: For now, `UserInfo` is mostly a stub, I still need to figure out how do we want to shape it (hence my skipping the authorization tests)

tests/unit/backend/layers/business/test_business.py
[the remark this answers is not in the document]
ebezzi: I originally thought about doing it, but I ended up preferring this method. That said, most of the time we're calling the database provider to get a version, and the business layer will have a semantically equivalent function, so maybe it's not a bad idea.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -0,0 +1,778 @@
...
+        self.assertIsNotNone(fetched_version.published_at)
+        self.assertEqual(fetched_version.metadata, version.metadata)
+
+    def test_get_collection_unpublished_fail(self):
+        """
+        An unpublished collection should not be obtained using get_collection
+        Instead, `get_collection_version` should be used
+        """
+        version = self.initialize_private_collection()
+        fetched_version = self.business_logic.get_collection(version.collection_id, self.user_info)
```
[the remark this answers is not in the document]
ebezzi: Those are all great points. I will defer the tombstone test for now.

tests/unit/backend/layers/business/test_business.py
ebezzi: I'm not sure I'm following. This state is only pertinent to `TestGetAllCollections`, and it's immutable, so I don't see what can cause issues. Also, inlining the state in each test (rather than the unit) won't work because it will keep adding collections to the database. I apologize if I didn't get your concern.

backend/layers/business/business.py
ebezzi: Unfortunately no, `owner` cannot be changed via an update.

tests/unit/backend/layers/business/test_business.py
ebezzi: You're correct, but I still think these don't hurt (as you specified).

tests/unit/backend/layers/business/test_business.py
atolopko-czi: Suggested code changes on https://github.com/chanzuckerberg/single-cell-data-portal/pull/3505

tests/unit/backend/layers/business/test_business.py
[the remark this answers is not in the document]
atolopko-czi: provided the method rename change as suggestion on PR https://github.com/chanzuckerberg/single-cell-data-portal/pull/3505

tests/unit/backend/layers/business/test_business.py
ebezzi: 👍

Issue #3490, opened 2022-10-25
Add backoffs + retry logic to flakey functional test POST requests


nayib-jose-gloria:
There are a few POST calls in our functional tests that can cause a false negative test failure due to timeouts related to competing resources during large uploads or a backlogged upload queue, or transient server errors.  Currently, POST calls are not retried by default, as POST requests are not designed to be idempotent. Given that our solution for this is to manually re-run the functional tests GHA until they pass, we should refactor these to use our regular backoff + retry config for functional test requests.

Example: https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/3316320791/jobs/5478101287

nayib-jose-gloria:
closed: already addressed by https://github.com/chanzuckerberg/single-cell-data-portal/pull/3291

PR #3498, opened 2022-10-26
fix: apply functional test retry config towards POST requests

nayib-jose-gloria:

- #3490

## Changes
- Updated Retry config object in BaseFunctionalTest class to apply retry logic on POST requests

## Notes for Reviewer
By default, the Retry config object we pass into the requests.Session for our functional tests only applies towards a limited set of HTTP methods (see docs [here](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry)), which does not include 'POST' requests because they are not idempotent by default. As a result, we are not retrying POST requests in our functional tests, which increases flakiness on transient server/gateway errors and/or busy queues. Because our solution in these cases of false negatives is to retry the entire functional test suite anyway until it works, it should be fine to add 'POST' requests to the whitelist of retry config allowed methods (some of our POST requests, such as the WMG query endpoints, are idempotent anyway). In the worst-case scenario, we upload 7 repeat collections/datasets in the dev or staging environment (as functional tests don't run in prod), which we also do for smoke tests.

Please flag if you think this is dangerous, or would prefer a separate set of retry config values be applied to POST requests (i.e. lower max retries, longer backoff, etc.)

### DIFF
--- a/tests/functional/backend/common.py
+++ b/tests/functional/backend/common.py
@@ -31,7 +31,12 @@ class BaseFunctionalTestCase(unittest.TestCase):
         cls.deployment_stage = os.environ["DEPLOYMENT_STAGE"]
         cls.config = CorporaAuthConfig()
         cls.session = requests.Session()
-        retry_config = Retry(total=7, backoff_factor=2, status_forcelist=[500, 502, 503, 504])
+        retry_config = Retry(
+            total=7,
+            backoff_factor=2,
+            status_forcelist=[500, 502, 503, 504],
+            allowed_methods=frozenset({"DELETE", "GET", "HEAD", "OPTIONS", "PUT", "TRACE", "POST"}),
+        )
         cls.session.mount("https://", HTTPAdapter(max_retries=retry_config))
         token = cls.get_auth_token(cls.config.test_account_username, cls.config.test_account_password)
         cls.curator_cookie = cls.make_cookie(token)
--- a/tests/functional/backend/common.py
+++ b/tests/functional/backend/common.py
@@ -31,11 +31,13 @@ class BaseFunctionalTestCase(unittest.TestCase):
         cls.deployment_stage = os.environ["DEPLOYMENT_STAGE"]
         cls.config = CorporaAuthConfig()
         cls.session = requests.Session()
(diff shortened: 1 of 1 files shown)


tests/functional/backend/common.py
```diff
@@ -31,7 +31,12 @@ def setUpClass(cls):
         cls.deployment_stage = os.environ["DEPLOYMENT_STAGE"]
         cls.config = CorporaAuthConfig()
         cls.session = requests.Session()
-        retry_config = Retry(total=7, backoff_factor=2, status_forcelist=[500, 502, 503, 504])
+        retry_config = Retry(
+            total=7,
+            backoff_factor=2,
+            status_forcelist=[500, 502, 503, 504],
+            allowed_methods=frozenset({"DELETE", "GET", "HEAD", "OPTIONS", "PUT", "TRACE", "POST"}),
```
atolopko-czi, line 38: Would add a comment, similar to this PR's description, justifying why its okay to include the non-idempotent POST here.  Also worth noting that we're just adding POST to Retry's DEFAULT_ALLOWED_METHODS.  Could write that as `DEFAULT_ALLOWED_METHODS.add("POST")`, but not sure that's better.

Bento007: You don't need frozenset. Let's only keep the methods we actually use to avoid creating confusion. Remove `TRACE` or `OPTIONS`.
```suggestion
            allowed_methods={"DELETE", "GET", "HEAD", "PUT" "POST"},
```

atolopko-czi (approved):
> Please flag if you think this is dangerous, or would prefer a separate set of retry config values be applied to POST requests (i.e. lower max retries, longer backoff, etc.)

I think this is a reasonable solution (and not dangerous, afaict), though I haven't reviewed all the places where POST requests are being made in our tests. I suppose there could be places where a POST request is processed, but the response is never sent, and somehow this means the test is doomed to fail even if the next POST retry is successful. But if that happens, it means we'd just have to rewrite the test to be more robust.

The retry/backoff args seems ok.  I think it's either 2 or 4 mins of total waiting time (depending how retries are being counted).  Longer waits for retries seem slightly better than having to re-run the GHA manually.

PR #3505, opened 2022-10-28
business layer unit tests PR suggestions as code changes

atolopko-czi:
CR feedback, as code change suggestions, for PR https://github.com/chanzuckerberg/single-cell-data-portal/pull/3463

- suggest replacing "visibility" terminology with published/unpublished, since I think it's clearer and reduces number of business layer concepts by 1; API layer can keep "visibility" concept, as needed
- suggest keeping unit test setup state inlined for get_collection w/filtering tests
- a couple method name rename suggestions and a doc comment or two.
- adds business layer exceptions stub classes
- minor formatting (newlines mostly, to keep my IDE happy)

### DIFF
--- a/backend/layers/business/exceptions.py
+++ b/backend/layers/business/exceptions.py
@@ -1,0 +1,23 @@
+class BusinessException(Exception):
+    # TODO maybe useful as a base class, maybe not
+    pass
+
+
+class CollectionCreationException(BusinessException):
+    pass
+
+
+class CollectionUpdateException(BusinessException):
+    pass
+
+
+class InvalidLinkException(BusinessException):
+    pass
+
+
+class DatasetIngestException(BusinessException):
+    pass
+
+
+class CollectionPublishException(BusinessException):
+    pass
--- a/tests/unit/backend/layers/business/test_business.py
+++ b/tests/unit/backend/layers/business/test_business.py
@@ -1,14 +1,17 @@
-from datetime import datetime
 import unittest
-import copy
+from datetime import datetime
 from unittest.mock import Mock
-from backend.corpora.common.providers.crossref_provider import CrossrefException, CrossrefProvider
-from backend.layers.business.business import BusinessLogic, CollectionQueryFilter, UserInfo
-from backend.layers.common.entities import CollectionMetadata, CollectionVersion, DatasetMetadata, DatasetStatus, Link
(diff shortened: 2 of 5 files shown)


backend/layers/business/exceptions.py
```diff
@@ -0,0 +1,23 @@
+class BusinessException(Exception):
+    # TODO maybe useful as a base class, maybe not
```
ebezzi, line 2: Good idea regardless

tests/unit/backend/layers/business/test_business.py
```diff
@@ -172,111 +172,146 @@ def test_create_collection_unauthorized_fail(self):
...
-        self.collection2 = self.initialize_private_collection(owner="test_user_2")
-        self.collection3 = self.initialize_public_collection(published_at=datetime(2022, 1, 1))
-        self.collection4 = self.initialize_public_collection(published_at=datetime(2021, 1, 1))
-        self.collection5 = self.initialize_public_collection(published_at=datetime(2020, 1, 1))
-        return super().setUp()
 
-    def test_get_all_collections_ok(self):
+class TestGetAllCollections(BaseBusinessLogicTestCase):
+
+    def test_get_all_collections_unfiltered_ok(self):
```
ebezzi, line 235: I still think this isn't gonna work. The created collections will belong to the lifetime of `TestGetAllCollections` and aren't reset after each test case, so each subsequent test case will have to rely on what's being created by the previous tests, which doesn't make a lot of sense. If we want to do this, we need to split `TestGetAllCollections` into multiple units. Am I wrong?

ebezzi (changes requested):

tests/unit/backend/layers/business/test_business.py
atolopko-czi: yes, sorry, I meant to comment that every test should re-init the persistence layer.  that should be cheap with our mock persistence layer, right?

atolopko-czi: `BaseBusinessLogicTestCase` could do that in a setup method perhaps

ebezzi: I am not sure using a `.reset()` method is a great idea - what about splitting into separate test cases?

Issue #3401, opened 2022-10-06
chore: Delete "n_cells" column from the expression summary cube


atarashansky:
Scenario:

Let's say I am querying the expression summary cube by:
 - a cell type
 - a tissue
 - an organism
 - all genes

The result is the following:
<img width="1018" alt="image" src="https://user-images.githubusercontent.com/16548075/194371374-8dd63f6d-848e-4fa7-8161-b4bf5d2d5886.png">

Now, let's say I wish to group by genes and sum to get the total number of cells, expression, etc:
<img width="602" alt="image (1)" src="https://user-images.githubusercontent.com/16548075/194371545-5edef260-754d-4a4c-9f41-d7ccaacb16e9.png">

In parallel, I did the same query for the cell counts cube, grouped by the cell type, and summed:
<img width="470" alt="image (2)" src="https://user-images.githubusercontent.com/16548075/194371652-01ad107c-4d52-45a5-ab3b-b27486ee7834.png">

As you can see, each row in the second image has a different value for "n_cells" and these values are less than the total number of cells identified in the cell counts cube (third image).

This is because of [this line](https://github.com/chanzuckerberg/single-cell-data-portal/blob/400cc28df8429a47b38bf9415b7cf26f0f4afedd/backend/corpus_asset_pipelines/summary_cubes/expression_summary/load.py#L49). Basically, if for a particular combination of filters a gene is not detected whatsoever, it is omitted from the dataframe.

As a result, when constructing the dot plot dataframe in the WMG backend (where we aggregate by genes, tissues, and cell types), the "n_cells" column is incorrect (compare to `n_cells_cell_type`):
<img width="668" alt="image" src="https://user-images.githubusercontent.com/16548075/194372613-a16f5810-78e4-447b-8b2e-7080f982229d.png">

In fact, the "n_cells" column isn't even used anywhere, so it is both misleading and irrelevant. Thus, we should get rid of it from the `expression_summary` cube.

atolopko-czi:
Yes, it can and should be deleted! I believe "misleading and irrelevant" captures its essence. :)

PR #3508, opened 2022-10-28
feat(FMG): Adding new data artifacts to the WMG snapshot for marker gene computation.

atarashansky:
- #3520 

This PR adds new data artifacts to the WMG snapshot for marker gene computation. Please see the [Schema changes](https://docs.google.com/document/d/1gDx3Ad3b0uvvsXzaRFsKAeBAUkzH_EPrqwZv66c7k3E/edit#heading=h.nn44ej8p8r1b) section in the tech spec for more details.

This PR also addresses #3401 by deleting the "n_cells" attribute from `expression_summary` cube. This attribute was completely unused and so it was safe to delete.


backend/wmg/data/schemas/cube_schema.py
```diff
@@ -42,13 +71,26 @@
...
+expression_summary_fmg_domain = tiledb.Domain(
+    [
+        tiledb.Dim(name=cube_indexed_dim, domain=None, tile=None, dtype="ascii", filters=filters)
+        for cube_indexed_dim in expression_summary_fmg_indexed_dims
+    ]
+)
+
 # The cube attributes that comprise the core data stored within the cube.
 expression_summary_logical_attrs = [
-    tiledb.Attr(name="n_cells", dtype=np.uint32, filters=filters),
```
atolopko-czi, line 47: note: call this change out in the PR, since it's modifying the existing WMG expr summary cube schema

backend/wmg/data/schemas/cube_schema.py
```diff
@@ -42,13 +71,26 @@
...
     tiledb.Attr(name="sum", dtype=np.float32, filters=filters),
 ]
 
+expression_summary_fmg_logical_attrs = [
+    tiledb.Attr(name="sum", dtype=np.float32, filters=filters),
+    tiledb.Attr(name="sqsum", dtype=np.float32, filters=filters),
+    tiledb.Attr(name="nnz", dtype=np.uint64, filters=filters),
+    tiledb.Attr(name="nnz_thr", dtype=np.uint64, filters=filters),
+]
+
```
atolopko-czi, line 93: could we place all `fmg` schema-related together? possibly in a new schema file? might allow us to remove the `fmg` infix and name the module something like `marker_genes_cube_schema.py` for clarity

backend/wmg/data/schemas/cube_schema.py
```diff
@@ -11,6 +11,12 @@
     "organism_ontology_term_id",
 ]
 
+expression_summary_fmg_indexed_dims = [
```
Bento007, line 14: The [spec](https://docs.google.com/document/d/1gDx3Ad3b0uvvsXzaRFsKAeBAUkzH_EPrqwZv66c7k3E/edit#heading=h.nn44ej8p8r1b) says this is the order `[organism_ontology_term_id, tissue_ontology_term_id, cell_type_ontology_term_id]`

backend/wmg/data/schemas/cube_schema.py
Bento007: If you move them to a new file, also copy over some of the comments that explain what the variables are for.

backend/wmg/pipeline/integrated_corpus/job.py
```diff
@@ -42,25 +48,35 @@ def build_integrated_corpus(dataset_directory: List, corpus_path: str):
...
-                h5ad_file_path, corpus_path
-            )  # TODO Can this be parallelized? need to be careful handling global indexes but tiledb has a lock I think
-            gc.collect()
+            try:
+                dataset_id, gene_ids = process_h5ad_for_corpus(h5ad_file_path, corpus_path)
+                # (mdunitz) TODO Can the above be parallelized? need to be careful handling
+                # global indexes but tiledb has a lock I think
+                gc.collect()
+            except ValueError:
+                # this is needed for a test to pass, otherwise "process_h5ad_for_corpus"
```
Bento007, line 62: I'm guessing it's an unpacking error. You'll need to update the Mock to return None None. Probably [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/6a423183c255737d2a44e40447a91d0ece041a41/tests/unit/wmg_processing/test_load_corpus.py#L68)

```@patch("backend.wmg.pipeline.integrated_corpus.job.process_h5ad_for_corpus", new=Mock(return_value=(None, None))```

File: backend/wmg/pipeline/integrated_corpus/job.py [hunk 5]
```diff
@@ -72,7 +88,7 @@ def process_h5ad_for_corpus(h5ad_path: str, corpus_path: str):
     """
     dataset_id = should_load_dataset(h5ad_path, corpus_path)
     if not dataset_id:
-        return
+        return None, None
```
Bento007, line 91: Update the type hinting and the docstring to describe what is returned

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -277,6 +305,32 @@ def create_expression_summary_cube(
...
+    coords,
+    dim_values,
+    expression_summary_fmg_vals_fn: Callable[[List[tuple]], Dict[str, List]] = random_expression_summary_fmg_values,
+) -> str:
+    cube_dir = f"{data_dir}/expression_summary_fmg"
+    tiledb.Array.create(cube_dir, expression_summary_fmg_schema, overwrite=True)
+
+    with tiledb.open(cube_dir, mode="w") as cube:
+        logical_attr_values = expression_summary_fmg_vals_fn(coords)
+        assert all(
```
Bento007, line 319: Is there a reason this assertion would fail?

Bento007 (changes requested):

backend/wmg/data/schemas/cube_schema.py
atarashansky: good catch, spec was outdated.

tests/unit/backend/wmg/fixtures/test_snapshot.py
atarashansky: no idea, just copied what was written for `create_expression_summary_cube`.

backend/wmg/pipeline/integrated_corpus/job.py
atarashansky: This caused an error when I tried it 

Required argment "consolidate" missing

Bento007: Ah, that is because the test is exting 3 parameters but it only receiving 2 since we added `new=Mock(...)` for `process_h5ad_for_corpus`. Revert the patch line and at the top of the test function body do:
```
mock_process_h5ad.return_value = None, None
```

tests/unit/backend/wmg/fixtures/test_snapshot.py
Bento007: That's a lot of repeated code. You can generalize the original `create_expression_summary_cube` to handle both cases and avoid the copy-paste.

Bento007 (changes requested):

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/load.py
```diff
@@ -0,0 +1,77 @@
...
+    gene_ids: pd.DataFrame,
+    cube_index: pd.DataFrame,
+    other_cube_attrs: list,
+    cube_sum: np.ndarray,
+    cube_sqsum: np.ndarray,
+    cube_nnz: np.ndarray,
+    cube_nnz_thr: np.ndarray,
+):
+    """
+    Build the cube in memory, calculating the gene expression value for each combination of attributes
```
Bento007, line 19: I see a lot of duplication between the two `build_in_mem_cube`. Have you tried factoring out reusable portions?

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/transform.py
```diff
@@ -0,0 +1,123 @@
...
+    Reduce the expression data stored in the integrated corpus by summing it by gene for each cube row (unique combo
+    of cell attributes)
+    """
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        cfg = {
+            "py.init_buffer_bytes": 512 * MB,
+            "py.exact_init_buffer_bytes": "true",
+        }
+        with tiledb.open(f"{tdb_group}/{INTEGRATED_ARRAY_NAME}", ctx=create_ctx(config_overrides=cfg)) as expression:
+            iterable = expression.query(return_incomplete=True, order="U", attrs=["rankit"])
```
Bento007, line 63: this variable it too general. Maybe query_results.

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/transform.py
```diff
@@ -0,0 +1,123 @@
...
+            future = None
+            for i, result in enumerate(iterable.df[:]):
+                logger.info(f"reduce integrated expression data, iter {i}")
```
Bento007, line 66: nit
```suggestion
                logger.info(f"reduce integrated expression data, {i=}")
```

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/transform.py
```diff
@@ -0,0 +1,123 @@
...
+                if future is not None:
+                    future.result()  # forces a wait
+                future = executor.submit(
```
Bento007, line 69: Is this actually executing in parallel? It looks like we are waiting for the previous thread to return before starting the next. Are there some dependencies between the threads that prevent them from running at the same time?

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/load.py
atarashansky: I tried, but it's a bit difficult because it makes the required input arguments overly verbose. I will think about it more to see if we can get a clean abstraction. Otherwise, I don't see the harm in having similar functions living in isolated modules. It's the most straightforward for grokking the code, imo.

tests/unit/backend/wmg/fixtures/test_snapshot.py
atarashansky: As I pointed out in another comment, I don't think generalizing repeated code into a function that can handle both cases is necessarily better if it makes function input overly verbose. There are a lot of schema-related variables ("_fmg") that would need to be parametrized as input arguments.

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/transform.py
atarashansky: I'm not exactly sure if this is going row-by-row in the target array for the reduction. It looks like it might not be. In that case, we want the wait to avoid the race condition. I'm not sure why it's being executed in a thread-pool to begin with.

backend/wmg/pipeline/summary_cubes/expression_summary_fmg/load.py
Bento007: As long as making a change in one function will not require to make the same change in the other.

Bento007 (approved):
LGTM

Issue #3454, opened 2022-10-18
Check if we can move post deployed tests to run pre deployed too


tihuan:
Still needs to be tested in deployed envs:
1. WMG

Can be moved to run pre deployed:
1. DOI? And all other logged in collection tests?

PR #3516, opened 2022-10-31
test(smoketest): 3454 Update test annotations

tihuan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/3454
- #3454

### Reviewers
**Functional:** 
@nayib-jose-gloria 

**Readability:** 
@seve 
---

## Changes
1. Update test skipping with Playwright's [test annotation](https://playwright.dev/docs/test-annotations#conditionally-skip-a-test) syntax, so we can document why specific tests are skipped in pre-deployed or post-deployed 

1. Update some tests to run in more deployed environments

### DIFF
--- a/frontend/tests/features/collection/collection.test.ts
+++ b/frontend/tests/features/collection/collection.test.ts
@@ -1,21 +1,21 @@
 import { expect, Page, test } from "@playwright/test";
+import HTTP_STATUS_CODE from "src/common/constants/HTTP_STATUS_CODE";
 import { ROUTES } from "src/common/constants/routes";
 import { Collection } from "src/common/entities";
-import { API_URL } from "src/configs/configs";
 import { sortByCellCountDescending } from "src/components/Collection/components/CollectionDatasetsGrid/components/DatasetsGrid/common/util";
 import { INVALID_DOI_ERROR_MESSAGE } from "src/components/CreateCollectionModal/components/Content/common/constants";
+import { API_URL } from "src/configs/configs";
 import { BLUEPRINT_SAFE_TYPE_OPTIONS, TEST_URL } from "tests/common/constants";
 import {
-  describeIfDeployed,
-  describeIfDevStaging,
   goToPage,
+  isDevStagingProd,
   login,
   tryUntil,
 } from "tests/utils/helpers";
 import { getTestID, getText } from "tests/utils/selectors";
 import datasets from "../fixtures/datasets";
 
-const { describe } = test;
+const { describe, skip } = test;
 
 const TEST_COLLECTION: CollectionFormInput = {
   contact_email: "TEST@example.com",
@@ -38,7 +38,12 @@ type CollectionFormInput = Pick<
 >;
 
 describe("Collection", () => {
-  describeIfDeployed("Logged In Tests", () => {
+  describe("Logged In Tests", () => {
+    skip(
+      !isDevStagingProd,
+      "Currently push-test runs against dev BE, so login doesn't work for local containers"
+    );
+
     test("creates and deletes a collection", async ({ page }) => {
       const timestamp = Date.now();
       await login(page);
@@ -94,7 +99,9 @@ describe("Collection", () => {
     });
   });
 
-  describeIfDevStaging("Deployed Env Tests", () => {
+  describe("Deployed Env Tests", () => {
+    skip(!isDevStagingProd, "BE DOI endpoints only work in dev, staging, prod");
+
     describe("invalid DOIs", () => {
       test("doesn't create a collection with a DOI in an invalid format", async ({
         page,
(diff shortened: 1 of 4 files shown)


frontend/tests/features/collection/collection.test.ts
```diff
@@ -38,7 +38,7 @@ type CollectionFormInput = Pick<
 >;
 
 describe("Collection", () => {
-  describeIfDeployed("Logged In Tests", () => {
+  describe("Logged In Tests", () => {
```
tihuan, line 41: all the tests nested in this test block can be run in all environments, since we're not publishing a collection

frontend/tests/features/collection/collection.test.ts
```diff
@@ -94,7 +94,9 @@ describe("Collection", () => {
     });
   });
 
-  describeIfDevStaging("Deployed Env Tests", () => {
+  describe("Deployed Env Tests", () => {
+    skip(!isDevStagingProd, "BE DOI endpoints only work in dev, staging, prod");
```
tihuan, line 98: replace `describeIfDevStaging`, but does the same thing. The bonus is that now we can document why the tests are skipped, which also shows up in the test report

frontend/tests/features/collection/collection.test.ts
```diff
@@ -180,13 +182,19 @@ async function showCreateForm(page: Page) {
...
 /**
  * Submit create invalid collection form.
  * @returns Form submit invalid parameters error response (400).
  */
 async function submitCreateFormInvalid(page: Page) {
   return await Promise.all([
-    page.waitForResponse(response => response.url() === `${API_URL}/dp/v1/collections` && response.status() === 400),
+    page.waitForResponse(
+      (response) =>
+        response.url() === collectionEndpoint &&
```
tihuan, line 195: small refactor to keep the code DRY!

frontend/tests/features/collection/collection.test.ts
seve: Won't this lead to failures in `e2e-test` in `push-tests.yml`?

frontend/tests/features/collection/collection.test.ts
seve: what was the initial reason for us avoiding running these tests on prod?

frontend/tests/features/collection/collection.test.ts
tihuan: What kind of failures? The push tests seem to pass for this PR already. Maybe I'm missing something 😆

frontend/tests/features/collection/revision.test.ts
```diff
@@ -2,17 +2,24 @@ import { ElementHandle, expect, Page, test } from "@playwright/test";
...
+const { describe, skip } = test;
+
 const COLLECTION_ROW_ID = "collection-row";
 
-describeIfDeployed("Collection Revision", () => {
+describe("Collection Revision", () => {
+  skip(
+    !isDevStaging,
+    "We only seed published collections for revision test in dev and staging"
+  );
```
seve, line 21: There is a test coverage diff introduced here, these tests will no longer run on rdev and happy

seve: (this actually goes for all tests where we changed `describeIfDeployed` to `skip(!isDevStaging,...)`

seve (changes requested):

frontend/tests/features/collection/collection.test.ts
tihuan: Hmm good question! Since the tests don't publish collections, I'm not sure why they can't run in prod 🤔 

@frano-m do you remember why these tests don't run in prod? Thank you!

frontend/tests/features/collection/revision.test.ts
tihuan: Yeah the change is intentional because revision depends on existing published collections. which I think BE only seeds published collections in dev and staging, thus not available in happy local container or rdev!

frontend/tests/features/collection/collection.test.ts
seve: Push tests are run on a local FE instance with a deployed backend, wouldn't the test runner hit the deployed Auth0 and then get redirected to the deployed FE?

tihuan: ah sorry my bad 🤦 We're only running the tests in dev, staging, and prod as shown in line 42. Just updated the original comment!

seve (approved):

frontend/tests/features/collection/collection.test.ts
tihuan: hmm that's a good point! Maybe so we won't pollute the prod database in any case 🤔 But even if it does, it'd be pretty easy to clean up all private collections associated with the test account once in a while too, and it shouldn't slow down other users' experience too, since it's all private collections?

I'll merge the PR for now 👌 

Thanks all!

Issue #3411, opened 2022-10-07
Review existing tests and create test taxonomy


nayib-jose-gloria:
Create a taxonomy classifying all our test types, and documenting differences in which tests are run per environment. It should cover both single-cell-explorer and single-cell-data-portal. 

Should include at least the following test types and basic definitions for what each is meant to provide coverage for: unit tests, smoke tests (push-tests), smoke tests (post-deployment), functional tests, performance tests, and datadog tests. 

If there are environment-specific tests within any of those test types, flag them and why that is the case. 

The taxonomy should be written such that it can be used as a 'rule of thumb' guide for contributing to test coverage, and to help determine whether we have any redundancies, gaps, or tests that aren't consistent with our taxonomy in our testing ecosystem.

seve:
Assigned to myself with plans to work on this and the blocked issue when I return next week.

PR #3517, opened 2022-10-31
docs(testing): create github workflow taxonomy

seve:
- #3411 

## Changes
- add
   - Create workflow README that explains steps and the drilled down makefile targets + shell scripts 
- remove
- modify
  - Add section to e2e test README regarding skipped tests

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,115 @@
+# Workflow Taxonomy
+
+## Where and what tests are run
+
+---
+
+### Workflow: `deploy-happy-stack.yml`
+
+#### Ran on:
+
+- `deployment`: (deployments are triggered by `push-tests`)
+
+#### Jobs:
+
+- ##### `upgrade`: TODO
+- ##### `e2e-test`:
+  - Runs playwright against deployed frontend app uploading any artifacts
+    - ```bash
+            if [ "${DEPLOYMENT_STAGE}" == "stage" ]; then
+            export DEPLOYMENT_STAGE=staging
+          fi
+          if [ "${DEPLOYMENT_STAGE}" != "prod" ]; then
+            pip3 install -r scripts/smoke_tests/requirements.txt
+            python3 -m scripts.smoke_tests.setup
+          fi
+          cd frontend
+          npm ci
+          npx playwright install --with-deps
+          cp src/configs/${DEPLOYMENT_STAGE}.js src/configs/configs.js
+          DEBUG=pw:api npm run e2e-${DEPLOYMENT_STAGE}
+      ```
+- ##### `functional-test`:
+  - If environment is not `prod` start up backend container and run `make local-functional-test`
+  - `make local-functional-test` -> `python3 -m unittest discover --start-directory tests/functional --top-level-directory . --verbose`
+- ##### `performance-test`:
+  - If environment is `prod` run `make prod-performance-test`
+  - `make prod-performance-test` -> `python3 -m unittest discover --start-directory tests/performance --top-level-directory . --verbose`
+
+### Workflow: `lint-pr.yml`
+
+#### Ran on:
+
+- `pull_request_target`
+
+  - opening, editing, and syncing a pull request
+
+  #### Summary: Runs a GitHub action that lints the PR commit message
+
+### Workflow: `push-processing-base.yml`
+
+TODO
+
+### Workflow: `push-rdev.yml`
+
+TODO
+
+### Workflow: `push-tests.yml`
+
+#### Ran on:
+
+- `push` to `main`, `staging`, `prod`
+- `pull_request` targeting `"*"`
+
+#### Jobs:
+
+- ##### `lint`:
+
+  - Inside backend container: `make lint` -> `flake8 backend tests`
+  - Inside frontend container: (frontend/Makefile)`make lint` -> `npm run prettier-check & npm run lint`
+    - `npm run prettier-check` -> `prettier --check .`
+    - `npm run lint` -> `concurrently \"node_modules/.bin/next lint\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\" \"npm run type-check\"`
+      - `node_modules/.bin/next lint` -> runs eslint with config [next.js docs](https://nextjs.org/docs/basic-features/eslint)
+      - `node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'` -> runs stylelint with config
+      - `npm run type-check` -> `tsc --noEmit`
+
+- ##### `e2e-test`:
+
+  - Installs dependencies and runs frontend server pointing at dev deployment:
+    - `npm ci &&npx playwright install --with-deps && cp src/configs/dev.js src/configs/configs.js && npm run dev&`
+  - Runs e2e tests:
+    - `npm run e2e` ->`playwright test`
+      - runs all tests in `frontend/tests` using playwright config, no TEST_ENV provided defaults to `local`
+  - Pushes images to RCS
+
+- ##### `build-extra-images`: TODO
+
+- ##### `push-prod-images`: TODO
+
+- ##### `backend-unit-test`:
+
+  - Checks if containers need to be rebuilt based on diffs on docker + requirements files
+  - Runs tests in docker-compose: `make local-init-test-data & make all-local-unit-test-backend`
+    - `make local-init-test-data` -> `docker-compose $(COMPOSE_OPTS) run --rm -T backend /bin/bash -c "pip3 install awscli && cd /single-cell-data-portal && scripts/setup_dev_data.sh"`
+    - `make all-local-unit-test-backend` -> (inside backend container) `make container-unittest;`
+      - `make container-unittest` -> `DEPLOYMENT_STAGE=test PYTHONWARNINGS=ignore:ResourceWarning python3 \ -m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose;`
+
+- ##### `processing-unit-test`:
+
+  - Checks if containers need to be rebuilt based on diffs on docker + requirements files
+  - Runs tests in docker-compose
+    - `make local-init-test-data` -> `docker-compose $(COMPOSE_OPTS) run --rm -T backend /bin/bash -c "pip3 install awscli && cd /single-cell-data-portal && scripts/setup_dev_data.sh"`
+    - `make local-unit-test-processing` -> (inside processing container) `make processing-unittest;`
+      - `make processing-unittest` -> `DEPLOYMENT_STAGE=test PYTHONWARNINGS=ignore:ResourceWarning python3 \ -m unittest discover --start-directory tests/unit/processing_container --top-level-directory . --verbose;`
+
+- ##### `wmg-processing-unit-test`:
+
+  - Checks if containers need to be rebuilt based on diffs on docker + requirements files
+  - Runs tests in docker-compose:
+    - `make local-unit-test-wmg-processing` -> (inside wmg_processing container)`make wmg-processing-unittest`
+      - `make wmg-processing-unittest` -> `DEPLOYMENT_STAGE=test PYTHONWARNINGS=ignore:ResourceWarning python3 \ -m unittest discover --start-directory tests/unit/wmg_processing --top-level-directory . --verbose;`
+
+- ##### `push-image`: TODO
+- ##### `create_deployment`: TODO
+
+### Workflow: `scale-test.yml`
--- a/frontend/tests/README.md
+++ b/frontend/tests/README.md
@@ -47,7 +47,6 @@ All the E2E test commands can be found in `frontend/Makefile` and `frontend/pack
 
       2. To run against Deployed BEs,`frontend/src/configs/configs.js` should point to `API_URL: "https://api.cellxgene.dev.single-cell.czi.technology"`, or any other deployed BE API base url
 
-
 1. local -> local (app started with `npm run dev`)
 
    1. Make sure you have your local app running already on http://localhost:3000. If not, in `frontend/` directory, run `npm run dev`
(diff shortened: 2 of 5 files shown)


.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+            export DEPLOYMENT_STAGE=staging
+          fi
+          if [ "${DEPLOYMENT_STAGE}" != "prod" ]; then
+            pip3 install -r scripts/smoke_tests/requirements.txt
+            python3 -m scripts.smoke_tests.setup
+          fi
+          cd frontend
+          npm ci
+          npx playwright install --with-deps
+          cp src/configs/${DEPLOYMENT_STAGE}.js src/configs/configs.js
```
seve, line 29: In what case is this config necessary?

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+          DEBUG=pw:api npm run e2e-${DEPLOYMENT_STAGE}
+      ```
+- ##### `functional-test`:
+  - If environment is not `prod` start up backend container and run `make local-functional-test`
```
seve, line 33: Am I crazy or does this not actually run in the container that's started up? https://github.com/chanzuckerberg/single-cell-data-portal/blob/ee3e590bab9fb77fd814c28df8a13fc4d38b1c83/.github/workflows/deploy-happy-stack.yml#L218-L219

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+#### Jobs:
+
+- ##### `lint`:
+
+  - Inside backend container: `make lint` -> `flake8 backend tests`
+  - Inside frontend container: (frontend/Makefile)`make lint` -> `npm run prettier-check & npm run lint`
+    - `npm run prettier-check` -> `prettier --check .`
+    - `npm run lint` -> `concurrently \"node_modules/.bin/next lint\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\" \"npm run type-check\"`
+      - `node_modules/.bin/next lint` -> runs eslint with config [next.js docs](https://nextjs.org/docs/basic-features/eslint)
+      - `node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'` -> runs stylelint with config
```
seve, line 73: Isn't running this with `--fix` dangerous? We don't actually want to fix the lint failures in this workflow. Just return lint errors

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+      - `npm run type-check` -> `tsc --noEmit`
+
+- ##### `e2e-test`:
+
+  - Installs dependencies and runs frontend server pointing at dev deployment:
+    - `npm ci &&npx playwright install --with-deps && cp src/configs/dev.js src/configs/configs.js && npm run dev&`
+  - Runs e2e tests:
+    - `npm run e2e` ->`playwright test`
+      - runs all tests in `frontend/tests` using playwright config, no TEST_ENV provided defaults to `local`
+  - Pushes images to RCS
```
seve, line 83: Why push images as part of the e2e-test step? Seems like an overextension.

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
+# Workflow Taxonomy
+
+## Where and what tests are run
```
seve, line 3: Is this document straight forward on how to skim? I tried to drill down as far as it made sense to and expose what was actually being called by these workflows.

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+  - Inside frontend container: (frontend/Makefile)`make lint` -> `npm run prettier-check & npm run lint`
+    - `npm run prettier-check` -> `prettier --check .`
+    - `npm run lint` -> `concurrently \"node_modules/.bin/next lint\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\" \"npm run type-check\"`
+      - `node_modules/.bin/next lint` -> runs eslint with config [next.js docs](https://nextjs.org/docs/basic-features/eslint)
+      - `node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'` -> runs stylelint with config
+      - `npm run type-check` -> `tsc --noEmit`
+
+- ##### `e2e-test`:
+
+  - Installs dependencies and runs frontend server pointing at dev deployment:
```
tihuan, line 78: nit: since we have FE and BE servers, maybe the following will be more explicit?

```suggestion
  - Installs dependencies and runs frontend server pointing at dev env BE API:
```

.github/workflows/README.md
tihuan: My understanding is that Infra helped us set up this way, so we only push images that pass the tests. And thus actual deployment will be shorter since the images will already be available in ECS (instead of pushing images at deployment time)

.github/workflows/README.md
nayib-jose-gloria: src/configs/configs.js defines which backend API base URL to use for any backend calls (local vs. a deployed env) + a set-up config for plausible analytics

.github/workflows/README.md
tihuan: it's pretty straight forward to skim 🎉  Thanks for doing this 😄 

I guess my only concern is that if any of the code implementation changes, we'd need to update the docs as well, so we'd sort of have two sources of truth. But I don't think this part of code changes often, so probably not a huge concern?

.github/workflows/README.md
nayib-jose-gloria: It does, the make command does some set-up and runs the actual tests in the docker container. 

I do wonder what the point of running this post-deploy is if it's being run against a docker container anyway, though...

.github/workflows/README.md
tihuan: Hmm great point! Maybe we should remove `--fix` here and add `--fix` to both `next lint` and `stylelint` in `npm run format`?

Thanks for catching this!

frontend/tests/README.md
```diff
@@ -129,3 +128,103 @@ All the E2E test commands can be found in `frontend/Makefile` and `frontend/pack
       1. For more information, visit [Playwright Inspector](https://playwright.dev/docs/debug)
 
       ![Playwright Inspector](https://user-images.githubusercontent.com/6309723/196775284-81bd4853-b4c6-45c4-827b-22e97b97ee06.png)
+
+---
+
+# Test/Env Differential
+
+Where tests are skipped vs. run in different environments.
```
tihuan, line 136: I have a [PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3516) that annotates e2e tests that run in different envs, not sure if it's worth mentioning that here?

frontend/tests/utils/helpers.ts
```diff
@@ -7,6 +7,9 @@ const { describe, skip } = test;
 
 export const TIMEOUT_MS = 3 * 1000;
 
+// (seve): We use TEST_ENV to describe the environment that playwright is running against. Sometimes the FE tests are run against a local instance of the app which points at a deployed instance of the backend.
+// We might want to rename some of these functions to better convey this.
+
```
tihuan, line 12: This function is removed in [this PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3516)!

.github/workflows/README.md
seve: but the frontend is never started, so there's no reading of the config? @nayib-jose-gloria @tihuan

.github/workflows/README.md
nayib-jose-gloria: I have a similar concern; while nice to have the code snippets near the docs, I actually think it'd be easier to maintain and similarly clear if we replaced code snippets with descriptions of what each type of test is meant to accomplish and whether it's run in a deployed env vs. a container

.github/workflows/README.md
seve: ah I misread the Dockerfile, was looking at the wrong target

tihuan: Huh yeah, could have been a copy/paste error. Probably should run functional test against the deployed BE 😆 Great catch!

nayib-jose-gloria: Might be good to include that implementation detail here for clarity

tihuan: Ah interesting! Would it make sense to drop `local-` in `local-functional-test`?

.github/workflows/README.md
seve: This should probably be brought up to a larger visibility for the team: 

I think we can still push images in this workflow, but made it a separate run that relies on a positive e2e test run. Wouldn't that still meet that same functionality while allowing e2e tests to run quicker and potentially fail quicker?

Ideally, we'd refactor/shrink these workflow jobs to contain only the context and operations relative to a single operation (lint, functional test, smoke test, build, push, etc.)

.github/workflows/README.md
nayib-jose-gloria: I believe its playwright uses it as well for certain tests, at least the API_URL

nayib-jose-gloria: we can probably refactor that so it isn't the case, but as-is we'd need to change a few things

.github/workflows/README.md
tihuan: You mean updating the `if` in the `Push images` block to only push images when e2e tests succeed?

```
- name: Push images
        // Add extra condition to only push when success?
        if: (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/prod' ) 
        run: |
          echo "HAPPY_COMMIT=$(git rev-parse --verify HEAD)" >> envfile
          echo "HAPPY_BRANCH=$(git branch --show-current)" >> envfile
          happy push --docker-compose-env-file envfile --aws-profile "" --tag sha-${GITHUB_SHA:0:8} frontend backend
```

.github/workflows/README.md
tihuan: Ah test URL is configured through `TEST_ENV`, so it doesn't use `API_URL` I think!

```ts
const TEST_ENV_TO_TEST_URL = {
  dev: "https://cellxgene.dev.single-cell.czi.technology",
  happy: "http://frontend.corporanet.local:3000",
  local: "http://localhost:3000",
  localProd: "http://localhost:9000",
  prod: "https://cellxgene.cziscience.com",
  rdev: process.env.RDEV_LINK || "",
  staging: "https://cellxgene.staging.single-cell.czi.technology",
};
```

nayib-jose-gloria: here's one example: https://github.com/chanzuckerberg/single-cell-data-portal/blob/d5c6377d407957db32e82bf5152cea943680d02b/frontend/tests/features/collection/collection.test.ts#L4

nayib-jose-gloria: those TEST_ENV_TO_TEST_URL values also aren't the base URLs you'd use to wait for a response from a particular backend API endpoint, which is how we fixed the flakey invalid DOI tests

.github/workflows/README.md
nayib-jose-gloria: Yeah, that's probably appropriate

frontend/tests/README.md
```diff
@@ -129,3 +128,103 @@ All the E2E test commands can be found in `frontend/Makefile` and `frontend/pack
...
+    <td>skip</td>
+    <td>skip</td>
+    <td>describe</td>
+  </tr>
+
+  <tr>
+    <td>
+      <ul>
+         <li>`./features/wheresMyGene.test.ts`</li>
+         <li>"invalid DOIs"</li>
```
nayib-jose-gloria, line 201: I believe Invalid DOIs are skipped in prod also

.github/workflows/README.md
seve: it would be pulling Push images out of the `e2e-test` job and into its own job with a shape similar to:

```
push-happy-backend-frontend
    needs:
      - backend-unit-test
      - e2e-test
    runs-on: ubuntu-20.04
    if: (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/prod' )
    steps:
      ...
      - name: Push images
        run: |
          echo "HAPPY_COMMIT=$(git rev-parse --verify HEAD)" >> envfile
          echo "HAPPY_BRANCH=$(git branch --show-current)" >> envfile
          happy push --docker-compose-env-file envfile --aws-profile "" --tag sha-${GITHUB_SHA:0:8} frontend backend
       ...
```

While writing this out, I noticed that we're also pushing the backend image here.

.github/workflows/README.md
seve: That makes sense to me!

.github/workflows/README.md
tihuan: Ahh I see! Sorry totally missed that!

What if we remove `API_URL` import in tests and update the assertion to be:

```ts
const collectionEndpointPath = `dp/v1/collections`; // <-- removed API_URL

/**
 * Submit create invalid collection form.
 * @returns Form submit invalid parameters error response (400).
 */
async function submitCreateFormInvalid(page: Page) {
  return await Promise.all([
    page.waitForResponse(
      (response) =>
        response.url().includes(collectionEndpointPath) &&  // <-- update to match partial path
        response.status() === HTTP_STATUS_CODE.BAD_REQUEST
    ),
    page.click(getTestID("create-button")),
  ]);
}
```

tihuan: If that looks good I can add the change in this PR!

https://github.com/chanzuckerberg/single-cell-data-portal/pull/3516

.github/workflows/README.md
```diff
@@ -0,0 +1,115 @@
...
+  - Inside frontend container: (frontend/Makefile)`make lint` -> `npm run prettier-check & npm run lint`
+    - `npm run prettier-check` -> `prettier --check .`
+    - `npm run lint` -> `concurrently \"node_modules/.bin/next lint\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\" \"npm run type-check\"`
+      - `node_modules/.bin/next lint` -> runs eslint with config [next.js docs](https://nextjs.org/docs/basic-features/eslint)
+      - `node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'` -> runs stylelint with config
+      - `npm run type-check` -> `tsc --noEmit`
+
+- ##### `e2e-test`:
+
+  - Installs dependencies and runs frontend server pointing at dev env BE API:
```
nayib-jose-gloria, line 78: nit: mind adding 'local' frontend server? just for easy cross-reference to the TEST_ENV table in the tests readme

frontend/tests/README.md
```diff
@@ -129,3 +128,103 @@ All the E2E test commands can be found in `frontend/Makefile` and `frontend/pack
...
+<col >
+<col >
+<col >
+<col >
+<col >
+</colgroup>
+<thead>
+  <tr>
+    <th></th>
+    <th colspan="7">TEST_ENV (defaults to local)</th>
```
nayib-jose-gloria, line 152: Great work on this table, this will be really useful to have handy

What do you think of also adding a blurb below or up-top briefly summarizing why certain tests are Dev/Staging Only or Deployed Env only? I think it'd be useful for anyone adding/debugging new tests who is unfamiliar

frontend/tests/README.md
tihuan: this PR will run the test in prod now 👌 

https://github.com/chanzuckerberg/single-cell-data-portal/pull/3516/files

.github/workflows/README.md
nayib-jose-gloria: yeah, that looks good to me!

frontend/tests/README.md
seve: I think it makes more sense to have those at the test level and not here, but maybe a brief summary on potential reasons would work here

.github/workflows/README.md
seve: Do these GH code snippets work in GH markdown? Is that a good middle ground, if so?

.github/workflows/README.md
seve:
```suggestion
  - Installs dependencies and runs local frontend server pointing at dev env BE API:
```

.github/workflows/README.md
tihuan: Awesome! Since that PR is now merged, I'll create a separate PR to address that today 👍 Thank you!

tihuan: PR out! https://github.com/chanzuckerberg/single-cell-data-portal/pull/3531 PTAL thank you 🙏

.github/workflows/README.md
seve: I'm not actually sure if dropping local here makes sense, since it runs the test in “local dev” based on the happy team's definition

.github/workflows/README.md
```diff
@@ -89,25 +79,25 @@ TODO
...
-  - Runs tests in docker-compose: `make local-init-test-data & make all-local-unit-test-backend`
-    - `make local-init-test-data` -> `docker-compose $(COMPOSE_OPTS) run --rm -T backend /bin/bash -c "pip3 install awscli && cd /single-cell-data-portal && scripts/setup_dev_data.sh"`
-    - `make all-local-unit-test-backend` -> (inside backend container) `make container-unittest;`
-      - `make container-unittest` -> `DEPLOYMENT_STAGE=test PYTHONWARNINGS=ignore:ResourceWarning python3 \ -m unittest discover --start-directory tests/unit/backend --top-level-directory . --verbose;`
+  - Runs tests in docker-compose: https://github.com/chanzuckerberg/single-cell-data-portal/blob/6a423183c255737d2a44e40447a91d0ece041a41/.github/workflows/push-tests.yml#L214-L216
+    - https://github.com/chanzuckerberg/single-cell-data-portal/blob/6a423183c255737d2a44e40447a91d0ece041a41/Makefile#L106-L107
+    - https://github.com/chanzuckerberg/single-cell-data-portal/blob/6a423183c255737d2a44e40447a91d0ece041a41/Makefile#L182-L185
+      - https://github.com/chanzuckerberg/single-cell-data-portal/blob/6a423183c255737d2a44e40447a91d0ece041a41/Makefile#L27-L30
 
 - ##### `processing-unit-test`:
```
nayib-jose-gloria, line 87: I think one-liner descriptions for all the test types like 'Sets up docker container to run through a dataset submission job and validates behavior" would be helpful

.github/workflows/README.md
nayib-jose-gloria: yeah I think I've changed my mind on this, ultimately we can obscure details like that as long as we do a good job illustrating the point here

.github/workflows/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Workflow: `deploy-happy-stack.yml`
+
+#### Ran on:
+
+- `deployment`: (deployments are triggered by `push-tests`)
+
+#### Jobs:
+
+- ##### `upgrade`: TODO
```
nayib-jose-gloria, line 15: Mind opening an issue to fill out these remaining job steps and link it in the Improve Testing Confidence Documentation epic?

nayib-jose-gloria (commented):
left a couple notes, looks good!

nayib-jose-gloria (approved):

tihuan (approved):
Wow super easy to read, especially with the code snippets for context! Super awesome work ⭐ Thanks so much, Seve!!

PR #3524, opened 2022-11-01
chore: business layer tests for update collection

ebezzi:


tests/unit/backend/layers/business/test_business.py
```diff
@@ -343,17 +379,21 @@ def test_update_collection_wrong_params_fail(self):
...
         Updating a collection version that is published should fail
         """
         version = self.initialize_published_collection()
+        body = CollectionMetadataUpdate(
+            name="new collection name",
+            description="new collection description",
+            contact_name=None,
+            contact_email=None,
+            links=None,
+        )
```
danieljhegeman, line 393:
```suggestion
        body = CollectionMetadataUpdate(
            name="new collection name",
            description="new collection description",
            contact_name="new contact name",
            contact_email="new_email@czi.com",
        )
```

danieljhegeman: Otherwise you are testing 'missing parameters' at the same time as 'update published'

tests/unit/backend/layers/business/test_business.py
```diff
@@ -310,29 +310,65 @@ def test_update_collection_ok(self):
...
+
+        # These three fields should remain the same
+        self.assertEqual(updated_version.metadata.contact_name, version.metadata.contact_name)
+        self.assertEqual(updated_version.metadata.contact_email, version.metadata.contact_email)
+        self.assertEqual(updated_version.metadata.links, version.metadata.links)
+
 
     def test_update_collection_wrong_params_fail(self):
         """
         Updating a collection with a payload with missing fields should fail
```
danieljhegeman, line 362: this appears to describe the opposite feature of the test directly above

danieljhegeman (approved):
I have to admit, one feature of TDD that I hadn't considered before is that it may require you to start to implement classes before you know exactly how you're going to use them in the actual code.

But this is not really a real problem. Things can always be amended.

🚀

PR #3532, opened 2022-11-03
fix: re-add config.js copy before running post-deploy e2e tests

nayib-jose-gloria:

### Reviewers
**Functional:** 
@tihuan  
**Readability:** 

---

## Changes
- copy config.js over from <deployment_stage>.js before running post-deploy e2e tests

## Notes for Reviewer
- post deploy e2e [tests failed](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/3386128595) due to imported dependency on API_URL from config.js, which we removed in [this commit](https://github.com/chanzuckerberg/single-cell-data-portal/commit/e1a2d7487a2a1d05e6567f87f981f0b3def4c39d). 
- Re-adding config.js for simplicity, it doesn't really affect clarity or performance so I think we should just add it back in vs. trying to remove dependency on it entirely. If you disagree however, let me know

### DIFF
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -140,7 +140,6 @@ jobs:
           cd frontend
           npm ci
           npx playwright install --with-deps
-          cp src/configs/${DEPLOYMENT_STAGE}.js src/configs/configs.js
           DEBUG=pw:api npm run e2e-${DEPLOYMENT_STAGE}
       - uses: actions/upload-artifact@v2
         if: always()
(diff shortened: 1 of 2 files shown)


tihuan:
Hmm yeah looks like some of our tests have transient dependency on API_URL file thus the failure 🤔 TIL!

Thanks for the fix, Nayib! 🙏

.github/workflows/deploy-happy-stack.yml
```diff
@@ -140,6 +140,7 @@ jobs:
           cd frontend
           npm ci
           npx playwright install --with-deps
+          cp src/configs/${DEPLOYMENT_STAGE}.js src/configs/configs.js
```
tihuan, line 143: Can we add a comment to explain that our test files have transient dependency on the configs.js file please? So a few months from now we don't delete it again 😆 Thank you!

PR #3536, opened 2022-11-03
chore: Add api thrash script

tihuan:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/scripts/thrash_api.py
+++ b/scripts/thrash_api.py
@@ -1,5 +1,8 @@
 # Authored by Bruce Martin
-# https://czi-sci.slack.com/archives/C023Q1APASK/p1667438044981199?thread_ts=1667431712.857239&cid=C023Q1APASK
+# The script is used for checking if our /curation/* endpoints stay performant
+# after being hit concurrently in a short period of time
+# Context: https://czi-sci.slack.com/archives/C023Q1APASK/p1667438044981199?thread_ts=1667431712.857239&cid=C023Q1APASK
+# Note: The Slack link will expire around 11/2/2023 or whatever CZI's Slack retention policy entails
 
 import argparse
 import concurrent.futures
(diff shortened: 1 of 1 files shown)


scripts/thrash_api.py
```diff
@@ -0,0 +1,130 @@
+# Authored by Bruce Martin
+# https://czi-sci.slack.com/archives/C023Q1APASK/p1667438044981199?thread_ts=1667431712.857239&cid=C023Q1APASK
```
nayib-jose-gloria, line 2: Since Slack messages will eventually expire (and this is a public repo), I think we should replace this link with a brief description of what this script is for

tihuan: Great idea! Just added a comment 👌 I kept the slack link because our public channels retain messages indefinitely unlike our DMs, so that way CZI members can gain more context. What do you think?

Thank you!

atolopko-czi: I think it's just 1 year for public?

tihuan: @nayib-jose-gloria PTAL again thank you!

nayib-jose-gloria: I don't think anything was added since my last review @tihuan--did you push your local commits?

nayib-jose-gloria (approved):

scripts/thrash_api.py
tihuan: Oh gosh sorry you right lol Just pushed!

@atolopko-czi Oh! I thought it was permanent 🤦 You're right that I'm unable to find any messages older than a year. Thanks for the headsup!

Maybe I'll add a comment to remove the Slack link once it reaches a year old?

tihuan: Added extra comment!

> # Note: The Slack link will expire around 11/2/2023 or whatever CZI's Slack retention policy entails

Issue #3509, opened 2022-10-28
Investigate reoccuring delayed image tagging issue in production deployments


nayib-jose-gloria:
Example: https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/3346659267/jobs/5543738872

Current solution is to retry--consider adding a wait as a stopgap? Investigate root cause and see if there's a more permanent solution.

PR #3537, opened 2022-11-03
fix: wait until prod image tag is pushed before creating deployment

nayib-jose-gloria:

- #3509

## Changes
- wait until prod image is pushed/tagged before creating deployment, as it was consistently causing the first prod upgrade attempt to fail + require a manual retry 
- rename push-images step to add-image-tags to better reflect what its actually doing (adding branch-<main/staging/prod> tag to latest pushed docker images)
- update push-prod-images to only push prod images for prod branch merges. Still runs in dev/staging as a no-op due to limitations with GHA workflow conditional statements

## Notes for Reviewer
- Unfortunately, GHA does not support conditional "needs" statements, so I chose to implement this wait for all environments. Alternatively, I can make a separate, duplicate 'create_deployment' job called "create_prod_deployment" and differentiate behavior per environment. However, that seemed like it might sacrifice workflow clarity to save about 1 extra min of deployment time. 
- Another alternative: do we even need to run push-prod-images on dev/staging? Can we change it to a prod-only step? Currently it runs in all envs.

### DIFF
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -144,7 +144,7 @@ jobs:
 
   push-prod-images:
     needs:
-      - push-image
+      - add-image-tags
     runs-on: ubuntu-20.04
     if: (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/prod')
     steps:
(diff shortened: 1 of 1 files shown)


tihuan (approved):
Wow nice! Thanks for making the deploy more stable, Nayib 🤩🙏! 

Should we add a comment to highlight that the retag image is running on all environments, so we can keep an eye on it for when/if GHA adds conditional needs?

Thanks so much!

danieljhegeman (approved):
>Another alternative: do we even need to run push-prod-images on dev/staging? Can we change it to a prod-only step? Currently it runs in all envs.

I don't see why we would need to run it on dev/staging!


Thanks for making this fix. This issue was a headache.

PR #3556, opened 2022-11-05
refactor: organize Data Portal API handlers

danieljhegeman:
https://github.com/chanzuckerberg/single-cell/issues/350

I've applied the 'handler file path' structure that @Bento007 applied to good effect in our Curation API work. Crucially, I have chosen to (for now) _keep all API paths the same_ so that 
1) no FE changes have to be made, and 
2) no rogue usage of the portal API is disrupted (as some curators may still be doing)

The changes here are to _API handlers_ - their file paths, and their code sharing. This is almost entirely focused on reorganizing the Portal API, which was the most disorganized; the Curation API had largely followed the standard I implement here (though some tweaks were required), and the WMG and Gene Info API's are so small, and not really built around "resources" in any meaningful sense yet, that they are not worth messing with. LET ME KNOW IF YOU DISAGREE ON THIS POINT.

The actual standard is to put handlers in a file path structure that mirrors the API path, with the file itself being named after the last portion of the API path (A route of `../the/last_portion` results in handler being in `../the/last_portion.py`) EXCEPT in cases where subsequent directories would conflict in name with a handler file, such as when both `../this/path` and also `../this/path/exist`, in which case handlers for the former reside in `../this/path/actions.py`.

All handler methods are named according to their HTTP method: GET,  POST, etc.

Together, the above rules for organizing handlers means that it is extremely easy, when looking at a given piece of code, to know exactly what endpoint it is tied to, and it also makes the sharing of logic (in distributed `common.py` files) much more... well, logical.

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


Bento007 (approved):
Excellent job!

PR #3562, opened 2022-11-07
fix(FMG): add backwards compatibility to the WMG snapshot while waiting for schema migration to complete

atarashansky:
#3508 
The linked PR introduced new data artifacts to the snapshot. However, the processing pipeline has been frozen until the schema migration completes. As a result, we need to add some error handling to make sure that the WMG snapshot loads properly even if certain artifacts are missing (specifically, `expression_summary_fmg` and `dataset_to_gene_ids.json`).


backend/wmg/data/snapshot.py
```diff
@@ -106,7 +110,11 @@ def _load_primary_filter_data(snapshot_identifier: str) -> Dict:
 
 
 def _load_dataset_to_gene_ids_data(snapshot_identifier: str) -> Dict:
-    return json.loads(_read_s3obj(f"{snapshot_identifier}/{DATASET_TO_GENE_IDS_FILENAME}"))
+    try:
+        return json.loads(_read_s3obj(f"{snapshot_identifier}/{DATASET_TO_GENE_IDS_FILENAME}"))
+    except:
```
atarashansky, line 115: Anyone know what error to catch here?

nayib-jose-gloria: I believe its botocore.exceptions.ClientError

nayib-jose-gloria: assuming this is meant to return a 404 from aws

nayib-jose-gloria (commented):
looks good, just missing more specificity in that try-except block

nayib-jose-gloria (approved):

Issue #3519, opened 2022-11-01
remove `--fix` in lint and add `--fix` to both `next lint` and `stylelint` in `npm run format`


seve:
Hmm great point! Maybe we should remove `--fix` here and add `--fix` to both `next lint` and `stylelint` in `npm run format`?

Thanks for catching this!

_Originally posted by @tihuan in https://github.com/chanzuckerberg/single-cell-data-portal/pull/3517#discussion_r1010648681_

tihuan:
Thanks for creating this ticket, @seve ! Added to e2e test epic!

PR #3564, opened 2022-11-07
refactor: only formatting code on npm run format instead of npm run lint

ashin-czi:
- #3519

### Reviewers
**Functional:** @seve @nayib-jose-gloria @tihuan 

**Readability:** @seve @nayib-jose-gloria @tihuan 

---

## Changes
- add
- remove
    - `--fix` from `npm run lint` and moved to `npm run format` command
- modify


seve (approved):

frontend/package.json
```diff
@@ -95,7 +95,7 @@
     "build": "NODE_OPTIONS=\"--max_old_space_size=2048\" next build",
     "start": "next start",
     "develop": "next dev",
-    "format": "npx prettier --write .",
+    "format": "concurrently \"npx prettier --write .\" \"node_modules/.bin/next lint --fix\" \"node_modules/.bin/stylelint --fix '**/*.{js,ts,tsx,css}'\"",
```
tihuan, line 98: Ooh thanks for adding `concurrently` here! I think this is a great speedup with probably a small chance of different commands competing to write to the same file? Don't think it'll happen any time soon, but maybe good for us to keep in mind 😆 !

frontend/src/components/common/Filter/common/style.ts
```diff
@@ -35,7 +35,6 @@ export const SelectionIcon = styled(Icon)`
       width: 14px;
     }
   }
-}
```
tihuan, line 38: woah we've had this invalid closing bracket for a while lol 😆 Thanks for fixing this!

tihuan (approved):
LGTM! Thanks so much for fixing this, @ashin-czi 🙏 ✨ 🙏 !

frontend/package.json
ashin-czi: hmmm should we run the commands sequentially then? I think the small speed up might not be worth any collisions 😅

PR #3575, opened 2022-11-08
test(smoketest): 3466-clean-up-e2e-test-commands

tihuan:
- #TICKET_NUMBER

### Reviewers
**Functional:** 
@seve 

**Readability:** 
@nayib-jose-gloria 
---

## Changes
1. Remove commands we don't use
2. Originally I added a few test commands to build the FE app in prod mode and run tests against it, but since we don't have Server Side Rendering (SSR) like CZ Meta, we don't really need to run e2e tests against prod mode FE. Will probably need to revisit if we start having SSR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,14 +58,6 @@ prod-performance-test:
 local-backend:
 	$(MAKE) local-server -C ./backend/api_server
 
-.PHONY: smoke-test-prod-build
-smoke-test-prod-build:
-	$(MAKE) smoke-test-prod-build -C ./frontend
-
-.PHONY: smoke-test-with-local-backend
-smoke-test-with-local-backend:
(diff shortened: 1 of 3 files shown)


seve:
I put so much work into defining these test runners. I am so glad to see them leaving 😈

----

Adding some keywords to boost GitHub SEO in the case that someone goes looking for these:

Run smoke test in container happy smoke test e2e docker local e2e

seve (approved):

tihuan:
Oy sorry 😆 ! Honestly I was debating if we should just keep them, so we have the commands when we have SSR. But ultimately we have them in the commit history anyway, so we can always refer back to find them!

tihuan:
The SEO keywords is a brilliant idea 💡 !!

PR #3577, opened 2022-11-09
feat: api layer

ebezzi:


backend/corpora/common/authorizer.py
```diff
@@ -10,8 +10,9 @@
 from .utils.jwt import jwt_decode, get_unverified_header
 
 auth0_session_with_retry = requests.Session()
-retry_config = Retry(total=3, backoff_factor=1, status_forcelist=CorporaAuthConfig().retry_status_forcelist)
-auth0_session_with_retry.mount("https://", HTTPAdapter(max_retries=retry_config))
+# TODO: these read the configuration on initialization and cause problems with tests
```
ebezzi, line 13: Will reinstate later - we need a way to mock the configuration here

PR #3581, opened 2022-11-10
fix: Fix final wmg processing pipeline success message function

atarashansky:
`stats` is a dictionary of dataset count, gene count, and cell count. It was previously passed as a single argument by mistake. It should be `**stats` .

### DIFF
--- a/backend/wmg/pipeline/cube_pipeline.py
+++ b/backend/wmg/pipeline/cube_pipeline.py
@@ -81,7 +81,7 @@ def main():
     # todo pass in validate_cubes as env arg
     try:
         snapshot_path, stats = load_data_and_create_cube("datasets", ".")
-        success_message = gen_wmg_pipeline_success_message(snapshot_path, stats)
+        success_message = gen_wmg_pipeline_success_message(snapshot_path, **stats)
         notify_slack(success_message)
     except Exception as e:
         logger.exception("Pipeline failed")


backend/wmg/pipeline/cube_pipeline.py
```diff
@@ -81,7 +81,7 @@ def main():
     # todo pass in validate_cubes as env arg
     try:
         snapshot_path, stats = load_data_and_create_cube("datasets", ".")
-        success_message = gen_wmg_pipeline_success_message(snapshot_path, stats)
+        success_message = gen_wmg_pipeline_success_message(snapshot_path, **stats)
```
nayib-jose-gloria, line 84: I think we also need to update the `gen_wmg_pipeline_success_message` method signature to make all the stats args into keyword parameters. Right now they're positional but dictionaries don't maintain order and ** passes as keyword args

nayib-jose-gloria: we can do `gen_wmg_pipeline_success_message(snapshot_path: str, *, dataset_count: int, cell_count: int, gene_count: int) -> dict:`

nayib-jose-gloria: nvm, apparently this isn't necessary after all--positional arguments can act as keyword arguments without the * notation

nayib-jose-gloria (approved):

Issue #3415, opened 2022-10-07
Assess GHA test workflows and address redundancies or issues


nayib-jose-gloria:
Address any redundant or problematic steps in the workflow that triggers after a merge to main.

Known concerns:
- If possible, skip unit tests / smoke tests from 'Push Tests' in main if commits were merged in from a branch (since branches also require passing these same tests in order to merge) and save on deployment time. 
     - We do not want to make dev deployments more susceptible to bugs, so we don't want to outright remove 'push tests' from running in main, in case of a force push direct to main (@seve: we probably can, we're smart enough to know direct pushes to main aren't safe, and we do them in desperation). We may be able to get around this by setting stricter branch protection rules--if so, please flag any branch protection rule changes in this issue. 
- Rename 'Push Test' smoke tests vs. 'Post-Deployment' smoke tests to clarify that a different suite of tests are run in each case
-Only run push tests relative to certain files being changed (@seve: I saw some “check file diff” conditionals in the backend unit test job definitions, reference that when working on this)
- seve: we should pull linting out of containers

Concerns to investigate: 
- Currently, we run post-deployment tests in parallel to optimize for time. We should investigate whether we are degrading test performance or causing increased false negative test failures by running intensive tests in parallel in a deployed environment. If so, we should unparalellize. 
   - (@seve: As long as the parallelization affects are linear, I don't see an issue with having degraded test performance as long as they're documented and accounted for in our tests numbers.)
   
Related Issue: https://github.com/chanzuckerberg/single-cell-data-portal/issues/3526

nayib-jose-gloria:
Changes addressed or made unnecessary due to faster tests in https://github.com/chanzuckerberg/single-cell-data-portal/pull/3788

PR #3587, opened 2022-11-14
feat: refactor push tests to reduce redundant steps

nayib-jose-gloria:

- #3415

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- For each test step, check changed files to determine whether tests need to be run. Will save considerable time in deploying/testing PRs, as each suite of tests is only relevant to changes on a well-defined subset of the codebase.
- Separate frontend/backend docker image pushing step from test suite step
- Misc renames and comments for clarity
- Remove branch conditional relic from feature/schema-3.0.0 work
- Changed checks for any_changed files to any_modified, because any_modified also covers deletions while any_changed does not ([source](https://github.com/marketplace/actions/changed-files#outputs))

## QA steps (optional)
- Dummy commits/reverts to this PR (will squash + merge) to test triggers

### DIFF
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -24,10 +24,10 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.7
+      - name: Set up Python 3
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
       - name: Lint Python
         run: |
           pip3 install -r requirements.txt
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -23,15 +23,16 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     steps:
+      - uses: actions/checkout@v2
       - name: Set up Python 3.7
         uses: actions/setup-python@v2
         with:
           python-version: 3.7
-      - uses: actions/checkout@v2
       - name: Lint Python
         run: |
-           pip3 install flake8
-           make lint
+        python3 -m pip install --upgrade pip
+        pip3 install flake8
+        make lint
       - uses: actions/setup-node@v2
         name: Set up npm
         with:
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -23,25 +23,25 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     steps:
-      - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v1
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-west-2
-          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
-          role-duration-seconds: 1800
-      - name: Login to ECR
-        uses: docker/login-action@v1
-        with:
-          registry: ${{ secrets.ECR_REPO }}
+          python-version: 3.7
       - uses: actions/checkout@v2
       - name: Lint Python
         run: |
-          docker-compose run --no-deps --rm backend make -C /single-cell-data-portal lint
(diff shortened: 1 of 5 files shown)


.github/workflows/push-tests.yml
```diff
@@ -75,26 +75,67 @@ jobs:
...
+      - name: Get changed frontend or backend files
+        id: changed-files
+        uses: tj-actions/changed-files@v23.1
+        with:
+          files: |
+            frontend/src/**
+            frontend/tests/**
+            frontend/playwright*
+            frontend/package*
+            frontend/next*
```
tihuan, line 87: Added a few more files I think we need to watch 😄 Thanks!

```suggestion
            frontend/next*
            frontend/playwright*
            frontend/package*
            frontend/doc-site*
            frontend/Dockerfile
            frontend/Makefile
            frontend/container_init.sh
            frontend/stylelint.config.js
            frontend/tsconfig.json
```

.github/workflows/push-tests.yml
```diff
@@ -75,26 +75,67 @@ jobs:
...
           npm ci
           npx playwright install --with-deps
           cp src/configs/dev.js src/configs/configs.js
           npm run dev&
       - name: Run E2E tests
+        if: steps.changed-files.outputs.any_modified == 'true'
         run: DEBUG=pw:api npm run e2e
       - uses: actions/upload-artifact@v2
-        if: always()
+        # always() ensures this step still runs when e2e tests are cancelled early; steps.changed-files check ensures it only runs if tests did run at all
```
tihuan, line 100: wow really important comment! TIL thank you!

tihuan (approved):
Love the idea of checking diff for test runs 💡 !! Thanks for adding that 🤩 👏 

I added more stuff to check for diff. PTAL!

.github/workflows/push-tests.yml
```diff
@@ -24,25 +23,26 @@ jobs:
...
-          registry: ${{ secrets.ECR_REPO }}
       - uses: actions/checkout@v2
+      - name: Set up Python 3
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
       - name: Lint Python
         run: |
-          docker-compose run --no-deps --rm backend make -C /single-cell-data-portal lint
+          pip3 install -r requirements.txt
```
seve, line 33: Is there a way we can reduce the # of installs here since we're just doing a simple lint?

.github/workflows/push-tests.yml
seve: What files that are not included in this allow list, but are in `/frontend`, do we expect to result in high #'s of “erroneous” test runs? Is it worth having such a verbose allow-list?

seve (changes requested):

.github/workflows/push-tests.yml
tihuan: Hmm good question. Basically you're asking if we should just watch `frontend/*`?

I guess any file changes that don't affect the build we could skip E2E tests? Like *.md. Maybe there's something else 🤔

nayib-jose-gloria: I was under the impression we don't need to run e2e tests on doc site updates, Makefile updates, lint styling config update

Either way, FWIW, I'm going to wait on further work on this PR regardless because we'll need to refactor it for the portal redesign anyway and that structure is going to simplify this workflow

.github/workflows/push-tests.yml
nayib-jose-gloria: I'll play around with it once I get back to work on this, there are at least a few that we need or else flake8 complains

.github/workflows/push-tests.yml
tihuan: Ah yeah we currently don't have e2e tests for the doc site, but I was hoping to run the tests still, so when we add doc site tests, it will trigger e2e tests. But that could be over engineering on my end!

I added Makefile to it in case when command changes and break e2e, we'll catch it in the PR

Lint config changes probably don't need to trigger tests, since it only matters when the lint affects actual production code. So my bad for including it 😆

seve: There's still some .mdx changes that can make it through the linter and still break `next build`. I think its worth testing on those in the meantime.

nayib-jose-gloria:
All changes are addressed (or made uneccessary due to faster tests) by https://github.com/chanzuckerberg/single-cell-data-portal/pull/3788

Issue #3543, opened 2022-11-04
(FMG): Build query function for communicating with marker gene endpoint


atarashansky:
This function should accept:
1. clicked cell type --> `cellType`
2. clicked cell type's tissue --> `tissue`
3. current organism --> `organism`
4. selected filters on left sidebar --> `selectedFilters`

It should the following arguments for the POST request:
1) `context_filters` --> looks like the following:
```
{
organism_ontology_term_id: organism,
tissue_ontology_term_ids: [tissue],
...selectedFilters
}
```

2) `target_filters` --> looks like the following:
```
{
organism_ontology_term_id: organism,
cell_type_ontology_term_ids: [cellType],
tissue_ontology_term_ids: [tissue],
...selectedFilters
}
```

3) `n_markers` --> 10. This number should be saved in `globals.ts`.

These arguments should be in the body of the POST request to the `/findmarkers` endpoint.

PR #3591, opened 2022-11-15
feat(FMG): add `/markers` request and first run of request UI

seve:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3543

---

Properly requests marker gene info from backend on click of cell type. Sends marker genes to heatmap.

Note to reviewers:

* At time of coding, both the React Query and Redux logic were built as scaffolding to support pivots based on UX feedback. At time of PR, we've established work to be done to show a marker gene info panel, which will most likely require a new piece of state in Redux store. ([shown in this Figma](https://www.figma.com/file/fgDQf6RSNyPeZpZoZRph0H/Alpha-Find-Marker-Genes?node-id=2%3A179321&t=5qzIj3I2goVeoAfj-4))
* With the above info, I would appreciate feedback on handling state that is stored in both Tanstack Query + Redux (seems a bit redundant). Do we avoid Query in this case and move towards raw fetching for this info?


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -691,3 +703,103 @@ export function aggregateCollectionsFromDatasets(
...
+interface HardcodedMarkerGeneRequest extends MarkerGeneRequestBody {
+  n_markers: 10;
+  test: "ttest";
+}
+
+export function generateMarkerGeneBody(
+  cellTypeID: string,
+  tissueID: string,
+  organismID: string
+): HardcodedMarkerGeneRequest | null {
```
tihuan, line 724: hmm I might be wrong, but this doesn't seem to return `null`?

```suggestion
): HardcodedMarkerGeneRequest {
```

frontend/src/views/WheresMyGene/components/HeatMap/hooks/useDeleteGenesAndCellTypes.ts
```diff
@@ -34,6 +34,7 @@ export function useDeleteGenesAndCellTypes(): {
       if (!dispatch) return;
 
       dispatch(toggleCellTypeIdToDelete(cellTypeMetadata));
+      // use the mutation callback here
```
tihuan, line 37: are we missing something here or is this comment out of date?

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -346,7 +346,8 @@ export function createYAxisOptions({
             const { name, depth = 0 } = deserializeCellTypeMetadata(
               value as CellTypeMetadata
             );
-
+            // TODO: We will have to add a ternary based on hover state too allow for room of the Marker Genes button
```
tihuan, line 349:
```suggestion
            // TODO: We will have to add a ternary based on hover state to allow for room of the Marker Genes button
```

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
```diff
@@ -551,6 +552,7 @@ export function deserializeCellTypeMetadata(
   };
 }
 
+// seve: looks like this isn't used anywhere
```
tihuan, line 555: ooh yeah dead code! We can remove this. Thank you!

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
tihuan, line 349: Maybe I need to catch up on the context, but UX already showed that inline button for adding marker genes is hard to discover the feature, so I'm assuming we're not doing that anymore thus the comment is no longer relevant?

tihuan (commented):
Code looks good! Can you add videos/screenshots to give reviewers an idea what we should expect to see? Thank you!

tihuan:
@seve If React Query is storing the state already, why do we need to duplicate that in Redux store again?

frontend/src/views/WheresMyGene/components/HeatMap/utils.ts
seve: Missed this file when cleaning up, ty!

seve: that's right!

seve:
>  @seve If React Query is storing the state already, why do we need to duplicate that in Redux store again?
 @tihuan 
We're going to be creating a bit of a more complex store in Redux in the near future. Marker genes are going to be treated as a separate collection of genes from the normal user selected genes in WMG.

Even in the current version that this PR is introducing, we want to have the ability to display/remove the genes from the heatmap which is currently controlled by our Redux solution.

tihuan:
Got it! Probably need more discussion. But right now I'm wondering if we could still keep the marker genes in React Query and store client state in redux store, and derive the presentation state by combining the remote + client state

For example, React Query returns Gene A, B, C, and the user removes Gene B. Then the final presentation state will filter out B (client state) from React Query remote state, this way we have single source of truth for both remote and client state?

seve:
> Got it! Probably need more discussion. But right now I'm wondering if we could still keep the marker genes in React Query and store client state in redux store, and derive the presentation state by combining the remote + client state
> 
> For example, React Query returns Gene A, B, C, and the user removes Gene B. Then the final presentation state will filter out B (client state) from React Query remote state, this way we have single source of truth for both remote and client state?
> 

I think that all makes sense to me, I'll just shrink the Tanstack Query logic from a mutation to a query. (Since I've already begun work on the second iteration, I'll do that there)

tihuan:
Perfect! Thanks so much, @seve 🚀 🌔 !

atarashansky:
After poking around an [rdev stack](https://rdev-dsadgat-main-frontend.rdev.single-cell.czi.technology/gene-expression?fmg=true), I have a couple notes.

1. What happens when we "unclick" a cell type? I think the UI tries to add the marker genes again to `selectedGenes`. I don't know what the behavior should be atm. This probably won't be a problem when we implement the target design.
2. I noticed that when adding the marker genes to `selectedGenes`, it does not check for duplicates. This throws a validation error and the tool crashes. (To reproduce, click on a cell type twice). We cannot have duplicates in `selectedGenes`.
3. We should add a try/catch and ignore genes that don't have any gene symbols in `primary_filter_dimensions.json`.

seve:
This is holding until #3635 lands

atarashansky (approved):
Rubber stamping

Issue #3164, opened 2022-08-23
How does a user select an organ from another system when a system is already selected? 


brianraymor:
@jahilton scenario that was raised during the **August 23** review for _filter tissues with ontology awareness_:

1. The user is initially offered a set of available filters for systems, organs, and tissues:

![Screen Shot 2022-08-23 at 4.15.26 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/8ab27db0-c9ed-4891-9e31-bdaea7bf3cf6)

2. The user selects one system (cardiovascular). The items in the organ and tissue columns are now filtered to that system. 

![Screen Shot 2022-08-23 at 4.10.24 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/e46593ad-12f2-456e-8f16-5d90ab1eb41e)

3. If the user intended to select additional organs (kidney) not included in the cardiovascular system, then the user must either:
- Unselect the cardiovascular system to show other items in the organ column.
- Select a second system (renal) that includes the organ (kidney)

Either is a bit awkward and/or non-intuitive.

@hthomas-czi proposed that the search feature could temporarily display all matching items in system, organ, and tissue. For example the user could start to type "kid" to quickly add "kidney" to the organ column for selection.

PR #3602, opened 2022-11-17
feat: How does a user select an organ from another system when a syst…

NoopDog:
…em is already selected? (#316)

- #TICKET_NUMBER
#3164

### Reviewers
**Functional:** 
@tihuan  

---

## Changes
Previously, SelectCategoryValueViews were filtered before sending to the FilterCategoryViewPanels to remove any SelectCategoryValueViews that were filtered out by selections in parent panels. For example, in the “Tissue” filter,  if a system is selected, then only organs and tissues under the system are passed through to the FilterCategoryViewPanels.

To support the updated functionality, all SelectCategoryValueViews are now sent to the  FilterCategoryViewPanels with a “visible” flag to indicate if they are filtered out by selections in parent panels. 

The FilterCategoryViewPanels shows only “visible” SelectCategoryValueViews unless an “omnisearch” search term is present in the search box. If a search term is present, the FilterCategoryViewPanel shows all SelectCategoryValueViews matching the search term regardless of the visibility state of the SelectCategoryValueViews.

### DIFF
--- a/frontend/src/common/hooks/useCategoryFilter/common/multiPanelOntologyUtils.ts
+++ b/frontend/src/common/hooks/useCategoryFilter/common/multiPanelOntologyUtils.ts
@@ -331,8 +331,13 @@ function buildOntologyPanelCategoryViews(
     // If panel is filtered (that is, there are selected values in this panel's parents), only include values in this
     // panel that are descendants of the selected parent values. If there are no selected values in parent panels,
     // include all values in this panel.
+
+    // Get all UI Nodes configured for this panel.
     const panelUINodes = listPanelUINodes(panelIndex, categoryFilterUIState);
-    const includeUINodes = selectedParentUINodes.length
+
+    // Get the UINodes to be displayed in the panel based on the
+    // selected nodes in parent panels.
+    const visibleUINodeIds = selectedParentUINodes.length
       ? filterPanelUINodes(
           panelUINodes,
           selectedParentUINodes,
@@ -341,8 +346,14 @@ function buildOntologyPanelCategoryViews(
         )
       : panelUINodes;
 
-    // Collect the select category values for the values to be included.
-    const selectCategoryValues = includeUINodes
+    // Create a set of visibleNodeIds to use in the next step where
+    // we mark each node as visible or not.
+    const visibleSet = new Set(
+      visibleUINodeIds.map((node) => node.categoryValueId)
+    );
+
+    // Collect the SelectCategoryValues for the UINodes to be included.
+    const selectCategoryValues = panelUINodes
       .map((uiNode: MultiPanelUINode) =>
         categoryValuesByValue.get(uiNode.categoryValueId)
       )
@@ -352,11 +363,13 @@ function buildOntologyPanelCategoryViews(
         ): selectCategoryValue is SelectCategoryValue => !!selectCategoryValue
       );
 
-    // Build views for each selected category value to be included.
+    // Build the SelectCategoryValueViews for all configured SelectCategoryValues, mark the
+    // ones not filtered out by parent panels as visible=true.
     const panelViews: SelectCategoryValueView[] = buildSelectCategoryValueViews(
       config,
       selectCategoryValues,
-      ontologyTermLabelsById
+      ontologyTermLabelsById,
+      visibleSet
     );
 
     // Apply non-specific labels to values that appear in this panel as well as parent panels.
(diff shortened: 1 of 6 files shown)


frontend/src/common/hooks/useCategoryFilter/common/multiPanelOntologyUtils.ts
```diff
@@ -341,8 +346,19 @@ function buildOntologyPanelCategoryViews(
         )
       : panelUINodes;
 
-    // Collect the select category values for the values to be included.
-    const selectCategoryValues = includeUINodes
+    // Create a set of visibleNodeIds to use in the next step where
+    // we mark each node as visible or not.
+    const visibleSet = visibleUINodeIds.reduce(function (
```
tihuan, line 351: nit: prefer () => {} for anonymous function 🙏 Thank you!

tihuan: discussion: for constructing the `visibleSet`, what do you think about the following 😄 ?

```ts
const visibleSet = new Set(visibleUINodeIds.map((node) => node.categoryValueId))
```

I'm hoping this would improve readability 🤞

frontend/src/common/hooks/useCategoryFilter/common/multiPanelOntologyUtils.ts
```diff
@@ -331,8 +331,13 @@ function buildOntologyPanelCategoryViews(
     // If panel is filtered (that is, there are selected values in this panel's parents), only include values in this
     // panel that are descendants of the selected parent values. If there are no selected values in parent panels,
     // include all values in this panel.
+
+    // Get all UI Nodes configured for this panel.
```
tihuan, line 335: thanks for all the informational comments!

frontend/src/components/common/Filter/components/FilterContent/components/FilterViews/components/FilterSinglePanelCategoryView/index.tsx
```diff
@@ -29,10 +30,11 @@ function filterViews(
 ): SelectCategoryValueView[] {
   if (searchValue) {
     return views.filter(({ label }) =>
-      label.toLowerCase().includes(searchValue)
+      label.toLowerCase().includes(searchValue.toLowerCase())
```
tihuan, line 33: just curious, was this a bug that wasn't caught before or a new behavior?

Thank you!

tihuan (approved):
LGTM!! Great flexible solution 💡 ! And thanks so much for the comments too ⭐ 🙏 

Just some discussions, otherwise LGTM!

frontend/src/components/common/Filter/components/FilterContent/components/FilterViews/components/FilterSinglePanelCategoryView/index.tsx
NoopDog: This is sneaking in a fix for a bug. I will create a new ticket and make it its own PR.

frontend/src/common/hooks/useCategoryFilter/common/multiPanelOntologyUtils.ts
NoopDog: Thank you for reading them!

frontend/src/common/hooks/useCategoryFilter/common/multiPanelOntologyUtils.ts
NoopDog: Will fix.

NoopDog: Updated thx.

frontend/src/components/common/Filter/components/FilterContent/components/FilterViews/components/FilterSinglePanelCategoryView/index.tsx
NoopDog: Reverted.

tihuan: Thanks so much, @NoopDog ! 

Sorry I didn't mean to ask you to create a ticket and a fix in a separate PR, since that seems to be a higher cost than the fix itself in this case 😆 Was just genuinely curious what the code change was for lol 

But I do appreciate you taking the time to do it, since technically it's probably the best practice to have focused PRs!

PR #3603, opened 2022-11-17
feat: processing layer

ebezzi:


backend/layers/processing/upload_failures/app.py
```diff
@@ -0,0 +1,52 @@
...
+    This functions updates the processing status for a given dataset uuid to failed
+    """
+    try:
+        business_logic.update_dataset_version_status(
+            DatasetVersionId(dataset_id), 
+            DatasetStatusKey.PROCESSING, 
+            DatasetProcessingStatus.FAILURE
+        )
+    # If dataset not in db dont worry about updating its processing status
+    except AttributeError:
```
nayib-jose-gloria, line 40: I believe it will throw a sqlalchemy.exc.ResultNotFound exception in this case. We can either catch for that, update the business layer to catch+throw an AttributeError, or update the persistence layer to handle it differently

Issue #3605, opened 2022-11-17
Bug: Omnisearch in multi panel ontology filters are not case insensitive


NoopDog:
## Describe the bug
In the multi-pannel ontology filters, "Omni Search" bar search terms are not properly case insensitive.

## To Reproduce
For example, in the cell type filter, typing "T" or "B" in the search bar returns no matches.

<img width="1264" alt="image" src="https://user-images.githubusercontent.com/3239697/202543696-48871e5e-b241-4791-acd5-f49a1331c643.png">

Matches are, however, returned for "t" and "b."

<img width="1259" alt="image" src="https://user-images.githubusercontent.com/3239697/202543757-42ed974a-db68-46f6-b63c-60c716731d23.png">

## Expected behavior
I am expecting instead that the search ignores the case of both the ontology term label and the search term so that searching for "T" or "t" will return "T cell" as a match.

## Environment
All

PR #3608, opened 2022-11-18
fix: omnisearch in multi panel ontology filters are not case insensit…

NoopDog:
…ive (#3605)

- #TICKET_NUMBER
#3605
### Reviewers
**Functional:** 
@tihuan 

---

## Changes
Lowercased the omni-search search string before comparing it with the lowercase ontology term name.


tihuan (approved):
LGTM! Thanks so much again, Dave 🤩🙌🔥

Issue #3598, opened 2022-11-16
Move local Docker dev network from HTTP to HTTPS


danieljhegeman:
- Add reverse proxy with TLS enabled (requires key and cert) to our docker compose suite
- Route local request to frontend server through the reverse proxy to the frontend server
- redirect all HTTP requests to frontend server to HTTPS

PR #3618, opened 2022-11-22
chore: move local FE and BE servers to HTTPS

danieljhegeman:
- #3598 

### Reviewers
**Functional:** 
@tihuan @Bento007 
**Readability:** 
@atarashansky 
---

## Changes
- Introduce [next-dev-https](https://www.npmjs.com/package/next-dev-https) for our local dev server for serving the FE
- Move BE to HTTPS as well (required for compatibility)
- Share cert/key generated by the oidc container (legacy code from Jessica) with the FE and the BE containers so that (it continues to be true that) **only one** cert has to be added to the local keychain.


danieljhegeman:
Had taken a second to investigate some 404's that were appearing but the same errors are present in `main` with http so it's nothing new -- just an artifact of a call to an Auth0 domain that fails in local scenario

PR #3635, opened 2022-11-25
feat: implement fmg endpoint and processing pipeline

atarashansky:
- #3540 
- #3541 

### Reviewers
**Functional:** 
@atolopko-czi 

**Readability**
@Bento007 

**Algorithmic**
@pablo-gar 

This PR adds a new endpoint (`/markers`) to the Gene Expression API. It queries a new data artifact in the snapshot (`marker_genes`) which is a TileDB array containing precomputed marker genes. This cube is now created as part of the weekly cube generation process (see `backend/wmg/pipeline/summary_cubes/marker_genes.py`).

## Notable changes
- **Marker gene calculation module** - marker gene calculation logic: `backend/wmg/api/calculate_markers.py`
- **Query criteria** - marker genes cube and expression summary fmg cube: `backend/wmg/api/query.py`
- **API endpoint** - `/markers`: `backend/wmg/api/v1.py` and `backend/wmg/api/wmg-api.yml`
- **Data artifacts** - new data artifacts have been added to `WmgSnapshot`: `backend/wmg/data/snapshot.py`
- **WMG processing pipeline step** - `create_marker_genes_cube`: `backend/wmg/pipeline/summary_cubes/marker_genes.py`
- **Test snapshot for marker gene tests** - `tests/unit/backend/wmg/fixtures/test-fmg-snapshot`
- **Marker gene calculation tests** - `tests/unit/backend/wmg/test_marker_gene_calculation.py`
- **Marker gene query tests** - `tests/unit/backend/wmg/test_query.py`
- **Marker gene cube creation test** - `tests/unit/wmg_processing/test_marker_gene_cube_creation.py`

## Note to reviewers
If there is any missing test coverage, please let me know and I will add the corresponding tests.


backend/wmg/data/transform.py
```diff
@@ -206,7 +206,7 @@ def _cell_type_ordering_per_tissue(data_cells: pd.DataFrame, target_cells: List[
 
 
 @log_func_runtime
-def generate_primary_filter_dimensions(snapshot_path: str, corpus_name: str, snapshot_id: int):
```
atarashansky, line 209: `corpus_name` was never set to anything different than the `snapshot_path` so I removed it. This simplifies the access pattern in the processing pipeline, as well.

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -254,16 +252,12 @@ def create_cubes(
         data_dir, coords, dim_values, expression_summary_vals_fn=expression_summary_vals_fn
     )
 
-    expression_summary_fmg_cube_dir = create_expression_summary_fmg_cube(
-        data_dir, coords, dim_values, expression_summary_fmg_vals_fn=expression_summary_fmg_vals_fn
```
atarashansky, line 258: I decided to not generate fake data artifacts for anything related to marker genes.

Instead, tests will use a test fixture snapshot containing subsetted, real data. Now, we can test the marker gene cube generation as well as the associated marker gene calculation code.

atarashansky: A side benefit is that any test that uses `create_temp_wmg_snapshot` is much faster, now. Backend unit test times are cut in half after this reversion.

atolopko-czi (commented):
CR still in progress. Posting a few minor requests for comments (literally! not a RFC!) as PR "Comment". 🤯

backend/wmg/pipeline/cube_pipeline.py
```diff
@@ -47,14 +46,14 @@ def load_data_and_create_cube(
...
-    snapshot_path = f"{snapshot_path}/{snapshot_id}"
-    corpus_path = f"{snapshot_path}/{corpus_name}"
+    corpus_path = f"{snapshot_path}/{snapshot_id}"
 
     integrated_corpus.run(path_to_h5ad_datasets, corpus_path, extract_data)
     try:
         summary_cubes.run(corpus_path, validate_cube)
     except CubeValidationException as e:
-        upload_artifacts_to_s3(snapshot_path, "latest_validation_failed_snapshot")
+        # TODO: should we split integrated/obs/var into separate directory (not under snapshot ID)?
```
atolopko-czi, line 55: Maybe, but not worth any effort, since ultimately we hope to build WMG off Cell Census data, replaxcing these intermediate data artifacts. For now, it helps to have these linked in case we need to troubleshoot.

backend/wmg/api/calculate_markers.py
```diff
@@ -0,0 +1,627 @@
...
+            elif isinstance(v, list):
+                v = HList(v)
+            new_kwargs[k] = v
+
+        return func(*new_args, **new_kwargs)
+
+    return wrapped
+
+
+@_make_hashable
```
atolopko-czi, line 57: worth a comment about why this is needed for memoization done by lru_cache.

tests/unit/wmg_processing/test_marker_gene_cube_creation.py
```diff
@@ -0,0 +1,33 @@
+from backend.wmg.pipeline.summary_cubes.marker_genes import create_marker_genes_cube
+import unittest
+from tests.unit.backend.wmg.fixtures import FIXTURES_ROOT
+import os
+import tiledb
+import shutil
+
+TEST_SNAPSHOT = "test-fmg-snapshot"
+
+# creates the marker gene cube de novo and compares to the existing fixture
```
atolopko-czi, line 10: Worth commenting also that the the marker computation logic is being tested elsewhere and that this test _only_ needs to verify the snapshot cube building is working as expected.   (Before I realized that, I wrote a comment about how this was not a sufficient test to ensure marker computation logic is not really being tested by this.)

tests/unit/wmg_processing/test_marker_gene_cube_creation.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class MarkerGeneCubeCreationTest(unittest.TestCase):
```
atolopko-czi, line 13: This expression_summary_fmg cube is implicitly being tested by this, but it may be worth having a similar test that asserts it is being built correctly. At some point the system will depend upon that data artifact directly, right? So its existence will no longer be an "implementation detail" of this marker gene cube. If you agree, it's fair to take that on as another story; but maybe easy enough to do now.

atolopko-czi (changes requested):
Completed a full pass. Overall, looks good! Added a number of (mostly minor) in-line comments and these:

* Can you add some documentation or a script that is used to build the new snapshot fixture? It should be as easy as possible for a future engineer to recreate this fixture. Tests that make use of it should comment that "If you need to update the fixture due to changes in expected test values, please refer to...".
* Maybe as a separate issue/PR, but is there anything that can be added to `backend.wmg.data.validation.validation.Validation` to validate the FMG data artifacts?
* Out-of-scope for this PR, but in keeping with the ETL structure of the pipeline job classes: 
  * Should `backend.wmg.pipeline.summary_cubes.expression_summary_fmg.job._load()` be in `load.py` (and not `job.py`)?
  * Should `backend.wmg.pipeline.summary_cubes.expression_summary_fmg.load.build_in_mem_cube()` bin `transform.py`?

backend/wmg/pipeline/summary_cubes/marker_genes.py
```diff
@@ -0,0 +1,99 @@
+import logging
+
+import pandas as pd
+import tiledb
+import numpy as np
+import gc
+from backend.wmg.data.schemas.marker_gene_cube_schema import marker_genes_schema
+from backend.wmg.data.snapshot import CELL_COUNTS_CUBE_NAME, MARKER_GENES_CUBE_NAME
+from backend.wmg.data.utils import create_empty_cube, log_func_runtime
+from backend.wmg.api.calculate_markers import get_markers
```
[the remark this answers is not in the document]
atarashansky: Is the reverse allowable? Can API import from pipeline? Eventually, both API and pipeline will need to use this module.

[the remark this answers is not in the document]
atolopko-czi: The TileDB schemas should probably be the only shared code between the API and the pipeline. The schema becomes the contract between them.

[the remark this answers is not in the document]
atolopko-czi: If this shared computation code, as I believe will be the case here, probably worth splitting out into a sibling package.

tests/unit/backend/wmg/fixtures/test_snapshot.py
atarashansky: I had not tested `expression_summary_fmg` generation, as we are testing neither `cell_counts` or `expression_summary` generation. Those tests exists but they were stubbed (i.e. just a `pass`).

`expression_summary_fmg` is used in two ways for tests:
1. test marker gene cube creation.
2. test query results (`test_query.py`).

backend/wmg/pipeline/summary_cubes/marker_genes.py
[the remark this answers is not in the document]
atarashansky: Gotcha. I will relocate to `pipeline` for now, and we can revisit in the future when it's needed on-the-fly by the API.

tests/unit/wmg_processing/test_marker_gene_cube_creation.py
atarashansky: We currently do not have any tests that `expression_summary_fmg` is being built correctly, as we don't have any tests for `expression_summary` or `cell_counts`, either. Adding tests just for `expression_summary_fmg` does not make sense as we may as well add tests for `expression_summary` and `cell_counts`, but that seems like it makes more sense to do as a separate story.

atarashansky:
> Completed a full pass. Overall, looks good! Added a number of (mostly minor) in-line comments and these:
> 
> * Can you add some documentation or a script that is used to build the new snapshot fixture? It should be as easy as possible for a future engineer to recreate this fixture. Tests that make use of it should comment that "If you need to update the fixture due to changes in expected test values, please refer to...".

Done! It's now in `scripts/generate_fmg_test_snapshot.py`.

> * Maybe as a separate issue/PR, but is there anything that can be added to `backend.wmg.data.validation.validation.Validation` to validate the FMG data artifacts?

I think @pablo-gar and I will work on the validation in a separate story.

> * Out-of-scope for this PR, but in keeping with the ETL structure of the pipeline job classes:
>   
>   * Should `backend.wmg.pipeline.summary_cubes.expression_summary_fmg.job._load()` be in `load.py` (and not `job.py`)?
>   * Should `backend.wmg.pipeline.summary_cubes.expression_summary_fmg.load.build_in_mem_cube()` bin `transform.py`?

I copied the structure of the `summary_cubes.expression_summary` module, but your suggestions make sense in the context of ETL structure!

tests/unit/wmg_processing/test_marker_gene_cube_creation.py
atolopko-czi: I've created an issue to _figure out_ what remaining tests should be written, which should cover any missing tests from this effort. However, I would recommend adding a test stub for now to capture what you think _should_ be tested (it could just be a single unimplemented test).

atolopko-czi (approved):
Thanks for the final changes! LGTM from a high-level perspective (overall design, code structure), but did not review some details like math and algorithms, with the understanding that Pablo et.al. may review that directly or indirectly (via data validation efforts).

One final comment herein to add some test stubs for future implementation.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+    with np.errstate(divide="ignore", invalid="ignore"):
+        p_query = n1 / (n1 + n2)
+        size = nnz_thr1 + nnz_thr2
+        pvals = stats.binom.sf(nnz_thr1 + 1, size + 2, p_query)
+        mean_n = (n1 + n2) / 2
+        pn1 = n1 / mean_n
+        pn2 = n2 / mean_n
+        effects = np.log2((nnz_thr1 + pn1) / (n1 + 2 * pn1)) - np.log2((nnz_thr2 + pn2) / (n2 + 2 * pn2))
+
+    pvals[effects < 0] = np.nan
```
pablo-gar, line 510:
```suggestion
```
not needed, `stats.binom.sf` is a one-sided test

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+        n_target,
+        n_context,
+        target_data_nnz,
+        genes,
+        groups_indices_context,
+        genes_indices_context,
+        genes_indices_target,
+    ) = _prepare_indices_and_metrics(target_filters, context_filters, corpus=corpus)
+    target_data_nnz_thr = np.zeros((1, len(genes)))
+    target_data_nnz_thr[0, genes_indices_target] = list(target_agg["nnz_thr"])
```
pablo-gar, line 555: Is there a way to not get these values (`nnz_thr`) during cube generation and rather to have them as a parameter here? It would be extremely beneficial for parametrization and in a future where we allow users to select their own threshold in the standalone tool.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+
+    n_markers - int, optional, default 10
+        Number of top markers to return. If None, all marker genes with effect size > 0 are returned.
+    """
+    zero_out = nnz.flatten() < min_num_expr_cells
+    effects[:, zero_out] = 0
+    pvals[:, zero_out] = 1
+    # aggregate
+    effects = np.nanpercentile(effects, percentile * 100, axis=0)
+    pvals = np.array([stats.combine_pvalues(x[np.invert(np.isnan(x))] + 1e-300)[-1] for x in pvals.T])
```
pablo-gar, line 394: under what circumstances do we have `nan` here?

pablo-gar (changes requested):
I have requested some changes, and clarifications. 

There's places of the t test that don't follow the method processing spec, a few stat bugs (that are creating artifacts for the final p-value calculation) and some questions I have.

Let me known when you have addressed them and I can take another look at it.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+        var2_n = var2 / n2
+        sum_var_n = var1_n + var2_n
+        dof = sum_var_n**2 / (var1_n**2 / (n1 - 1) + var2_n**2 / (n2 - 1))
+        dof[np.isnan(dof)] = 1
+        tscores = (mean1 - mean2) / np.sqrt(sum_var_n)
+        effects = (mean1 - mean2) / np.sqrt(((n1 - 1) * var1 + (n2 - 1) * var2) / (n1 + n2 - 1))
+
+    tscores[np.isnan(tscores)] = 0
+    pvals = stats.t.sf(tscores, dof)
+    pvals[effects < 0] = np.nan
```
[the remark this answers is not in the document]
pablo-gar: And actually these are converted later on to 10e-300, right?

https://github.com/chanzuckerberg/single-cell-data-portal/blob/df0ef4715e54af30be2d782e5991aa35dc2befb3/backend/wmg/pipeline/summary_cubes/calculate_markers.py#L394

backend/wmg/pipeline/summary_cubes/calculate_markers.py
pablo-gar: And actually these are converted later on to 1e-300, right?

https://github.com/chanzuckerberg/single-cell-data-portal/blob/df0ef4715e54af30be2d782e5991aa35dc2befb3/backend/wmg/pipeline/summary_cubes/calculate_markers.py#L394

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+
+        for i in [mean1, meansq1, mean2, meansq2]:
+            i[np.isnan(i)] = 0
+
+        var1 = meansq1 - mean1**2
+        var1[var1 < 0] = 0
+        var2 = meansq2 - mean2**2
+        var2[var2 < 0] = 0
+
+        var1_n = var1 / n1
```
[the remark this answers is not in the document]
atarashansky: https://www.deriveit.net/ps/expected_value_and_variance/varDef.html

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+        var2_n = var2 / n2
+        sum_var_n = var1_n + var2_n
+        dof = sum_var_n**2 / (var1_n**2 / (n1 - 1) + var2_n**2 / (n2 - 1))
```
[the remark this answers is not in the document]
atarashansky: Which part is different? 
> var1_n = sd_exp1^2 /n_cells1

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+        dof[np.isnan(dof)] = 1
+        tscores = (mean1 - mean2) / np.sqrt(sum_var_n)
```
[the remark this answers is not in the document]
atarashansky: It's the same.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+
+    percentile - float, optional, default 0.15
+        The percentile of effect sizes to select as the representative effect size.
+
+    n_markers - int, optional, default 10
+        Number of top markers to return. If None, all marker genes with effect size > 0 are returned.
+    """
+    zero_out = nnz.flatten() < min_num_expr_cells
+    effects[:, zero_out] = 0
+    pvals[:, zero_out] = 1
```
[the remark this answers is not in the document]
atarashansky: These genes are effectively deleted from the results as the entire column of comparisons is zero'd out.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
atarashansky: Unfortunately, no. It's impossible to calculate it after cube generation unless we want to read from the integrated array and do a reduction on the fly (very inefficient).

backend/wmg/pipeline/summary_cubes/calculate_markers.py
atarashansky: If tscore for a particular group/gene combo is NaN. And tscore is NaN if population size for a particular group/gene combo is 0. These comparisons are excluded from aggregation.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+        var1 = meansq1 - mean1**2
+        var1[var1 < 0] = 0
+        var2 = meansq2 - mean2**2
+        var2[var2 < 0] = 0
+
+        var1_n = var1 / n1
+        var2_n = var2 / n2
+        sum_var_n = var1_n + var2_n
+        dof = sum_var_n**2 / (var1_n**2 / (n1 - 1) + var2_n**2 / (n2 - 1))
+        dof[np.isnan(dof)] = 1
```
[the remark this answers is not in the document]
atarashansky: Oh, this was a mistake!

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -0,0 +1,620 @@
...
+    """
+    with np.errstate(divide="ignore", invalid="ignore"):
+        mean1 = sum1 / n1
+        meansq1 = sumsq1 / n1
+
+        mean2 = sum2 / n2
+        meansq2 = sumsq2 / n2
+
+        for i in [mean1, meansq1, mean2, meansq2]:
+            i[np.isnan(i)] = 0
```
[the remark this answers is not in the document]
atarashansky: Vestigial, thanks for catching. Will leave those values as just nan, now. Nan values will be completely ignored for aggregation step.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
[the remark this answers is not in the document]
pablo-gar: Similar to the other one I got thrown off by `sum_var_n`

backend/wmg/pipeline/summary_cubes/calculate_markers.py
pablo-gar: Got it, then should we omit the associated effect sizes as well for the percentile calculation in line 393? Are they already handled properly?

atarashansky: [np.nanpercentile](https://numpy.org/doc/stable/reference/generated/numpy.nanpercentile.html) ignores NaN's, so these nan comparisons are properly excluded.

backend/wmg/api/query.py
```diff
@@ -22,9 +22,28 @@ class WmgQueryCriteria(BaseModel):
     sex_ontology_term_ids: List[str] = Field(default=[], unique_items=True, min_items=0)
 
 
+class FmgQueryCriteria(BaseModel):
+    organism_ontology_term_id: str  # required!
+    tissue_ontology_term_ids: List[str] = Field(default=[], unique_items=True, min_items=0)
```
Bento007, line 27: ooo sorry I missed this. Don't use default=[]. You need to do `Field(default_factory=list)`. This is because we want a new list to be created as the default. If you don't do this, then the same list will be used as default for all instances of `FmgQueryCriteria`. Apply this same logic every where you are using a `dict` or `list` as the default value.

atarashansky: Should this change be applied to `WmgQueryCriteria` as well?

atarashansky: Also, does this matter if the criteria will never be modified in place?

Bento007: > Should this change be applied to WmgQueryCriteria as well?
yes for both

> Also, does this matter if the criteria will never be modified in place?
That makes it less of a problem, but could lead to some difficult debugging in the future if we don't fix it now.

Issue #3652, opened 2022-11-30
bug(wmg): tissues are not scrollable when loaded without the heatmap


atarashansky:
To reproduce:
 - Go here: https://cellxgene.cziscience.com/gene-expression
 - Select lung tissue
 - Try to scroll (you can't)

ashin-czi:
introduced another bug with this fix so resolving that now

PR #3653, opened 2022-11-30
fix: cant scroll after selecting tissue

ashin-czi:
- #3652

### Reviewers
**Functional:** @atarashansky 

---

## Changes
- add
- remove
- modify
    - raising yAxisWrapper zindex after hiding getting started step 1 so that scroll works for the axis


atarashansky (approved):
this looks good. this would also let the cell types be clickable in the y axis, right?

ashin-czi:
> this looks good. this would also let the cell types be clickable in the y axis, right?

Yup!

PR #3656, opened 2022-11-30
fix: dockerfile.backend build process fixed so pycharm test running works

Bento007:
…

Signed-off-by: Trent Smith <trent.smith@chanzuckerberg.com>

- #TICKET_NUMBER

### Reviewers

**Readability:** 

---

## Changes
- modify dockerfile.backend build process fixed so pycharm test running works.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -245,7 +245,6 @@ services:
         - HAPPY_BRANCH
         - HAPPY_TAG
     restart: always
-    command: [ "./container_init.sh" ]
     depends_on:
       - localstack
       - database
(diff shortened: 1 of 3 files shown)


danieljhegeman:
Interesting - why was this failing with Pycharm?

docker-compose.yml
```diff
@@ -245,7 +245,6 @@ services:
         - HAPPY_BRANCH
         - HAPPY_TAG
     restart: always
-    command: [ "./container_init.sh" ]
```
danieljhegeman, line 248: yeah this was superfluous given that a `dockerfile` is defined above right?

Bento007:
Something about not being able to find "./container_init.sh"

Bento007:
hold on maybe this does work fine. I was running this with the redesigned branch it was failing, but it seems to be working now on main.

danieljhegeman:
Ah yes you might've needed to rebase. It's possible you were working off of `1220be39` on main but you needed my subsequent commit `8e9dc8ed` which followed a few hours later to fix exactly that error - not able to find `container_init.sh`

My fault on that. Unlucky on the timing. 🙋‍♂️ 😕 😞

Bento007:
change not needed

Issue #3506, opened 2022-10-28
chore(GeneExpression): Create smoke test to ensure source data functions properly


atarashansky:
Bugs in the source data component have made it into prod a couple of times because it was uncovered by smoke tests. Smoke tests covering the following actions must be created:
 - Source data with no secondary filters selected
 - Source data with dataset secondary filter selected
 - Source data with any non-dataset secondary filter selected

PR #3672, opened 2022-12-05
chore: created smoke test to ensure source data functions properly

atarashansky:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3506

This test ensures that the source data sidebar pops up properly.
It also checks that at least one element with class `MuiListItem` appears (indicating that source data is present).


frontend/tests/features/wheresMyGene.test.ts
```diff
@@ -142,6 +143,69 @@ describe("Where's My Gene", () => {
...
+      },
+      { page }
+    );
+    await clickUntilSidebarShowsUp(getSourceDataButton, page);
+    await expect(page).toHaveSelector(getText("After filtering cells with low coverage (less than 500 genes expressed)"));
+    const sourceDataList = await page.$("[class*=MuiList-root]");
+    if (!sourceDataList) throw Error("no source data displayed");
+    
+    const sourceDataListItems = await sourceDataList?.$$(".MuiListItem-root")
+    expect(sourceDataListItems?.length).toBe(17);
```
tihuan, line 179: Should we just assert that the length is more than 0, since asserting `17` is pretty specific and will break when data changes?

frontend/tests/features/wheresMyGene.test.ts
```diff
@@ -142,6 +143,69 @@ describe("Where's My Gene", () => {
...
+    const selectedDatasetsBefore = await datasetSelector.$$(".MuiChip-root");
+    await expect(selectedDatasetsBefore.length).toBe(0);
+    await clickUntilOptionsShowUp(getDatasetSelector, page);
+    await selectFirstOption(page);
+    await clickUntilSidebarShowsUp(getSourceDataButton, page);
+    const sourceDataListAfter = await page.$("[class*=MuiList-root]");
+    if (!sourceDataListAfter) throw Error("no source data displayed after selecting dataset filter");
+    
+    const sourceDataListAfterItems = await sourceDataListAfter?.$$(".MuiListItem-root")
+    expect(sourceDataListAfterItems?.length).toBe(2);
```
tihuan, line 206: Same for this. Maybe just assert > 0?

frontend/tests/features/wheresMyGene.test.ts
```diff
@@ -142,6 +143,69 @@ describe("Where's My Gene", () => {
...
+    await tryUntil(
+      async () => {
+        const canvases = await page.$$("canvas");
+        expect(canvases.length).not.toBe(0);
+      },
+      { page }
+    );
+    await clickUntilSidebarShowsUp(getSourceDataButton, page);
+    await expect(page).toHaveSelector(getText("After filtering cells with low coverage (less than 500 genes expressed)"));
+    const sourceDataList = await page.$("[class*=MuiList-root]");
```
tihuan, line 175: Wondering if we need `tryUntil()` here to wait for `sourceDataList` to be available instead of throwing error and retrying the whole test too soon?

tihuan (commented):
Looks really good! Thanks for the robust test 🔥 Just some minor discussion points 😄 PTAL thank you!

frontend/tests/features/wheresMyGene.test.ts
atarashansky: I think this one will always be reliably 2 because we are selecting 1 dataset in the filtering panel, so we should only have 1 source data.

tihuan: Sounds good! Should we add a little comment to explain why it's 2 instead of 1, if we only select one dataset?

PR #3675, opened 2022-12-05
feat: revise at for curation api

Bento007:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
Add revised_at and validation message to backend.layers.common.entity

- change UserInfo.user_id to a property. Simplifies accessing the variable.
- use pydantic.datalclass. This gives us type checking for free.
- Moved errors in BusinessException to be an instance variable. Making it is class variable will have unintended consequences.
- Fixed pointer in persistence_mock.py
- Populating revised_at in persistence_mock.py


backend/layers/common/entities.py
```diff
@@ -181,6 +180,7 @@ class DatasetVersion:
     metadata: DatasetMetadata
     artifacts: List[DatasetArtifact]
     created_at: datetime
+    revised_at: Optional[datetime]  # The last time this Dataset Version was Published
```
ebezzi, line 183: If we put `revised_at` here, this is functionally equivalent to `published_at` since a collection version can only be published once. If you want to get `revised_at` for a _canonical_ collection, we have two options:
1. Save the field to the canonical table similarly to `originally_published_at`, and update it each time we publish a version (revision)
2. Infer it as the `published_at` of the most recent version

backend/layers/common/entities.py
```diff
@@ -83,13 +81,14 @@ class DatasetStatus:
     rds_status: Optional[DatasetConversionStatus]
     h5ad_status: Optional[DatasetConversionStatus]
     processing_status: Optional[DatasetProcessingStatus]
+    validation_message: Optional[str] = None
```
ebezzi, line 84: 👍

backend/layers/persistence/persistence_mock.py
```diff
@@ -167,23 +177,29 @@ def get_all_versions_for_collection(self, collection_id: CollectionId) -> Iterab
                 yield self._update_version_with_canonical(collection_version)
 
     # MAYBE
-    def finalize_collection_version(self, collection_id: CollectionId, version_id: CollectionVersionId, published_at: Optional[datetime]) -> None:
+    def finalize_collection_version(
+        self, collection_id: CollectionId, version_id: CollectionVersionId, published_at: Optional[datetime]
+    ) -> None:
 
-        now = datetime.utcnow()
+        published_at = published_at if published_at else datetime.utcnow()
```
ebezzi, line 184: 👍

backend/layers/common/entities.py
Bento007: So there are two timestamp we care about in a dataset revision.
1. `dataset.published_at` which is the timestamp when the dataset was **first** published, which may not be the same as `collection.published_at` if the dataset was added after the collection was first published.
2. `dataset.revised_at` which is the timestamp when a revision for a dataset was last published. If the dataset were ever changed/reuploaded, this value would be different from both `collection.published_at` and `dataset.published_at`.

I didn't see a value in the database table to represent `revised_at`

Issue #3569, opened 2022-11-07
Bug: when i download the image for gene expression the dot plot scrolls


signechambers1:
Steps to reproduce:

1. https://cellxgene.cziscience.com/gene-expression
2. Add blood, lung as tissues
3. Add DPM1, FGR, CFH as genes
4. Scroll to the bottom of the dot plot
5. Click download image
6. Dot plot auto scrolls to the top of the dot plot

Desired behavior: dot plot does not scroll when you click download.

ashin-czi:
`  &.CLONED {
    overflow: hidden;
    height: fit-content;
    width: fit-content;
    & ${Wrapper}, ${Container} {
      **height: fit-content;       <--------------- this is what's causing the jump because the height is being changed on the page**
      width: fit-content;
      overflow: hidden;
    }
  }`

Figuring out how to change this behavior

ashin-czi:
finding this a bit more complicated than expected 
@seve @tihuan might need a second pair of eyes on this

tihuan:
Huh interesting! Will take a look later today. Thanks so much, @ashin-czi !

PR #3684, opened 2022-12-07
fix(wmg): 3569 download heatmap scroll fix

tihuan:
- #3569 

This is a workaround to save the scrollTop and then restore it after download is done

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


tihuan:
Continue with https://github.com/chanzuckerberg/single-cell-data-portal/pull/3691

Issue #3658, opened 2022-12-01
feat(FMG): Add info button to cell type labels on x-axis


seve:
https://www.figma.com/file/fgDQf6RSNyPeZpZoZRph0H/Alpha-Find-Marker-Genes?node-id=393%3A269940&t=RcyHyj1Zf9HSKcK1-4

![image](https://user-images.githubusercontent.com/8716829/204934438-c0466ff4-843b-4843-90c0-c267a9c97809.png)

Cell type labels need an info button that call the soon-to-be implemented reducer action `addCellInfoCellType` 

https://github.com/chanzuckerberg/single-cell-data-portal/blob/897476d5190daf9ab062a719f908741240fb0037/frontend/src/views/WheresMyGene/common/store/reducer.ts#L370

ashin-czi:
@seve Is this feature still needed? I think we talked about closing it out but I might have misheard

PR #3687, opened 2022-12-07
feat(FMG): add info buttons in-line to the y axis labels

atarashansky:
#3658 

echarts does not allow HTML elements to be added to text labels. The solution is to use MutationObserver to wait for the <g> text elements to load in the Y-axis SVG, use scoped query selection to select all these text elements, get their pixel coordinates, and place the info buttons accordingly.

Functional
@tihuan 

Readability
@seve


frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -159,10 +202,44 @@ export default memo(function YAxisChart({
...
       />
+      {yAxisInfoCoords &&
+        yAxisInfoCoords.map((coord, i) => {
+          const content = cellTypeMetadata[i];
+          return (
+            <div
+              id={`${content}`}
+              key={`${content}`}
+              style={{
+                position: "absolute",
```
tihuan, line 217: let's extract this as a styled component instead of using inline style, so it's more readable 🙏

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -159,10 +202,44 @@ export default memo(function YAxisChart({
...
+                    tissueID,
+                  });
+                }
+              }}
+            >
+              <Image
+                src={InfoSVG.src}
+                width="9"
+                height="9"
+                alt="display marker genes"
```
tihuan, line 238: nice alt 🔥 ! Should we go the extra mile and add cell type name if it's easy 😆 ?

tihuan (approved):
LGTMMM 🚀 ! Go Alec ⭐ ⭐ ⭐ ⭐ ⭐ !

PR #3694, opened 2022-12-07
chore: redesign run tests in GHA

ebezzi:


.github/workflows/push-tests.yml
```diff
@@ -222,16 +222,10 @@ jobs:
...
-        run: |
-          echo "DOCKER_REPO=${DOCKER_REPO}" > .env.ecr
-          make local-rebuild-backend
-      - name: Run tests in docker-compose
+      - name: Run unit tests
         run: |
           echo "DOCKER_REPO=${DOCKER_REPO}" > .env.ecr
-          make local-init-test-data
-          make all-local-unit-test-backend
+          make redesign-unit-test
```
Bento007, line 228: why aren't we running these tests in the docker image? This would fix the gevent monkey wrench problem.

ebezzi: They are run in Docker. Check the makefile rule.

Bento007 (approved):

Makefile
```diff
@@ -157,6 +157,10 @@ local-shell: ## Open a command shell in one of the dev containers. ex: make loca
 .PHONY: local-unit-test
 local-unit-test: local-unit-test-backend local-unit-test-processing  local-unit-test-wmg-processing# Run all backend and processing unit tests in the dev environment, with code coverage
 
+.PHONY: redesign-unit-test
```
nayib-jose-gloria, line 160: since we're doing a feature branch release anyway, can we delete all the deprecated commands and rename this to just `unit-test `or something?

.github/workflows/push-tests.yml
```diff
@@ -222,16 +222,10 @@ jobs:
             **/Dockerfile
             requirements*.txt
             **/requirements*.txt
-      - name: Check if containers need to be rebuilt
-        if: steps.changed-files.outputs.any_changed == 'true'
-        run: |
-          echo "DOCKER_REPO=${DOCKER_REPO}" > .env.ecr
-          make local-rebuild-backend
-      - name: Run tests in docker-compose
+      - name: Run unit tests
```
nayib-jose-gloria, line 225: Is adding an integration test step part of the scope of this PR, or should it be added in a different PR?

ebezzi: I would say different PR. This should be merged as soon as possible to avoid further PRs to break tests.

nayib-jose-gloria (approved):

PR #3696, opened 2022-12-08
feat(FMG): update integrated corpus

atarashansky:
There is the concern that the filtering step we do in which we zero out rankit values with corresponding raw counts < 2 may be affecting the "correctness" of the statistical tests used to identify marker genes. Ideally, we should be identifying marker genes using _unfiltered_ data. The tradeoff is that the `integrated_corpus` and the `expression_summary_fmg` array will become much larger, thereby increasing the time it will take for the processing pipeline to run, though the magnitude of this increase is hard to estimate.

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- the integrated corpus contains a new attribute "rankit_unfiltered" with no thresholding. "rankit" is the same, thresholded data.


atarashansky:
double check if cell counts are summing up by presence/absence of expression elements and if presence of 0's can infalte the count

PR #3633, opened 2022-11-23
feat: Generate Allure test report for backend tests

Bento007:
- #3193

### Reviewers
**Functional:** @ashin-czi 

**Readability:** @danieljhegeman 

---

## Changes
- use pytest as the test runner. This is so we can generate an Allure report
-  generate allure test reports for backend tests

PR #3704, opened 2022-12-08
fix: update portal API layer tests to run against persistence mock and as integration tests

nayib-jose-gloria:

- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- updated portal API tests and updated portal API / business logic to pass tests as required
- refactored persistence session management --> moved session management logic into the persistence implementation file
- updated business and portal API tests to be able to easily run against the persistence mock or local postgres instance, using env var as a flag  

## QA steps (optional)
- Ran portal API tests locally against database provider mock and against postgres docker containerized instance 

## Notes for Reviewer
- TODO: updating persistence interface
- TODO: make runnable in docker
- TODO: Review in-comment todos and test logic, determine if all of it is actually needed for portal API tests and if proper test coverage is there 
- Note: Please double check logic around DatasetArtifactTypes--specifically whether we should test against H5AD (uppercase) or h5ad (lowercase) and in which layers that distinction exists. Can we / Should we standardize? 
- Note: 3 unit tests are currently skipped because they test canonical collection tombstoning logic, which has not been implemented yet. TODO in a different PR


backend/layers/api/portal_api.py
```diff
@@ -331,7 +331,7 @@ def delete_collection(self, collection_id: str, token_info: dict):
         Deletes a collection version from the persistence store.
         """
         version = self.business_logic.get_collection_version(CollectionVersionId(collection_id))
-        if not UserInfo(token_info).is_user_owner_or_allowed(version.owner):
+        if version is None or not UserInfo(token_info).is_user_owner_or_allowed(version.owner):
```
ebezzi, line 334: Nit, but maybe it's better to split those into two ifs

backend/layers/api/portal_api.py
```diff
@@ -566,7 +568,7 @@ def get_dataset_identifiers(self, url: str):
             raise NotFoundHTTPException()
 
         # Retrieves the URI of the cxg artifact
-        s3_uri = next(a.uri for a in dataset.artifacts if a.type == "CXG")
+        s3_uri = next(a.uri for a in dataset.artifacts if a.type == "cxg")
```
ebezzi, line 571: I think we can use `DatasetArtifactType.CXG`

backend/layers/business/business.py
```diff
@@ -356,8 +355,15 @@ def get_dataset_artifact_download_data(
 
         file_name = artifact_url.path[1:]
         file_type = artifact.type
-        file_size = self.s3_provider.get_file_size(artifact.uri)
-        presigned_url = self.s3_provider.generate_presigned_url(artifact.uri)
+        try:
```
ebezzi, line 358: I'm not sure we should do this. If these fail, it's likely an AWS issue and we should probably escalate the error.

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -157,7 +150,7 @@ def test__get_collection_id__ok(self, mock_dt):
             "access_type": "WRITE",
             "contact_email": "john.doe@email.com",
             "contact_name": "john doe",
-            "created_at": 1234,
+            "created_at": mock.ANY,
```
ebezzi, line 153: Probably a good idea. Mocking the timestamp isn't really the goal of this test 👍

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -329,7 +322,8 @@ def test__get_collection__ok(self):
     # ✅
     def test__get_collection_id__403_not_found(self):
         """Verify the test collection exists and the expected fields exist."""
-        test_url = furl(path="/dp/v1/collections/AAAA-BBBB-CCCC-DDDD", query_params=dict(visibility="PUBLIC"))
+        fake_id = self._generate_id()
```
ebezzi, line 325: 👍

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -611,40 +605,6 @@ def test__can_retrieve_created_collection(self):
         response = self.app.get(test_url.url, headers=no_cookie_headers)
         self.assertEqual("READ", json.loads(response.data)["access_type"])
 
-    # 🔴 used to be done by the database - we need to decide where this goes now
```
ebezzi, line 614: This is still TODO, I would leave the test and add a strip method to the business logic.

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -1185,12 +1071,11 @@ def test__update_collection_same_doi_does_not_update_metadata(self):
 
 class TestCollectionsCurators(NewBaseTest):
 
-    # 🔴 not quite sure why this fails - a non curator should not have access_type of WRITE
```
ebezzi, line 1188: Note that this test (in main) passes with an access type of `WRITE`, hence my comment. I agree that `READ` makes more sense but we shouldn't break the compatibility.

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -1341,33 +1226,34 @@ def test__post_dataset_asset__file_SERVER_ERROR(self):
...
         self.assertEqual(500, response.status_code)
         body = json.loads(response.data)
+        print(body)
         self.assertEqual("An internal server error has occurred. Please try again later.", body["detail"])
 
     # ✅
     def test__post_dataset_asset__dataset_NOT_FOUND(self):
-        # TODO: we're requiring UUIDs
-        test_url = furl(path="/dp/v1/datasets/test_user_id/asset/test_dataset_artifact_id")
+        fake_id = self._generate_id()
```
ebezzi, line 1242: 👍 thanks

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -1732,28 +1620,27 @@ def setUp(self):
     def tearDown(self):
         super().tearDown()
 
-    # 💛 review later
     def test__get_status__200_for_non_owned_dataset_as_super_curator(self):
-        processing_status = DatasetStatusUpdate("upload_status", DatasetUploadStatus.WAITING)
+        processing_status = DatasetStatusUpdate("upload", DatasetUploadStatus.WAITING)
```
ebezzi, line 1624: I believe we can use `DatasetStatusKey.UPLOAD` here

tests/unit/backend/layers/common/base_api_test.py
```diff
@@ -101,9 +102,12 @@ def setUp(self):
         super().setUp()
         os.environ.setdefault("APP_NAME", "corpora-api")
 
-        database_provider = DatabaseProvider()
-        database_provider._drop()
-        database_provider._create()
+        database_provider = DatabaseProviderMock()
```
ebezzi, line 105: Maybe we can use a flag here?

backend/layers/business/business.py
nayib-jose-gloria: yeah I made this change to pass a test, but honestly thinking about it, I should change the test instead

tests/unit/backend/layers/api/test_portal_api.py
nayib-jose-gloria: Looks like Trent fixed this in main a couple of weeks ago! https://github.com/chanzuckerberg/single-cell-data-portal/pull/3633

PR #3708, opened 2022-12-09
feat: curation API for datasets

ebezzi:


backend/layers/api/portal_api.py
```diff
@@ -48,7 +49,7 @@
 )
 
 
-class PortalApi:
+class PortalApi(ApiCommon):
```
ebezzi, line 52: Ignore this for now

tests/unit/backend/api_server/base_api_test.py
```diff
@@ -87,6 +87,14 @@ def get_cxguser_token(user="owner"):
 class BaseAuthAPITest(BaseAPITest):
     def setUp(self):
         super().setUp()
+
+        # TODO: this can be improved, but the current authorization method requires it
```
ebezzi, line 91: @Bento007 do you know what's going on here? Looks like the mock below isn't enough since this line:
```x = assert_authorized_token(token, CorporaAuthConfig().curation_audience)```
is calling the config in the parameters, which happens before the mock is called.

File: backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py [hunk 3]
```diff
@@ -1,41 +1,163 @@
...
+
+    return dataset
+
+
 def get(collection_id: str, dataset_id: str = None):
-    db_session = g.db_session
-    if not db_session.query(DbCollection.id).filter(DbCollection.id == collection_id).first():
+    business_logic = get_business_logic()
+
+    # TODO: double lookup
```
ebezzi, line 83: Here we need to decide if we want to always do a double lookup or if those methods only make sense when using either the canonical or the version id. I can look at the notebooks and see what makes sense, or just add the double lookup anywhere.

tests/unit/backend/api_server/base_api_test.py
Bento007: this is how I'm doing it https://github.com/chanzuckerberg/single-cell-data-portal/blob/9e0a2cb7ecc9b009da4567670fc118cc4355fd7c/tests/unit/backend/api_server/base_api_test.py#L90-L94
I'm letting the config value resolve to the default value and mock the whole `assert_asuthorized_token` function

backend/layers/api/portal_api.py
Bento007: I added this function https://github.com/chanzuckerberg/single-cell-data-portal/blob/9e0a2cb7ecc9b009da4567670fc118cc4355fd7c/backend/portal/api/curation/v1/curation/collections/common.py#L287-L298
We can punt on the double look-up decision and be inefficient for now for the sake of time.

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
Bento007: I added this function https://github.com/chanzuckerberg/single-cell-data-portal/blob/9e0a2cb7ecc9b009da4567670fc118cc4355fd7c/backend/portal/api/curation/v1/curation/collections/common.py#L287-L298
We can punt on the double look-up decision and be inefficient for now for the sake of time.

ebezzi: I agree. I will use that function.

Bento007 (changes requested):
a few tings to address

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -1,41 +1,161 @@
...
-        collection_id,
-        token_info,
-        body.get("url", body.get("link")),
-        dataset_id,
-    )
-    return "", 202
+    # TODO: deduplicate from ApiCommon. We need to settle the class/module level debate before can do that
+    url = body.get("url", body.get("link"))
+    business_logic = get_business_logic()
+    dataset_version = business_logic.get_dataset_version(DatasetVersionId(dataset_id))
```
Bento007, line 130: Lin 130-136 is duplicated logic from a the delete handler. We should factor is out.

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -1,41 +1,161 @@
...
+    business_logic = get_business_logic()
+    user_info = UserInfo(token_info)
+
+    # TODO: deduplicate from ApiCommon. We need to settle the class/module level debate before can do that
+    dataset_version = business_logic.get_dataset_version(DatasetVersionId(dataset_id))
+    if dataset_version is None:
+        raise ForbiddenHTTPException(f"Dataset {dataset_id} does not exist")
+
+    collection_version = business_logic.get_collection_version(CollectionVersionId(collection_id))
+    # If the collection does not exist, it means that the dataset is orphaned and therefore we cannot
```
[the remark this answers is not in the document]
ebezzi: I think this isn't supposed to happen, but I added it anyway. For what it's worth, `get_collection_version` returns an `Optional` so pylance will force you to add the `if is not None` condition, or you'll get a warning.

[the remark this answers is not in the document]
Bento007: Is this a condition we can test for?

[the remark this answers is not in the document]
ebezzi: I don't think it's necessary. Also the test should rely on a database state that's impossible to achieve. If you prefer, we can add a TODO to double check or specify that the check is only to prevent a possible edge case that should not be possible in the system.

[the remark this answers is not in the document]
ebezzi: Actually nvm, this was a comment taken from the portal API. In this case, this will simply be raised if a wrong collection_id is specified (i.e., a collection_id that doesn't contain the specified dataset_id). I removed the comment.

Bento007 (approved):

PR #3736, opened 2022-12-13
chore: restore processing tests

ebezzi:
Restores processing tests so they run on a separate GHA workflow and using the `processing` container.


backend/layers/business/entities.py
```diff
@@ -1,6 +1,6 @@
 from typing import List, Optional
 
-from pydantic.dataclasses import dataclass
+from dataclasses import dataclass
```
ebezzi, line 3: This is necessary for now since `pydantic` isn't declared in the dependencies. We should do that in a separate PR.

tests/unit/backend/layers/common/base_test.py
```diff
@@ -0,0 +1,200 @@
...
+
+    dataset_version_id: str
+    dataset_id: str
+    explorer_url: str
+    collection_version_id: str
+    collection_id: str
+    artifact_ids: List[str]
+
+
+class NewBaseTest(unittest.TestCase):
```
nayib-jose-gloria, line 55: Can we rename this to BaseTest? To remove the 'new' since it will stop being 'new' eventually and I'd rather refactor the name out now.

tests/unit/backend/layers/common/base_api_test.py
```diff
@@ -90,66 +41,9 @@ def _mock_assert_authorized_token(self, token: str, audience: str = None):
...
-    sample_dataset_metadata: DatasetMetadata
-
-    @classmethod
-    def setUpClass(cls) -> None:
-        cls.run_as_integration = True if os.environ.get("INTEGRATION_TEST", "false").lower() == "true" else False
-        if cls.run_as_integration:
-            cls.database_provider = DatabaseProvider(database_uri="postgresql://postgres:secret@localhost")
-            cls.database_provider._drop_schema("persistence_schema")
-
+class NewBaseAPITest(BaseAuthAPITest, NewBaseTest):
```
nayib-jose-gloria, line 44: Can we rename this to BasePortalAPITest? To remove the 'new' from the name pre-emptively, and also to clarify that this is specific to the PortalAPI tests (assuming curation API will use its own subclass, as the setUp seems portalAPI specific). Otherwise, BaseAPITest is fine if it'll be adapted for both directly.

ebezzi: Yes, we absolutely need to do this.

PR #3740, opened 2022-12-13
feat: add canonical collection tombstoning to portal redesign

nayib-jose-gloria:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- added canonical collection tombstoning logic to API, business, and persistence layers.
- added and adapted tests
- removed duplicate field caught in orm

## QA steps (optional)
- Tested against portal API and business layer unit tests (against mock and postgres instance)

## Notes for Reviewer

### DIFF
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -142,7 +142,7 @@ class PortalApi:
         return {k: v for k, v in body.items() if v is not None}
 
     # Note: `metadata` can be none while the dataset is uploading
-    def _dataset_to_response(self, dataset: DatasetVersion, is_tombstoned):
+    def _dataset_to_response(self, dataset: DatasetVersion, is_tombstoned: bool):
         return self.remove_none(
             {
                 "assay": None
@@ -227,14 +227,11 @@ class PortalApi:
         version = self.business_logic.get_published_collection_version(CollectionId(collection_id))
         if version is None:
             version = self.business_logic.get_collection_version(CollectionVersionId(collection_id))
-        else:
-            if version.canonical_collection.tombstoned:
-                raise GoneHTTPException()
+            if version is None:
+                raise ForbiddenHTTPException()
 
-        if version is None:
-            raise ForbiddenHTTPException()  # TODO: maybe remake this exception
-
-
+        if version.canonical_collection.tombstoned:
+            raise GoneHTTPException()
 
         user_info = UserInfo(token_info)
         access_type = "WRITE" if user_info.is_user_owner_or_allowed(version.owner) else "READ"
@@ -343,6 +340,7 @@ class PortalApi:
             version = self.business_logic.get_collection_version_from_canonical(resource_id)
             if version is None:
                 raise ForbiddenHTTPException()
+
         if not UserInfo(token_info).is_user_owner_or_allowed(version.owner):
             raise ForbiddenHTTPException()
 
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -6,6 +6,7 @@ from flask import Response, jsonify, make_response
 
 from backend.common.utils.http_exceptions import (
     ConflictException,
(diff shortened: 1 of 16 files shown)


backend/layers/api/portal_api.py
```diff
@@ -141,7 +142,7 @@ def remove_none(self, body: dict):
         return {k: v for k, v in body.items() if v is not None}
 
     # Note: `metadata` can be none while the dataset is uploading
-    def _dataset_to_response(self, dataset: DatasetVersion):
+    def _dataset_to_response(self, dataset: DatasetVersion, is_tombstoned):
```
ebezzi, line 145:
```suggestion
    def _dataset_to_response(self, dataset: DatasetVersion, is_tombstoned: bool):
```

backend/layers/api/portal_api.py
```diff
@@ -225,13 +227,16 @@ def get_collection_details(self, collection_id: str, token_info: dict):
         version = self.business_logic.get_published_collection_version(CollectionId(collection_id))
         if version is None:
             version = self.business_logic.get_collection_version(CollectionVersionId(collection_id))
+        else:
```
ebezzi, line 230: Nit: else is probably not needed here

backend/layers/api/portal_api.py
```diff
@@ -329,15 +334,25 @@ def get_collection_index(self):
...
-            raise ForbiddenHTTPException()
+            resource_id = CollectionId(collection_id)
+            version = self.business_logic.get_collection_version_from_canonical(resource_id)
+            if version is None:
+                raise ForbiddenHTTPException()
         if not UserInfo(token_info).is_user_owner_or_allowed(version.owner):
             raise ForbiddenHTTPException()
 
-        self.business_logic.delete_collection_version(CollectionVersionId(collection_id))
+        if isinstance(resource_id, CollectionVersionId):
```
ebezzi, line 349: I am not sure about using `isinstance` in this context. That said, I can't come up with a better solution myself.

backend/layers/business/business.py
```diff
@@ -423,9 +425,8 @@ def create_collection_version(self, collection_id: CollectionId) -> CollectionVe
         Also ensures that the collection does not have any active, unpublished version
         """
 
-        try:
-            all_versions = self.database_provider.get_all_versions_for_collection(collection_id)
-        except Exception:  # TODO: maybe add a RecordNotFound exception for finer grained exceptions
+        all_versions = self.database_provider.get_all_versions_for_collection(collection_id)
+        if not all_versions:
```
ebezzi, line 429: 👍

backend/layers/persistence/persistence_mock.py
```diff
@@ -159,7 +165,9 @@ def add_collection_version(self, collection_id: CollectionId) -> CollectionVersi
         return new_version_id
 
     def delete_collection_version(self, version_id: CollectionVersionId) -> None:
-        # Can only delete an unpublished collection
+        collection_version = self.collections_versions.get(version_id.id)
+        if collection_version.published_at:
```
ebezzi, line 169: I generally prefer an explicit `is not None`. Even if this is a super edge case, a collection published in January 1st, 1970 at midnight would result as non published 😄

ebezzi (approved):

backend/layers/api/portal_api.py
nayib-jose-gloria: we can replace with a flag if you prefer. 

Or we can rework the underlying business logic method to fetch the collection/collection version before deleting and handle the logic there. The only issue is we'd be introducing a 2nd set of lookups, since we need the lookup here to get the version owner for the auth logic handling

ebezzi: Let's leave it as is

ebezzi (approved):

Bento007 (approved):

PR #3761, opened 2022-12-15
chore: restore cloudfront invalidation

ebezzi:


backend/layers/thirdparty/cloudfront_provider_interface.py
```diff
@@ -0,0 +1,3 @@
+class CloudfrontProviderInterface:
```
nayib-jose-gloria, line 1: nit: technically Cloudfront is the implemented provider, so if we want to name a generic provider interface it should be CDNProvider or similar

backend/layers/thirdparty/cloudfront_provider_interface.py
```diff
@@ -0,0 +1,3 @@
+class CDNProviderInterface:
```
nayib-jose-gloria, line 1: need to rename the filepath as well to match

nayib-jose-gloria (approved):

PR #3778, opened 2022-12-15
feat(FMG): display marker genes as separate heatmaps

atarashansky:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3777


frontend/src/views/WheresMyGene/components/HeatMap/components/XAxisChart/index.tsx
```diff
@@ -67,19 +65,11 @@ export default function XAxisChart({ geneNames }: Props): JSX.Element {
...
-          ref={xAxisRef}
-        />
-        <MaskWrapper>
-          <CellCountLabel>Cell Count</CellCountLabel>
-          <XAxisMask />
-        </MaskWrapper>
-      </div>
-    </XAxisWrapper>
+    <XAxisContainer
+      data-test-id="gene-labels"
```
atarashansky, line 69: need to add group name to make unique test ids

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -5,15 +5,15 @@ import {
...
+import { CellType, SORT_BY, Tissue, Genes } from "../types";
 export interface PayloadAction<Payload> {
   type: keyof typeof REDUCERS;
   payload: Payload;
 }
 export interface State {
   cellTypeIdsToDelete: CellTypeMetadata[];
   genesToDelete: string[];
-  selectedGenes: string[];
+  selectedGenes: Map<string, Genes>;
```
seve, line 16: It'll probably be more legible if you make `groupName` on the `removeSelectedGenes` action an enum and then make `selectedGenes` a mapping of Genes based on that `groupName` enum.

seve: Empty string is a bit magic in the case of pre-fmg gene behavior with the `.get("")` going on

frontend/src/views/WheresMyGene/common/types.ts
```diff
@@ -9,6 +9,9 @@ export interface Organism {
 /** tissue name */
 export type Tissue = string;
 
+/* array of genes */
```
seve, line 12: gene name or id?

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveImage/index.tsx
```diff
@@ -108,12 +110,16 @@ export default function SaveImage({
       }
       link.click();
       link.remove();
-      track(EVENTS.WMG_DOWNLOAD_COMPLETE, { file_type: fileType , tissues: selectedTissues.toString(), genes: selectedGenes.toString() });
+      track(EVENTS.WMG_DOWNLOAD_COMPLETE, {
+        file_type: fileType,
+        tissues: selectedTissues.toString(),
+        genes: selectedGenes.toString(),
```
seve, line 116: This output is going to be a bit diff than it has historically, might want to run this past amanda

seve (commented):
prelim review

PR #3795, opened 2022-12-19
fix: missing 'e' in explorer_url + dataset id in meta endpoint

ebezzi:


metakuni (approved):

atolopko-czi (approved):
Looks reasonable to me.  As I understand it, API response for _unpublished_ collection versions will include (unpublished) datasets, which must be represented by their dataset _version_.

PR #3809, opened 2022-12-20
fix: version_id in meta endpoint

ebezzi:


atolopko-czi (approved):
Discussed in Slack. LGTM

PR #3814, opened 2022-12-20
feat: PATCH collection curator API redesign

ebezzi:


backend/layers/common/validation.py
```diff
@@ -22,7 +22,11 @@ def _verify_collection_metadata_fields(
 
     def check(key):
         value = getattr(metadata, key)
-        if check_existence and not value:
+        if check_existence and value is None:
```
ebezzi, line 25: I am not a fan of this method, but this should check two things:
1. The field must exist if `check_existence` is true (namely, if this is a collection _creation_ where all the fields are required)
2. If a field exists, it cannot be falsey (namely, an empty string).

Maybe it would be better to split this behavior into two separate methods, rather than two if conditions.

backend/portal/api/curation/v1/curation/collections/collection_id/actions.py
```diff
@@ -41,31 +41,55 @@ def get(collection_id: str, token_info: dict) -> Response:
...
+
+    # TODO: dedup
+    def _link_from_request(body: dict):
+        return Link(
+            body.get("link_name"),
+            body["link_type"],
+            body["link_url"],
+        )
+
+    if body.get("links") is not None:
```
nayib-jose-gloria, line 72: nit: we already have a None check on body['links'] above

backend/layers/business/business.py
```diff
@@ -242,6 +245,11 @@ def update_collection_version(self, version_id: CollectionVersionId, body: Colle
                     value.strip_fields()
                 setattr(new_metadata, field, value)
 
+        # Issue all updates
+        if unset_publisher_metadata:
+            self.database_provider.save_collection_publisher_metadata(version_id, None)
+        if set_publisher_metadata is not None:
```
nayib-jose-gloria, line 251: nit: i believe its not possible for both to be true as written, but its easier to quickly understand the intention if we make this an elif

backend/layers/business/business.py
```diff
@@ -221,13 +220,17 @@ def update_collection_version(self, version_id: CollectionVersionId, body: Colle
...
+
         if old_doi and new_doi is None:
             # If the DOI was deleted, remove the publisher_metadata field
-            self.database_provider.save_collection_publisher_metadata(version_id, None)
+            unset_publisher_metadata = True
         elif (new_doi is not None) and new_doi != old_doi:
             # If the DOI has changed, fetch and update the metadata
-            publisher_metadata = self._get_publisher_metadata(new_doi, errors)
-            self.database_provider.save_collection_publisher_metadata(version_id, publisher_metadata)
+            set_publisher_metadata = self._get_publisher_metadata(new_doi, errors)
```
nayib-jose-gloria, line 233: nit: set_publisher_metadata sounds more like a flag name next to the unset_publisher_metadata flag. Maybe this can be updated_publisher_metadata instead? But I don't feel super strongly, just an idea

backend/layers/common/validation.py
```diff
@@ -22,7 +22,11 @@ def _verify_collection_metadata_fields(
...
+            # if checks_existence is true, value cannot be None since it must be required
+            errors.append({"name": key, "reason": "Cannot be blank."})
```
nayib-jose-gloria, line 27: lets give this error reason a different value than the one in the next block (or vice-versa). We should distinguish the error paths

nayib-jose-gloria (commented):
a few nits, but this mostly looks good

backend/layers/business/business.py
ebezzi: I went for `publisher_metadata_to_set`. Let me know if that makes sense

nayib-jose-gloria (approved):

PR #3815, opened 2022-12-20
fix: functional tests portal redesign compatibility

nayib-jose-gloria:
## Changes
- re-introduced 'original_id' to datasets API response object, using canonical dataset id. Needed to access the original dataset associated with a private revision of a published dataset 
- remove duplicated 'delete' path in portal-api definition yaml
- raise a descriptive error if curator_name is not provided for a collection

### DIFF
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -182,6 +182,7 @@ class PortalApi:
                 "organism": None
                 if dataset.metadata is None
                 else self._ontology_term_ids_to_response(dataset.metadata.organism),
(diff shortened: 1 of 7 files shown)


backend/layers/api/portal_api.py
```diff
@@ -304,6 +305,9 @@ def create_collection(self, body: dict, user: str):
         if doi_node := doi.get_doi_link_node(body, errors):
             if doi_url := doi.portal_get_normalized_doi_url(doi_node, errors):
                 doi_node["link_url"] = doi_url
+        curator_name = body.get("curator_name")
+        if curator_name is None:
```
ebezzi, line 309: 👍

PR #3818, opened 2022-12-21
chore: update local wmg snapshot generation

atarashansky:


danieljhegeman (approved):
thanks @atarashansky 🚀

PR #3821, opened 2022-12-21
fix: send slack alerts on failing post-deploy tests

nayib-jose-gloria:
## Changes
- slack alerts will now send on post-deploy test failures; previously the syntax was set up to depend on post-deploy test success to run the 'status' step that sends alerts to slack + checked the 'status' job status rather than the overall workflow status
- Brought in third party GH action to do overall workflow status check

## Notes for Reviewer
- Link to GHA third party solution docs for [reference](https://github.com/martialonline/workflow-status)


atolopko-czi (approved):
Looks sane, thanks for the fix!

Issue #3689, opened 2022-12-07
feat(FMG): hide i icon when cell type has <25 cells


atarashansky:
marker genes can't be found if cell counts is < 25, so we should disable the functionality by hiding the i button

signechambers1:
Suggested language: Marker genes are unavailable for cell types with less than 25 cells.

signechambers1:
Suggested language: Marker genes are unavailable for cell types with less than 25 cells.

ashin-czi:
> Suggested language: Marker genes are unavailable for cell types with less than 25 cells.

Was something like this what you had in mind?
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/109984998/208993775-27a606ed-992a-4d45-9e9a-fa1591e118cd.png">

signechambers1:
Looks good to me! fyi @karenliang0

signechambers1:
good catch by @karenliang0, @ashin-czi im sorry, we had a convo in slack yesterday and I forgot to update this ticket. We had actually decided to remove the i icon for cell types with less then 25 cells instead of having the panel because we think it is less misleading to users.

PR #3825, opened 2022-12-21
feat: marker genes unavailable for cell types <25

ashin-czi:
- #3689

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
@@ -212,6 +212,7 @@ export default memo(function YAxisChart({
         isMarkerGenes &&
         yAxisInfoCoords.map((coord, i) => {
           const content = cellTypeMetadata[i];
+          const cellType = isMarkerGenes ? deserializeCellTypeMetadata(content) : null;
           return (
             <InfoButtonWrapper
               id={content}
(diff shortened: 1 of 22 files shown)


atarashansky (approved):
I think it looks right, but the diff is atrocious xD.

ashin-czi:
Based on [this comment](https://github.com/chanzuckerberg/single-cell-data-portal/issues/3689#issuecomment-1363253025), we're just going to be hiding the info icon entirely based on the cell type count

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -212,18 +212,19 @@ export default memo(function YAxisChart({
...
               left={coord[0]}
               top={coord[1]}
               onClick={() => {
-                if (isMarkerGenes) {
-                  const cellType = deserializeCellTypeMetadata(content);
+                if (cellType) {
                   generateMarkerGenes(cellType, tissueID);
                 }
               }}
+              style={cellType && cellType.total_count < 25 ? {visibility: "hidden"} : {}}
```
tihuan, line 227: should we just return `null` instead of `visibility: "hidden"` 😄 ? `hidden` is good if we expect to unhide the DOM element after user interaction, but that doesn't seem to be the case here, so we can save some memory and CPU cycle on rendering unused elements 👌 !

tihuan (approved):
Awesomee! LGTM 🚀 Thanks so much, @ashin-czi !!

PR #3827, opened 2022-12-22
chore: redesign dataset submissions lambda

ebezzi:


nayib-jose-gloria (commented):
I believe there's a missing ingest_dataset call method signature to update in portal_api.upload_from_link

backend/layers/business/business.py
```diff
@@ -343,10 +343,18 @@ def ingest_dataset(
...
                 )
 
-            # TODO: this method could very well be called `add_dataset_version`
-            new_dataset_version = self.database_provider.replace_dataset_in_collection_version(
-                collection_version_id, existing_dataset_version_id
-            )
+            # If a dataset is "empty", we will not replace it but instead reuse the existing version id.
+            # Make sure that the conditions are not relaxed or this will break immutability.
+            if (
+                dataset_version.status.processing_status == DatasetProcessingStatus.INITIALIZED
```
nayib-jose-gloria, line 349: question: can a dataset ever have processing status INITIALIZED and also have artifacts and be published? I would assume processing being SUCCESS are prerequisites to those two other clauses being true

obviously this will work either way and is safer in case an edge case is introduced, so maybe we should stick with it anyway, but just curious if we can safely simplify the logic here

ebezzi: No, because as soon as we start processing a dataset, the status becomes `PENDING`, so the other conditions are indeed superfluous. But since this is a very strong constraint of our system, I think this is still useful to future-proof the code.

nayib-jose-gloria (approved):

PR #3834, opened 2022-12-23
chore: double lookup for GET dataset in curation API

ebezzi:


nayib-jose-gloria (approved):

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -37,7 +37,10 @@
 def get(collection_id: str, dataset_id: str = None):
     business_logic = get_business_logic()
 
+    # Double lookup
     collection_version = business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
```
nayib-jose-gloria, line 41: nit: can we use the common util method get_infered_collection_version_else_forbidden? ignore if we want to return a 404 instead

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -37,7 +37,10 @@
 def get(collection_id: str, dataset_id: str = None):
     business_logic = get_business_logic()
 
+    # Double lookup
```
dsadgat, line 40: nit - this comment would be more beneficial if it explained why we need the double lookup, as opposed to just saying that we are implementing it. That way in the future someone will know not to refactor / why it was put here in the first place

dsadgat (approved):

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
ebezzi: I think we need to refactor those methods but for now let's keep the existing behavior

Issue #3668, opened 2022-12-05
Standardize ID naming and generation in persistence layer


nayib-jose-gloria:
Translate ORM and common entity IDs to use 'id' name for primary keys (i.e. version_id -> id for CollectionVersion). 

Generate UUIDs as part of ID objects, rather than as a static method in the DatabaseProvider (as is current)

Refactor ORM tables to use class variables convention

Refactor table names to differentiate from business entities 

Fix spots in the persistence layer where transactionality is not being implemented properly

Update migration script accordingly

PR #3835, opened 2022-12-23
refactor: persistence layer clean-up

nayib-jose-gloria:
- #3668 

## Changes
- refactor + standardize entity ID generation as an ID class method, create entity ID base dataclass for common functionalities 
- refactored orm to use class attribute-based declarative convention + renamed field names to meet standard conventions (i.e. version_id in CollectionVersion --> id) 
- update migration script + persistence layer with new field names accordingly 
- updated finalize_collection_version method to use 1 session for transactional safety 
- renamed ORM objects to use -Table suffix
- Alembic migration script for db field name changes

## Notes for Reviewer
- had to change 'metadata' field in CollectionVersion and DatasetVersion to collection_metadata and dataset_metadata because 'metadata' is a reserved class attribute name in sqlalchemy for the ORM class attribute-based declarative syntax  
- TODO: Change business entity names to correspond as well? Not necessary but maybe it'd be easier to understand. OR, go the opposite direction and change the few business entities with 'id' fields (CanonicalCollection, DatasetArtifact) to use more specific names in order to better distinguish between the business and form models


backend/layers/persistence/orm.py
```diff
@@ -10,71 +10,61 @@
...
-        Column("created_at", DateTime),
-        Column("datasets", ARRAY(UUID(as_uuid=True))),
-    )
+class CollectionVersionTable:
+
+    __tablename__ = "CollectionVersion"
+
+    id = Column(UUID(as_uuid=True), primary_key=True)
+    collection_id = Column(UUID(as_uuid=True))
+    collection_metadata = Column(JSON)
```
Bento007, line 30: In order to be able to use `metadata` you'll need to switch to [declarative-mapping](https://docs.sqlalchemy.org/en/14/orm/mapping_styles.html#declarative-mapping)
```python
from sqlalchemy import Column, Integer, String, ForeignKey
from sqlalchemy.orm import declarative_base

# declarative base class
Base = declarative_base()

# an example mapping using the base
class User(Base):
    __tablename__ = "user"

    id = Column(Integer, primary_key=True)
    name = Column(String)
    fullname = Column(String)
    nickname = Column(String)
```

or you can try mapping the column like this: 
```suggestion
    collection_metadata = Column("metadata", JSON)
```

Bento007 (commented):
Now that we've already done the manual migration, you'll need to generate a migration script using [these instructions](https://github.com/chanzuckerberg/single-cell-data-portal/tree/main/backend/database). Otherwise you can see my comment bellow for a way to avoid doing the migration.

backend/layers/common/entities.py
```diff
@@ -98,43 +100,34 @@ def empty():
 
 
 @dataclass
-class CollectionId:
+class EntityId:
```
ebezzi, line 103: Great idea. I initially went for the same approach but I discarded it (but I can't remember the reason - maybe something with dataclasses?). Anyway if it works this is certainly great.

backend/layers/persistence/orm.py
ebezzi: I agree it would be preferable to keep `metadata` and use Trent's solution.

nayib-jose-gloria: I'm getting a similar error when I try this: sqlalchemy.exc.InvalidRequestError: Attribute name 'metadata' is reserved for the MetaData instance when using a declarative base class.

Bento007: The second solution will work for sure, but then we will have different naming. Not the end of the world IMO.

nayib-jose-gloria: FWIW we'd still need to do a migration to fix the id naming (i.e. for DatasetVersions, renaming dataset_version_id -> id, etc.) so at that point I think I'd rather migrate the metadata names as well and keep internal consistency with the naming. 

Alternatively, with imperative mapping (the 'Table' object mapping we're currently doing in main) we can keep the metadata names. Still requires migration for the Id rename though, which I do think is worth fixing.

DEV_ENV.md
```diff
@@ -51,11 +51,12 @@ The dev environment is initialized with AWS Secrets/S3 data in the [scripts/setu
...
-| Command                      | Description                                   | Notes |
-| ---------------------------- | --------------------------------------------- | ----- |
-| `make local-unit-test`       | Run backend tests in the dev environment      |       |
-| `make local-functional-test` | Run functional tests in the dev environment   |       |
-| `make local-smoke-test`      | Run frontend/e2e tests in the dev environment |       |
+| Command                      | Description                                                                                  | Notes |
+|------------------------------|----------------------------------------------------------------------------------------------| ----- |
+| `make local-unit-test`       | Run backend tests in the local dockerized environment, against mock of persistence layer     |       |
+| `make local-integration-test`      | Run backend tests in the local dockerized environment, against dockerized database instance. |       |
+| `make local-functional-test` | Run functional tests in local dockerized environment                                         |       |
```
atolopko-czi, line 58: `Run [backend] functional tests`?

backend/database/versions/34_2be441104b48_standardize_db_fields.py
```diff
@@ -0,0 +1,60 @@
...
+revision = '34_2be441104b48'
+down_revision = '33_c5aaf6e2ca9e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('Collection', sa.Column('tombstone', sa.BOOLEAN(), nullable=True), schema='persistence_schema')
+    op.drop_column('Collection', 'tombstoned', schema='persistence_schema')
```
Bento007, line 22: Where are we copying the old column to the new column? I think running this migration will result in data loss. You need to [alter_column](https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.BatchOperations.alter_column)

Bento007 (changes requested):
Change the migration script to avoid data loss.

backend/Makefile
```diff
@@ -75,11 +75,11 @@ db/local/load-data:
 
 db/local/load-schema:
     # Imports the corpora_dev.sqlc schema (schema ONLY) into the corpora_test database
-	# Usage: make db/local/load-schema INFILE=<file>
+    # Usage: make db/local/load-schema INFILE=<file>
     $(eval DB_PW = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/test/database --region us-west-2 | jq -r '.SecretString | match(":([^:]*)@").captures[0].string'))
-	PGPASSWORD=${DB_PW} pg_restore --schema-only --clean --no-owner --host 0.0.0.0 --dbname corpora $(INFILE)
-	# Also import alembic schema version
-	PGPASSWORD=${DB_PW} pg_restore --data-only --table=alembic_version --no-owner --host 0.0.0.0 --dbname corpora $(INFILE)
+	PGPASSWORD=${DB_PW} pg_restore --schema-only --schema=persistence-schema --clean --no-owner --host 0.0.0.0 --username corpora --dbname corpora $(INFILE)
```
Bento007, line 80: For now don't we want both schemas, until we depreciate corpora_orm.py

nayib-jose-gloria: I wasn't able to get it to work without adding this arg, I kept getting strange pg_restore errors. For example:
```
pg_restore: error: could not execute query: ERROR:  extension "aws_s3" does not exist
Command was: DROP EXTENSION aws_s3;
pg_restore: error: could not execute query: ERROR:  extension "aws_commons" does not exist
Command was: DROP EXTENSION aws_commons;
```
I figured limiting this to the persistence_schema is fine, given that this command is only used to load the schema into the test DB to detect diffs after a test migration. By the time this is in main, we won't have plans to support schema changes in corpora_orm anymore. And if we have to revert for whatever reason, this would be reverted as well.

Bento007 (approved):

backend/Makefile
```diff
@@ -75,11 +75,11 @@ db/local/load-data:
 
 db/local/load-schema:
     # Imports the corpora_dev.sqlc schema (schema ONLY) into the corpora_test database
-	# Usage: make db/local/load-schema INFILE=<file>
+    # Usage: make db/local/load-schema INFILE=<file>
     $(eval DB_PW = $(shell aws secretsmanager get-secret-value --secret-id corpora/backend/test/database --region us-west-2 | jq -r '.SecretString | match(":([^:]*)@").captures[0].string'))
```
Bento007, line 79: make sure this is a tab not spaces. It showed up as space in my IDE and wouldn't run.

PR #3841, opened 2023-01-03
chore: update alembic migration to use new database

Bento007:
- https://github.com/chanzuckerberg/single-cell-data-portal/issues/3872
### Reviewers

**Readability:** @ebezzi

---

## Changes
- Upgrade SQLAlchemy for the processing container so it matches the other containers
- Upgrade Alembic for local development.
- factor out "persistance_schema" into a constant to avoid naming issues.
- Update alembic to support both the new and legacy ORM for now. The legacy ORM can be removed in a future migration.
- Add an Alembic migration for the redesign database changes. This is will not affect the legacy database.
- Modify `migrate_redesign_write` so it no longer create the new database tables. This should be handled by the automated migration processes in GHA.
- Fixe the legacy ORM to match the actual database.
- Update `create_db` to use the new ORM.
- updated sqlalchemy in the processing container to be current with other containers.
- Add Make recipe `db/check` to make it easier to detect if the ORM is different from the database. 
- Add a try..except block in `DatabaseProvider._drop_schema` so prevent breaking if the schema does not exist.
- Add support for creating both the legacy and new database.
- removing CORPORA_LOCAL_DEV from setup_dev_data.sh. 

## QA steps (optional)
- copied the database from dev and ran `make db/test_migration`

## Notes for Reviewer
- @ebezzi We will need to roll back the database so the auto migration script can run. Then we will need to rerun the data migration scripts. Alternatively we will need to manually modify the `alembic_version` in that database to match the latest migration. This can be done by connecting to the DB locally and running `alembic stamp 33_`.


backend/database/versions/33_95ac900d53ef_redesign.py
```diff
@@ -0,0 +1,99 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "Collection",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("version_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("originally_published_at", sa.DateTime(), nullable=True),
+        sa.Column("tombstoned", sa.BOOLEAN(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+        schema="persistence_schema",
```
ebezzi, line 28: I believe the schema`persistence_schema` also needs to be created, although I am not sure.

Bento007: Added it

backend/layers/persistence/persistence.py
```diff
@@ -45,27 +46,31 @@
...
-    def __init__(self, database_uri: str = None, schema_name: str = "persistence_schema") -> None:
+    def __init__(self, database_uri: str = None, schema_name: str = SCHEMA_NAME) -> None:
         if not database_uri:
             database_uri = CorporaDbConfig().database_uri
         self._engine = create_engine(database_uri, connect_args={"connect_timeout": 5})
         self._session_maker = sessionmaker(bind=self._engine)
+        self._schema_name = schema_name
         try:
-            self._create_schema(schema_name)
+            self._create_schema(self._schema_name)
```
nayib-jose-gloria, line 56: remove arg here, now that _create_schema method signature has changed

backend/database/versions/33_c5aaf6e2ca9e_redesign.py
```diff
@@ -0,0 +1,89 @@
...
+        sa.Column("published_at", sa.DateTime(), nullable=True),
+        sa.PrimaryKeyConstraint("dataset_id"),
+        schema="persistence_schema",
+    )
+    op.create_table(
+        "DatasetArtifact",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("type", sa.Enum("RAW_H5AD", "H5AD", "RDS", "CXG", name="datasetartifacttype"), nullable=True),
+        sa.Column("uri", sa.String(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
```
nayib-jose-gloria, line 58: primary key is 'dataset_id'

nayib-jose-gloria (commented):
one minor fix needed, otherwise this looks good!

backend/layers/persistence/persistence.py
Bento007:
```suggestion
            self._create_schema()
```

nayib-jose-gloria (approved):

Bento007:
Updated the alembic head for dev environment

Bento007:
Updated the alembic head for staging environment

PR #3843, opened 2023-01-04
chore: add filter for local.h5ad to wmg extract function

ebezzi:
Reintroduces the filter for `local.h5ad` file names. Even if we support a raw h5ad type, this isn't enforced so this check is still necessary.

### DIFF
--- a/tests/unit/backend/layers/business/test_business.py
+++ b/tests/unit/backend/layers/business/test_business.py
@@ -190,13 +190,13 @@ class BaseBusinessLogicTestCase(unittest.TestCase):
         is a complex process which happens asynchronously, and cannot be easily mocked.
         """
         self.database_provider.add_dataset_artifact(
-            dataset_version_id, DatasetArtifactType.H5AD.value, "s3://fake-bucket/artifact.h5ad"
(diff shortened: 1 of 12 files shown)


tests/unit/backend/layers/business/test_business.py
```diff
@@ -775,6 +775,9 @@ def test_get_dataset_artifacts_ok(self):
         self.assertCountEqual(
             [a.type for a in artifacts], [DatasetArtifactType.H5AD, DatasetArtifactType.CXG, DatasetArtifactType.RDS]
         )
+        self.assertCountEqual(
+            [a.get_file_name() for a in artifacts], ["local.h5ad", "local.cxg", "local.rds"]
```
atarashansky, line 779: Should we also check for equality of values instead of just number of files?

atarashansky (approved):
LGTM! Just a super minor question in the tests

tests/unit/backend/layers/business/test_business.py
ebezzi: Turns out that `assertCountEqual` actually [checks if two iterables have the same elements regardless of order](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual), not just the count.

PR #3850, opened 2023-01-04
chore: update docs for redesign

ebezzi:
Documentation backfill for previous functions.

Also changes the typing of some function to reflect the latest changes.


backend/layers/api/portal_api.py
```diff
@@ -430,6 +456,9 @@ def publish_post(self, collection_id: str, body: object, token_info: dict):
         return make_response({"collection_id": version.collection_id.id, "visibility": "PUBLIC"}, 202)
 
     def upload_from_link(self, collection_id: str, token_info: dict, url: str, dataset_id: str = None):
+        """
```
nayib-jose-gloria, line 459: I believe this method is used only for dropbox links (at least per the error handling), while the `upload_link` method below is used for presigned s3 URLs. If this is the case, should be noted in the comment

That said, it probably could be consolidated. But we don't need to do that here

nayib-jose-gloria (approved):
one suggestion, otherwise looks good

atolopko-czi (commented):
Thanks for including me as reviewer, but 2 reviewers is enough. :-)

backend/layers/processing/process_download_validate.py
```diff
@@ -25,6 +25,18 @@
...
+    Base class for handling the `Download and Validate` step of the step function.
+    This will:
+    1. Download the original artifact from the provided URI
+    2. Run the cellxgene-schema validator
+    3. Save and upload a labeled copy of the original artifact (local.h5ad)
+    4. Upload a copy of the original artifact (raw.h5ad)
+    5. Persist the dataset metadata on the database
+    6. Determine if a Seurat conversion is possible (it is not if the X matrix has more than 2**32-1 nonzero values)
+    If this step completes successfully, ProcessCxg and ProcessSeurat will start in parallel.
+    If this step fails, the handle_failures lambda will be invoked
```
Bento007, line 38: You are not using `.`s consistently in your documentation.
```suggestion
    If this step fails, the handle_failures lambda will be invoked.
```

Bento007 (approved):
In general you are not putting period at the end of the docstring. Make the rest of these changes consistent with that pattern.

Issue #3853, opened 2023-01-04
Find Marker Genes panel: Add tool tip to Marker Score


signechambers1:
We'd like to add a tool tip with the question mark icon to Marker Score. The tool tip should read: "Marker Score indicates the strength and specificity of a gene as a marker. It is the 5th percentile of the effect sizes when comparing the expressions in a cell type of interest to each other cell type in the tissue. [Learn more here](https://cellxgene.cziscience.com/docs/04__Analyze%20Public%20Data/4_2__Gene%20Expression%20Documentation/4_2_5__Find%20Marker%20Genes)"

[Marker score tooltip design
](https://www.figma.com/file/blyEMZDsi7EMopyyIX4XAg/Q1-2023-Gene-Expression-Improvements?node-id=121%3A377895&t=4ZcJ1IZJ1uUi9uPF-1)

atarashansky:
Effect size has been renamed to Marker Score in [this PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3855). The tool-tip will be added later, presumably in the subsequent deployment cycle.

pablo-gar:
@atarashansky, @signechambers1  asked me to provide text for the tool tip here it is:

> Marker Score represents indicates the strength and specificity of a gene as a marker, it is the 5th percentile of effect size when comparing its expression in the cell type against each of all other cell types. [Learn more](link to docs)

PR #3855, opened 2023-01-04
chore(FMG): change the effect size label in the marker gene panel to "marker score"

atarashansky:
- #3853

This changes the effect size label in the marker gene panel to "Marker Score" which is a more accurate description of the numbers being shown.

A separate PR will add the tooltip next week.


hthomas-czi (approved):
LGTM!

Issue #3849, opened 2023-01-04
Bug: Application error when trying to get to a dot plot on Gene Expression


signechambers1:
Steps to reproduce:

1. Prod, latest version of chroms
2. Select blood, TSPAN6
3. Select dataset 49 year old males - Fresh PBMCs (2 days post intubation)
4. See error messages attached
![Screen Shot 2023-01-04 at 10.26.57 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/9fcf5738-378e-4694-b82a-3d96aa619fc0)

seve:
@tihuan to add details

tihuan:
Looks like `deserializeCellTypeMetadata()` is getting `undefined` for argument `cellTypeMetadata`. Not sure if we're now expecting passing `undefined` to `deserializeCellTypeMetadata`, if so, we need to handle that case in the helper function now. If we are not expecting passing `undefined`, then we need to fix that!

seve:
This is no longer occurring in dev, I believe the issue with the bug was related to the removal or cell types when narrowing the dataset scope

signechambers1:
@seve I'm still seeing this issue in prod, is that expected?

signechambers1:
Ah I see should be out with the prod deploy this week. Thanks!

Issue #3800, opened 2022-12-19
Batch delete gene requests to app isn't overwhelmed


signechambers1:
From Alec: If users use the gene selector to unselect many (e.g. 30) genes in a row (instead of refreshing the page), that will trigger 30 queries to be sent to the BE. This will DOS the site. I accidentally shut down staging for 5 minutes because I unselected like ~50 genes.

tihuan:
@seve is the plan to add `Apply` `Cancel` to `QuickSelect` like how SDS Dropdown works?

![Screenshot 2023-01-04 at 9.12.48 AM.png](https://images.zenhubusercontent.com/5ebed8780d2d150fc281ad55/7f3356e6-14a8-4f0c-a3dd-05d7d289c096)?

seve:
The plan is to only send the request once the dropdown has closed

tihuan:
Got it!

PR #3859, opened 2023-01-04
fix(wmg): don't send requests if only genes or tissues have been removed

seve:
- #3800 
### Reviewers
**Functional:**  @tihuan


frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -216,9 +214,12 @@ export default memo(function YAxisChart({
       {yAxisInfoCoords &&
         isMarkerGenes &&
         !FMG_EXCLUDE_TISSUES.includes(tissue) &&
+        cellTypeMetadata.length > 0 &&
```
seve, line 217: fixes #3849?

tihuan: This quick fix stops the runtime error, but  do we know why `cellTypeMetadata` can be an empty array or undefined here? In my mind every cell type should always have serialized metadata, so `cellTypeMetadata` should always be available 🤔

tihuan (approved):
Thanks for the great solution, Seve! LGTM 🚀

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
seve: Just for posterity I followed the thread earlier when I encountered the bug and saw that `cellTypeMetadata` is empty when `cellTypes` is empty

tihuan: Thanks for digging into this! 

Interesting, the surrounding code was introduced when the FMG info button was added, how come we're only seeing the bug now? And does the runtime error only happen for certain datasets or all of them? 

IIRC the info buttons were rendering just fine before the holidays, so not sure what code change caused the runtime error now 🤔

tihuan: CC @atarashansky in case you have more info here! Thanks!

seve: fix is void from upstream changes on main

PR #3860, opened 2023-01-05
fix: run `npm run format`

seve:


Bento007:
Would it be worth adding as a commit hook to automate this during commits.

PR #3869, opened 2023-01-05
fix: downloaded image being cut off

ashin-czi:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
    - remove `width: fit-content;` in `globalStyles.ts` as it makes the chart jump/disappear on download and not show up in the generated image
- modify
    - increasing hardcoded width to not make legend be cut off


ashin-czi:
**Before**:
lung (21)

**After**: 
lung (22)

atarashansky:
@ashin-czi I tightened up the whitespace around the heatmap and fixed a bug from before that caused heatmaps that were too large to be cut off.

Issue #3623, opened 2022-11-22
(FMG): Add documentation for fmg to the doc site


atarashansky:

signechambers1:
Couple notes:
* Include that marker genes are unavailable for cell types with <25 cells and for blood

signechambers1:
@atarashansky i think this is in progress?

PR #3870, opened 2023-01-05
chore(fmg): fmg documentation

atarashansky:
- #3623 

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


frontend/src/pages/docs/[...slug].tsx
```diff
@@ -8,6 +8,8 @@ import {
   MDXRemoteProps,
   MDXRemoteSerializeResult,
 } from "next-mdx-remote";
+import rehypeMathJax from 'rehype-mathjax';
+import remarkMath from "remark-math";
```
Comment (seve) [inline frontend/src/pages/docs/[...slug].tsx:12]: Shouldn't there be a package.json diff?

seve (changes requested):

frontend/src/pages/docs/[...slug].tsx
Comment (atarashansky) [reply frontend/src/pages/docs/[...slug].tsx:12]: Definitely added it to the commit, not sure why it's not there...trying again

signechambers1:
@atarashansky would you mind adding the additional caveat that we do not calculate markers for cell types with <25 cells or blood?

Otherwise looks good to me. Tagging @ambrosejcarr for review as well.

atarashansky:
> @atarashansky would you mind adding the additional caveat that we do not calculate markers for cell types with <25 cells or blood?
> 
> Otherwise looks good to me. Tagging @ambrosejcarr for review as well.

It's already in there, the last bullet point!

ambrosejcarr (approved):
LGTM. Thanks for flagging me for review.

Issue #3802, opened 2022-12-19
Bring the latest set of marker genes to the front of the dot plot


signechambers1:
Currently, if you have one set of marker genes, and then add a second with only 3 new unique marker genes, only 3 genes will be added.

The desired behavior is that all 25 will be ordered as the latest set of marker genes. This will rearrange prior ordering of marker genes, but will make it easier for the user to figure out what is happening.

PR #3874, opened 2023-01-05
feat(fmg): shift marker genes to front if already selected when adding

seve:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3802

Changed reducer logic to filter out existing genes and then insert all at front when adding marker genes to the heatmap

https://user-images.githubusercontent.com/8716829/210895827-63e94b90-f113-4e7b-88f4-3df2c988f694.mov


atarashansky (approved):
LGTM!

PR #3877, opened 2023-01-06
fix(wmg): y axis is now properly sticky

atarashansky:
Y axis stopped being sticky when scrolling horizontally on a very large heatmap. The fix involved (a) changing the y axis/chart container to be inline instead of flex and (b) ensuring that both y axis and chart had the same height.

### DIFF
--- a/frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
@@ -1,7 +1,6 @@
 import cloneDeep from "lodash/cloneDeep";
 import { memo, useContext, useMemo, useRef, useState } from "react";
 import { EMPTY_ARRAY } from "src/common/constants/utils";
-import { useResizeObserver } from "src/common/hooks/useResizeObserver";
 import {
   generateTermsByKey,
   OntologyTerm,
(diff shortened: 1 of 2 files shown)


frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
```diff
@@ -161,10 +158,8 @@ export default memo(function HeatMap({
           <XAxisMask />
           <XAxisChart geneNames={sortedGeneNames} />
         </XAxisWrapper>
-        <FlexRow>
-          <YAxisWrapper
-            height={(chartWrapperRect?.height || 0) - X_AXIS_CHART_HEIGHT_PX}
-          >
+        <InlineRow>
```
tihuan, line 161: I think we can just use div since by default it's inline! Or maybe we no longer need it at all?

tihuan (approved):
LGTM! Thanks so much for your patience on fixing this hairy issue 🙏

frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
atarashansky: True! It's not needed period.

PR #3879, opened 2023-01-06
chore(wmg): rewrite x axis using react

atarashansky:
This PR rewrites the X axis to use react so that we have implementation parity between X and Y axes.

One caveat is that the Y axis uses buttons whereas the X axis uses divs with click listeners. I had trouble orienting text inside the buttons sideways using CSS. 

@tihuan Do you think this is problematic (div vs button)?


atarashansky:
- I converted the div elements to buttons with an inner div element containing the text
 - I used @ashin-czi 's very handy `formatCellLabel` (now `formatLabel`) function to perform inner truncation of gene labels that are too long.

atarashansky:
Tagging @seve and @ashin-czi in review for visibility to this change.

frontend/src/views/WheresMyGene/components/HeatMap/components/XAxisChart/index.tsx
```diff
@@ -1,75 +1,67 @@
...
+  X_AXIS_CHART_HEIGHT_PX,
+  Y_AXIS_CHART_WIDTH_PX,
+} from "../../utils";
+import {
+  XAxisWrapper,
+  XAxisContainer,
+  XAxisLabel,
+  GeneButtonStyle,
+} from "./style";
+import { formatLabel } from "../YAxisChart";
```
tihuan, line 15: low pri nit: ideally if we could extract `formatLabel` to a common util folder (maybe `frontend/src/views/WheresMyGene/components/HeatMap/utils.ts`?, then XAxis component wouldn't need to depend on YAxis component at all!

tihuan (approved):
Looking great! Just a non-blocking comment 👼 Thanks again for the great work!

PR #3881, opened 2023-01-07
fix(wmg): Select Tissues/Genes dropdown not triggering onClick event

tihuan:
- #TICKET_NUMBER

### Reviewers
**Functional:** 
@atarashansky 

**Readability:** 
@ainfeld 
---

## Changes
This triggers the onClick analytics event when selecting a previously unselected option again

How it broke:
After migrating to MUIv5, MUI adds their own `onClick()` handler now, which overrides our custom one, thus the need to merge the two now

## QA steps
1. Select a tissue option, you should see a Plausible response
2. In the GIF below, you can see that the warning message count is incrementing, which is a warning that Plausible received the event but is not reporting it to the server because this is a local dev environment

demo

### DIFF
--- a/frontend/src/views/WheresMyGene/components/Filters/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/Filters/index.tsx
@@ -6,7 +6,6 @@ import {
 } from "czifui";
 import isEqual from "lodash/isEqual";
 import {
-  Fragment,
   memo,
   ReactElement,
   useCallback,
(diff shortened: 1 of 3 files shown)


frontend/src/views/WheresMyGene/components/Filters/index.tsx
```diff
@@ -241,13 +240,15 @@ export default memo(function Filters({ isLoading }: Props): JSX.Element {
...
+    if (areFiltersDisabled) {
+      return (
+        <Tooltip title="Please select an organism, tissue and at least one gene to use these filters.">
+          {children}
+        </Tooltip>
+      );
+    }
 
-    return (
-      <Wrapper title="Please select an organism, tissue and at least one gene to use these filters.">
```
tihuan, line 247: This triggers React warning that `title` attr is not supposed to be passed to `Fragment`

frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
```diff
@@ -100,7 +100,13 @@ export default memo(function Chart({
   useEffect(() => {
     const { current } = ref;
 
-    if (!current || isChartInitialized) {
+    if (
+      !current ||
+      isChartInitialized ||
+      // (thuang): echart's `init()` will throw error if the container has 0 width or height
```
tihuan, line 106: this eliminates echarts error message

atarashansky (approved):
thanks so much for fixing this!

PR #3882, opened 2023-01-08
fix(wmg): fix echarts initialization

atarashansky:
Currently, echarts initialization is broken for the chart introduced by [this PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3881). The fix is to add heatmapWidth and height to the dependencies in the useEffect so that when they update, the initialization triggers. For some reason, the update to the reference is not enough to trigger useEffect to run again.


tihuan:
Thanks so much for noticing and fixing this, Alec 🙏 ! Sorry for breaking it too 😆 

PTAL my proposed fix, so we don't add deps to `useEffect` we don't use: https://github.com/chanzuckerberg/single-cell-data-portal/pull/3884

Thank you!

PR #3884, opened 2023-01-09
fix(wmg): echarts init

tihuan:
- #TICKET_NUMBER

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
alternative to https://github.com/chanzuckerberg/single-cell-data-portal/pull/3882

@atarashansky found that I broke echarts init in this [PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3882) (Thanks so much 🙏 !) And this is my proposed fix to avoid adding dependencies we don't use in a hook


atarashansky (approved):
this is cleaner! thank you!

Issue #3803, opened 2022-12-19
Only include normal ("healthy") cells in calculating marker genes


signechambers1:
Right now, all cells are used to calculate marker genes. We want to only include normal ("healthy") cells in calculating marker genes to better meet user expectations.

signechambers1:
@ashin-czi are you taking on this issue or not? If not, please move back to Disorder. If so, please assign yourself.

PR #3886, opened 2023-01-09
chore(fmg): marker gene calculation only uses healthy cells

atarashansky:
- #3803


backend/wmg/pipeline/summary_cubes/marker_genes.py
```diff
@@ -52,10 +52,12 @@ def create_marker_genes_cube(corpus_path: str):
                 "tissue_ontology_term_ids": [tiss],
                 "cell_type_ontology_term_ids": [ct],
                 "organism_ontology_term_id": organism,
+                "disease_ontology_term_ids": ["PATO:0000461"],
```
atarashansky, line 55: PAT0:0000461 is the ontology term ID for `normal` cells.

backend/wmg/data/query.py
```diff
@@ -169,14 +169,10 @@ def retrieve_top_n_markers(query_result, test, n_markers):
     attrs = [f"p_value_{test}", f"effect_size_{test}"]
     col_names = ["p_value", "effect_size"]
     markers = query_result[["gene_ontology_term_id"] + attrs].rename(columns=dict(zip(attrs, col_names)))
+    markers = markers[markers["effect_size"].astype("str") != "nan"]
```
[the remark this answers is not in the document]
atolopko-czi: Why filter nan as a string as opposed to, say, `notna()`? If there's a good reason, worth capturing as a comment.

backend/wmg/pipeline/summary_cubes/marker_genes.py
atolopko-czi: worth making that clear via an appropriately named constant

atolopko-czi (approved):
Looks reasonable, with a couple nits.

PR #3887, opened 2023-01-09
fix(wmg): update cl-basic.obo link

atarashansky:
[This failed test in wmg-processing-unit-test](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/3874147388/jobs/6605696195) (see screenshot attached) seems to be due to the ontology site 404-ing.
The culprit: http://purl.obolibrary.org/obo/cl-basic.obo

[Relevant slack thread](https://czi-sci.slack.com/archives/C023Q1APASK/p1673273314865159)


nayib-jose-gloria (approved):
ty!

Issue #3845, opened 2023-01-04
Aggregate summary cube values across all IS_A relationships


seve:
Specifically, the summary cube values will be aggregated across all of the IS_A relationships depicted in the cell ontology. Note that we will exclude other kinds of relationships from this aggregation. Consequently, the first step is to ensure that all nodes are reachable via IS_A edges (e.g. the IS_A subgraph contains a Minimum Spanning Tree).

PR #3889, opened 2023-01-09
feat(wmg): aggregate expressions and cell counts across cell type descendants

atarashansky:
- #3845 

## Changes
- added two new attributes to expression_summary (`nnz_rollup` and `sum_rollup`) and one new attribute to cell_counts (`n_cells_rollup`). `*_rollup` now refers to stats that have been aggregated across cell type descendants. This will allow us to toggle back and forth between raw and aggregated counts.
- added a new module (`rollup.py`) which handles the aggregation of metrics across descendants.
- `query` endpoint now accepts optional `is_rollup` flag which is `True` by default. If False, original expression summary counts (not rolled up) are returned.

## Tests
 - Tests have been updated to match new cube schemas.


backend/wmg/pipeline/integrated_corpus/job.py
```diff
@@ -48,7 +48,11 @@ def build_integrated_corpus(dataset_directory: List, corpus_path: str):
     """
     with tiledb.scope_ctx(create_ctx()):
         dataset_count = len(os.listdir(dataset_directory))
-        dataset_gene_mapping = {}
+        if os.path.exists(f"{corpus_path}/{DATASET_TO_GENE_IDS_NAME}.json"):
```
atarashansky, line 51: If we resume integrated corpus build, we should start from the extant json.

backend/wmg/pipeline/summary_cubes/cell_count.py
```diff
@@ -63,6 +65,13 @@ def create_cell_count_cube(corpus_path: str):
     """
     obs = extract(corpus_path)
     df = transform(obs)
+
+    cell_types = list(df["cell_type_ontology_term_id"])
+    n_cells = np.array(list(df["n_cells"]))
+    (n_cells_agg,) = aggregate_across_cell_type_descendants(cell_types, [n_cells])
+    df["n_cells"] = n_cells_agg
```
atarashansky, line 72: aggregate counts are default, raw counts are now `n_cells_raw`

backend/wmg/pipeline/summary_cubes/expression_summary/job.py
```diff
@@ -20,13 +20,25 @@
...
     """
     dims, vals = build_in_mem_cube(
-        gene_ontology_term_ids, cube_index, expression_summary_non_indexed_dims, cube_sum, cube_nnz
+        gene_ontology_term_ids,
+        cube_index,
+        expression_summary_non_indexed_dims,
+        cube_sum,
+        cube_nnz,
+        cube_sum_raw,
+        cube_nnz_raw,
```
atarashansky, line 41: `*_raw` contains the original (before this PR) counts. `cube_sum/nnz` now contains the aggregate counts.

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -83,11 +83,18 @@ def random_expression_summary_values(coords):
...
         "nnz": randint(size=len(coords), low=0, high=100),
         "sum": random(size=len(coords)) * 10,
+        "nnz_raw": randint(size=len(coords), low=0, high=100),
+        "sum_raw": random(size=len(coords)) * 10,
     }
 
 
 def all_ones_expression_summary_values(coords):
-    return {"nnz": np.ones(len(coords)), "sum": np.ones(len(coords))}
+    return {
```
atarashansky, line 92: The cubes generated for the test snapshot do not reflect any aggregation of raw counts because the generators are currently just random sampling or setting values to all ones. Updating this to perform aggregation is difficult because the cell types being used are not real and we don't know the relationships between the fake cell types. 

I think a unit test needs to be added to test the aggregator function (in `aggregate.py`) separately, and that should be enough to ensure that the counts are aggregating properly.

backend/wmg/pipeline/summary_cubes/cell_count.py
seve: Might be worth noting that as a comment somewhere. Raw is a bit of an overloaded term

backend/wmg/pipeline/integrated_corpus/job.py
seve: There's no case where this file exists and we're trying to **regenerate** integrated corpus and not **resume**?

atarashansky: There's no such case, currently. In that case, you would need to delete the current integrated corpus folder and start again. The above update was to match current behavior.

seve (commented):
Forgot to give this a gray approval yesterday, I would wait for AT on this one

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+        List of numpy arrays with the same shape as the input arrays, but
+        with values aggregated across descendants for each cell type (row).
+    """
+    onto = Ontology(CL_BASIC_PERMANENT_URL)
+
+    def descendants_for_node(cl):
+        return list(onto[cl].subclasses().to_set().ids)
+
+    descendants = []
+    for i in cell_types:
```
atolopko-czi, line 32: nit
```suggestion
    for cell_type in cell_types:
```

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+        try:
+            descendants.append(descendants_for_node(i))
+        except KeyError:
```
atolopko-czi, line 35: nit: move exception handling into `descendents_for_node`, then you can assign descendents using a Python comprehension, if that's your style. :)

backend/wmg/pipeline/summary_cubes/cell_count.py
```diff
@@ -63,6 +65,13 @@ def create_cell_count_cube(corpus_path: str):
     """
     obs = extract(corpus_path)
     df = transform(obs)
+
+    cell_types = list(df["cell_type_ontology_term_id"])
+    n_cells = np.array(list(df["n_cells"]))
```
atolopko-czi, line 70: nit
```suggestion
    n_cells = df["n_cells"].to_numpy()
```

backend/wmg/pipeline/summary_cubes/cell_count.py
```diff
@@ -63,6 +65,13 @@ def create_cell_count_cube(corpus_path: str):
...
+    (n_cells_agg,) = aggregate_across_cell_type_descendants(cell_types, [n_cells])
```
atolopko-czi, line 71: nit: since this method is sometimes called with an `arrays_to_sum` of size one, you could make this a var args, `*arrays_to_sum`, avoiding having to wrap the to-be-aggregated lists in an explicit list.

backend/api_server/requirements.txt
```diff
@@ -6,6 +6,7 @@ gunicorn[gevent] >=20.1.0, <21.0.0
 numba # required for where's my gene
 numpy==1.21.2  # required for where's my gene
 pandas==1.4.3 # required for where's my gene
+pronto==2.4.4
```
atolopko-czi, line 9: just checking that does `owlready2` does not provide the descendents extraction functionality we need, in case we can avoid adding this new pkg.

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+    Aggregate values for each cell type across its descendants in the input arrays.
+
+    Parameters
+    ----------
+    cell_types : list
+        List of cell types corresponding to each row of the input arrays
+
+    arrays_to_sum : list
+        List of numpy arrays to aggregate across descendants. Each array
+        should have the same length as cell_types.
```
atolopko-czi, line 18: worth adding an `assert` for this in the method

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+        except KeyError:
+            descendants.append([i])
+    for i, children in enumerate(descendants):
+        descendants[i] = list(set(children).intersection(cell_types))
+    indexer = pd.Series(index=cell_types, data=range(len(cell_types)))
+
+    summed_arrays = []
+    for array in arrays_to_sum:
+        summed = np.zeros_like(array)
+        flag = summed.ndim == 1
```
atolopko-czi, line 44: Confused...when do we have ndim > 1?  It would be nice to eliminate this complexity, so trying to understand when it's needed. Either way, would rename `flag` to `is_multi_dim` (e.g.).  (That inverts the semantics, but no matter...)

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+        return list(onto[cl].subclasses().to_set().ids)
+
+    descendants = []
+    for i in cell_types:
+        try:
+            descendants.append(descendants_for_node(i))
+        except KeyError:
+            descendants.append([i])
+    for i, children in enumerate(descendants):
+        descendants[i] = list(set(children).intersection(cell_types))
```
atolopko-czi, line 38: This is removing any descendent cell types that are not cell types in the WMG data, right? If so, worth a comment.

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+    for array in arrays_to_sum:
+        summed = np.zeros_like(array)
+        flag = summed.ndim == 1
+        if flag:
+            summed = summed[:, None]
+        for i, children in enumerate(descendants):
+            summed[i] += array[indexer[children]].sum(0)
+        if flag:
+            summed = summed.flatten()
+        summed_arrays.append(summed)
```
atolopko-czi, line 51: you might be able to implement all of this in Pandas using something like:
```
cell_types.join(pd.DataFrame(array)).join(descendants).explode().groupby('cell_type').sum()
```
Disclaimer: Consider this pseudo-code. It may be entirely wrong. And it is certainly untested. Also, this is not taking into account the multi-dim consideration, so not sure if this possible.

It's shorter, but is it any clearer? Hard call. 😄

backend/wmg/pipeline/summary_cubes/aggregate.py
```diff
@@ -0,0 +1,52 @@
...
+    indexer = pd.Series(index=cell_types, data=range(len(cell_types)))
+
+    summed_arrays = []
+    for array in arrays_to_sum:
+        summed = np.zeros_like(array)
+        flag = summed.ndim == 1
+        if flag:
+            summed = summed[:, None]
+        for i, children in enumerate(descendants):
+            summed[i] += array[indexer[children]].sum(0)
```
atolopko-czi, line 48:
```suggestion
            summed[i] += array[indexer[children]].sum(axis=0)
```

atolopko-czi (changes requested):
Minor nits and refactoring suggestions. The only comment I feel strongly about is re:use of the "raw" naming, as Seve initially pointed out (suggest "rollup", see comment).

backend/wmg/pipeline/summary_cubes/aggregate.py
atarashansky: Maybe I'm not fluent in `pandas-operations-space` but this is pretty hard to grok for me. I added a couple comments to the `indexer` operation to improve readability

atolopko-czi: No worries, it's definitely more on the relational db/SQL-style of computing.

atolopko-czi (approved):
Latest changes LGTM! 👏

Issue #3805, opened 2022-12-20
Add tool tip with link to documentation


signechambers1:
- [ ] Add tool tip, as designed [here](https://www.figma.com/file/fgDQf6RSNyPeZpZoZRph0H/Alpha-Find-Marker-Genes?node-id=376%3A260378&t=TbCqMTpEuXaO3z8L-0)
- [ ] Make Marker genes header one size bigger

PR #3893, opened 2023-01-09
fix: Documentation Tooltip on Icon Hover Only

ashin-czi:
- #3805

### Reviewers
**Functional:** @seve @atarashansky @tihuan 

**Readability:** @seve @atarashansky @tihuan 

---

## Changes
- add
- remove
- modify
    - "Marker Genes" header no longer a button


ashin-czi:
Waiting on question mark icon SVG from @karenliang0 

@seve Your analytics changes might be affected by these changes

frontend/src/views/WheresMyGene/components/CellInfoSideBar/QuestionMarkIconSvg.tsx
```diff
@@ -0,0 +1,7 @@
+const QuestionMarkIconSvg = (): JSX.Element => (
```
atarashansky, line 1: you should be able to import the svg directly instead of writing raw svg in tsx. checkout the reset button after deleting genes/cell types or the info button next to cell type labels for an example.

atarashansky (changes requested):

frontend/src/views/WheresMyGene/components/CellInfoSideBar/QuestionMarkIconSvg.tsx
atarashansky: Check out what I did [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/42d0b23e4610d2b64c2b9c2caaec74e679236586/frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/style.ts#L103).

If the color change is gray-scale, then we can mess with the brightness prop to do it. For consistency with what is done for the info buttons, I would actually just apply the same style and turn the button black on hover.

ashin-czi: Cool! I just made that change, check it out when you can!

atarashansky (approved):
LGTM! Thanks!

Issue #3563, opened 2022-11-07
Analytics (update): Add analytics to +marker genes button in FMG


atarashansky:
### Overview
**Tool**: The FMG feature in Gene Expression (Find Marker Genes).
- Feature (Optional): The `+marker genes` button.

**Description**: With the new marker gene feature coming, we wish to track the frequency of clicks and which cell types & tissues the buttons were clicked for.

### New events to add 
`WMG_FMG_INFO_CLICKED`: `{combination: "cell type, tissue"}`
- To fire when a user clicks any of the info icons

`WMG_FMG_COPY_GENES_CLICKED`
- To fire when a user clicks the `copy genes` button in the panel 

`WMG_FMG_ADD_GENES_CLICKED`
- To fire when a user clicks the `add to dot plot` button in the panel, make sure the `WMG_HEATMAP_LOADED` event fires as well if this is the first time the dot plot is rendering  

`WMG_FMG_QUESTION_BUTTON_HOVER`
- To fire when a user hovers over the `question mark button` next to `Marker Genes` in the panel and the tool tip pops up for at least 2 seconds 

`WMG_FMG_DOCUMENTATION_CLICKED`
- To fire when a user clicks the `Click to read more about the identification method button` in the tool tip that pops up when the user hovers over the question mark button 

#### People to cc
**PM**: @signechambers1 
**DS**: @ainfeld 

Privacy policy: https://cellxgene.cziscience.com/privacy

PR #3896, opened 2023-01-09
feat(fmg): add analytics

seve:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3563

Adds analytics to functionalities added with find marker genes

Note: The documentation action is most likely going to be moved/changed, make sure to maintain the analytics when that happens


frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
```diff
@@ -49,14 +52,26 @@ function CellInfoSideBar({
...
   }, [data]);
 
   const handleDisplayGenes = useCallback(() => {
     if (!data || !dispatch) return;
     const genes = Object.keys(data.marker_genes);
     dispatch(addSelectedGenes(genes));
+    track(EVENTS.WMG_FMG_ADD_GENES_CLICKED);
   }, [data, dispatch]);
 
+  const [hoverTime, setHoverTime] = useState({ start: -1, end: -1 });
```
tihuan, line 65: discussion: What do you think of using just one value, reportTooltipHoverEpochTime?

```ts
const [reportTooltipHoverEpochTime, setReportTooltipHoverEpochTime] = useState(0);

...

hoverStart={() => setReportTooltipHoverEpochTime(Date.now + 2 * 1000)}
hoverEnd={handleHoverEnd}

function handleHoverEnd() {
  if (Date.now >= reportTooltipHoverEpochTime) {
    track(...)
  }
}
```

seve: No strong feelings, except that without context this looks like some magic number stuff:
`setReportTooltipHoverEpochTime(Date.now + 2 * 1000)}`

tihuan: Yeah great call-out: maybe extract 2 * 1000 as a constant like so?

`const TOOLTIP_INTERACTION_MS = 2 * 1000`

seve: I'm actually going to move it to handleHoverEnd as that's more connotative with the calculation being done

frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
```diff
@@ -62,15 +62,13 @@ function CellInfoSideBar({
...
   }, [data, dispatch]);
 
-  const [hoverTime, setHoverTime] = useState({ start: -1, end: -1 });
+  const [hoverStartTime, setHoverStartTime] = useState(0);
 
-  useEffect(() => {
-    const { start, end } = hoverTime;
-    if (start !== -1 && end !== -1 && end - start > 2 * 1000) {
+  const handleHoverEnd = useCallback(() => {
+    if (Date.now() - hoverStartTime > 2 * 1000) {
```
tihuan, line 68: super nit: cognitively adding is easier to comprehend than subtracting, should we do `Date.now() > hoverStartTime + 2 * 1000` ? Personally it reads better (current time is larger than hoverStartTime plus 2 seconds vs. current time minus hoverStartTime is larger than 2 seconds)

Also should we extract magic number 2 * 1000 as a constant?

seve: I believe the standard for calculating time duration between saved point and now is: ```Date.now() - savedTime```, I can pull this out into a named variable if that helps

seve:
```
Date.now() > hoverStartTime + 2 * 1000
```

Requires much more cognitive load for me than
```
Date.now() - savedDate > 2 * 1000
```

because I instantly pattern match 
```
Date.now() - savedDate
``` 
to `duration`

tihuan: Ahh got it, I see what you're seeing now. Thanks for explaining. No need to extract a variable `duration` thank you!

tihuan (approved):

Issue #3883, opened 2023-01-08
bug(wmg): choosing secondary filters no longer trigger queries


atarashansky:
[This PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3859) broke secondary filters. Selecting a secondary filter after loading lung + MALAT1 no longer triggers a query and updates the heatmap.

PR #3898, opened 2023-01-10
fix(wmg): still check for equality on non-checked filters when checking to clobber queries

seve:
- #3883 

Was naively returning true for all filter attributes that weren't being checked, we still needed to check to see if equality was there.


atarashansky (approved):
Thanks for the fix! LGTM

PR #3907, opened 2023-01-10
feat: unpublished collections can be looked up by canonical id

ebezzi:
### Reviewers
**Functional:** 
@nayib-jose-gloria 

**Readability:** 

---

Allows lookup by canonical `collection_id` for unpublished collections.

### DIFF
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -234,15 +234,22 @@ def _collection_to_response(collection: CollectionVersionWithDatasets, access_ty
 
 def get_collection_details(collection_id: str, token_info: dict):
     """
-    Retrieves the collection information. Will look up for a published collection first,
-    and then looks up for a collection version
+    Retrieves the collection information. Will operate the following lookups, moving to the next one
+    only if the previous returns None.
+    Returns None only if no lookup was successful.
+    1. A published collection that has `collection_id` as the canonical id
+    2. A collection version with `collection_id` as the version_id (published or not)
+    3. An unpublished collection that has `collection_id` as the canonical id. This matches the case
+       where a
     """
     # TODO: this logic might belong to the business layer?
     version = get_business_logic().get_published_collection_version(CollectionId(collection_id))
     if version is None:
         version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
-        if version is None:
-            raise ForbiddenHTTPException()
+    if version is None:
+        version = get_business_logic().get_collection_version_from_canonical(CollectionId(collection_id))
+    if version is None:
+        raise ForbiddenHTTPException()
 
     if version.canonical_collection.tombstoned:
         raise GoneHTTPException()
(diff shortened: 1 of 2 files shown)


backend/layers/api/portal_api.py
```diff
@@ -234,15 +234,22 @@ def _collection_to_response(collection: CollectionVersionWithDatasets, access_ty
...
     """
-    Retrieves the collection information. Will look up for a published collection first,
-    and then looks up for a collection version
+    Retrieves the collection information. Will operate the following lookups, moving to the next one
+    only if the previous returns None.
+    Returns None only if no lookup was successful.
+    1. A published collection that has `collection_id` as the canonical id
+    2. A collection version with `collection_id` as the version_id (published or not)
+    3. An unpublished collection that has `collection_id` as the canonical id. This matches the case
+       where a
```
nayib-jose-gloria, line 243: docstring note here is incomplete

ebezzi: 🤦

backend/layers/api/portal_api.py
```diff
@@ -234,15 +234,23 @@ def _collection_to_response(collection: CollectionVersionWithDatasets, access_ty
...
+    only if the previous returns None.
+    Returns None only if no lookup was successful.
+    1. A published collection that has `collection_id` as the canonical id
+    2. A collection version with `collection_id` as the version_id (published or not)
+    3. An unpublished collection that has `collection_id` as the canonical id. This matches the case
+       where a collection doesn't have any published version yet, but we want to access it
+       via its permalink
     """
     # TODO: this logic might belong to the business layer?
     version = get_business_logic().get_published_collection_version(CollectionId(collection_id))
```
nayib-jose-gloria, line 247: `get_collection_version_from_canonical` already calls `get_published_collection_version` first, then tries to fetch a collection version from canonical if  `get_published_collection_version` is None. Order shouldn't matter here, so we can just switch this call to `get_collection_version_from_canonical` rather than adding a third call

ebezzi: Good catch. I also changed the docstring to reflect the correct lookup order.

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -260,9 +260,55 @@ def test__get_collection__ok(self):
...
+        test_url = furl(path=f"/dp/v1/collections/{version.collection_id}")
+        response = self.app.get(test_url.url, headers=dict(host="localhost"))
+        self.assertEqual(200, response.status_code)
+        body = json.loads(response.data)
+        self.assertEqual(body["visibility"], "PRIVATE")
+
     def test__get_collection_id__403_not_found(self):
-        """Verify the test collection exists and the expected fields exist."""
+        """
+        GET /collections/:collection_id returns 403 if a valid is specified
```
nayib-jose-gloria, line 310: nit: this docstring should say if an INvalid id is specified--though this is a pretty straightforward test anyway, probably doesn't need a docstring to explain it

nayib-jose-gloria (approved):

PR #3944, opened 2023-01-11
feat: newly unpublished collections return canonical collection_id in API

ebezzi:
### Reviewers
**Functional:** 
@nayib-jose-gloria @Bento007 

**Readability:** 

---
With this change, newly unpublished collections will work with canonical ids rather than version ids. Also introduces double lookups for all update operations, since they will need to work with collection_ids as well.

### DIFF
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -1,5 +1,5 @@
 import itertools
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 from urllib.parse import urlparse
 
 from flask import Response, jsonify, make_response
@@ -243,6 +243,15 @@ def _collection_to_response(collection: CollectionVersionWithDatasets, access_ty
         }
     )
 
+def lookup_collection(collection_id: str):
+    """
+    Look up a collection by either its version id or its canonical id
+    """
+    version = get_business_logic().get_collection_version_from_canonical(CollectionId(collection_id))
+    if version is None:
+        version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
+    return version
+
 
 def get_collection_details(collection_id: str, token_info: dict):
     """
@@ -256,9 +265,7 @@ def get_collection_details(collection_id: str, token_info: dict):
     3. A collection version with `collection_id` as the version_id (published or not)
     """
     # TODO: this logic might belong to the business layer?
-    version = get_business_logic().get_collection_version_from_canonical(CollectionId(collection_id))
-    if version is None:
-        version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
+    version = lookup_collection(collection_id)
     if version is None:
         raise ForbiddenHTTPException()
 
@@ -401,10 +408,7 @@ def update_collection(collection_id: str, body: dict, token_info: dict):
     """
 
     # Ensure that the version exists and the user is authorized to update it
-    # TODO: this should be extracted to a method, I think
-    version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
-    if version is None:
-        version = get_business_logic().get_collection_version_from_canonical(CollectionId(collection_id))
+    version = lookup_collection(collection_id)
     if version is None or not UserInfo(token_info).is_user_owner_or_allowed(version.owner):
         raise ForbiddenHTTPException()
 
@@ -435,10 +439,7 @@ def publish_post(collection_id: str, body: object, token_info: dict):
     """
     Publishes a collection
     """
-
-    version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
-    if version is None:
-        version = get_business_logic().get_collection_version_from_canonical(CollectionId(collection_id))
+    version = lookup_collection(collection_id)
     if version is None or not UserInfo(token_info).is_user_owner_or_allowed(version.owner):
         raise ForbiddenHTTPException()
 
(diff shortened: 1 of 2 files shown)


backend/layers/api/portal_api.py
```diff
@@ -195,12 +195,24 @@ def _dataset_to_response(dataset: DatasetVersion, is_tombstoned: bool, is_in_pub
...
+    """
     if collection.canonical_collection.originally_published_at is not None and collection.published_at is None:
+        # In this case, the collection version is a revision of an already published collection.
+        # We should expose version_id as the collection_id
         revision_of = collection.collection_id.id
+        collection_id = collection.version_id.id
+    elif collection.canonical_collection.originally_published_at is None and collection.published_at is None:
+        # In this case, we're dealing with a freshly created collection - we should expose the canonical id here,
+        # since the curators will circulate the permalink immediately. `revision_of` is also null.
+        # Note that this case is effectively equivalent to a published collection
```
nayib-jose-gloria, line 209: nit: we could just put this in an `else` statement and keep the comment for clarity. But both should work

backend/layers/api/portal_api.py
```diff
@@ -423,11 +437,13 @@ def publish_post(collection_id: str, body: object, token_info: dict):
     """
 
     version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
+    if version is None:
```
nayib-jose-gloria, line 440: non-blocking suggestion: I do think if we're going to be doing this double lookup so often, we should probably capture it in a double_lookup function similar to the curation API

nayib-jose-gloria (approved):

backend/layers/api/portal_api.py
ebezzi: It was worth doing, I pushed it. Thanks!

backend/layers/api/portal_api.py
```diff
@@ -195,12 +195,24 @@ def _dataset_to_response(dataset: DatasetVersion, is_tombstoned: bool, is_in_pub
 
 
 def _collection_to_response(collection: CollectionVersionWithDatasets, access_type: str):
-    collection_id = collection.collection_id.id if collection.published_at is not None else collection.version_id.id
-
+    """
+    Converts a CollectionVersion to a format that can be used as an API response. The returned id
```
Bento007, line 199: What does `The returned id` mean?

backend/layers/api/portal_api.py
Bento007: We will need this same logic for the curation API. It might make sense to extract this logic to a shared location.

As for the above comments, I have no strong opinion either way.

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -458,8 +458,10 @@ def test__post_collection_ignores_metadata_if_crossref_exception(self):
         )
         self.assertEqual(201, response.status_code)
         collection_id = json.loads(response.data)["collection_id"]
-        collection = self.business_logic.get_collection_version(CollectionVersionId(collection_id))
-        self.assertIsNone(collection.publisher_metadata)
+        # TODO: this endpoint should also return `version_id`
+        collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
+        if collection is not None:
+            self.assertIsNone(collection.publisher_metadata)
```
Bento007, line 464: When is the if logic tested?

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -504,9 +508,11 @@ def test__post_collection_adds_publisher_metadata(self):
         )
         self.assertEqual(201, response.status_code)
         collection_id = json.loads(response.data)["collection_id"]
-        collection = self.business_logic.get_collection_version(CollectionVersionId(collection_id))
-        self.assertIsNotNone(collection.publisher_metadata)
-        self.assertEqual(collection.publisher_metadata, generate_mock_publisher_metadata())
+        # TODO: this endpoint should also return `version_id`
+        collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
+        if collection is not None:
+            self.assertIsNotNone(collection.publisher_metadata)
```
Bento007, line 514: When is the if logic tested?

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -364,10 +363,11 @@ def test__post_collection_normalizes_doi(self):
...
         self.assertEqual(201, response.status_code)
         collection_id = json.loads(response.data)["collection_id"]
-        collection = self.business_logic.get_collection_version(CollectionVersionId(collection_id))
-        print(collection)
-        doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
-        self.assertEquals(doi, "https://doi.org/10.1016/foo")
+        # TODO: this endpoint should also return `version_id`
+        collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
+        if collection is not None:
+            doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
```
Bento007, line 369: When is the if logic tested?

Bento007 (commented):
The if statements in the tests looks like we are unsure what the behavior of `get_collection_version_from_canonical` will be. Why are the if statements even there?

tests/unit/backend/layers/api/test_portal_api.py
ebezzi: the `if` is required otherwise pylance will complain since `collection` can be an optional`. Unfortunately adding an assertion `.isNotNone` won't help, so you will find some of those ifs in tests.

backend/layers/api/portal_api.py
ebezzi: Sorry that was a leftover. Deleting it

tests/unit/backend/layers/api/test_portal_api.py
Bento007: Are there flags like `#noqa` for pylance to ignore a line. Otherwise, this makes our tests more confusing by adding a forked path in the testing logic that may or may not be hit.

Bento007 (approved):

PR #3957, opened 2023-01-12
chore: cxg links are permalinks for new unpublished collections

ebezzi:
CXG links are using canonical dataset ids for collections that are unpublished but not revisions.

### Reviewers
**Functional:** 
@Bento007 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
+++ b/backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
@@ -51,13 +51,13 @@ def get(collection_id: str, dataset_id: str = None):
     # A canonical url should be only used in two cases:
     # 1. the collection version is unpublished but it's not a revision
     # 2. the collection version is published
-    should_use_canonical_url = (
+    use_canonical_url = (
         collection_version.canonical_collection.originally_published_at is None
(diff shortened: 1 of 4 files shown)


Bento007 (commented):
We need to make the equivalent change in the curation API

Bento007 (changes requested):
Add a test to verify the new behavior.

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -48,7 +48,17 @@ def get(collection_id: str, dataset_id: str = None):
     if version is None:
         raise NotFoundHTTPException("Dataset not found")
 
-    response_body = reshape_dataset_for_curation_api(version, collection_version.published_at is not None)
+    # A canonical url should be only used in two cases:
+    # 1. the collection version is unpublished but it's not a revision
+    # 2. the collection version is published
+    should_use_canonical_url = (
```
Bento007, line 54: should seems wishy washy and extra. Change to:
```suggestion
    use_canonical_url = (
```

Bento007 (changes requested):

Bento007 (approved):

PR #3968, opened 2023-01-13
fix: support null metadata and double lookup for write datasets

ebezzi:
### Reviewers
**Functional:** 
@Bento007 

**Readability:** 

---

1. In Curation API's `GET collection`, support the case where `metadata = None`. This is necessary if the endpoint is queried before the dataset is fully ingested
2. Add support for double lookup in operations that involve updating datasets: PUT and DELETE. Previously this wasn't required as they were only accessed via the version_id. The double lookup here isn't ideal but it will work since the only change is for unpublished, not revised collection, for which only a version exists.

### DIFF
--- a/backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
+++ b/backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
@@ -26,7 +26,6 @@ from backend.layers.common.entities import (
     CollectionVersionId,
     CollectionVersionWithDatasets,
     DatasetVersion,
-    DatasetVersionId,
 )
 from backend.portal.api.curation.v1.curation.collections.common import (
     get_infered_dataset_version,
@@ -63,14 +62,24 @@ def get(collection_id: str, dataset_id: str = None):
 
 
 def _get_collection_and_dataset(
-    business_logic: BusinessLogicInterface, collection_id: CollectionVersionId, dataset_id: DatasetVersionId
+    business_logic: BusinessLogicInterface, collection_id: str, dataset_id: str
 ) -> Tuple[CollectionVersionWithDatasets, DatasetVersion]:
-    dataset_version = business_logic.get_dataset_version(dataset_id)
-    if dataset_version is None:
+    """
+    Get collection and dataset by their ids. Will look up by both version and canonical id for both.
+    """
+
+    collection_version = business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
+    if collection_version is None:
+        collection_version = business_logic.get_collection_version(CollectionVersionId(collection_id))
+    if collection_version is None:
         raise ForbiddenHTTPException()
 
-    collection_version = business_logic.get_collection_version(collection_id)
-    if collection_version is None:
+    # Extract the dataset from the dataset list.
+    try:
+        dataset_version = next(
+            d for d in collection_version.datasets if d.version_id.id == dataset_id or d.dataset_id.id == dataset_id
+        )
+    except Exception:
         raise ForbiddenHTTPException()
 
     return collection_version, dataset_version
@@ -80,9 +89,7 @@ def delete(token_info: dict, collection_id: str, dataset_id: str = None):
     business_logic = get_business_logic()
     user_info = UserInfo(token_info)
 
-    collection_version, dataset_version = _get_collection_and_dataset(
-        business_logic, CollectionVersionId(collection_id), DatasetVersionId(dataset_id)
-    )
+    collection_version, dataset_version = _get_collection_and_dataset(business_logic, collection_id, dataset_id)
 
     if not user_info.is_user_owner_or_allowed(collection_version.owner):
         raise ForbiddenHTTPException("Unauthorized")
@@ -105,9 +112,7 @@ def put(collection_id: str, dataset_id: str, body: dict, token_info: dict):
     url = body.get("url", body.get("link"))
     business_logic = get_business_logic()
 
-    collection_version, _ = _get_collection_and_dataset(
-        business_logic, CollectionVersionId(collection_id), DatasetVersionId(dataset_id)
-    )
+    collection_version, dataset_version = _get_collection_and_dataset(business_logic, collection_id, dataset_id)
 
     if not UserInfo(token_info).is_user_owner_or_allowed(collection_version.owner):
         raise ForbiddenHTTPException()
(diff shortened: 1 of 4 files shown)


tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -961,18 +961,29 @@ def test__delete_dataset(self):
...
-                )
-
-                test_url = (
-                    f"/curation/v1/collections/{dataset.collection_version_id}/datasets/"
-                    f"{dataset.dataset_version_id}"
-                )
-                headers = auth() if callable(auth) else auth
-                response = self.app.delete(test_url, headers=headers)
-                self.assertEqual(expected_status_code, response.status_code)
+                with self.subTest("with version_ids"):
```
Bento007, line 964: nested subtests are not supported by unittest or pytest You'll need to add these details to the out subtest. That should also reduce the code duplication

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1191,17 +1202,31 @@ def test__new_from_link__OK(self, *mocks):
...
-        body = {"link": self.good_link}
-        headers = self.make_owner_header()
-        response = self.app.put(
-            f"/curation/v1/collections/{dataset.collection_version_id}/datasets/{dataset.dataset_version_id}",
-            json=body,
-            headers=headers,
-        )
-        self.assertEqual(202, response.status_code)
+        with self.subTest("with version_ids"):
+            dataset = self.generate_dataset(
```
Bento007, line 1206: can you deduplicated these two test blocks.

Bento007 (changes requested):

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -63,14 +62,24 @@ def get(collection_id: str, dataset_id: str = None):
...
+        collection_version = business_logic.get_collection_version(CollectionVersionId(collection_id))
+    if collection_version is None:
+        raise ForbiddenHTTPException()
+
+    # Extract the dataset from the dataset list.
+    try:
+        dataset_version = next(
+            d for d in collection_version.datasets if d.version_id.id == dataset_id or d.dataset_id.id == dataset_id
+        )
+    except Exception:
```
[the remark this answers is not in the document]
ebezzi: Yes, thanks!

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -63,14 +62,24 @@ def get(collection_id: str, dataset_id: str = None):
 
 
 def _get_collection_and_dataset(
-    business_logic: BusinessLogicInterface, collection_id: CollectionVersionId, dataset_id: DatasetVersionId
+    business_logic: BusinessLogicInterface, collection_id: str, dataset_id: str
```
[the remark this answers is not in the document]
ebezzi: I saw that dataset_id can be None, but why would that be the case? `delete` is the only case where `dataset_id` can be None but what would the endpoint do in that case?

tests/unit/backend/layers/api/test_curation_api.py
ebezzi: It's not easy to add them to the main subtest since the ids cannot be determined before the dataset is generated, and I'd avoid to add an if. I just added a function to remove some of the duplication.

backend/portal/api/curation/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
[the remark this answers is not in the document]
ebezzi: I am fairly confident that the `dataset_id = None` was an oversight. I will remove it.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -959,31 +959,29 @@ def test__delete_dataset(self):
             (None, "none", 401),
             (self.make_not_owner_header, "not_owner", 403),
         ]
+
+        def _delete(collection_id, dataset_id):
+            test_url = f"/curation/v1/collections/{collection_id}/datasets/{dataset_id}"
+            headers = auth() if callable(auth) else auth
+            return self.app.delete(test_url, headers=headers)
+
         for auth, auth_description, expected_status_code in auth_credentials:
```
Bento007, line 968:
```suggestion
    for auth, auth_description, expected_status_code in auth_credentials:
        dataset = self.generate_dataset(
            statuses=[DatasetStatusUpdate(DatasetStatusKey.UPLOAD, DatasetUploadStatus.UPLOADING)],
            publish=False,
        )
        tests = [
            ("with version ids", dataset.collection_id, dataset.dataset_id), 
            ("with canonical ids",dataset.collection_version_id,dataset.dataset_version_id)
        ]
        for description, collection_id, dataset_id in tests:
            with self.subTest(f"{auth_description} {expected_status_code} {description}"):
                # with version ids

                response = _delete(dataset.collection_version_id, dataset.dataset_version_id)
                self.assertEqual(expected_status_code, response.status_code)
```

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1202,31 +1200,27 @@ def test__new_from_link__OK(self, *mocks):
...
-                f"/curation/v1/collections/{dataset.collection_version_id}/datasets/{dataset.dataset_version_id}",
+                f"/curation/v1/collections/{collection_id}/datasets/{dataset_id}",
                 json=body,
                 headers=headers,
             )
             self.assertEqual(202, response.status_code)
 
-        with self.subTest("with collection_ids"):
+        with self.subTest("with version_ids"):
             dataset = self.generate_dataset(
```
Bento007, line 1214: nit: Do you need to generate a new dataset for each test?

Bento007 (changes requested):

tests/unit/backend/layers/api/test_curation_api.py
ebezzi: I don't think that's gonna work because the dataset needs to be generated before each deletion.

tests/unit/backend/layers/api/test_curation_api.py
ebezzi: I think it's better if each test is completely independent, including subtests.

tests/unit/backend/layers/api/test_curation_api.py
Bento007: ah duh. Then I'd make this into two separate tests.

Bento007 (approved):

Bento007 (approved):

PR #3980, opened 2023-01-14
fix: Portal Get /collections

Bento007:
- #TICKET_NUMBER

### Reviewers
**Functional:** @ebezzi 

**Readability:** 

---

## Changes
- We need to return the canonical collection_id for a new unpublished collections when list all the collections. This technically isn't user facing problem but it would mean we no longer have working functional test if we don't fix this way. Since we do not openly disclose the version ID we wont be able to determine if a collection was successfully listed by get collection since creating a collection only return the canonical collection ID.

## QA steps (optional)
Updated the unit test to reflect the ID that should be returned when listing collections.

### DIFF
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -73,18 +73,18 @@ def get_collections_list(from_date: int = None, to_date: int = None, token_info:
         )
 
     collections = []
-    for c in itertools.chain(all_published_collections, all_owned_collections):
+    for version in itertools.chain(all_published_collections, all_owned_collections):
         collection = {
-            "visibility": "PRIVATE" if c.published_at is None else "PUBLIC",
-            "owner": c.owner,
-            "created_at": c.created_at,
+            "visibility": "PRIVATE" if version.published_at is None else "PUBLIC",
+            "owner": version.owner,
+            "created_at": version.created_at,
         }
-        if c.is_unpublished_version():
-            collection["id"] = c.version_id.id
+        if version.is_unpublished_version():
+            collection["id"] = version.version_id.id
         else:
-            collection["id"] = c.collection_id.id
-        if not c.is_published():
-            collection["revision_of"] = c.collection_id.id
+            collection["id"] = version.collection_id.id
+        if not version.is_published():
+            collection["revision_of"] = version.collection_id.id
         collections.append(collection)
 
     result = {"collections": collections}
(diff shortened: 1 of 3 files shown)


backend/layers/api/portal_api.py
```diff
@@ -75,12 +75,15 @@ def get_collections_list(from_date: int = None, to_date: int = None, token_info:
     collections = []
     for c in itertools.chain(all_published_collections, all_owned_collections):
         collection = {
-            "id": c.version_id.id if c.published_at is None else c.collection_id.id,
             "visibility": "PRIVATE" if c.published_at is None else "PUBLIC",
             "owner": c.owner,
             "created_at": c.created_at,
         }
-        if c.published_at is None:
+        if c.published_at is not None or c.canonical_collection.originally_published_at is None:
```
ebezzi, line 82: Not something we need for this PR, but at this point I feel like this condition should be added as a method to `CollectionVersion` (e.g. `.is_new_unpublished()`) since it is ubiquitous

ebezzi (approved):
✅

backend/layers/api/portal_api.py
Bento007: I'll put a TODO since adding it will require more unit tests.

backend/layers/common/entities.py
```diff
@@ -244,6 +244,30 @@ class CollectionVersionBase:
     created_at: datetime
     canonical_collection: CanonicalCollection
 
+    def is_published(self) -> bool:
+        """
+        This collection version has been published.
+        """
+        if self.published_at is not None:
```
ebezzi, line 251: Can probably do: `return self.published_at is not None`

backend/layers/api/portal_api.py
```diff
@@ -79,11 +79,11 @@ def get_collections_list(from_date: int = None, to_date: int = None, token_info:
             "owner": c.owner,
             "created_at": c.created_at,
         }
-        if c.published_at is not None or c.canonical_collection.originally_published_at is None:
-            collection["id"] = c.collection_id.id
-        else:
+        if c.is_unpublished_version():
```
atolopko-czi, line 82: Is `c` is a collection or a collection version? If a Collection, it sounds odd to have a method on it with "version" in the name.    If a CollectionVersion, I would change var naming to indicate "version"

atolopko-czi (approved):
One minor naming confusion on my part, but overall a nice readability improvement.

PR #3986, opened 2023-01-17
fix: handle null metadata in curation API GET collection (take 2)

ebezzi:
### Reviewers
**Functional:** 
@Bento007 

**Readability:** 

---

## Changes
- In the `GET /collections/:collection_id` endpoint, handle additional cases where the endpoint breaks when the metadata is null
- Add a unit test for this case

### DIFF
--- a/backend/portal/api/curation/v1/curation/collections/common.py
+++ b/backend/portal/api/curation/v1/curation/collections/common.py
@@ -187,11 +187,12 @@ def reshape_dataset_for_curation_api(
         ds["revision_of"] = None if is_published else dataset_version.canonical_dataset.dataset_version_id.id
         ds["revision"] = 0  # TODO this should be the number of times this dataset has been revised and published
         ds["title"] = ds.pop("name", None)
-        ds["is_primary_data"] = is_primary_data_mapping.get(ds.pop("is_primary_data"), [])
         ds["explorer_url"] = generate_explorer_url(dataset_version, use_canonical_url)
         ds["tombstone"] = False  # TODO this will always be false. Remove in the future
-        if ds["x_approximate_distribution"]:
-            ds["x_approximate_distribution"] = ds["x_approximate_distribution"].upper()
+        if dataset_version.metadata is not None:
+            ds["is_primary_data"] = is_primary_data_mapping.get(ds.pop("is_primary_data"), [])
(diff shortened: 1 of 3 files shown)


backend/portal/api/curation/v1/curation/collections/common.py
```diff
@@ -187,11 +187,12 @@ def reshape_dataset_for_curation_api(
         ds["revision_of"] = None if is_published else dataset_version.canonical_dataset.dataset_version_id.id
         ds["revision"] = 0  # TODO this should be the number of times this dataset has been revised and published
         ds["title"] = ds.pop("name", None)
-        ds["is_primary_data"] = is_primary_data_mapping.get(ds.pop("is_primary_data"), [])
         ds["explorer_url"] = generate_explorer_url(dataset_version, use_canonical_url)
         ds["tombstone"] = False  # TODO this will always be false. Remove in the future
-        if ds["x_approximate_distribution"]:
-            ds["x_approximate_distribution"] = ds["x_approximate_distribution"].upper()
+        if dataset_version.metadata is not None:
```
Bento007, line 192: This state should never occur for datasets created using the our endpoints.

scripts/setup_dev_data.sh
```diff
@@ -103,7 +103,7 @@ snapshot_identifier='dummy-snapshot'
 wmg_bucket="wmg-test"
 wmg_config_secret_name="corpora/backend/test/wmg_config"
 python3 -m tests.unit.backend.wmg.fixtures.test_snapshot ${tmp_snapshot_dir}
-${local_aws} s3api create-bucket --bucket ${wmg_bucket} &>/dev/null || true
+${local_aws} s3api create-bsucket --bucket ${wmg_bucket} &>/dev/null || true
```
Bento007, line 106: is this a typo?

Bento007 (changes requested):

scripts/setup_dev_data.sh
ebezzi: 🤦

ebezzi:
```suggestion
${local_aws} s3api create-bucket --bucket ${wmg_bucket} &>/dev/null || true
```

Bento007 (approved):

PR #4008, opened 2023-01-19
feat: cronjob to remove old rdev stacks

jakeyheath:
This PR adds a CI GIthub Action that runs on a schedule (every hour) and removes old rdev stacks. An old rdev stack, by default, is a stack that has not been updated in two weeks. This stale period is configurable. If you'd like to update the cronjob period, simply change the schedule at the top of this action. This should help clean up unused stacks over time.

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.github/workflows/clean_happy_rdev.yml
+++ b/.github/workflows/clean_happy_rdev.yml
@@ -1,0 +1,28 @@
+name: Clean up stale happy stacks every hour
+
+on:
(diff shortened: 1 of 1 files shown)


metakuni:
Whoa, this is awesome! Thanks, @jakeyheath ! 🙇

Issue #4007, opened 2023-01-19
Adapt RevisionStatusTag to redesigned data model


nayib-jose-gloria:
[RevisionStatusTag](https://github.com/chanzuckerberg/single-cell-data-portal/blob/41af03dd63690210a245c00c05b38989a1769207/frontend/src/components/Collection/components/CollectionDatasetsGrid/components/Row/DatsetRow/components/RevisionStatusTag/index.tsx#L22-L31
) is set based on the dataset 'original ID' field, which is now deprecated (has no meaning in the new data model, which keeps the original ID of datasets across revisions) and 'published' field, which is currently static.

Update RevisionStatusTag logic accordingly, meeting parity with previous functionality 

Currently, new or updated datasets in revisions will NOT display a RevisionStatusTag (NEW, UPDATED, DELETED).

PR #4009, opened 2023-01-19
fix: Adapt RevisionStatusTag to use redesigned data model

nayib-jose-gloria:

- #4007

### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
-  implement logic to determine 'published' field in portal API response  
-  update dataset portal API response translator method to optionally accept arg for collection revision created timestamp (logic: if dataset is in an unpublished revision and its created time is after the revision created time, it is also unpublished even if the canonical dataset has been published before). 
- add optional 'updated' boolean in portal API dataset response, to share with frontend whether a dataset in a revision is updated from its published counterpart or not 
- update Frontend to use new data model to interpret whether revised datasets need a NEW, UPDATED, or DELETED RevisionStatusTag (this feature was previously depending on fields that were deprecated or unimplemented in redesign)
- refactor explorer URL generate method name + is_in_published arg, since the flag was no longer about whether a dataset was in a published collection (we were setting to True for unpublished, non-revision collections)

## QA steps (optional)
- tests + rdev

### DIFF
--- a/backend/layers/api/portal_api.py
+++ b/backend/layers/api/portal_api.py
@@ -201,7 +201,9 @@ def _dataset_to_response(
             "suspension_type": None if dataset.metadata is None else dataset.metadata.suspension_type,
             "tissue": None if dataset.metadata is None else _ontology_term_ids_to_response(dataset.metadata.tissue),
             "tombstone": is_tombstoned,
-            "updated": True if is_in_revision and published is False else None,
+            "updated": True
+            if is_in_revision and dataset.canonical_dataset.published_at and published is False
+            else None,
             "updated_at": dataset.created_at,
             "x_approximate_distribution": None
             if dataset.metadata is None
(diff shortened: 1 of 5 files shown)


backend/layers/api/portal_api.py
```diff
@@ -147,14 +148,24 @@ def remove_none(body: dict):
...
-def _dataset_to_response(dataset: DatasetVersion, is_tombstoned: bool, is_in_published_collection: bool = False):
+def _dataset_to_response(
+    dataset: DatasetVersion, is_tombstoned: bool, is_in_revision: bool = False, revision_created_at: datetime = None
+):
     dataset_id = dataset.version_id.id
     # Only return `dataset_deployments` if a CXG artifact is available. This is to prevent the "Explore"
     # button to show up while a dataset upload is in progress
     if any(a for a in dataset.artifacts if a.type == DatasetArtifactType.CXG):
-        dataset_deployments = [{"url": explorer_url.generate(dataset, is_in_published_collection)}]
+        use_canonical = False if is_in_revision else True
```
ebezzi, line 158: Technically this could be: `use_canonical = not is_in_revision`, but I am also ok with the explicit form.

ebezzi (approved):

Issue #3996, opened 2023-01-18
consortia values must be sorted during validation of  PATCH /v1/collections/...


brianraymor:
If consortia is set to:

` 'consortia': ['CZ Biohub', 'BRAIN Initiative']`,

in `update_collection(new_collection_id, new_collection_form_metadata)` then these values are not sorted during validation. As a result, the UX incorrectly displays unordered data:

![Screen Shot 2023-01-17 at 4.39.36 PM.png](https://images.zenhubusercontent.com/5b632709929b6b4421540ba4/b6089eeb-e0eb-4ecf-a4e0-ca5b678f9ce5)

If I edit the collection in the UX, make no changes, and save, then the cursor simply spins forever.

PR #4015, opened 2023-01-20
fix: sort consortia on collection create or edit (#3996)

NoopDog:
- #TICKET_NUMBER
#3996 

### Reviewers
**Functional:** 
@ebezzi 

---

## Changes
- sorted collection during creation and edit of consortia
- added tests for portal and curation APIs

### DIFF
--- a/backend/layers/common/entities.py
+++ b/backend/layers/common/entities.py
@@ -223,6 +223,14 @@ class CollectionMetadata:
         if self.consortia is not None:
             self.consortia = [consortium.strip() for consortium in self.consortia]
 
+    def sortConsortia(self):
+        if self.consortia is not None:
+            self.consortia = sorted(self.consortia)
+
+    def sanitize(self):
+        self.strip_fields()
+        self.sortConsortia()
+
 
 @dataclass
 class CanonicalCollection:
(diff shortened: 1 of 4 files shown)


backend/layers/common/entities.py
```diff
@@ -223,6 +223,14 @@ def strip_fields(self):
         if self.consortia is not None:
             self.consortia = [consortium.strip() for consortium in self.consortia]
 
+    def sortConsortia(self):
+        if self.consortia is not None:
+            self.consortia = sorted(self.consortia)
```
ebezzi, line 228: Nit: you can do `self.consortia.sort()` and will sort in place

backend/layers/common/entities.py
```diff
@@ -223,6 +223,14 @@ def strip_fields(self):
         if self.consortia is not None:
             self.consortia = [consortium.strip() for consortium in self.consortia]
 
+    def sortConsortia(self):
```
ebezzi, line 226: Use snake case here: `sort_consortia`

ebezzi (changes requested):

backend/layers/common/entities.py
NoopDog: Updated, thanks.

backend/layers/common/entities.py
NoopDog: Updated!

ebezzi (approved):

PR #4016, opened 2023-01-20
fix: check for Revision Changes accounts for deleted datasets in new data model + refactor Collection checks

nayib-jose-gloria:

Bug: Certain revisions were not rendering in the UI even though backend updates went through and curation API worked for equivalent endpoints

## Changes
- check for revision changes at Collection-level in same way as Datasets (dependent on data field type) to avoid edge case bugs when adding new fields 
- update dataset-level changes check to return True when revision datasets are missing a published dataset (i.e. mark collection as changed if dataset from published is deleted in revision).

## QA steps (optional)
- checked revisions failing to render in dev in local frontend (pointed at dev backend). Renders correctly locally w/ changes

## Notes for Reviewer
- The discovered issue was that, before the redesign, deleted datasets in revisions were simply marked as tombstoned but still existed in the collection dataset map. Revision changes check assumed all datasets from published collection would be in revision dataset map, which was no longer the case with the new data model in which deleted datasets in revisions are removed. Issue caused a null pointer

### DIFF
--- a/frontend/src/common/utils/checkForRevisionChange.ts
+++ b/frontend/src/common/utils/checkForRevisionChange.ts
@@ -10,13 +10,10 @@ const IGNORED_COLLECTION_FIELDS = [
   "updated_at",
   "revisioning_in",
   "revision_of",
-  "id",
   "revision_diff",
   "published_at",
   "datasets",
   "genesets",
-  "links",
-  "consortia",
 ] as Array<keyof Collection>;
 const IGNORED_DATASET_FIELDS = [
   "created_at",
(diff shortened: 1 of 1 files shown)


frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -78,6 +110,7 @@ function checkDatasetsForChanges(
   // Check dataset fields for differences
   return Array.from(publishedDatasets.values()).some((publishedDataset) => {
     const revisedDataset = revisedDatasets.get(publishedDataset.id) as Dataset;
+    if (revisedDataset == undefined) return true;
```
ebezzi, line 113: I am not an expert but I recall that using the triple equal should be the correct behavior here

frontend/src/common/utils/checkForRevisionChange.ts
```diff
@@ -37,6 +34,40 @@ function checkListForChanges(
...
+) {
+  if (publishedCollection[collectionKey] instanceof Array) {
+    // entry is an array
+    if (
+      checkListForChanges(
+        publishedCollection[collectionKey] as Array<unknown>,
+        revisedCollection[collectionKey] as Array<unknown>
+      )
+    ) {
+      console.log(collectionKey);
```
ebezzi, line 50: Can remove these.

ebezzi (approved):

PR #4043, opened 2023-01-23
chore(dag): add feat flag to rollup

atarashansky:


seve (approved):

seve (approved):
I'd drop the query param from the feature flag, since this isn't a feature flag. Just hard code the string in `get()`

frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -617,6 +619,7 @@ function useWMGQueryRequestBody(options = { includeAllFilterOptions: false }) {
         tissue_ontology_term_ids,
       },
       include_filter_dims: true,
+      is_rollup: get("dag") === BOOLEAN.TRUE,
```
seve, line 622:
```suggestion
      is_rollup: get("is_rollup") === BOOLEAN.TRUE,
```

PR #4061, opened 2023-01-25
feat(dag): Revert pipeline changes and instead roll up expressions on the fly per query

atarashansky:
## Reason for Change
- #3845 

Previously, in #3889, I updated the processing pipeline to roll-up the expression counts. However, it's actually not possible to precompute the rolled up expressions while respecting the secondary filters. To accurately display rolled up expressions, we must first know which subset of the corpus we are looking at (where the subset is defined by the secondary filters and the tissue). The existing approach of naively rolling-up the expressions across all the unique groups in the expression_summary cube is not accurate because we will start double counting when the query aggregates according to the secondary filters & tissue. 

As a result, the most straightforward strategy is to calculate the rolled-up expressions on the fly. For the small dataframes that are being returned by the query, the rollup computation time should be negligible.

**In summary:**
 - To accurately calculate rolled up expressions, we first need to know which secondary filters (if any) are selected. 
 - Then, we can compute the rolled up expressions per tissue. 
 - Thus, rolled up expressions cannot be precomputed and we must compute them on the fly. 
 - The roll-up operation should add a negligible amount of time to the query.

## Changes
- Revert #3889 
- Add on-the-fly rollup to the query API.
- Update the rollup function in `rollup.py` to accept a tidy dataframe and infer the dimensions across which the numeric columns will be aggregated.

## Testing
[RDEV STACK](https://atar-dyn-dag-frontend.rdev.single-cell.czi.technology/gene-expression?is_rollup=true)

Quick performance comparison
 - Go to lung
 - Find marker genes for ciliated cell
 - Add to dotplot
 - Query response time:
    - With rollup - 15.5s
    - Without rollup - 15s


backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,130 @@
...
+        List of lists of descendants for each cell type in the input list.
+    """
+
+    global onto
+    if not onto:
+        onto = owlready2.get_ontology(CL_BASIC_PERMANENT_URL_OWL)
+        onto.load()
+
+    # cache finding descendants per cell type
+    @lru_cache(maxsize=None)
```
ebezzi, line 34: Any reason for `maxsize=None`? Since this runs on a webserver, caches should always be bounded.

ebezzi: That said, I understand that the number of values is limited and it's probably not an issue.

backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,130 @@
...
+    def _descendants(cell_type):
+        cell_type_iri = cell_type.replace(":", "_")
+        entity = onto.search_one(iri=f"http://purl.obolibrary.org/obo/{cell_type_iri}")
```
ebezzi, line 37: I think you could save the ontology file locally and do the lookups on a local file, and that might increase performance.

ebezzi (approved):

backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,130 @@
...
+        descendants_per_cell_type.append(descendants)
+
+    for i, children in enumerate(descendants_per_cell_type):
+        # remove descendent cell types that are not cell types in the WMG data
+        descendants_per_cell_type[i] = list(set(children).intersection(cell_types))
+
+    return descendants_per_cell_type
+
+
+def rollup_across_cell_type_descendants(df, cell_type_col="cell_type_ontology_term_id"):
```
atarashansky, line 56: This module was relocated, so the diff is hard to parse. The main update was to rewrite this function to accept a tidy dataframe with `N` non-numeric and `M` numeric columns. The `N` non-numeric columns must contain cell type ontology term IDs, and are treated as the dimensions of an multi-dimensional numpy array. The numeric data is slotted into this array and rolled up along the first axis (which will always correspond to the cell type ontology term IDs). The resulting rolled up array is reshaped back into the tidy dataframe and returned.

This ensures that cell types are only rolled up within each combination of other dimensions (e.g. tissue, gene, organism).

backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,130 @@
...
+    summed = summed[tuple(dim_indices)]
+    dtypes = numeric_df.dtypes
+    for col, array in zip(numeric_df.columns, summed.T):
+        df[col] = array.astype(dtypes[col])
+
+    return df
+
+
+@nb.njit(parallel=True, fastmath=True, nogil=True)
+def _array_summer(array, summed, descendants_indexes, linear_indices):
```
atolopko-czi, line 126: nit: `_sum_arrays`

tests/unit/backend/wmg/test_expression_rollup.py
```diff
@@ -1,49 +1,42 @@
 import unittest
 import numpy as np
-from backend.wmg.pipeline.summary_cubes.rollup import rollup_across_cell_type_descendants
+import pandas as pd
+from backend.wmg.api.rollup import rollup_across_cell_type_descendants
```
atolopko-czi, line 4: either this test file should be moved to tests/unit/backend/wmg/api or the module under test should be moved to tests/unit/backend/wmg

backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,137 @@
+import owlready2
+import pandas as pd
+import numpy as np
+import numba as nb
+from functools import lru_cache
+from backend.wmg.data.constants import CL_BASIC_PERMANENT_URL_OWL
+
+
+# ontology object
+onto = None
```
atolopko-czi, line 10: nit: spell it out (avoid abbreviating to an actual word)

backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Returns
+    -------
+    descendants_per_cell_type : list
+        List of lists of descendants for each cell type in the input list.
+    """
+
+    global onto
+    if not onto:
+        onto = owlready2.get_ontology(CL_BASIC_PERMANENT_URL_OWL)
```
atolopko-czi, line 30: we might want to load this eagerly at server startup to avoid the load time on the first request that triggers this

backend/wmg/api/rollup.py
atolopko-czi: most of this is useful as a comment in the source (from `The non-numeric columns must contain cell type ontology term IDs,...` on)

backend/wmg/api/v1.py
```diff
@@ -194,26 +189,34 @@ def agg_cell_type_counts(cell_counts: DataFrame) -> DataFrame:
...
 
 
-def get_dot_plot_data(query_result: DataFrame, cell_counts: DataFrame) -> Tuple[DataFrame, DataFrame]:
+def get_dot_plot_data(query_result: DataFrame, cell_counts: DataFrame, is_rollup: bool) -> Tuple[DataFrame, DataFrame]:
     # Get the dot plot matrix dataframe and aggregated cell counts per cell type
     cell_counts_cell_type_agg = agg_cell_type_counts(cell_counts)
     cell_counts_tissue_agg = agg_tissue_counts(cell_counts)
     dot_plot_matrix_df = build_dot_plot_matrix(query_result, cell_counts_cell_type_agg, cell_counts_tissue_agg)
+
+    if is_rollup:
```
atolopko-czi, line 209: nit: to keep this method simple, maybe move this conditional block to its own method and then have the caller do:
```
dot_plot_matrix_df, cell_counts_cell_type_agg = get_dot_plot_data(...)
if rollup:
    dot_plot_matrix_df, cell_counts_cell_type_agg = rollup(dot_plot_matrix_df, cell_counts_cell_type_agg)
```
maybe this helps simplify tests as well

backend/wmg/api/v1.py
```diff
@@ -194,26 +189,34 @@ def agg_cell_type_counts(cell_counts: DataFrame) -> DataFrame:
...
+        if dot_plot_matrix_df.shape[0] > 0:
+            dot_plot_matrix_df = rollup_across_cell_type_descendants(dot_plot_matrix_df)
+        # make the cell counts dataframe tidy
```
atolopko-czi, line 212: nit: newline

backend/wmg/api/rollup.py
```diff
@@ -0,0 +1,137 @@
...
+        descendants_per_cell_type.append(descendants)
+
+    for i, children in enumerate(descendants_per_cell_type):
+        # remove descendent cell types that are not cell types in the WMG data
+        descendants_per_cell_type[i] = list(set(children).intersection(cell_types))
+
+    return descendants_per_cell_type
+
+
+def rollup_across_cell_type_descendants(df, cell_type_col="cell_type_ontology_term_id"):
```
atolopko-czi, line 54:
```suggestion
def rollup_across_cell_type_descendants(df, cell_type_col="cell_type_ontology_term_id") -> pd.DataFrame:
```

backend/wmg/api/v1.py
```diff
@@ -194,26 +189,34 @@ def agg_cell_type_counts(cell_counts: DataFrame) -> DataFrame:
...
+        for col in cell_counts_cell_type_agg.index.names:
+            cell_counts_cell_type_agg[col] = cell_counts_cell_type_agg.index.get_level_values(col)
+        cell_counts_cell_type_agg = rollup_across_cell_type_descendants(cell_counts_cell_type_agg)
+        # clean up columns that were added to the dataframe to make it tidy
+        for col in cell_counts_cell_type_agg.index.names:
+            del cell_counts_cell_type_agg[col]
```
atolopko-czi, line 218: is this equivalent?
`cell_counts_cell_type_agg.drop(columns = cell_counts_cell_type_agg.index.names)`

backend/wmg/api/rollup.py
atarashansky: Yeah, I could technically precompute all the descendants for all the cell types and the resulting object wouldn't be that large.

atolopko-czi (commented):
The overall motivation for this and code updates make sense, but I haven't done a deep review, so not explicitly approving.  Added a handful of nits to consider.

On the PR or issue you might want to provide a short before/after performance comparison for the affect endpoint(s), in case API request times are increasing too much.

atarashansky:
> The overall motivation for this and code updates make sense, but I haven't done a deep review, so not explicitly approving. Added a handful of nits to consider.
> 
> On the PR or issue you might want to provide a short before/after performance comparison for the affect endpoint(s), in case API request times are increasing too much.

thanks andrew! agree with all your feedback and made relevant updates. Also added a cursory performance comparison to the PR. Will do a bit more comparisons for different scenarios (e.g. many tissues selected) and report those shortly

atolopko-czi (commented):
If there is a specific part you would like me to focus on for a review, please highlight. I won't otherwise have time to review.

PR #4062, opened 2023-01-25
[WIP] chore: reintroduce dataset revision number

ebezzi:
* This PR is a work in progress *

## Changes
- Reintroduces the revision count for datasets

## Testing steps
- Will link an rdev collection

PR #4065, opened 2023-01-26
[WIP] chore: reintroduce dataset revision number (approach 2)

ebezzi:
Similar to https://github.com/chanzuckerberg/single-cell-data-portal/pull/4062, but uses a different approach (stores the counter in the canonical dataset table)


ebezzi:
Closing as we decided to deprecate the field.

Issue #3846, opened 2023-01-04
Update marker gene algorithm


seve:
Simultaneously, the marker gene algorithm will need to be updated. Cell types can no longer be compared to their descendants or direct ancestors, as the populations being compared must be totally disjoint. #3846

PR #4073, opened 2023-01-27
feat(fmg): ignore comparisons with overlapping cell populations based on ontology

atarashansky:
- #3846 

## Implementation notes
 - Now that we are rolling up expression counts across the descendants of each cell type, the marker gene algorithm must be updated to ensure that we never compare populations with overlapping cells. 
 - `calculate_markers.py` needed to be refactored a bit to extract the expression metrics for the target population directly from the query results containing the context populations (instead of using tiledb to query for the target population directly). Otherwise, we would not be able to calculate the rolled up expressions for the target population.
 - Here, this is accomplished with a new function `are_cell_types_colinear` in `rollup.py`. This function checks if there is any overlap between the ancestors of cell type 1 and the descendants of cell type 2, and vice-versa. If there is no overlap, then the cell types do not have any overlapping cells.
 - `are_cell_types_colinear` is used in `calculate_markers.py` to exclude comparisons that have overlapping cells (setting their effect sizes and p-values to `nan`).
 - **Edge case**: Rollup does not occur if cell types are not specified as a filter in `FmgQueryCriteria` for the target population.
 - **Edge case**: The `max_items` of `cell_type_ontology_term_ids` in `FmgQueryCriteria` is now set to 1. The rollup makes it more complicated to allow multiple cell types in the target population, and seeing as WMG currently only gives marker genes for individual cell types, accounting for this edge case is out of scope.

## Testing
 - Unit tests and test fixtures have been updated.
 - The target population's expected sum in the `test__query_tiledb` test (in `test_calculate_marker_genes.py`) did not change, indicating that the alternative way of getting the target population expression metrics is equivalent to what was done previously.


backend/wmg/api/rollup.py
```diff
@@ -108,7 +143,38 @@ def rollup_across_cell_type_descendants(df, cell_type_col="cell_type_ontology_te
...
+    # extract numeric data and write back into the dataframe
+    summed = summed[tuple(dim_indices)]
+    dtypes = numeric_df.dtypes
+    for col, array in zip(numeric_df.columns, summed.T):
+        df[col] = array.astype(dtypes[col])
+
+    return df
+
+
+def rollup_across_cell_type_descendants_array(array_to_sum, cell_types) -> np.ndarray:
```
atarashansky, line 158: The logic here was factored out into a separate function so that the marker gene calculation code can perform data roll-up on a numpy array directly (instead of needing to reshape it into a tidy dataframe).

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -53,50 +58,20 @@ def wrapped(*args, **kwargs):
...
+
 
-# this decorator function is required to allow lru_cache to
 # memoize the function as dicts, lists, and WmgSnapshot objects
 # are not typically hashable.
 @_make_hashable
 @lru_cache(maxsize=None)
-def _query_tiledb(
-    filters: dict,
+def _query_tiledb_context_memoized(
```
atarashansky, line 68: This refactor makes it so that we only need to do one _memoized_ tiledb query (for the context population). The data for the target population is by definition present in the context data, so we can extract the target data from the context query result.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -145,12 +115,120 @@ def _query_tiledb(
...
     n_cells = cell_counts_query.groupby(gb_dims).sum()["n_total_cells"]
 
-    keep_dataset_ids = group_by_dims is None and "dataset_id" in filters.keys()
-    if genes is None:
-        genes = list(agg.index.levels[0])
+    genes = list(agg.index.levels[0])
+    n_cells_per_gene, n_cells_index = _calculate_true_n_cells(n_cells, genes, dataset_to_gene_ids, keep_dataset_ids)
+
+    # roll up across cell type descendants
+    if "cell_type_ontology_term_id" in agg.index.names:
```
atarashansky, line 122: if `cell_type_ontology_term_id` is a group-by dimension, then it was present in the target population filters. In that case, we want to roll-up data across cell type descendants.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -145,12 +115,120 @@ def _query_tiledb(
...
+        for name in agg.index.names:
+            agg[name] = agg.index.get_level_values(name)
+        agg = rollup_across_cell_type_descendants(agg)
+        # remove tidy
+        agg.drop(columns=agg.index.names, inplace=True)
+
+    return agg, n_cells_per_gene, n_cells_index, genes
+
+
+def _query_tiledb_context(
```
atarashansky, line 133: This function wraps `_query_tiledb_context_memoized`. The reason I did not memoize this function directly is because the context query requires both target_filters and context_filters to know what the group-by dimensions should be. The `cell_type_ontology_term_id` in target_filters will change frequently (whereas the group-by dims will often stay the same) which would break memoization if it were an argument to the memoized function. `_query_tiledb_context_memoized` only needs the context_filters so its memoization will not be broken as often.

In the future, if this code were to be used to calculate marker genes for arbitrary filters, this could probably be further refactored to get better memoization, but that is out of scope for now.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -145,12 +115,120 @@ def _query_tiledb(
...
+    return _query_tiledb_context_memoized(context_filters, group_by_dims, keep_dataset_ids, corpus=corpus)
+
+
+def _query_tiledb_target(
+    target_filters: dict,
+    context_filters: dict,
+    context_agg: pd.DataFrame,
+    n_cells_per_gene_context: np.array,
+    n_cells_index_context: pd.MultiIndex,
+):
```
atarashansky, line 178: This function extracts the data for the target population from the context query results.

ashin-czi (approved):
approved!

PR #4078, opened 2023-01-27
fix(portal redesign): published_at and revised_at for a collection must not be updated when only the collection metadata is updated in a revision

nayib-jose-gloria:

## Reason for Change
- #4012

## Changes
- updated publish collection version logic in business layer to re-use previous version's `published_at` for a revision if no changes to dataset version ID list is detected between the revision to be published and the current published version (i.e. no additions, deletions, or changed dataset version IDs)
- revised_at is an API response level field only, and its value is dependent on the CollectionVersion published_at, so changing this logic will persist the change to revised_at as well.

## Testing steps
- unit tests + checked API response in rdev instance

### DIFF
--- a/tests/unit/backend/layers/business/test_business.py
+++ b/tests/unit/backend/layers/business/test_business.py
@@ -1037,7 +1037,7 @@ class TestCollectionOperations(BaseBusinessLogicTestCase):
 
     def test_publish_version_with_updated_metadata_ok(self):
         """
-        Publishing a collection with updated metadata should succeed
+        Publishing a collection with updated metadata should succeed, but not update published_at
         """
         published_collection = self.initialize_published_collection()
         new_version = self.business_logic.create_collection_version(published_collection.collection_id)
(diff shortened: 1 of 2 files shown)


tests/unit/backend/layers/business/test_business.py
```diff
@@ -1222,6 +1231,7 @@ def test_get_all_collections_published_does_not_retrieve_old_versions(self):
         """
         first_version = self.initialize_published_collection()
         second_version = self.business_logic.create_collection_version(first_version.collection_id)
+        self.business_logic.remove_dataset_version(second_version.version_id, second_version.datasets[0].version_id)
```
Bento007, line 1234: The reason for this line is not obvious. Could you add a comment or change the code to be more explicit.

Bento007 (changes requested):

nayib-jose-gloria:
Tracking here that Emanuele suggested (offline) that we retain the Collection Version `published_at` as always being updated on all publishes, and instead we could introduce 'revised_at' as a db field on the Canonical Collection so we don't lose information about individual Collection Versions. At the API level, these Collection Version `published_at` timestamps are not exposed and only the `originally_published_at` from the canonical collection is passed as the collection 'published_at' timestamp, as expected in the attached issue.

Bento007 (approved):
I'd double check that this is the desired behavior with @brianraymor

PR #4083, opened 2023-01-30
feat: updating date formatting for collections sitemap

sethfeingold:


frontend/src/pages/collections-sitemap.xml/index.tsx
```diff
@@ -1,16 +1,41 @@
 import { GetServerSideProps } from "next";
 import { getServerSideSitemap, ISitemapField } from "next-sitemap";
 
-// function to convert epoch/unix timestamp to YYYY-MM-DD format
+// function to convert epoch/unix timestamp to YYYY-MM-DD HH:MM +00:00 format
 function unixToYYYYMMDD(date: string) {
```
tihuan, line 5: maybe we need to update the function name now that it outputs HH:MM?

Also Github Copilot spits out the following code, not sure if it helps!

```ts
function convertTimestamp(timestamp) {
  const date = new Date(timestamp);
  const year = date.getFullYear();
  const month = date.getMonth() + 1;
  const day = date.getDate();
  const hour = date.getHours();
  const minute = date.getMinutes();
  const second = date.getSeconds();
  const offset = date.getTimezoneOffset();
  const offsetHours = Math.abs(Math.floor(offset / 60));
  const offsetMinutes = Math.abs(offset % 60);
  const timezoneOffset = `${offset < 0 ? "+" : "-"}${offsetHours
    .toString()
    .padStart(2, "0")}:${offsetMinutes.toString().padStart(2, "0")}`;

  return `${year}-${month.toString().padStart(2, "0")}-${day
    .toString()
    .padStart(2, "0")} ${hour.toString().padStart(2, "0")}:${minute
    .toString()
    .padStart(2, "0")}:${second.toString().padStart(2, "0")} ${timezoneOffset}`;
}
```

sethfeingold: Updated the function name to include HHMM!

Tried the Copilot code but every timestamp showed up as being from year 1970 with negative HH:MM value 😨  so I left my logic in there haha

sethfeingold: also merged in latest from main, which I think included a fix (material UI was preventing build from compiling) @tihuan

tihuan: Non blocking:

Awesome thanks for updating the function name!

Interesting 🤔 What's the value you pass to the copilot generated function?

I tried `convertTimestamp(Date.now())` and it seemed to work?

Screenshot 2023-02-01 at 10 10 51 AM

sethfeingold: It gets passed either the revised date (if it exists) and if not it defaults to the published date of the collection.

`const fields: ISitemapField[] = collections.map((collection) => ({
    loc: `https://www.cellxgene.cziscience.com/collections/${collection.id}`,
    lastmod: collection.revised_at
      ? unixToYYYYMMDD(collection.revised_at)
      : unixToYYYYMMDD(collection.published_at),
  }));`

tihuan: I see! Can you share the code you tried that got 1970? So curious what happened 😆 Thanks so much!

sethfeingold: Sure thing! @tihuan pasting below, but essentially pulling in all of the collection data from [this api url](https://api.cellxgene.cziscience.com/dp/v1/collections/index) and using the data to grab revised/published dates, passing those dates into the function

tihuan: ah I see! I'm guessing maybe `const date = new Date(timestamp);` needs to be `new Date(parseFloat(date) * 1000)` like your existing code!

sethfeingold: Oh man, thank you for the extra set of eyes! 🤦🏻 can't believe I pasted over that part 😆 Should be all set now!! 👍🏻

tihuan: Ahaha happens to the best of us! Thanks for verifying the fix 🎉

tihuan (approved):
LGTM!! Thanks so much again for updating this, Seth ❤️🙏

Issue #4074, opened 2023-01-27
Analytics (update): Additional link to documentation from Marker Score tool tip 


signechambers1:
### Overview
**Tool**: Gene Expression
- Feature (Optional): Find Marker Genes

**Description**: We are adding a tool tip to Marker Score that explains the score and links out to documentation. We'd like to be able to track people who click on this link and land on documentation.

### Updates needed
`WMG_FMG_QUESTION_BUTTON_HOVER` : `{label: <fill in which question mark the user hovered over>, either "marker genes" or "marker score"}`
- Current: Currently it only fires when a user hovers over the question mark next to `marker genes`
- Requested change: Fire when a user hovers over the question mark next to `marker genes` or `marker score` and update the plausible event to include the new payload mentioned above

`WMG_FMG_DOCUMENTATION_CLICKED` : `{label: <fill in which aspect user clicked to documentation for>, either "marker genes" or "marker score"}`
- Current: Currently it only fires when a user hovers over the question mark next to `marker genes` and clicks the `Click to read more about the identification method button` in the tool tip that pops up 
- Requested change: Fire when a user clicks either the `Click to read more about the identification method button` or the `Learn more about marker score button` in the tool tip that pops up when a user hovers over the question mark next to `marker genes` or `marker score` and update the plausible event to include the new payload mentioned above

#### People to cc
**PM**: @signechambers1 
**DS**: @ainfeld 

Privacy policy: https://cellxgene.cziscience.com/privacy

PR #4090, opened 2023-01-31
chore(fmg): add tooltip to marker score and update analytics for info tooltips

atarashansky:
## Reason for Change
- #3853 
- #4074 
- Update analytics on marker score and marker genes tooltips.
## Changes
- added new payload to the tooltip hover and documentation clicked events indicating which tooltip was used.

## Testing steps
- Tested with a local instance pointing to dev.

### DIFF
--- a/frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
@@ -58,11 +58,17 @@ function CellInfoSideBar({
 
   const [hoverStartTime, setHoverStartTime] = useState(0);
 
-  const handleHoverEnd = useCallback(() => {
+  const handleFmgHoverEnd = useCallback(() => {
     if (Date.now() - hoverStartTime > 2 * 1000) {
-      track(EVENTS.WMG_FMG_QUESTION_BUTTON_HOVER);
+      track(EVENTS.WMG_FMG_QUESTION_BUTTON_HOVER,{label: "marker genes"});
     }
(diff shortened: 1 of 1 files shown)


seve (changes requested):
I have a feeling we're going to be doing checking like this pretty often. Let's start the transition to a util function with something like:
```
useCallback((event, payload = {}) => {
    if (Date.now() - hoverStartTime > 2 * 1000) {
      track(event);
    }
  }, [hoverStartTime]);
  ```

**Review by ashin-czi — APPROVED** [review]
LGTM! tested locally

File: frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx [hunk 1]
```diff
@@ -22,6 +22,13 @@ export interface CellInfoBarProps {
   cellInfoCellType: Exclude<State["cellInfoCellType"], null>;
   tissueName: string;
 }
+function useHandleHoverEnd(event: EVENTS, hoverStartTime: number, payload = {}) {
+  return useCallback(() => {
+    if (Date.now() - hoverStartTime > 2 * 1000) {
+      track(event, payload);
+    }
+  }, [hoverStartTime]);
+}
```
seve, line 31:
```suggestion 
useHandleHoverEnd = useCallBack((event: EVENTS, hoverStartTime: number, payload = {}) => {
    if (Date.now() - hoverStartTime > 2 * 1000) {
      track(event, payload);
    }
  }, [hoverStartTime])
```

You might need to format that ^

seve (approved):

PR #4099, opened 2023-02-01
fix(fmg): marker gene calculation ignores cell types that have no healthy cells

atarashansky:
The healthy cell filter broke the marker gene calculation because there are some cell types for which we have no healthy cells.

This PR rectifies this issue by moving the healthy cell filter directly inside the marker gene calculation module and raising a custom error if the query result dataframe is empty.

## Testing
Successfully completed a full dry run of marker gene cube generation on the prod snapshot in an ec2 instance.


backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -101,6 +103,17 @@ def _query_tiledb_context_memoized(
     query = q.expression_summary_fmg(criteria)
     cell_counts_query = q.cell_counts(criteria)
 
+    healthy_filter_es = query["disease_ontology_term_id"].astype("str") == NORMAL_CELL_DISEASE_ONTOLOGY_TERM_ID
```
atarashansky, line 106: The healthy cell filters were moved here. Previously, we were filtering out unhealthy cells from the expression summary fmg cube. That broke parity between the expressions used for calculating marker genes and the cell counts, since cell counts cube was not filtered as it is being shared between WMG and FMG. 

Now, we are explicitly filtering both cell counts and expression summary down to just the healthy cells in the marker gene calculation code.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -101,6 +103,17 @@ def _query_tiledb_context_memoized(
...
+    query = query[healthy_filter_es]
+
+    healthy_filter_cc = (
+        cell_counts_query["disease_ontology_term_id"].astype("str") == NORMAL_CELL_DISEASE_ONTOLOGY_TERM_ID
+    )
+    cell_counts_query = cell_counts_query[healthy_filter_cc]
+
+    if query.shape[0] == 0 or cell_counts_query.shape[0] == 0:
```
atarashansky, line 114: If either of the filtered dataframes (expression summary or cell counts) are empty, raise an error and move on to the next cell type.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -233,6 +246,9 @@ def _query_target(
         level_values = np.array(list(n_cells_index_context.get_level_values(level)))
         filt_ncells = np.logical_and(filt_ncells, np.in1d(level_values, target_filters[level + "s"]))
 
+    if not np.any(filt) or not np.any(filt_ncells):
```
atarashansky, line 249: An extra safety check - if for whatever reason the target population is not found in the context, raise an error and move on to the next cell type.

backend/wmg/pipeline/summary_cubes/calculate_markers.py
```diff
@@ -513,6 +529,8 @@ def _post_process_stats(
         )
         effects[is_colinear] = np.nan
         pvals[is_colinear] = np.nan
+        pvals[:, np.all(np.isnan(effects), axis=0)] = 1
+        effects[:, np.all(np.isnan(effects), axis=0)] = 0
```
atarashansky, line 533: For all-nan columns, just set effect size and p-values to 0 to avoid numpy warnings.

backend/wmg/pipeline/summary_cubes/marker_genes.py
```diff
@@ -19,10 +20,17 @@ def extract_tissue_celltype_organism(corpus_path: str) -> pd.DataFrame:
     with tiledb.open(f"{corpus_path}/{CELL_COUNTS_CUBE_NAME}") as array:
         yield (
             array.query(
-                attrs=["cell_type_ontology_term_id"], dims=["organism_ontology_term_id", "tissue_ontology_term_id"]
+                attrs=["cell_type_ontology_term_id"],
```
atarashansky, line 23: linting changes

backend/wmg/pipeline/summary_cubes/marker_genes.py
```diff
@@ -57,10 +65,17 @@ def create_marker_genes_cube(corpus_path: str):
...
-                target, context, corpus=corpus_path, test="binomtest", percentile=0.3, n_markers=None
-            )
+            try:
+                t_markers = get_markers(
+                    target, context, corpus=corpus_path, test="ttest", percentile=0.05, n_markers=None
+                )
+                b_markers = get_markers(
+                    target, context, corpus=corpus_path, test="binomtest", percentile=0.3, n_markers=None
+                )
+            except MarkerGeneCalculationException as e:
```
atarashansky, line 75: try/except so pipeline doesn't fail if no cells match query criteria

dsadgat: this explanation would go well as a comment in the code

backend/wmg/pipeline/summary_cubes/calculate_markers.py
dsadgat: Context would be nice in a method comment in the code

ashin-czi (approved):

PR #4102, opened 2023-02-01
fix(fmg): update analytics on tooltips to fire properly

atarashansky:
## Reason for Change
- Analytics were broken after CR in #4090 (oops!)
## Changes
- wrap useCallback in a function so it properly behaves as an event
- replace onMouseEnter/Leave logic with onOpen/Close of the tooltip itself. That way mouse doesn't need to hover over the question mark button specifically.

## Testing steps
- Tested locally and console.log'd the events firing with the appropriate payloads.

### DIFF
--- a/frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
@@ -57,17 +57,23 @@ function CellInfoSideBar({
   }, [data, dispatch]);
 
   const [hoverStartTime, setHoverStartTime] = useState(0);
-  
-  const useHandleHoverEnd = (event: EVENTS, payload = {}) => {
-      return useCallback(() => {
-        if (Date.now() - hoverStartTime > 2 * 1000) {
-          track(event, payload);
-        }
-      }, [hoverStartTime])
-  }
 
-  const handleFmgHoverEnd = useHandleHoverEnd(EVENTS.WMG_FMG_QUESTION_BUTTON_HOVER, {label: "marker genes"});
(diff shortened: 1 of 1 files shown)


frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
```diff
@@ -152,6 +155,8 @@ function CellInfoSideBar({
                 width="default"
                 className="fmg-tooltip-icon"
                 arrow={true}
+                onOpen={() => setHoverStartTime(Date.now())}
```
ashin-czi, line 158: good catch

ashin-czi (approved):

frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
```diff
@@ -58,14 +58,22 @@ function CellInfoSideBar({
...
-      track(event, payload);
-    }
-  }, [hoverStartTime])
+  const useHandleHoverEnd = (event: EVENTS, payload = {}) => {
+    return useCallback(() => {
+      if (Date.now() - hoverStartTime > 2 * 1000) {
+        track(event, payload);
+      }
+    }, [hoverStartTime]);
+  };
```
seve, line 67: Does event and payload need to be added to the dependency list for `useCallback`?

atarashansky: no i dont think so. when the function is called, it returns a fresh callback each time, so we never have a situation where the event and payload change but the callback doesn't.

seve: Wait, what's the point of useCallback here? We're circumventing the memoization that is serves.

PR #4107, opened 2023-02-02
fix(wmg): Select filter dispatch options too more than needed

tihuan:
## Reason for Change
1. Fix bug where select options in the left panel trigger empty updates unnecessarily

## Changes
- Add code to detect if options haven't changed or has just been initialized, don't dispatch changes

## Testing steps
1. Add `console.log('-----dispatch options', options) in this [line](https://github.com/chanzuckerberg/single-cell-data-portal/blob/3823eacdb0db18cfcaff63860cfdd34122cf9fd1/frontend/src/views/WheresMyGene/components/Filters/index.tsx#L159):

```ts
// DEBUG
console.log('-----dispatch options', options)  // <--- should not trigger more dispatches than needed

dispatch(
          selectFilters(
            key,
            options.map((option) => (option as unknown as { id: string }).id)
          )
        );
```


tihuan:
Code cov error is from backend!

ashin-czi (approved):
LGTM! Tested locally

PR #4108, opened 2023-02-02
fix: add additional pre-commit checks for python code

Bento007:
## Reason for Change
To encourage clean python code with every commit. I notice these same checks were being done in [cell-census](https://github.com/chanzuckerberg/cell-census/blob/main/tools/scripts/formatting.sh) and thought we could benefit from those changes.

## Changes
- Replace flake8 with [ruff](https://pypi.org/project/ruff/)
- Add autoflake to remove unused python variables and imports
- Add isort to sort python imports
- Add flake8-bugbear for bug checking
- Add flake8-comprehensions for code simplification checks
- Add flake8-simplify for code simplification checks
- Apply suggested changes so linter passes

## Testing steps
- "Unit tested only" No functional changes have been made.


atolopko-czi (changes requested):
Some questions for discussion

atolopko-czi (approved):
Thanks for the code quality improvements!

Issue #2622, opened 2022-06-01
Filter by suspension type


LuckyMD:
See [suspension type](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/3.0.0/schema.md#suspension_type) in the 3.0.0 schema. 

1. The suspension type filter MUST be available as both a collection and dataset filter.
2. The plausible event MUST be added.

@ainfeld proposed:

```

event name: FILTER_SELECT_SUSPENSION_TYPE
payload: {suspension_type: <fill in value>}

```

---

Original request:

### Need
It would be incredibly useful to be able to sort also by cells vs nuclei in the data portal. I've been looking for lung data to use as example query data to map onto the HLCA, and the only human, 10X data from the lung that only consists of 1 dataset is from lungMap, and that's all single-nucleus data, which might not be the best test case.

Just to add a second feature that would be useful (but harder to do I'm sure), would be the ability to load only the lung data from datasets with multiple tissues.

### Approach

**Describe the solution you'd like**
An additional filter option for cell vs nucleus

**Describe alternatives you've considered**
The alternative would be go to through all dataset descriptions and then come back to download the data that fits the bill.

### Definition of Done
Some automatic filter for sc vs sn to immediately gauge relevant datasets.

---

jahilton:
First step to a filter is a required metadata field...
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-curation/227

LuckyMD:
I'm slowly learning how zenhub works ^^.

PR #4111, opened 2023-02-02
fix: removed na from suspension type filter. #2622

MillenniumFalconMechanic:
## Reason for Change
- #2622
- Remove "na" option from suspension type filter as per [Slack](https://clevercanary.slack.com/archives/C02CF7MQPGV/p1675267437217269).

## Changes
- Removed "na" option from suspension type filter values.

## Testing steps
- For both collections and datasets, verify there is no "na" option in the suspension type filter.


tihuan (approved):
LGTM 🫡👌

PR #4118, opened 2023-02-02
fix: set UPLOAD_STATUS to NA when creating empty dataset 

nayib-jose-gloria:
#4120

## Reason for Change
-  Frontend [assumes UPLOAD_STATUS = 'WAITING' means the dataset is in the Queue](https://github.com/chanzuckerberg/single-cell-data-portal/blob/c30e37a444f6189a03642c90256574ee789f8fd1/frontend/src/components/Collection/components/CollectionDatasetsGrid/components/Row/DatsetRow/index.tsx#L175), and uses that [to set 'Queued...' tag for UI.](https://github.com/chanzuckerberg/single-cell-data-portal/blob/41af03dd63690210a245c00c05b38989a1769207/frontend/src/components/Collection/components/CollectionDatasetsGrid/components/Row/DatsetRow/components/UploadStatus/index.tsx#L23-L26) However, currently, we set UPLOAD_STATUS = 'WAITING' both when we queue datasets and when we create an empty dataset without a file upload. This erroneously sets the 'Queued...' tag on the frontend for datasets that are empty and not queued. 

## Changes
- when creating an empty dataset, set UPLOAD_STATUS = 'NA' rather than 'WAITING' so that 'WAITING' is only set if a datafile is being ingested and passed into the upload queue

## Testing steps
- unit tests

## Notes for Reviewer
- looking at the code, it doesn't seem that there are any downstream dependencies on the UPLOAD_STATUS being set to 'WAITING' for an empty dataset. I think it makes sense to define 'WAITING' as waiting in the queue and 'NA' as no upload yet, but let me know about any edge cases or if this doesn't make intuitive sense.
- should not affect API response processing_status field for curation API, which doesn't depend on upload_status


Bento007 (approved):
Magnificent!

PR #4126, opened 2023-02-03
chore: enable happy shell into the backend container

atarashansky:
Currently we are unable to use `happy shell` to get inside the backend containers. This would be very helpful for debugging certain issues.


jakeyheath (approved):
LGTM!

alexlokshin-czi (approved):

PR #4158, opened 2023-02-08
chore: update html sitemap margin and font size on mobile

sethfeingold:


tihuan (approved):
Quick question, is `sitemap.tsx` the same as `html-sitemap.tsx`? If so, maybe worth removing `html-sitemap.tsx`? Thank you!

Otherwise LGTM!

sethfeingold:
> 

good catch! Thank you! I had deleted that/renamed it locally - thought it would've updated the remote branch but apparently kept both files around. Now deleted! 😄

tihuan (approved):

Issue #4159, opened 2023-02-08
curator_name MUST be set on collection creation in the Discover API


brianraymor:
It's unclear whether this is a regression. 

The curator_name MUST be set on collection creation to enable Lattice to understand which curators are working on collections that have not been published. Discover UX is currently working on a sort by curator_name feature too. 

Tests:

1. Create a collection using the UX

`curator_name: "Brian Raymor"`

---

1. Create a collection using the API in Dev

`'curator_name': '',`

2. Add a dataset 
3. Publish

`'curator_name': '',`

---

PR #4164, opened 2023-02-09
feat: set curator_name when creating Collection via Discover API

danieljhegeman:
## Reason for Change

- #4159 

## Changes

- add logic to pull `curator_name` from access token
    - leverage new `curator_name` claim (`name` field from user object in Auth0); associated infra PR to follow 
    - use flask `g` request context to hold access token

## Testing steps
- verified in rdev
- unit test updated to test for `curator_name` as well as metadata fields used when creating the Collection

## Notes for Reviewer
- might be having trouble with a mock added for the unit tests where the test is successful when invoked via `python -m unittest ...` but not via `make local-unit-test-backend`

### DIFF
--- a/backend/curation/api/v1/curation/collections/actions.py
+++ b/backend/curation/api/v1/curation/collections/actions.py
@@ -1,5 +1,6 @@
 from flask import jsonify, make_response, request, g
 
+from backend.common.authorizer import get_payload_from_request_context
 from backend.common.corpora_orm import CollectionVisibility, ProjectLinkType
 from backend.common.utils.http_exceptions import ForbiddenHTTPException, InvalidParametersHTTPException
 from backend.curation.api.v1.curation.collections.common import reshape_for_curation_api
@@ -66,8 +67,8 @@ def post(body: dict, user: str):
         body.get("consortia", []),
     )
 
-    print(f"\nDJH {g.access_token_payload}\n")
-    curator_name = g.access_token_payload["curator_name"]
+    # Get 'curator_name' from access token
+    curator_name = get_payload_from_request_context()["curator_name"]
 
     try:
         version = get_business_logic().create_collection(user, curator_name, metadata)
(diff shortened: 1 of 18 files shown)


Bento007 (commented):
was curator_name added to the auth0 access token?

danieljhegeman:
> was curator_name added to the auth0 access token?

@Bento007 [associated PR](https://github.com/chanzuckerberg/single-cell-infra/pull/714)

danieljhegeman:
[Apply at same time](https://github.com/chanzuckerberg/single-cell-infra/pull/714)

backend/curation/api/v1/curation/collections/actions.py
```diff
@@ -66,8 +67,11 @@ def post(body: dict, user: str):
         body.get("consortia", []),
     )
 
+    # Get 'curator_name' from access token
+    curator_name = get_payload_from_request_context()["curator_name"]
+
     try:
-        version = get_business_logic().create_collection(user, "", metadata)
+        version = get_business_logic().create_collection(user, curator_name, metadata)
```
Bento007, line 74: this can be simplified by adding `token_info: dict` as a parameter to the post. [Connexion magic](https://connexion.readthedocs.io/en/latest/security.html) will pass the decoded token_info into the function. Then you wont have to store the token_info in `flask.g`

Bento007 (changes requested):
Connexion can help you simplify getting `token_info`. This will remove the need to use `flask.g`, `get_payload_from_request_context`, and `get_payload_from_request_context`. You will still need to add `curator_name` to `_mock_assert_authorized_token`

backend/curation/api/v1/curation/collections/actions.py
danieljhegeman: @Bento007 this is amazing. And it makes tons of sense. Everything I had found just mentioned strict passing of the str var through. Thanks!! Will make the change.

Bento007 (approved):

PR #4165, opened 2023-02-09
perf: Persistence Layer method performance upgrades

nayib-jose-gloria:
## Reason for Change

- #4151 

## Changes

- Refactored _get_datasets logic to reduce number of calls to DB when fetching multiple datasets and associated metadata (connection overhead introduces significant performance delays)
- Refactored some methods for resuability 
- Use map in implementation of fetching all collections instead of iterating over list to find particular collections

## Testing steps

- local benchmarking of endpoint performance before and after + unit tests for no functional regressions

## Notes for Reviewer
- local benchmarking showed significant improvement when fetching all datasets (for 1000 published datasets + 1000 unpublished, performance was about 10 faster--15s avg vs 1.4 s average of the business layer dataset index method)
- this method relies on keeping large portions of our dataset metadata corpus in-memory. For our current size and rate of growth, this seems sustainable, but is there any cell-census or similar work on the horizon that could change that? If so, are we ok with this approach for now and to re-evaluate later?


ebezzi (approved):
Looks great, thanks!

ebezzi:
> this method relies on keeping large portions of our dataset metadata corpus in-memory. 

I don't think that's an issue. If the corpus growth we will need to introduce pagination, but that needs a different infrastructure that will allow to run boolean queries on the corpus.

PR #4166, opened 2023-02-09
chore: increase health check interval to 25

ebezzi:
Required to fix the current update failures on the FE container.


tihuan (approved):
Thanks so much again, Emanuele!! ⭐ ⭐ ⭐

Issue #4176, opened 2023-02-10
bug(filter): select filters are using a categoryValueKey field rather than a payload field


MillenniumFalconMechanic:
## Describe the bug

Select filters are sending analytic events in the format `{categoryValueKey: ${value}}` rather than `{payload: ${value}}`.

## To Reproduce

1. Go to datasets
2. Click on the author filter
3. Select an author from the filter drop down
4. The event payload is `{categoryValueKey: ${authorName}}` 

## Expected behavior

The event payload in step four above should be `{payload: ${authorName}}` 

## Environment

* All environments and devices.

PR #4178, opened 2023-02-10
fix: corrected filter payload field. #4176.

MillenniumFalconMechanic:
## Reason for Change

- #4176

## Changes

- Corrected payload key for select filters to use the text "payload" if a payload key is not specified.
- Added tests to check generating the analytics event payload with and without a payload key.

## Testing steps

- "Existing" filters such as assay and author should send a payload in the format `{payload: ${value}}`.
- New filters such as suspension type should send a payload in the format `{suspension_type: ${value}}`.


tihuan (approved):
Ah sorry for assuming that analytics was okay with retiring the `payload` key 😞 I should have at least double checked. Thanks for adding the test and explicit comments!

Issue #4161, opened 2023-02-08
Marker genes do not update when you change the tissue for the same cell type


signechambers1:
Steps to reproduce:

1. Prod
2. Tissue = lung
3. Click i icon for macrophage, minimize panel
4. Change tissue = spleen
5. Select i icon for macrophage
6. Genes & marker score are exact same for spleen and lung

- [ ] Write a test

@atarashansky confirmed this is a UI bug, genes and scores are different on the back end.

Thanks to @GarabetY for finding this bug.

PR #4190, opened 2023-02-13
fix(fmg): 4161 Different tissues with the same cellType should displa…

tihuan:
…y different values

## Reason for Change

- #4161 

We need to add `tissueId` to the cache key, so React Query will fetch the response accordingly and not use the wrongly cached value

## Testing steps

1. Select lung + B cell
2. Observe marker genes
3. Select Speen + B cell
4. The marker genes should be different from the values at step 2

demo

### DIFF
--- a/frontend/src/common/queries/wheresMyGene.ts
+++ b/frontend/src/common/queries/wheresMyGene.ts
@@ -826,13 +826,18 @@ export function useMarkerGenes({
   }, [data]);
 
   return useQuery(
-    [USE_MARKER_GENES, cellTypeID, test],
+    /**
+     * (thuang): Add all arguments to `fetchMarkerGenes()` as dependencies,
+     * so React Query can cache responses correctly without running into
+     * issues like #4161
+     */
+    [USE_MARKER_GENES, cellTypeID, organismID, test, tissueID],
     async () => {
       const output = await fetchMarkerGenes({
         cellTypeID,
         organismID,
-        tissueID,
         test,
+        tissueID,
       });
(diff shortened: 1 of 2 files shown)


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -826,13 +826,18 @@ export function useMarkerGenes({
   }, [data]);
 
   return useQuery(
-    [USE_MARKER_GENES, cellTypeID, test],
+    /**
+     * (thuang): Add all arguments to `fetchMarkerGenes()` as dependencies,
+     * so React Query can cache responses correctly without running into
+     * issues like #4161
+     */
+    [USE_MARKER_GENES, cellTypeID, organismID, test, tissueID],
```
tihuan, line 834: @seve I added organismId to the cache key too, which I think makes sense to be part of the cache key. But please help verify that makes sense 🙏 Thanks!

frontend/src/views/WheresMyGene/components/CellInfoSideBar/index.tsx
```diff
@@ -35,9 +35,9 @@ function CellInfoSideBar({
   }
   const { isLoading, data } = useMarkerGenes({
     cellTypeID: cellInfoCellType.cellType.id,
-    tissueID: cellInfoCellType.tissueID,
     organismID: cellInfoCellType.organismID,
     test: testType,
+    tissueID: cellInfoCellType.tissueID,
```
tihuan, line 40: object key reorder

seve (approved):

PR #4193, opened 2023-02-14
fix: meta endpoint correctly looks up replaced datasets

ebezzi:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4192

## Changes

- Retrieve the correct dataset using the source of truth from the collection mappings.

## Testing steps

- Added a unit test specifically for this case.
- Since this cannot be easily tested in rdev, my recommendation is to do a prod->dev mirroring and use [this collection](https://cellxgene.cziscience.com/collections/f5af7a2f-ab4c-4728-829e-48efb9562105) - which prompted the issue - to verify the correct behavior.


backend/portal/api/portal_api.py
```diff
@@ -731,6 +731,13 @@ def get_dataset_identifiers(url: str):
     if collection is None:  # orphaned datasets - shouldn't happen, but we should return 404 just in case
         raise NotFoundHTTPException()
 
+    # If we look up by canonical_id (i.e. the if does not hold), we can run into issues if the collection
```
nayib-jose-gloria, line 734: this edge case can also occur anywhere else we try to look up a pre-publish dataset with its canonical ID. We should either: 1) (longer-term solution) update persistence layer to not map the canonical dataset to a version until publish, and return None (or most recent version) for pre-publish dataset canonical ID calls. 
or 2) (shorter-term solution) add this quick fix somewhere in the flow of `get_dataset_version_from_canonical`

nayib-jose-gloria: will follow up in https://github.com/chanzuckerberg/single-cell-data-portal/pull/4229

nayib-jose-gloria (approved):

PR #4195, opened 2023-02-14
chore: pin scanpy and matplotlib dependencies

ebezzi:
Required due to https://github.com/scverse/scanpy/issues/2411

Also pins scapy to 1.6.0 (the version currently used) since this is likely a library that will introduce potential breaking updates.

## Testing steps

- See if unit tests pass
- Do an upload in dev

### DIFF
--- a/backend/layers/processing/requirements.txt
+++ b/backend/layers/processing/requirements.txt
@@ -1,5 +1,6 @@
 cellxgene-schema==3.0.0
 dataclasses-json
+matplotlib==3.6.3 # 3.7.0 isn't compatible with scanpy: https://github.com/scverse/scanpy/issues/2411
 numba==0.56.2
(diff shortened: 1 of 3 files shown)


backend/layers/processing/requirements.txt
```diff
@@ -8,5 +9,6 @@ psycopg2-binary>=2.8.5
 pyarrow>=1.0
 pydantic>=1.9.0
 python-json-logger
+scanpy==1.6.0
```
nayib-jose-gloria, line 12: nit: for some reason this is pinned to 1.6.0 already, but its pinned in the processing dockerfile directly as a pip3 install scanpy==1.6.0 step. We should just keep one (probably the one here)

nayib-jose-gloria (approved):

Issue #3830, opened 2022-12-22
Fix failing dev deployment


atolopko-czi:
See https://czi-sci.slack.com/archives/C023Q1APASK/p1671664365669329

Fix:

Timmy: We should build the app and generate all the artifacts before pushing, so the cluster nodes don't need to spend time building static pages and randomly fails deployment due to timeout 

NOTE: We need to make sure the build step has access to `COMMIT_SHA` aka `HAPPY_COMMIT`, so the build id is consistent across all FE nodes

Tutorial:

https://dev.to/thakkaryash94/build-nextjs-application-using-github-workflow-and-docker-3foj

NOTES:
- [ ] Please reduce FE container resource from 8G to 2G in terraform, since we likely don't need more RAM than that for just running the app

tihuan:
Jess comment:

https://czi-sci.slack.com/archives/C024P2G2KU1/p1668552020425519

tihuan:
Jess is helping us with this now — the new approach is to build one image that can be used for dev/staging/prod and do text replacement per env, e.g., API_URL

tihuan:
Jess is helping us out with this one! Her branch is:

`jgadling-test-runtime-vars`

Currently waiting on her feedback of local FE container not working for me

ebezzi:
I tried the branch and the FE container seems to work fine, both locally and in rdev. Please re-test.

tihuan:
New PR!

tihuan:
For completeness, Emanuele was right that the branch did work, I just had to destroy all my container images locally and bump up RAM to 8GB for it to work on my machine 😆 👍

ebezzi:
This should now be working. It will be in prod on Feb 24th.

PR #4203, opened 2023-02-15
chore: build FE container at build time

ebezzi:
#3830

## Changes

- Builds the FE container at build time and not at runtime (like currently)
- Lower the memory requirement to 4G (minimum for the FARGATE configuration we have)
- Revert the healthcheck interval to 15s

## Testing steps
- https://ebezzi-fe-build-frontend.rdev.single-cell.czi.technology/
- Test the local application (it should work with `make local-sync`)
- Check that it works in dev and the container boots in a few seconds (through Cloudwatch)
- Check that the sitemaps are working (they're affected by this change)


seve:
Can we leave the `configs/*.js` files intact for development purposes? I'd prefer to not have to manually set up my `config.js` when running the dev server

frontend/Dockerfile
```diff
@@ -15,22 +15,21 @@ RUN apt-get update && apt-get install -y make wget \
...
 ENV PATH /opt/node_app/node_modules/.bin:$PATH
 
 # -- TODO, we should try turning this back on later.
 # ADD --chown=node . /corpora-frontend
 ADD . /corpora-frontend
 WORKDIR /corpora-frontend
+ADD ./src/configs/build.js src/configs/configs.js
+RUN mkdir -p node_modules
+RUN ln -sf /opt/node_app/node_modules/* /opt/node_app/node_modules/.bin ./node_modules/.
+RUN npm run build
```
ebezzi, line 32: @jgadling we had to replace `next build` with this so that the `postbuild` step could run. Any concerns?

jgadling: Nope, I just didn't know what the best way to build the static app was!

frontend/src/configs/local.js
```diff
@@ -1,11 +1,16 @@
-const { PLAUSIBLE_DATA_DOMAIN_STAGING } = require("./common");
-
 // (thuang): For local development, please copy the content of this file
 // to a new file named `configs.js` in this directory.
 
 const configs = {
+  // Dev
```
tihuan, line 5: @seve thanks for pointing this out!

I just added different endpoints in `local.js`, so we can still copy this to `configs.js` locally and run `npm run dev` without using `make local-sync`. So this should address our `npm run dev` use case now?

Thanks!

seve: Very much a nit, but the process of:
```
cp src/configs/dev.js src/configs/configs.js
npm run dev
```
is **way** faster than
```
cp src/configs/local.js src/configs/configs.js
code src/configs/configs.js
# uncomment one line and comment another line
npm run dev
```

tihuan: yeah I actually just changed the default in `local.js` to be pointing to the dev API

tihuan:
Also a huge shout out to @jgadling again for coming up with the solution AND the code to make this happen 🙏 ❤️ We're able to decrease Fargate memory from 8GB to 4GB because of this change too, so it's literally saving CZI money 💰 

And of course @ebezzi for pairing with me today to ironed out the sitemap gen issue I didn't catch earlier + changing our Infra config 🙏

Very grateful to be working with you both!

frontend/src/configs/local.js
seve: oh missed that part, sounds good!

seve (approved):

tihuan (approved):
LGTMMM 🎶 🎶 🎶

tihuan:
`backend-wmg-unit-test` and `wmg-processing-unit-test` failures are being taken care of and irrelevant to this PR. So I'll merge it now!

PR #4220, opened 2023-02-16
chore: restore js env files

ebezzi:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/3830

## Changes

- Reintroduce the configuration js files. This is necessary in order for the e2e tests to work. This introduces some config duplication, but I think it's acceptable. As an alternative, the file would have to be written at line 144 in deploy-happy-stack.yml based on the .env files, but that will require a new script since the tests are run outside Docker.


tihuan (approved):
LGTM!! Thanks so much, Emanuele 🙏 ❤️

Issue #4192, opened 2023-02-14
The meta endpoint should look up replaced datasets in unpublished collections correctly


ebezzi:
The meta endpoint is unable to perform a look up correctly in the following scenario:
1. A collection is not yet published
2. A dataset is replaced

PR #4229, opened 2023-02-17
fix: look up replaced datasets in unpublished collections correctly

nayib-jose-gloria:
## Reason for Change

- #4192 
- Datasets were mapping a canonical version on create; should only be done on publish
- Caused downstream issues in edge case: when fetching dataset version metadata by canonical dataset ID for a dataset that has never been published, it always returns the first dataset version it was mapped to even if its been since updated in the revision.
      - bug is resolved once dataset is published at least once
- most notably, above edge case occurs when checking explorer page for datasets in a private collection that have been added, then updated without a publish (see ticket for example)

## Changes

- update persistence layer to NOT map version ID to a canonical Dataset on dataset creation 
- update `get_dataset_version_from_canonical` business layer method to account for case where canonical ID is passed to retrieve a dataset version, but none is published yet--fetch latest created DatasetVersion

## Testing steps
-unit tests


backend/curation/api/v1/curation/collections/common.py
```diff
@@ -182,9 +183,11 @@ def reshape_dataset_for_curation_api(
         else:
             ds["revised_at"] = None
         ds["revision_of"] = (
-            None
-            if dataset_version.canonical_dataset.dataset_version_id == dataset_version.version_id
-            else dataset_version.dataset_id.id
+            dataset_version.dataset_id.id
+            if is_in_revision
```
ebezzi, line 187: nit: I would recommend using a non-inlined if here.

backend/layers/business/business.py
```diff
@@ -552,9 +552,23 @@ def get_dataset_version(self, dataset_version_id: DatasetVersionId) -> Optional[
...
         """
-        return self.database_provider.get_dataset_mapped_version(dataset_id)
+        dataset_version = self.database_provider.get_dataset_mapped_version(dataset_id)
+        if dataset_version is not None:
+            return dataset_version
+        else:
+            canonical_dataset = self.database_provider.get_canonical_dataset(dataset_id)
+            if not canonical_dataset:
+                return None
+            latest = datetime.datetime.fromtimestamp(0)
```
ebezzi, line 565: I would add a comment that explains why you do this (i.e. you want to lookup the latest version in case there are multiple versions in an unpublished collection).

Also, you could avoid the loop by doing `unpublished_dataset = max(dataset_versions, key=lambda d: d.created_at`, but I don't know if the functional style is preferred to the iterative style in our codebase.

ebezzi (approved):

backend/layers/business/business.py
nayib-jose-gloria: I'm going to stick to iterative for now since we do it in other parts of the codebase (i.e.`get_unpublished_collection_version_from_canonical`) but I don't feel strongly about it

will add the comment though

PR #4231, opened 2023-02-17
test(wmg): Fix heatmap await

tihuan:
Update test to wait for `canvas` element instead of SVG, since we've switched back to use canvas now!

Fixes failing tests in dev


danieljhegeman (approved):
can't believe there isn't an issue for this PR 😉

PR #4235, opened 2023-02-20
chore: updating freethink video section

sethfeingold:
simple update, freethink video section changed from 1-col to 2-col layout


tihuan (approved):
LGTM!

Issue #4204, opened 2023-02-15
feat(wmg): improve backend performance of WMG query endpoint


atarashansky:
The backend is currently a bit slow for WMG queries. In this story, the assignee should profile the backend, explore performance improvements, and implement them. This will likely involve TileDB optimizations.

PR #4239, opened 2023-02-21
feat(wmg): performance improvements

atarashansky:
## Reason for Change

- #4204 

## Changes

- add a new expression summary cube (`expression_summary_default`) that has been greatly simplified. the primary dimensions are gene, tissue, organism and the only secondary dimension is cell type. This cube is used for queries that do not specify any secondary filters.
- use the cell counts cube instead of the expression summary cube to determine the valid secondary filters given a set of primary filters.
- add timing logs to the response header of the query endpoint
- to help out functional performance tests, load the ontologies right off the bat at the global level instead of on the first function call
- set tile ordering to hilbert and capacity to 100,000 after empirical benchmarking

## Testing steps

- Tested and benchmarked in rdev.
- added a unit test to check that the default expression summary cube returns correct query results.
- added a unit test to ensure that both expression summary and expression summary default cubes have the same data in them (the latter is an aggregated version of the former).


backend/wmg/api/v1.py
```diff
@@ -28,31 +30,40 @@ def query():
 
     criteria = WmgQueryCriteria(**request["filter"])
 
-    snapshot: WmgSnapshot = load_snapshot()
-    q = WmgQuery(snapshot)
+    with ServerTiming.time("query and build response"):
+        snapshot: WmgSnapshot = load_snapshot()
+        q = WmgQuery(snapshot)
+        default = snapshot.expression_summary_default_cube is not None
```
atarashansky, line 36: add comment explaining why you need this if statement

backend/wmg/api/v1.py
```diff
@@ -93,12 +104,12 @@ def find_dim_option_values(criteria: Dict, snapshot: WmgSnapshot, dimension: str
     filter_options_criteria = criteria.copy(update={dimension + "s": []}, deep=True)
     # todo can we query cell_counts for a performance gain?
```
atarashansky, line 105: remove this comment as it is now addressed

ashin-czi (approved):

PR #4257, opened 2023-02-24
fix(wmg): Filters selected options not showing active state

tihuan:
## Reason for Change

- #4211 @ashin-czi please help add the ticket number thank you!

1. The filters are not showing the selected item as active state, because every time we change an option, React destroys all the children of `<TooltipWrapper />` and re-create new instances of them. This results in calculating memoized values for the first time, thus breaking active select state

## Changes

1. The root cause is because we're re-generating `TooltipWrapper` component in each render, thus all its children are considered first time showing up in the DOM tree, thus triggering useMemo() every time `TooltipWrapper`'s parent re-renders. The solution is simply removing the `TooltipWrapper` component and use the hidden API `disableHoverListener` and `disableFocusListener` to decide whether to show tooltip or not instead of relying on Fragment.

I also tried just passing `null` as `title` to the Tooltip component, which does suppress tooltip, but it still re-instantiates all children, which we don't want. So the hidden API is the best solution!

## Testing steps

1. Pull the PR and start up locally
3. Check all filters, the selected option should now show the active state for all filters

demo

### DIFF
--- a/frontend/src/views/WheresMyGene/components/Filters/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/Filters/index.tsx
@@ -282,11 +282,6 @@ export default memo(function Filters({ isLoading }: Props): JSX.Element {
 
         <Organism isLoading={isLoading} />
 
-        {/* DEBUG */}
-        {/* DEBUG */}
-        {/* DEBUG */}
-        <Foo />
-
         <Sort areFiltersDisabled={areFiltersDisabled} />
       </Wrapper>
     </Tooltip>
@@ -308,15 +303,4 @@ function sortOptions(a: DefaultMenuSelectOption, b: DefaultMenuSelectOption) {
     return 1;
   }
   return 0;
-}
-
-// DEBUG
-// DEBUG
-// DEBUG
-function Foo() {
-  useMemo(() => {
-    console.log("----------Foo calculating...");
-  }, []);
-
-  return null;
 }
--- a/frontend/src/views/WheresMyGene/components/Filters/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/Filters/index.tsx
@@ -8,7 +8,6 @@ import {
 import isEqual from "lodash/isEqual";
 import {
   memo,
-  ReactElement,
   useCallback,
   useContext,
   useEffect,
--- a/frontend/src/views/WheresMyGene/components/Filters/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/Filters/index.tsx
@@ -219,7 +219,14 @@ export default memo(function Filters({ isLoading }: Props): JSX.Element {
   );
 
   return (
-    <TooltipWrapper areFiltersDisabled={areFiltersDisabled}>
+    <Tooltip
+      title={
+        "Please select an organism, tissue and at least one gene to use these filters."
+      }
+      // (thuang): We need to disable the tooltip when filters are enabled
+      disableHoverListener={!areFiltersDisabled}
+      disableFocusListener={!areFiltersDisabled}
+    >
       <Wrapper>
         <div>
           <StyledComplexFilter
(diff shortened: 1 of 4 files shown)


frontend/src/views/WheresMyGene/components/Filters/index.tsx
```diff
@@ -44,6 +44,27 @@ const DropdownMenuProps = {
   getOptionSelected,
 };
 
+const ANALYTIC_MAPPING: {
```
tihuan, line 47: moved this constant out to the top, so we don't re-generate the object at each render

frontend/src/views/WheresMyGene/components/Filters/index.tsx
```diff
@@ -277,20 +284,8 @@ export default memo(function Filters({ isLoading }: Props): JSX.Element {
...
-  function TooltipWrapper({ children }: { children: ReactElement }) {
-    if (areFiltersDisabled) {
-      return (
-        <Tooltip title="Please select an organism, tissue and at least one gene to use these filters.">
-          {children}
-        </Tooltip>
-      );
-    }
-
-    return <>{children}</>;
```
tihuan, line 292: switching between `<></>` and `<Tooltip></Tooltip>` causes the all subtree to unmount and remount to the DOM, thus reinstantiating all the react components, including all the useMemo calls, since all the components are now considered being mounted in the DOM for the first time

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/Organism/index.tsx
```diff
@@ -77,6 +77,7 @@ export default function Organism({ isLoading }: Props): JSX.Element {
         onChange={handleOnChange as tempOnChange}
         InputDropdownProps={{ ...InputDropdownProps, disabled: isLoading }}
         data-test-id="add-organism"
+        value={organism}
```
tihuan, line 80: need to add `value` prop to show active state on the default value

frontend/src/views/WheresMyGene/components/InfoPanel/components/ColorScale/index.tsx
```diff
@@ -76,6 +80,7 @@ export default function ColorScale({ setIsScaled }: Props): JSX.Element {
         label={colorScaledOptionLabel}
         options={COLOR_SCALE_OPTIONS}
         InputDropdownProps={DEFAULT_INPUT_DROPDOWN_PROPS}
+        value={selectedSortBy}
```
tihuan, line 83: same here

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/Organism/index.tsx
tihuan: Great question! Those two filters are from `QuickSelect/index.tsx`, which under the hood uses SDS `MenuSelect` and we are already passing `value` prop to `MenuSelect`!

https://github.com/chanzuckerberg/single-cell-data-portal/blob/73b444ac8f37a85f4c41e4e6b49351580ce82e93/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx#L298

tihuan:
@ashin-czi since I created the PR I think I need your approval to merge 😆 Thank you!

ashin-czi (approved):
LGTM! simple and elegant fix!

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/Organism/index.tsx
tihuan: Oh wait, I misread your message 😆 You meant why CellTypes and Genes sort by options don't need `value`, not the QuickSelect for Genes and Tissues

I looked into this more. I think essentially hard coded options don't need `value` because the object reference stay the same, whereas `organisms` we filter them with `useMemo()`, so we need to pass `value` prop OR add the following as prop `DropdownMenuProps`:

```ts
const DropdownMenuProps = {
  isOptionEqualToValue: (
    option: DefaultMenuSelectOption,
    value: DefaultMenuSelectOption
  ) => option.name === value.name,
};
```

I decided to go with `DropdownMenuProps` since I think it's more intuitive to use!

tihuan: Actually I just found out that without `value`, it won't highlight the default option when you open the dropdown for the first time, so I changed back to using `value`

I tried added `value` to sortBy for genes/cell types and somehow it was causing an infinite loop, so will need to look into that later!

tihuan: just added `value` to sortBy genes/cell types too to make default active work too. Turned out it was just my Compare branch that was buggy 😆

PR #4261, opened 2023-02-24
chore: remove legacy db alembic migrations + remove legacy db functions from rdev start-up script [DO NOT MERGE]

nayib-jose-gloria:
## Reason for Change

- Starting up an rdev stack fails due to old alembic migrations referring to corpora_orm (legacy db) entities.
- general clean up (could avoid error above by just commenting out offending line, if we don't want to refactor)

## Changes

- removed old alembic migration scripts related to legacy db, effectively re-starting alembic chain with redesign schema migrations only
- cleaned up legacy db dependent functions in populate_db script, which is called by rdev. The removed functions are not called in our Makefile commands

## Testing steps
-rdev

## Notes for Reviewer
- Need to 'stamp' DBs in each env with the renumbered alembic migration script, before deploy
- will stamp before deploying, if approved


Bento007 (approved):
nice to have a clean repo again

PR #4271, opened 2023-02-27
chore: update high-level metrics of landing page

pablo-gar:
Updates the top level metrics in the landing page.

### DIFF
--- a/frontend/src/views/Landing/index.tsx
+++ b/frontend/src/views/Landing/index.tsx
@@ -51,9 +51,9 @@ const LandingPage = (): JSX.Element => {
   });
   const scrollSection5 = useRef<HTMLDivElement>(null!);
 
-  const [cellsHeroNum] = useState("33M+");
-  const [datasetsHeroNum] = useState("500+");
-  const [cellTypesHeroNum] = useState("550+");
+  const [cellsHeroNum] = useState("38M+");
+  const [datasetsHeroNum] = useState("700+");
+  const [cellTypesHeroNum] = useState("650+");
 
   const publications = [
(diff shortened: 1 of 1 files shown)


frontend/src/views/Landing/index.tsx
```diff
@@ -51,9 +51,9 @@ const LandingPage = (): JSX.Element => {
   });
   const scrollSection5 = useRef<HTMLDivElement>(null!);
 
-  const [cellsHeroNum] = useState("33M+");
-  const [datasetsHeroNum] = useState("500+");
-  const [cellTypesHeroNum] = useState("550+");
+  const [cellsHeroNum] = useState("38M+");
+  const [datasetsHeroNum] = useState("700+");
+  const [cellTypesHeroNum] = useState("650+");
```
brianraymor, line 57: Doublechecked that cells are the primary cell count and not the total cell count. Check.

Since you will be updating on a regular basis, is there a reason to round up/down to the nearest 50 instead of reporting the precise number? For example:

```
Number of primary cells: 38.1M
Number of datasets: 723
Number of unique cell types: 637
```

For example, cell types is not 650+. It could be 600+ or 625+ or just 637.

brianraymor (changes requested):

frontend/src/views/Landing/index.tsx
pablo-gar: I kept it consistent with the previous mode. I assumed it was a UX feature, but I have no problem providing more precise numbers.

brianraymor (approved):

Issue #2785, opened 2022-06-28
feat(scExpression): secondary filters should enable preemptive filtering of the data (BE ONLY)


atarashansky:
## Need
I ran into an issue where I wanted to use scExpression for a particular dataset, but I didn't know what tissues belonged to that data ahead of time. I had to select all the tissues, filter for the data, and then remove tissues with no data displayed (not knowing if they weren't displaying data because the tissues don't exist in the filtered dataset or simply because the genes were not expressed in them).

## Proposal
I'd like the filtering to be preemptive. 

1) The filter buttons should open a searchable dropdown menu of all available metadata in the corpus for each dimension/attribute.
2) Selecting a filtering option should remove all tissues that are not included in the selected filter

seve:
@atarashansky To update ticket based off of previous work

seve:
Alec marked as possibility to parallelize with validation work

seve:
Needs priority revisit

PR #4272, opened 2023-02-27
feat(wmg): filter relationships are precomputed and can be calculated without the selection of primary filters

atarashansky:
## Reason for Change

- #2785 
- This PR improves the speed of calculating the valid secondary filters given a particular set of query criteria. The new approach no longer requires the selection of primary filters first to determine the valid secondary filters. As a result, users will be able to select secondary filters _first_ to narrow down the set of available tissues (this will be implemented in a separate PR). 

## Changes

- Update the pipeline to generate a new data artifact containing the relationships between filter values found in the cell counts cube. This graph-like structure contains all the necessary information to calculate valid filter options given a set of query criteria.
- Update `find_dim_option_values` to traverse the filter relationships graph and find the intersection of valid filters given all the query criteria. This replaces the prior logic where we repeatedly queried the cell counts cube when excluding each criteria key.
- Update `build_filter_dims_values` to return valid tissue ontology term IDs. This will be used by the frontend to limit the tissues shown in the tissue dropdown (in a later PR).

## Testing steps

- Tested in `rdev`.
- Unit tests were added to test the valid secondary filter calculation logic. Specifically, these unit tests ensure equivalency between the _old_ and _new_ way of calculating these secondary filters.
- Due to the new way of calculating valid secondary filters given a set of query criteria, some tests needed to be temporarily commented out (specifically the secondary filter tests in `test_query.py`). It is not possible to calculate the filter relationships graph given the tight memory constraints in GHA due to the unnecessarily large size of the dynamically generated virtual snapshots. Because we are still unit testing the valid secondary filter calculation using the aforementioned new tests, it is safe to comment these failing tests out for now. Eventually, when we update the virtual test snapshot to be more wieldy, these tests can be added back.


tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -824,70 +827,7 @@ def test__query_request_with_filter_dims__returns_valid_filter_dims(
                 self.assertEqual(expected_development_stage_terms, dev_stage_terms_eth_2_dev_2)
                 self.assertEqual(expected_self_reported_ethnicity_term, eth_stage_terms_eth_2_dev_2)
                 self.assertEqual(dev_stage_terms_eth_2_dev_2, dev_stage_terms_eth_2_no_dev_filter)
-                self.assertNotEqual(eth_stage_terms_eth_2_dev_2, self_reported_ethnicity_terms_eth_2_no_dev_filter)
-
-            with self.subTest("Additional queries are not performed when the secondary dimensions are not set"), patch(
```
atarashansky, line 829: this test no longer relevant so it's removed.

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -194,6 +205,9 @@ def create_temp_wmg_snapshot(
         with tiledb.open(expression_summary_cube_dir, ctx=create_ctx()) as expression_summary_cube, tiledb.open(
             cell_counts_cube_dir, ctx=create_ctx()
         ) as cell_counts_cube:
+            # cc = cell_counts_cube.df[:] # this gives OOM error in GHA
+            # filter_relationships = create_filter_relationships_graph(cc)
```
ashin-czi, line 209: Should we keep these comments?

ashin-czi (approved):

PR #4276, opened 2023-02-27
chore: Remove redundant 2nd run of Push Tests upon merge to main

atarashansky:
I don't think we need to rerun tests upon adding to merge queue. I erroneously added this when adding merge queues as I thought it was required, but it's doubling the time it takes to merge into main because tests must run twice.

EDIT (from @nayib-jose-gloria): Merge Queue needs to run Push Tests because of our branch protection rules; we need to keep the branch protection rules in case a merge ahead in the queue causes a conflict or test failure in a subsequently queued PR. This new approach instead gets rid of the 2nd Push Test upon merge to main, which is redundant. However, we need to keep the image building and deployment triggering steps--so I moved those into a separate workflow file that only runs on merge to main/staging/prod (these steps were already filtered out from running on PR/merge queue Push Tests anyway, so this is actually a clearer division of what steps actually run in each scenario)


seve (approved):

.github/workflows/push-tests.yml
```diff
@@ -9,9 +9,6 @@ on:
   pull_request:
     branches:
       - "*"
-  merge_group:
```
nayib-jose-gloria, line 12: this file contains the step that triggers a deployment run (see `create_deployment` step), which in turn is dependent on `push-prod-images` to run, which itself is dependent on the push tests passing. So this actually created a stall in the merge queue (when it was at the head of the queue, it blocked dev deployments queued after because it wasn't triggering push tests and GHA never reached the create_deployment step). 

I think its doable to remove the double push tests, but it would require per-step conditions (i.e. only run test steps in branch X, auto-pass them in branch Y to unblock create_deployment) or refactoring to move create_deployment to a different file

nayib-jose-gloria (changes requested):
see comment, this was blocking the merge queue when it was at the head

.github/workflows/push-tests.yml
atarashansky: Do you know how to do this, @nayib-jose-gloria ?

nayib-jose-gloria: I can take a look tomorrow!

tihuan:
test test @hthomas-czi

tihuan:
hello again @hthomas-czi

tihuan:
test @hthomas-czi all activities enabled

seve (commented):
Can we rename the PR/squash commit message, otherwise looks good to me

nayib-jose-gloria (approved):

PR #4283, opened 2023-03-01
fix(wmg): Fix SVG Download

ashin-czi:
## Reason for Change

Fixes erroneous label names from displaying on SVG download

## Changes

- add
- remove
- modify
    - Query selectors for retrieving label text for svg generation

## Testing steps

1. Fill tissue and genes
2. Select Download
3. Select SVG
4. Press Download


tihuan (approved):
Ah great catch! Thank you!

Issue #4288, opened 2023-03-02
bug(wmg): secondary filters are not cross-filtering into each other


atarashansky:
To reproduce:
1. https://cellxgene.cziscience.com/gene-expression
2. select lung tissue
3. select MALAT1 gene
4. select pulmonary emphysema disease (secondary filter)
5. observe that datasets were not crossfiltered.
6. select dataset secondary filter (this specific dataset, screenshot before)
<img width="716" alt="image" src="https://user-images.githubusercontent.com/16548075/222553039-fdc6e221-eb50-4a8e-a550-9cc9f82e490f.png">
7. observe that there is no data and thus this combination of filters is invalid.

ashin-czi:
Just some initial observations from the frontend...
<img width="751" alt="image" src="https://user-images.githubusercontent.com/109984998/222558702-662a92ec-0859-4b66-88cf-5fb096906fe8.png">
<img width="514" alt="image" src="https://user-images.githubusercontent.com/109984998/222558915-df2334e1-d91f-47cc-904d-9c9475099d02.png">

metakuni:
Added the `P0` label based on @niknak33 's request to hot fix.

atarashansky:
@ashin-czi The response looks good. The FE is not updating from the response when the query is triggered by secondary filter interface.

PR #4290, opened 2023-03-02
fix(wmg): fix secondary filters not cross-filtering

atarashansky:
- #4288
This line (includeAllFilterOptions: true) was getting run at the end of every secondary filter trigger which overwrote the previous response with the crossfiltered options. I'm not sure why this broke _now_, but this does seem to fix the issue without any apparent bugs.

Thinking more about it, I do not think that there is ever a situation where we want to include all filter options _while a secondary filter is already selected_. Refreshing the page, all secondary filters will be removed. At that point, because no secondary filters are selected, _all_ filters will be available. Subsequently, we never want to include all secondary filters as we will always want to crossfilter.

## Testing
I tested this fix locally and tried a number of different filter permutations. The feature seems to be working now.


tihuan:
After digging more, I now remember the secondary filter is working as intended, so we can close this PR 🙏 

Full context:
https://czi-sci.slack.com/archives/C0247APK621/p1677888619567369?thread_ts=1677797196.603849&cid=C0247APK621

atarashansky:
> After digging more, I now remember the secondary filter is working as intended, so we can close this PR 🙏
> 
> Full context: https://czi-sci.slack.com/archives/C0247APK621/p1677888619567369?thread_ts=1677797196.603849&cid=C0247APK621

I'm going to leave this open for now until hearing from design and PM. Thanks, Timmy!

tihuan (approved):

PR #4293, opened 2023-03-03
fix(wmg): change join character to comma for shared links

atarashansky:
## Reason for Change

[Context](https://czi-sci.slack.com/archives/C0247APK621/p1677862105425899?thread_ts=1677860373.890269&cid=C0247APK621)

## Changes

- change join character from `-` to `,`
- disable share link when tissues or genes not selected
- add versioning
## Testing steps

- tested locally


danieljhegeman (approved):
yolo

Issue #4182, opened 2023-02-13
Remove code for using healthy cells for marker genes


signechambers1:
After discussion with @pablo-gar @atarashansky and our other SMEs we decided to use all cells to calculate marker genes. This ticket covers the work to remove the code.

Our rationale is: 
* Healthy cells are not necessarily healthy, there’s a lot of discretion in the annotation. However this risk is low as most single-cell biologist understand this nuance.
* We don’t believe marker genes will dramatically change upon inclusion of diseased cells, currently 1/3 are diseased. This is measurable and will be measured.
* SMEs think that more frequent than not user would prefer to see marker genes with all data.

PR #4295, opened 2023-03-03
feat(wmg): remove healthy cell filter for marker gene calculation

atarashansky:
## Reason for Change

- #4182 

## Changes

- remove healthy cell filter for marker gene calculation and update tests

## Testing steps

- Unit tested only


ashin-czi (approved):
LGTM and approved, but any chance that we'd want healthy cell filtering again in the future?

atarashansky:
> LGTM and approved, but any chance that we'd want healthy cell filtering again in the future?

Nope! Thanks, @ashin-czi !!!

Issue #4285, opened 2023-03-02
bug(wmg-share): Organism is not included as a parameter in the share url


seve:
@signechambers1 : im testing the share functionality in staging and if I select Organism = Mus musculus, copy the link, then paste the link into a browser the application errors (regardless of tissue and genes selected). I'm guessing organism isn't captured in the link generation, is this intended?

PR #4304, opened 2023-03-06
fix(wmg): add organism to share url

seve:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/4285

Adds organism to share URL (version 3)

Also adds organism to analytics payload

### DIFF
--- a/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/ShareButton/utils.tsx
+++ b/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/ShareButton/utils.tsx
@@ -10,18 +10,20 @@ import {
 
 export const generateAndCopyShareUrl = (
   filters: State["selectedFilters"],
+  organism: State["selectedOrganismId"],
   tissues: State["selectedTissues"],
   genes: State["selectedGenes"]
 ) => {
   // Create a URL that contains the selected filters, tissues, and genes as params in the URL
   // This URL can be shared with others to reproduce the same view
   const url = new URL(window.location.href);
+  url.searchParams.set("organism", organism ?? "");
   Object.entries(stripEmptyFilters(filters)).forEach(([key, value]) => {
     url.searchParams.set(key, value.join(","));
   });
   url.searchParams.set("tissues", tissues.join(","));
   url.searchParams.set("genes", genes.join(","));
-  url.searchParams.set("ver", "2");
+  url.searchParams.set("ver", "3");
 
   // Copy the URL to the clipboard
   navigator.clipboard.writeText(url.toString());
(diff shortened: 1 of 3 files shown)


frontend/src/views/WheresMyGene/components/GeneSearchBar/components/ShareButton/utils.tsx
```diff
@@ -67,7 +69,13 @@ export const loadStateFromQueryParams = (
   // Check for version
   let version = params.get("ver");
   if (!version) version = "1";
-  const delimiter = version === "1" ? "-" : ",";
+  const delimiter = version > "1" ? "," : "_";
```
tihuan, line 72:
```
let version = params.get("ver");
if (!version) version = "1";
```

can be compressed to `const version = params.get("ver") || "1"`

tihuan: Also just double checking, does version 1 use `"-"` or "_"` as delimiter? Can't tell from the code 😆 Thanks!

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/ShareButton/utils.tsx
```diff
@@ -10,18 +10,20 @@ import {
...
   // This URL can be shared with others to reproduce the same view
   const url = new URL(window.location.href);
+  url.searchParams.set("organism", organism ?? "");
   Object.entries(stripEmptyFilters(filters)).forEach(([key, value]) => {
     url.searchParams.set(key, value.join(","));
   });
   url.searchParams.set("tissues", tissues.join(","));
   url.searchParams.set("genes", genes.join(","));
-  url.searchParams.set("ver", "2");
+  url.searchParams.set("ver", "3");
```
tihuan, line 26: why are we bumping the version here? Is it because we're changing `delimiter` from `version === "1" ? "-" : ","` to `version > "1" ? "," : "_"`

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/ShareButton/utils.tsx
seve: oh my, thank you for catching this

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/ShareButton/utils.tsx
seve: no version bump needed!

tihuan (approved):
LGTM! Thanks Seve 🚀 👌

Issue #4300, opened 2023-03-03
Clicking Download resets the Datasets view to the top


brianraymor:
Reported by @Bento007 in [**#single-cell-eng**](https://czi-sci.slack.com/archives/C023Q1APASK/p1677880800623459).

1. Navigate to the Datasets view
2. Scroll down the page
3. Click Download for a dataset
4. The view is reset to the top of the page. The user loses their place.

PR #4318, opened 2023-03-09
fix: Clicking Download resets the Datasets view to the top (#4300)

frano-m:
## Reason for Change
The dataset download button, when actioned, causes the datasets table to scroll back to the top of the table. It is desirable to maintain the scroll position when downloading any dataset.

- #4300

## Changes

- Overflow styles on the datasets/collections table are relocated to an ancestor element.

## Testing steps

- Navigate to the Datasets view.
- Scroll down the page.
- Download any dataset.
- The table view should remain at the current scroll position.


tihuan (approved):
LGTM!! Thanks so much for the fix, Fran!

Issue #3621, opened 2022-11-22
Recreate wmg requirements.txt with primary dependencies only


atolopko-czi:
This [requirements.txt](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/wmg/pipeline/requirements.txt) file was generated using `pip freeze` and therefore includes all transitive dependencies, rather than just the primary package dependencies.

Recreate the file by identifying the primary packages only.  E.g.:
```
❯ python -m findimports --ignore-stdlib -p backend/wmg/pipeline | sort -u | grep -v backend
  anndata
  anndata._core
  numba
  numpy
  pandas
  scanpy
  scipy.sparse
  tiledb
```

Maintain the pinned versions for the primary packages.

atolopko-czi:
Attempting this as part of work for #3615

atolopko-czi:
@atarashansky The WMG requirements.txt have been updated on the PR, but a few unit test failures still need to be addressed.

Issue #3615, opened 2022-11-21
Data Portal - update BE dependencies


danieljhegeman:
- [ ] Update and pin BE dependencies
- [ ] Deploy and sanity check that nothing is broken
- [ ] Update [how-to-upgrade-dependencies.md](https://github.com/chanzuckerberg/single-cell/blob/main/how-we-work/how-to-upgrade-dependencies.md)

#5024 is scheduled to be implemented Jul 10 - Jul 21 so suggest addressing this update after then.

atolopko-czi:
Now P1, as @maniarathi has asked us to address sooner than later. (Also Explorer BE dependencies is already a P1.)

metakuni:
Some dependencies were updated here:
- https://github.com/chanzuckerberg/single-cell-data-portal/pull/4986

metakuni:
Labeled as `dp` and `data-viz` since the Portal repo contains both Portal and Gene Expression

metakuni:
See also #5024

metakuni:
See also #5024

metakuni:
Created #5075 to separate the Gene Expression BE dependency update work.

metakuni:
Assigning to @danieljhegeman to address during next DP on-call rotation starting 2023-12-11.

metakuni:
@Bento007 , since we've assigned [dependency file owners](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/.github/dependabot.yml), can we close this issue?

PR #4323, opened 2023-03-09
chore: update backend dependencies

atolopko-czi:
## Reason for Change

#3615 
#3621

## Changes

- wmg pipeline requirements.txt recreated to only include primary dependences; it was a `pip freeze` result previously, which not ideal for understanding or updating the primary dependencies


atolopko-czi:
recreating PR with single-issue association (address #3621 first via its own PR)

Issue #4324, opened 2023-03-10
Implement: `/datasets` index endpoint


danieljhegeman:
See Swagger API description for this endpoint in [this feature branch](https://github.com/chanzuckerberg/single-cell-data-portal/blob/dan/rdev-4277/backend/curation/api/curation-api.yml).

danieljhegeman:
Getting rid of `filename` and `original_id` as part of this work since they are subsets of the affected `dataset` component schema.

brianraymor:
>  List all public Datasets with full metadata and **additional Cell Census attributes**.

1. The description references the "cell census". I've requested multiple times that the cell census never appear in the documentation. 
2. The documentation is silent on the ordering of datasets in the response. There are no specific product requirements beyond documenting the order (which may be random) to set expectations (you need to sort by) for the data consumer. @atolopko-czi - does the cell census builder have any preference?

atolopko-czi:
> 2\. does the cell census builder have any preference?

there is no ordering requirement the census

atolopko-czi:
A possible idea/requirement has surfaced that we may want to consider further for this endpoint: being able to filter return datasets compliant up to a given maximum Dataset Schema version.  Some initial discussion [here](https://czi-sci.slack.com/archives/C0247APK621/p1678491096239889?thread_ts=1678478848.628269&cid=C0247APK621).

brianraymor:
The endpoint documentation does not explain the rationale for `dataset_id` and `dataset_version_id`. There are some definitions in [Surfacing Versions](https://docs.google.com/document/d/1wkbV67CI-jWgwjzDAtlpZ4qE3evZdHEaETaK0Y7RnSo/edit#heading=h.8yxeqpft8mig) that could be reused here?


```
dataset_id
A unique identifier of a Dataset.

dataset_version_id	
A unique identifier of a Dataset version.
```

Could be simple like:

```
dataset_id

The unique identifier that always references the latest published version of this dataset.

dataset_version_id	

The unique identifier that always identifies a specific published version of this dataset, 
even after subsequent versions of this dataset have been published via revisions.
```

I might include an example of dataset_revision_id (perhaps using uuid(s) to be more realistic and not suggest an incremental set of numbers.)

dataset_id is 1.
dataset_version_id is 2.

The dataset is updated and republished.
dataset_id is 1.
dataset_version id is 3.

The dataset is updated again and republished.
dataset_id is 1.
dataset_version_id is 4.

[...]

PR #4331, opened 2023-03-10
feat: add /v1/datasets index endpoint

danieljhegeman:
## Reason for Change
Introduce endpoint to receive full metadata for all published Datasets

- #4324 

## Testing steps

- unit tests
- confirmed in both rdev and dev
- [notebook PR](https://github.com/chanzuckerberg/single-cell-curation/pull/403)


backend/curation/api/v1/curation/datasets/actions.py
```diff
@@ -0,0 +1,33 @@
...
+def get():
+    """
+    Datasets index endpoint to retrieve full metadata for all public Datasets.
+    """
+    collections_with_datasets = get_business_logic().get_all_published_collections_with_datasets()
+
+    all_datasets_with_collection_name_and_doi = []
+
+    for collection in collections_with_datasets:
+        assert collection.datasets  # Any public Collection should have at least one Dataset
```
nayib-jose-gloria, line 17: I think this assert belongs in a test rather than in the API code. We already have checks at publish time that wouldn't let you publish collections with 0 datasets.

backend/curation/api/v1/curation/datasets/actions.py
```diff
@@ -0,0 +1,33 @@
...
+        for dataset in collection.datasets:
+            dataset_response_obj = reshape_dataset_for_curation_api(dataset, is_published=True, use_canonical_id=True)
+            collection_doi = None
+            for link in collection.metadata.links:
+                if link.type == CollectionLinkType.DOI.name:
+                    collection_doi = link.uri
+            dataset_response_obj.update(
```
nayib-jose-gloria, line 24: We should only need to build this object once per collection and add it to each dataset object, rather than rebuild every dataset.

nayib-jose-gloria, line 24: We should only need to build this object once per collection and add it to each dataset object, rather than rebuild every dataset.

backend/layers/persistence/persistence.py
```diff
@@ -507,6 +507,46 @@ def get_all_datasets(self) -> Iterable[DatasetVersion]:
...
+        for dataset in active_datasets:
+            if collection_datasets := datasets_by_collection_id.get(dataset.collection_id.id):
+                collection_datasets.append(dataset)
+            else:
+                datasets_by_collection_id[dataset.collection_id.id] = [dataset]
+
+        # Turn list of CollectionVersions into CollectionVersionsWithDatasets
+        collections_with_datasets: List[CollectionVersionWithDatasets] = []
+        for collection in active_collections_list:
+            print(f"collection with id {collection.collection_id}")
```
nayib-jose-gloria, line 531: remove print statement

backend/layers/persistence/persistence.py
```diff
@@ -507,6 +507,46 @@ def get_all_datasets(self) -> Iterable[DatasetVersion]:
...
+        active_collections = self.get_all_mapped_collection_versions()
+
+        dataset_version_ids = []
+        active_collections_list = list(active_collections)
+        for collection in active_collections_list:
+            dataset_version_ids.extend(collection.datasets)
+        active_datasets = self._get_datasets(dataset_version_ids)
+        datasets_by_collection_id = {}
+        for dataset in active_datasets:
+            if collection_datasets := datasets_by_collection_id.get(dataset.collection_id.id):
```
nayib-jose-gloria, line 523: I think we can simplify this loop block by initializing datasets_by_collection_id as a defaultdict(list) and then doing:
```
for dataset in active_datasets:
      datasets_by_collection_id[dataset.collection_id.id].append(dataset)
```

nayib-jose-gloria (commented):
left a few notes/suggestions + agree with Emanuele's comments to move the mapping logic out of the persistence layer and into the API layer using our existing business methods

backend/layers/persistence/persistence.py
```diff
@@ -507,6 +507,46 @@ def get_all_datasets(self) -> Iterable[DatasetVersion]:
             dataset_version_ids.extend(collection.datasets)
         return self._get_datasets(dataset_version_ids)
 
+    def get_all_published_collections_with_datasets(self) -> List[CollectionVersionWithDatasets]:
```
ebezzi, line 510: Why do we need another method in persistence? In fact, even the business layer has two methods that should give you all you need: `get_collections` and `get_all_published_datasets`. You can just call those two and join them by id. This will also guarantee that the /datasets index endpoint will return the same exact values than the portal API version.

ebezzi (changes requested):

backend/layers/persistence/persistence.py
nayib-jose-gloria: yeah this is a good point--fwiw, even this implementation as written is business logic mapping results from `get_all_mapped_collection_versions` and `_get_datasets` so it should be in the business layer

backend/curation/api/v1/curation/datasets/actions.py
danieljhegeman: I apologize. I pushed this change to rdev and forgot to push it to the feature branch 🤦‍♂️

backend/curation/api/v1/curation/datasets/actions.py
danieljhegeman: Totally agreed, will change

backend/layers/persistence/persistence.py
danieljhegeman: @ebezzi I'll look closer at those methods and see if I can refactor.

@nayib-jose-gloria I disagree based on the precedents set by other methods. There really is no "business" logic inherent to this API call. It is simply grabbing entities, and then there's a little api-layer logic which is restructuring the response object.

nayib-jose-gloria: @danieljhegeman I meant API layer, I edited my response after I sent it

danieljhegeman: Looked at the code again, and remembered why I chose to implement in a single persistence method:

using @ebezzi 's suggestion of `get_collections` and `get_all_published_datasets` would perform the 'collections' lookup twice, because it happens in both of those calls. My method streamlines this and only fetches all collections once.

Bento007: Rather than jumping from a list of collections with datasets to a list of datasets with collections, go straight to a list of datasets with collections from the persistence layer. The name would be `get_all_published_datasets_with_collection`.

Bento007: There has gotta be a query we can do that does most of this filtering on the database side.

danieljhegeman: So I'm trying to follow the precedents set in the code base already, which means not using sql alchemy joins (I'm assuming that's what you're suggesting @Bento007 ). I'll wait for consensus from @ebezzi and @nayib-jose-gloria.

Bento007: One more suggestion would be to modify `get_all_mapped_collection_versions` to optionally return `Iterable[CollectionVersionWithDatasets]` 

```
    def get_all_mapped_collection_versions(self, get_tombstoned: bool = False, datasets: bool =True) -> Union[Iterable[
        CollectionVersion], Iterable[CollectionVersionWithDatasets]]:
        .
        .
        .
            for version in versions:
                canonical_row = mapped_version_ids[version.id]
                canonical = CanonicalCollection(
                    CollectionId(str(canonical_row.id)),
                    CollectionVersionId(str(canonical_row.version_id)),
                    canonical_row.originally_published_at,
                    canonical_row.revised_at,
                    canonical_row.tombstone,
                )
                if datasets:
                    _datasets = self._get_datasets([DatasetVersionId(str(id)) for id in version.datasets])
                    yield self._row_to_collection_version_with_datasets(version, canonical, _datasets)
                else:
                    yield self._row_to_collection_version(version, canonical)

File: backend/layers/business/business.py [hunk 6]
```diff
@@ -403,6 +403,12 @@ def get_all_published_datasets(self) -> Iterable[DatasetVersion]:
         """
         return self.database_provider.get_all_datasets()
 
+    def get_all_published_collections_with_datasets(self) -> Iterable[CollectionVersionWithDatasets]:
+        """
+        Retrieves all the datasets from the database that belong to a published collection
+        """
+        return self.database_provider.get_all_published_collections_with_datasets()
```
nayib-jose-gloria, line 410: The persistence method should be called `get_all_mapped_collection_versions_with_datasets`, to be consistent with the distinction we've made in other methods between published collections (business concept) and mapped collection versions (schema concept). Mirrors 'get_all_mapped_collection_versions', and more clearly defines difference between that method and this one (and thus, the need for a different persistence method). 


note: while fields like 'is_published' exist in the db, those stay set to True once published even if the collection version is no longer the 'active' one. What we refer to in the business sense as 'published' usually means 'actively, currently published' which in the DB refers to whatever version is mapped to the canonical

backend/layers/persistence/persistence.py
danieljhegeman: Thanks for the suggestion @Bento007 . I definitely want to go with more code reuse. I think I would rather just always return both the collections and datasets. The user can ignore one, use the other, or use both. Slightly more readable/usable that way.

Bento007: you can also go the extra mile and add some caching over getting the collection and dataset_version

nayib-jose-gloria: @danieljhegeman in the spirit of reducing the number of needless calls though, there are use cases leveraging this now that have no need to make the db calls to populate the extra dataset info

danieljhegeman: Good point.

danieljhegeman: >you can also go the extra mile and add some caching over the collection and dataset versions
probably not since one of the general operating principles for the Discover API has been 'most recent data always'. So we'd need to track most-recently-published version in some way. A feature for another day, perhaps.

ebezzi (approved):

backend/layers/persistence/persistence.py
Bento007: you cache on the version_id and that would all be handled by the LRU cache, but it does not need to be part of this PR.

backend/layers/persistence/persistence.py
```diff
@@ -497,15 +498,46 @@ def get_all_versions_for_dataset(self, dataset_id: DatasetId) -> List[DatasetVer
...
+        datasets_by_collection_id = defaultdict(list)
+        # Construct dict of collection_id: [datasets]
+        [datasets_by_collection_id[d.collection_id.id].append(d) for d in mapped_datasets]
+
+        # Turn list of CollectionVersions into CollectionVersionsWithDatasets
+        collections_with_datasets: List[CollectionVersionWithDatasets] = []
+        for collection in mapped_collections:
+            dataset_versions = datasets_by_collection_id.get(collection.collection_id.id, [])
+            collections_with_datasets.append(
+                CollectionVersionWithDatasets(
```
Bento007, line 526: This parsing is repeated but slightly different. https://github.com/chanzuckerberg/single-cell-data-portal/blob/abd337c5c1c3e6b45005a433d29237e4e6c63969/backend/layers/persistence/persistence.py#L108-L122

We should reconcile the differences otherwise we will end up with two different shapes of `CollectionVersionWithDatasets`

Bento007 (changes requested):

Bento007:
The lru_cache is a simple decorator https://docs.python.org/3/library/functools.html#functools.lru_cache

backend/layers/persistence/persistence.py
danieljhegeman: What differences are you referring to? These are slightly different situations. The code you're linking to is for transforming a row object into a CollectionVersion. I'm converting a CollectionVersion to a CollectionVersionWithDatasets.

danieljhegeman:
> The lru_cache is a simple decorator https://docs.python.org/3/library/functools.html#functools.lru_cache

@Bento007 but does this have any mechanism for determining whether the relevant data have been updated? Previously we had intentionally introduced zero caching for the Discover API cc @brianraymor

Bento007:
The versions are immutable so there isn't a reason to check for up to date..

backend/layers/persistence/persistence.py
Bento007: I see now. 

suggestion: A helper function to reduce the risk of a translation error from CollectionVersion to CollectionVersionWithDatasets would be nice. There isn't currently a reuse case at the moment, but I can see it being handy for future APIs.

Bento007 (approved):

backend/layers/persistence/persistence.py
danieljhegeman: I'll introduce that as soon as we repeat this operation 👍

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1456,6 +1456,7 @@ def test_get_datasets_200(self):
                 + [dv.dataset_id.id for dv in published_collection_2.datasets]
             )
             received_dataset_ids = set()
+            print(response.json)
```
nayib-jose-gloria, line 1459: nit: remove print

nayib-jose-gloria (commented):
1 nit, but looks good. 

I think there's some changes needed from my PR to have the dataset API response be reshaped correctly. I'll move this 'common' code into a separate PR

PR #4333, opened 2023-03-10
fix(wmg): Dropdown Tracking Events

ashin-czi:
## Reason for Change

Tracking events for dropdown selections were being sent erroneously when component was being rendered. Added a check to prevent this from happening

https://czi-sci.slack.com/archives/C0247APK621/p1678401267042679

## Changes

- add
- remove
- modify
    - logic in onChange function to check comparison between current option and onChange value

## Testing steps


tihuan (approved):
LGTM! Thanks for fixing this 🙏

Issue #4281, opened 2023-03-01
Application errors when you add marker genes to dot plot after switching organism


signechambers1:
Steps to reproduce:

1. Prod
2. Tissue = Spleen
3. Click info icon for macrophage, add genes to dot plot
4. Select Mus Musculus for organism (dot plot goes blank, marker genes panel remains)
5. Add tissue = spleen
6. Click "Add to dot plot" in marker genes panel
7. Application errors (probably because you are adding human genes to mouse spleen)

Desired behavior: Marker genes panel should close when organism filter is applied (let me know if you think a different desired behavior makes more sense - this is just my working idea).

![Screenshot 2023-03-01 at 10.12.41 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/c8763f3d-1348-4462-9e8f-cdae1a63ad63))

seve:
> Desired behavior: Marker genes panel should close when organism filter is applied (let me know if you think a different desired behavior makes more sense - this is just my working idea).

This makes sense to me

PR #4344, opened 2023-03-14
fix(wmg-fmg): reset cell type info value on organism switch

seve:
#4281 

Also change some var names for readability


ashin-czi (approved):
LGTM! Thanks for working on this!

Issue #4329, opened 2023-03-10
Implement: `/datasets/{dataset_id}/versions`


danieljhegeman:

brianraymor:
What is the rationale for returning a dictionary rather than an array? 

`{'datasets': [{'assay': [{'label': "10x 3' v2",
`

An array would be similar to how we already return responses for similar patterns like GET collections.

PR #4352, opened 2023-03-15
feat: add endpoint to get dataset version history

nayib-jose-gloria:
## Reason for Change

- #4329

## Changes

- added /datasets/<dataset_id>/versions endpoint

## Testing steps

- Unit Tests

## Notes for Reviewer
- Could not just call `get_all_versions_for_dataset` in implementation because that method does not filter out DatasetVersions that were never published (i.e. intermediate changes during a revision--we create DatasetVersions for all changes that occur within a revision, but only those mapped to a CollectionVersion that got published are the published versions)

- Proposed a slightly different API response shape than the initial swagger, to simplify. Namely, I dropped the collection_version_id for each dataset version because I don't think it adds much extra value and it would somewhat complicate the business function return type (i.e. having to return List[CollectionVersionWithDatasets] vs List[DatasetVersion]) just for one field). Will change if desired


danieljhegeman:
> Could not just call get_all_versions_for_dataset in implementation because that method does not filter out DatasetVersions that were never published

@nayib-jose-gloria can't we call that method and then only return the dataset versions which have a non-null `published_at`?

nayib-jose-gloria:
> > Could not just call get_all_versions_for_dataset in implementation because that method does not filter out DatasetVersions that were never published
> 
> @nayib-jose-gloria can't we call that method and then only return the dataset versions which have a non-null `published_at`?

@danieljhegeman the db does not store dataset version published_at. 

They have a published_at on their canonical counterparts, but that would be populated as long as the canonical dataset has been published once, regardless of which version it was pointing to when it was published.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1391,6 +1392,49 @@ def test_get_datasets_nonexistent_collection_404(self):
             self.assertEqual(404, response.status_code)
 
 
+class TestGetDatasetIdVersions(BaseAPIPortalTest):
+    def test_get_dataset_id_versions_ok(self):
+        collection = self.generate_published_collection()
```
Bento007, line 1397: Add more documentation. It's obscure that your are testing that unpublished datasets should not appear in the response.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1391,6 +1392,49 @@ def test_get_datasets_nonexistent_collection_404(self):
...
+        self.assertNotIn(unpublished_dataset_revision_id, dataset_version_ids)
+
+    def test_get_dataset_id_version_404(self):
+        with self.subTest("Dataset with that UUID does not exist"):
+            test_url = f"/curation/v1/datasets/{str(uuid.uuid4())}/versions"
+            headers = self.make_owner_header()
+            response = self.app.get(test_url, headers=headers)
+            self.assertEqual(404, response.status_code)
+        with self.subTest("Dataset Exists, but not published so no public version history"):
+            dataset_id = self.generate_unpublished_collection(add_datasets=1).datasets[0].dataset_id
```
Bento007, line 1431: Add a test for something that is not UUID shaped.

Bento007 (changes requested):
A few things to add

danieljhegeman:
> @danieljhegeman the db does not store dataset version published_at.

@nayib-jose-gloria I think this is incorrect 😛 
```
                    Table "persistence_schema.CollectionVersion"
       Column        |            Type             | Collation | Nullable | Default 
---------------------+-----------------------------+-----------+----------+---------
 id                  | uuid                        |           | not null | 
 collection_id       | uuid                        |           |          | 
 collection_metadata | json                        |           |          | 
 owner               | character varying           |           |          | 
 curator_name        | character varying           |           |          | 
 publisher_metadata  | json                        |           |          | 
 published_at        | timestamp without time zone |           |          | 
 created_at          | timestamp without time zone |           |          | 
 datasets            | uuid[]                      |           |          | 
 ```

File: backend/curation/api/curation-api.yml [hunk 3]
```diff
@@ -560,6 +560,34 @@ paths:
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  collection_id:
+                    $ref: "#/components/schemas/collection_id"
+                  datasets:
+                    type: array
+                    items:
+                      $ref: "#/components/schemas/dataset"
```
danieljhegeman, line 587: This needs to be merged into a single object [for the sake of alphabetization](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1678918244130559). It should be like [dataset_version](https://github.com/chanzuckerberg/single-cell-data-portal/blob/dan/rdev-4277/backend/curation/api/curation-api.yml#L979-L1067).

danieljhegeman: If we're going to return `collection_id`, I really think we should also return `collection_version` idea, _especially_ for a `versions` endpoint.

nayib-jose-gloria: Its already alphabetized because I moved collection_id outside of the datasets array (its gonna be the same one for the whole list)

it'll be 
```
collection_id: ....
datasets: [
  ...,
  ...,
]
```

nayib-jose-gloria: re: collection_version_id--dataset versions can (and probably will) belong to more than 1 collection version. We can instead use that field to reference the collection version it was introduced in, but I would want to update the field name to be clear about that (and not accidentally imply it only exists in one collection version). Or, alternatively, provide a list of all the collection versions it does appear in. 

At that point, I do question how useful having it here is vs. providing the canonical collection ID and letting users get any follow up info by querying the collection version endpoint with the collection ID

danieljhegeman: >Its already alphabetized because I moved collection_id outside of the datasets array (its gonna be the same one for the whole list)

Thanks, I missed this. Though this is making me think: we really should try to implement a consistent approach to the shape of the entities _across_ the endpoints. So for any endpoint that returns a list of Dataset-like objects, they should be the same, or as similar as possible.

So, if you're set on moving shared collection-level values like `collection_id` and `collection_version_id` up, out of the Dataset object, then we should probably do the same for the Datasets index endpoint: instead of an array of Datasets, return an array of Collection objects with Datasets arrange underneath, like with the Collections index endpoint (but with _full_ Dataset metadata instead). The question then is: do we really even need to maintain separate Collections and Datasets index endpoints? I think we might want to just consolidate into one. They're very similar, and returning the other Collection metadata on top of what we're already returning for the Datasets index endpoint is trivial. cc @brianraymor

backend/curation/api/curation-api.yml
```diff
@@ -560,6 +560,34 @@ paths:
...
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  collection_id:
+                    $ref: "#/components/schemas/collection_id"
```
danieljhegeman, line 583: @nayib-jose-gloria per discussion today, pls put back into the dataset object 🙏

tests/unit/backend/layers/business/test_business.py
```diff
@@ -835,6 +835,34 @@ def test_get_dataset_version_from_canonical(self):
...
+            unpublished_collection_version_id, "http://fake.url", None, new_dataset_version_id
+        )
+
+        version_history = self.business_logic.get_prior_published_versions_for_dataset(dataset.dataset_id)
+        version_ids = [version.version_id for version in version_history]
+        # Check that only published datasets appear
+        self.assertEqual(len(version_history), 2)
+        self.assertIn(dataset.version_id, version_ids)
+        self.assertIn(new_dataset_version_id, version_ids)
+        self.assertNotIn(unpublished_dataset_version_id, version_ids)
```
danieljhegeman, line 864: nit / tell me how you feel: I've started building sets of ids to run a single `assertEqual` on: expected vs received. Fewer comparisons to keep track of since set takes care of number and inclusion/exclusion of relevant ids.

danieljhegeman: Also applies to the test further up

backend/layers/common/entities.py
```diff
@@ -201,6 +201,11 @@ class DatasetVersion:
     canonical_dataset: CanonicalDataset
 
 
+@dataclass
+class PublishedDatasetVersion(DatasetVersion):
+    collection_version_id: CollectionVersionId  # Pointer to collection version it was originally published under
```
danieljhegeman, line 206: I may circle back and create an entity class for DatasetIndex... looks good

backend/layers/business/business.py
```diff
@@ -578,6 +579,34 @@ def get_dataset_version_from_canonical(self, dataset_id: DatasetId) -> Optional[
...
+        Given a canonical dataset id, return all its DatasetVersions that have been part of published CollectionVersions
+        """
+        dataset_version = self.database_provider.get_dataset_mapped_version(dataset_id)
+        if not dataset_version:
+            return []
+        collection_versions = self.database_provider.get_all_versions_for_collection(dataset_version.collection_id)
+        published_version_history = []
+        found_version_ids = set()
+        # sort to ensure we always find earliest instance of a dataset version first when iterating
+        collection_versions = sorted(collection_versions, key=lambda cv: (cv.published_at is None, cv.published_at))
```
danieljhegeman, line 593: how does the sorting for a tuple key work?

nayib-jose-gloria: In general, tuple keys will prioritize sorting on the first entry and use the second entry as the tiebreaker sort

In this scenario that means it'll always sort unpublished versions to the end of the list (where published_at is None is True), then sort by published_at among the published versions

nayib-jose-gloria: for added context: the reason Trues go to the end is because its actually sorting on False vs. True alphabetically

nayib-jose-gloria: also necessary to sort while handling the None values, otherwise you'll get a Type Error on the sort

https://stackoverflow.com/questions/48234072/how-to-sort-a-list-and-handle-none-values-properly

danieljhegeman: >for added context: the reason Trues go to the end is because its actually sorting on False vs. True alphabetically

this is very... coincidental / orthogonal to the actual entity. Lol. Can you add a comment explaining this if you're going to use this tactic? I mean I'm not wholly opposed, but in the absence of an explicit lambda that shows the logic, it would be good to have a comment.

nayib-jose-gloria: tbf we're filtering out the Nones anyway so whether it gets sorted to the end or the start is actually irrelevant (the False vs. True stuff is more a fun fact/quirk of the implementation). The real purpose is to allow a list with a NoneType to be sorted by datetime. The other option is to iterate through the list once to filter out unpublished versions before sorting, which is maybe more clear but less efficient since we add an extra loop. 

tl;dr: I'll add a comment saying this is done to avoid a TypeError on sorting, the implementation details aren't super relevant

Issue #4353, opened 2023-03-16
fix(wmg): validation pipeline should not depend on having tissue_original_ontology_term_id as a dimension


atarashansky:
`tissue_original_ontology_term_id` was an unused dimension in the snapshot cubes. It was changed to be an attribute in the recent WMG test upgrade (#4336). 

The validation code that tests the tissue rollup is currently broken as the tests were written assuming the previous schema. As these tests are only run at the end of the processing pipeline, this regression was not caught until a recent [dev run](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Fdp$252Fdev$252Fdevstack$252Fwmg-processing/log-events/wmg-dev-devstack-wmg-processing$252Fdefault$252F323c0a8ef1514bfab5f4f816cbe780ff) failed.

Validation will be turned off temporarily until this ticket is resolved.

PR #4354, opened 2023-03-16
chore(wmg): disable validation until validation tests are updated to match new schemas

atarashansky:
## Reason for Change

See #4353 for context. This patch unblocks dev pipeline runs.


tihuan (approved):

ashin-czi (approved):
Thank you!!

Issue #4264, opened 2023-02-27
Add 'version_id' as a field to all Discover API responses returning Collections and/or Datasets


nayib-jose-gloria:
Simplify or remove logic around setting 'correct' ID in reshape_for_curation_api utility functions

Update curation-api OpenAPI accordingly. Clearly define use-case distinctions for version_id and id. 

Deprecate 'original_id' in the process

This change represents all endpoints that return collection and dataset responses returning 'dataset_id' or 'collection_id' in place of 'id'. It also adds 'dataset_version_id' and 'collection_version_id' to all endpoints with 'dataset_id' or 'collection_id' returned.

danieljhegeman:
This should be done after all other p0 endpoint implementations.

PR #4355, opened 2023-03-16
feat: Add 'version_id' as a field to all Discover API responses returning Collections and/or Datasets

nayib-jose-gloria:
## Reason for Change

- #4264
- #4266
- #4330

## Changes

- refactored reshape_dataset_for_curation_api and reshape_collection_for_curation_api functions to accommodate explicit 'collection_id'/'dataset_id' and collection_version_id/dataset_version_id in the API responses
- changed all Discover API instances of 'id' to either collection_id or dataset_id in API response
- updated open API spec

## Testing steps

- unit tests


danieljhegeman (approved):
Thanks Nayib! I didn't check to see if you caught all instances but I imagine we'll find them as we go if not.

PR #4202, opened 2023-02-14
feat(wmg): backend compare changes (#4198)

ashin-czi:
* feat(wmg): backend compare changes

Backend changes to retrieve aggregated and compare expressions from the backend through the query API

## Reason for Change

- #4115 
- If the reason for this PR's code changes are not clear in the issue, state value/impact

## Changes

- add
- remove
- modify
    - Modify existing query API to return aggregated and compare gene expressions

## Testing steps

- Either list QA steps or reasoning you feel QA is unnecessary
- Describe how you made sure to know that your changes worked. Should allow someone else to go verify your code without in depth knowledge.
- "Unit tested only", "tested in rdev by a, b, c, verifying feature worked by... ", "manually ran pipeline locally with these results: ..."
    - Will test in rdev and test locally pointing to rdev

## Notes for Reviewer

✨✨✨✨✨✨ FE CHANGES BELOW ✨✨✨✨✨✨

Front End Changes:

* feat(wmg): 
* #4112 
* #4113 
* #4116
* #4117

## Reason for Change

Add Compare FE

## Changes

1. Add analytics event `WMG_OPTION_SELECT_GROUP_BY `
2. Update analytics events related to share link to include compare value
3. Update `frontend/src/common/queries/wheresMyGene.ts` file to parse new BE response, where `term_id_labels` is now an object with stratified data and new key `order` instead of arrays
4. We basically now go through `term_id_labels`, iterate through all the tissues, within each tissue, sort all cell types by `order` key, and then within each cell type, sort compare options (e.g., aggregated, unknown, female, male) by the order of: aggregated first, then compareOptionId a to z, then unknown last, and then we add compare option name prefixed with `"    "` to mimic indentation instead of using css, so we don't need to change SVG download code, then we flatten all the compare option objects in a one dimensional sorted array like they'll show up in the heatmap y axis
5. Introduces new data model `CellTypeRow`, where `viewId` is a unique combo key of `cellTypeId + compareOptionId` as string template `cellTypeId$compareOptionId`. E.g., `"CL:0000003$PATO:0000383", "CL:0000003$aggregated"`. This is needed since `cellTypeId` alone is no longer enough to uniquely identify each cell type row in the heatmap given the compare options
6. Refactor Filter components like `ColorScale`, "Compare", "Organism", and "Sort By" to keep the code DRY, since there's lot of identical duplicate code to improve readability and maintainability
7. Add `compare` to `/query` request
8. Update share link to load `compare` 
[...]

PR #4361, opened 2023-03-17
fix(wmg): fixing compare and crossfiltering test

ashin-czi:
## Reason for Change

- Compare e2e test was failing for "multiethnic" compare option value
- Crossfiltering test was failing because disease wasn't done loading after selecting a dataset and retained same disease count after selecting a dataset

## Changes

- add
- remove
- modify
    - Handling multiethnic and similar cases from not showing up as "unknown" in the UI
    - After selecting a dataset, getting after tissue count before getting after disease count

## Testing steps

`npm run e2e -- tests/features/wheresMyGene.test.ts`

### DIFF
--- a/frontend/src/common/queries/wheresMyGene.ts
+++ b/frontend/src/common/queries/wheresMyGene.ts
@@ -807,7 +807,11 @@ function addCellTypeRowToResult({
      * (thuang): We manually indent 4 spaces instead of using CSS, so we don't
      * have to update SVG render function to mimic CSS padding
      */
-    const name = isAggregated ? rawName : `    ${rawName || "unknown"}`;
+    const name = isAggregated 
+      ? rawName 
+      : compareOptionId === COMPARE_OPTION_ID_FOR_UNKNOWN
+        ? `    ${rawName || COMPARE_OPTION_ID_FOR_UNKNOWN}`
+        : `    ${rawName || compareOptionId}`;
 
     if (isAggregated) {
       cellTypeName = rawName;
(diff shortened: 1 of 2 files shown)


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -807,7 +807,11 @@ function addCellTypeRowToResult({
      * (thuang): We manually indent 4 spaces instead of using CSS, so we don't
      * have to update SVG render function to mimic CSS padding
      */
-    const name = isAggregated ? rawName : `    ${rawName || "unknown"}`;
+    const name = isAggregated
+      ? rawName
+      : compareOptionId === COMPARE_OPTION_ID_FOR_UNKNOWN
+      ? `    ${rawName || COMPARE_OPTION_ID_FOR_UNKNOWN}`
```
tihuan, line 813: why is this needed again lol? Sorry I don't have the context 😆 Thank you!

ashin-czi: image

multiethnic is being set to "unknown" because the rawName is null from the BE response, similar to how the unknown compareOptionId is null for the name

tihuan: ah I thought you changed BE to use id as name already. Did we lose that in merge conflict resolution or something?

ashin-czi: hmmm you're right, I remember the conversation now

I'll take a look into it

ashin-czi: I think [this commit](https://github.com/chanzuckerberg/single-cell-data-portal/pull/4202/commits/a40bed6d2c82ae2410278855e8d4dcc9cc543cb0) from @atarashansky is what is passing multiethnic `name` back as null from the BE

Should we keep the name as null as it is or return the compare option value for `name` as before?

tihuan: I vote for returning the id as `name` since `null` is pretty useless 🙏

ashin-czi: Alec is working on a patch for this now!

PR #4362, opened 2023-03-17
feat: get collection version endpoint

nayib-jose-gloria:
## Reason for Change

- #4327

## Changes

- add collection version endpoint
- update reshape_for_curation_api function to accept a flag to account for wanting to use version IDs in the dataset explorer URLs for datasets in collection versions returned by this endpoint; otherwise, would return same canonical URL for all versions of datasets regardless of which version you queried. 

## Testing steps
- unit tests


backend/curation/api/v1/curation/collections/common.py
```diff
@@ -94,6 +96,8 @@ def reshape_for_curation_api(
             use_canonical_url = True
         revising_in = None
 
+    if use_dataset_version_explorer_urls is not None and use_dataset_version_explorer_urls:
```
danieljhegeman, line 99: ??

Why not just `if use_dataset_versions_explorer_urls:` ?

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -903,6 +903,51 @@ def test__get_collection_with_x_approximate_distribution_none__OK(self):
...
+        with self.subTest("Query endpoint with non-UUID"):
+            res = self.app.get("/curation/v1/collection_versions/bad-input-id", headers=self.make_owner_header())
+            self.assertEqual(403, res.status_code)
+        with self.subTest("Collection Version is part of tombstoned Collection"):
+            collection = self.generate_published_collection()
+            self.business_logic.tombstone_collection(collection.collection_id)
+            res = self.app.get(
+                f"/curation/v1/collection_versions/{collection.version_id}", headers=self.make_owner_header()
+            )
+            self.assertEqual(403, res.status_code)
```
danieljhegeman, line 948: Add sub test to check that querying for an published revision via this endpoint fails

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -903,6 +903,51 @@ def test__get_collection_with_x_approximate_distribution_none__OK(self):
...
+        # test correct dataset explorer url is used
+        explorer_url = res.json["datasets"][0]["explorer_url"]
+        self.assertTrue(explorer_url.endswith(f"{revision.datasets[0].version_id.id}.cxg/"))
+
+    def test_get_collection_version_403(self):
+        with self.subTest("Query endpoint with incorrect ID"):
+            res = self.app.get(
+                f"/curation/v1/collection_versions/{str(uuid.uuid4())}", headers=self.make_owner_header()
+            )
+            self.assertEqual(403, res.status_code)
```
danieljhegeman, line 932: I think this should be a 404...

danieljhegeman (changes requested):
see comments

tests/unit/backend/layers/api/test_curation_api.py
nayib-jose-gloria: do you mean unpublished revision? 

Assuming yes, does that mean we don't want the capability URL principle to be applied here?

danieljhegeman: > do you mean unpublished revision?

Yes 😇 

> Assuming yes, does that mean we don't want the capability URL principle to be applied here?

No. The id of an **_unpublished_** revision should be treated as a canonical id from a user perspective. That means, while it really is a version id on our end, users should only see it work in places where a canonical id works: 
`/collections/{collection_id}...` ✅ 
`/collection_versions/{collection_version_id}` ⛔

backend/curation/api/v1/curation/collection_versions/collection_version_id/actions.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def get(collection_version_id: str):
+    """
+    Fetch the full metadata of a published CollectionVersion, including its datasets
+    """
+    validate_uuid_else_forbidden(collection_version_id)
+    version = get_business_logic().get_collection_version(CollectionVersionId(collection_version_id))
+    if version is None or version.canonical_collection.tombstoned is True:
+        raise NotFoundHTTPException()
```
danieljhegeman, line 19: tombstones should produce a 403

danieljhegeman: have not done an exhaustive search for consistency, might be worth doing, but see [this](https://github.com/chanzuckerberg/single-cell-data-portal/blob/eefb7d71bb50fe81f434d652c51a09ec4bc344dd/backend/curation/api/v1/curation/collections/common.py#L302-L303) as example

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -49,50 +49,55 @@ def extract_doi_from_links(links: List[Link]) -> Tuple[Optional[str], List[dict]
...
+            else:
+                # If it's an unpublished, unrevised collection, then collection_url will point to the permalink
+                # (aka the link to the canonical_id) and the collection_id will point to version_id.
+                # Also, revision_of should be None, and the datasets should expose the canonical url
+                collection_url = f"{get_collections_base_url()}/collections/{collection_version.collection_id.id}"
+                revision_of = None
+                use_canonical_url = True
+            revising_in = None
+    else:
+        collection_url = f"{get_collections_base_url()}/collections/{collection_version.version_id.id}"
```
danieljhegeman, line 98: @nayib-jose-gloria why would we return a url for a historical version that 
1) currently doesn't do anything, and 
2) we will soon change to redirect to the published version

perhaps `2)` is the desired behavior? Thoughts, @brianraymor ?  I'm not saying there's a good answer given our current UI behavior.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -922,7 +922,7 @@ def test_get_collection_with_no_datasets(self):
         collection_version = self.generate_unpublished_collection(add_datasets=0)
         self._test_response(collection_version)
 
-    def test_get_colletion_with_dataset_no_metadata(self):
+    def test_get_collection_with_dataset_no_metadata(self):
```
danieljhegeman, line 925: ❤️

danieljhegeman (changes requested):
Tombstones need to return 403. Also see question about `collection_url` for historical versions.

backend/curation/api/v1/curation/collections/common.py
nayib-jose-gloria: I don't mind getting rid of it for this endpoint. I don't really think it makes sense in the context of querying for a particular version to get a URL out to the canonical collection, which may not match the version you're querying for. We also already provide the canonical collection_id in case you do want access to it.

tbh I linked out to the collection version url because we currently can access historical versions in the UI and I forgot we were going to make the change to redirect lol

danieljhegeman: I'm fine with leaving it and accepting the impending behavior of redirect. But also would like to just get @brianraymor 's approval for this.

nayib-jose-gloria: I also vaguely remember us discussing wanting to have historical versions available in the UI at some point in the future, but we can update this in the future if that ever does happen, we don't need to future-proof it now

danieljhegeman (approved):
@nayib-jose-gloria let's merge and we can circle back to edit the collection url as @brianraymor does or does not see fit. Thanks for getting the tombstone error code fix in.

PR #4368, opened 2023-03-20
chore(wmg): selected items in quick select pin after reopen

atarashansky:
## Reason for Change

- #4209

## Changes

- added a useMemo hook to make it so item options update only when the dropdown closes and opens.

## Testing steps

- tested locally

### DIFF
--- a/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
@@ -134,6 +134,7 @@ export default function QuickSelect<
   const [open, setOpen] = useState(false);
   const [input, setInput] = useState("");
   const [hasComma, setHasComma] = useState(false);
+  const [itemOptions, setItemOptions] = useState<T[]>(items);
 
   const useStyles = makeStyles((theme: SDSTheme) => {
     const colors = getColors({ theme });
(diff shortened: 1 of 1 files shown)


frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
```diff
@@ -252,10 +253,9 @@ export default function QuickSelect<
     }
   };
   const selectedAsArray = Array.isArray(selected) ? selected : [selected];
-
-  const itemOptionsWithSelectedOnTop = [
-    ...new Set([...selectedAsArray, ...items]),
-  ];
+  const itemOptionsWithSelectedOnTop = useMemo(() => {
+    return [...new Set([...selectedAsArray, ...items])];
+  }, [open]);
```
tihuan, line 258: This dependency list seems incomplete?

tihuan: Do you have the eslint extension installed to catch React hook bugs like this?

tihuan (changes requested):
Incomplete hook dependency!

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
atarashansky: It is purposefully incomplete to avoid `itemOptionsWithSelectedOnTop` from being updated when the selectedAsArray changes (though I think `items` is safe to add to the dependency list so I'll add that)

atarashansky: Nevermind, adding `items` to dependency list breaks the desired behavior because the list is getting recreated in memory somewhere upstream.

frontend/src/views/WheresMyGene/components/GeneSearchBar/components/QuickSelect/index.tsx
```diff
@@ -162,6 +163,14 @@ export default function QuickSelect<
 
   const classes = useStyles();
 
+  const toggleDropdownAndUpdateItemOptions = () => {
+    setOpen(!open);
```
tihuan, line 167: To be safe, let's do this instead:

```suggestion
    setOpen((open) => !open);
```

source: https://react.dev/learn/queueing-a-series-of-state-updates#updating-the-same-state-multiple-times-before-the-next-render

tihuan (approved):
Looks super good! Just one last thing 🔥 Thanks so much for the quick iterations, Alec!

PR #4370, opened 2023-03-20
chore(wmg+compare): return name as ID when no label is found in ontology

atarashansky:
# Reason for change
 - We want invalid IDs to not be `null` and just return the ID instead


ashin-czi (approved):
Thanks Alec!!!

tihuan:
Thanks for fixing this!

PR #4387, opened 2023-03-22
chore: Update model paths

MaximilianLombardo:
Updated model paths because of MLFlow fix to address some dependency issues, scanpy/leiden lib version also updated in the docs here


atolopko-czi (approved):
LGTM!

Issue #4347, opened 2023-03-15
chore(wmg): disentangle wmg processing pipeline from database business logic


atarashansky:
In #4337, we updated the pipeline requirements to be primary dependencies only. However, we did not account for code leakage between the WMG pipeline and the database business logic. The pipeline calls into the business logic to fetch the dataset S3 URIs. Consequently, WMG pipeline also requires business logic primary dependencies.

[Slack discussion](https://czi-sci.slack.com/archives/C023Q1APASK/p1678884251153219)

The best way to move forward is to have the WMG processing pipeline fetch the dataset URIs from the Discover API instead of the business logic. Something similar was done for the `fetch_dataset_metadata` function in the WMG API implemented here: #4133.

atolopko-czi:
@atarashansky can close this issue now?

atarashansky:
It was referenced by the PR but not implemented yet. Silly github overcommunication!

atarashansky:
Blocked until discover API changes for retrieving dataset metadata are online.

brianraymor:
@atarashansky - `datasets/index` is in dev if you want to test. CC: @danieljhegeman 

It would be good to get your feedback. Cell Census is also testing in their current sprint.

atarashansky:
@brianraymor Maybe I tested it a little too early today, but I noticed that `s3_uri` is not available in `dataset_assets`, like it is in the dataportal `datasets/index` endpoint. This is something I would need for the processing pipeline to be able to download the h5ad's to the EC2 instance. Do you know if adding that to `dataset_assets` is planned?

brianraymor:
We're not planning to expose that to data consumers. Why wouldn't the WMG builder use precisely the same code as the cell census builder for downloading the H5AD(s)?

atarashansky:
Disclaimer: I haven't looked too closely at how the cell census builder works (e.g. how it's getting s3 uris, retreiving metadata for filtering datasets, etc). What I can say is that WMG will eventually consume the integrated census array directly for building the summary cubes, so refactoring our pipeline and porting over the cell census builder logic now isn't worth it. In the short term, to complete the separation of WMG pipeline logic from database business logic, I will continue to parse the `/dp/v1/datasets/index` endpoint (which does contain s3_uris).

atarashansky:
Ok, looks like I can get the url to the h5ad directly, for example:
[...]

PR #4394, opened 2023-03-23
chore(wmg): disentangle business logic from wmg pipeline

atarashansky:
## Reason for Change

- #4347 
## Changes

- rewrote the `get_dataset_s3_uris` function in `extract.py` to get all necessary s3uri's from the data portal endpoint.
  - EDIT: updated this to use curation endpoint and changed function name to reflect that
- updated the test in `test_extract.py` to create a test fixture that resembles what is returned by the data portal endpoint. this fixture is passed into `get_dataset_s3_uris` to get the test s3 uris.
  - EDIT: updated and simplified this test to test the new function `get_dataset_asset_urls` (which replaces `get_dataset_s3_uris`)
- EDIT: updated `build_dataset_metadata_dict` function in `snapshot.py` to use the curation endpoint.

## Testing steps

- Tests pass.
- Local run of the WMG pipeline yields the same number of datasets as the previous implementation.


backend/wmg/pipeline/integrated_corpus/extract.py
```diff
@@ -37,26 +24,41 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
-            and any(assay.ontology_term_id in INCLUDED_ASSAYS for assay in dataset.metadata.assay)
-            and len(dataset.metadata.organism) < 2
-        ):
-            s3_uri = next(
-                a.uri
-                for a in dataset.artifacts
-                if a.type == DatasetArtifactType.H5AD and a.get_file_name() == "local.h5ad"
-            )
-            s3_uris[dataset.dataset_id.id] = s3_uri
+    if API_URL:
```
atolopko-czi, line 42: error if not defined?

backend/wmg/pipeline/integrated_corpus/extract.py
```diff
@@ -37,26 +24,41 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
+        dataset_metadata_url = f"{API_URL}/dp/v1/datasets/index"
```
atolopko-czi, line 43: Can Discover API be called instead? Thinking that GE as more of an external consumer of the data, rather than a part of the Data Portal web app (which should be the only caller of the DP API).

atolopko-czi: Also some significant changes are coming to the Discover API for retrieving datasets, so we might hold off on the PR until that's available.

atolopko-czi (changes requested):

backend/wmg/pipeline/integrated_corpus/extract.py
atarashansky: No error - an empty dict will be returned instead.

tests/unit/backend/wmg/pipelines/test_extract.py
```diff
@@ -45,37 +41,88 @@ def setUp(self):
...
+
+    def _convert_list_of_dataclass_to_list_of_dicts(self, list_of_dataclasses):
+        """
+        Convert a list of dataclasses to a list of dicts
+        """
+        if list_of_dataclasses is None:
+            return None
+        return [asdict(a) if a else a for a in list_of_dataclasses]
+
+    def convert_business_logic_data_to_endpoint_dict_format(self, dataset):
```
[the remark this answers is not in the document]
atarashansky: We have E2E tests for the source data, which will capture breaking changes to the data portal datasets/index endpoint.

backend/wmg/pipeline/integrated_corpus/extract.py
atarashansky: It appears that the curation API datasets endpoint does not provide the `s3_uri` information. This is required for the pipeline, so I think we need to stick with the `datasets/index` endpoint.

backend/wmg/pipeline/integrated_corpus/extract.py
```diff
@@ -37,33 +24,45 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
+                asset_url = next(a["url"] for a in dataset["assets"] if a["filetype"] == "H5AD")
+                asset_urls[dataset_id] = asset_url
+    return asset_urls
 
 
 def copy_datasets_to_instance(s3_uris: Dict, dataset_directory: str):
     """Copy given list of s3 uris to the provided path"""
     for dataset in s3_uris:
-        copy_command = ["aws", "s3", "cp", s3_uris[dataset], f"./{dataset_directory}/{dataset}/local.h5ad"]
+        copy_command = ["wget", s3_uris[dataset], "-O", f"./{dataset_directory}/{dataset}/local.h5ad"]
```
atarashansky, line 65: Question for reviewers - is wget okay to use? Any disadvantages here? Curl preferred? Why/why not?

brianraymor: https://daniel.haxx.se/docs/curl-vs-wget.html

backend/wmg/pipeline/integrated_corpus/extract.py
```diff
@@ -37,33 +24,45 @@ def get_X_raw(anndata_object: anndata.AnnData) -> Union[np.ndarray, sparse.spmat
...
+        "https://api.cellxgene.dev.single-cell.czi.technology"
+        if os.environ.get("DEPLOYMENT_STAGE") in ["test", "rdev"]
+        else os.getenv("API_URL")
+    )
+
+    asset_urls = dict()
+    if API_URL:
+        dataset_metadata_url = f"{API_URL}/curation/v1/datasets"
+        if datasets is None:
+            datasets = requests.get(dataset_metadata_url).json()
```
ebezzi, line 45: Almost a nit, but since this endpoint is quite heavyweight and likely not cached, we could add a `Retry` here.

ebezzi (approved):

Issue #4406, opened 2023-03-23
bug(WMG): missing test coverage


Bento007:
add test coverage to prevent https://github.com/chanzuckerberg/single-cell-data-portal/pull/4405 from happening in the future

PR #4414, opened 2023-03-24
chore(wmg): add unit test for no gene query edge case

atarashansky:
## Reason for Change

- #4406

## Changes

- added api tests to cover the edge case where no genes are selected.
  - one for 1 tissue
  - one for multiple tissues
- added an e2e test to cover the same edge case from the FE perspective

## Testing steps

- Tests pass

## Notes for reviewer
I confirmed that these tests fail as expected when the bug fixed by #4405 is introduced.

### DIFF
--- a/tests/unit/backend/wmg/api/test_v1.py
+++ b/tests/unit/backend/wmg/api/test_v1.py
@@ -60,31 +60,33 @@ def generate_expected_term_id_labels_dictionary(genes, tissues, cell_types, tota
     result["genes"] = []
     for gene in genes:
         result["genes"].append({gene: f"{gene}_label"})
+
     return result
 
 
 def generate_expected_expression_summary_dictionary(genes, tissues, cell_types, n, me, pc, tpc, compare_terms=None):
     result = {}
     for gene in genes:
-        result[gene] = {}
-        for tissue in tissues:
-            result[gene][tissue] = {}
-            for cell_type in cell_types:
-                result[gene][tissue][cell_type] = {}
-                result[gene][tissue][cell_type]["aggregated"] = {
-                    "n": n,
-                    "me": me,
-                    "pc": pc,
-                    "tpc": tpc,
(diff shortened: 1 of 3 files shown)


tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -60,31 +60,33 @@ def generate_expected_term_id_labels_dictionary(genes, tissues, cell_types, tota
...
-                }
-                if compare_terms:
-                    for term in compare_terms:
-                        result[gene][tissue][cell_type][term] = {
-                            "n": n // len(compare_terms),
-                            "me": me,
-                            "pc": pc,
-                            "tpc": tpc / len(compare_terms),
-                        }
+        if gene != ".":
```
atarashansky, line 70: by (a hacky) convention, gene will be "." when the FE is asking for cell type info with no genes selected.

tests/unit/backend/wmg/api/test_v1.py
```diff
@@ -251,6 +253,94 @@ def test__query_single_primary_dims__returns_200_and_correct_response(
...
+
+            # mock the functions in the ontology_labels module, so we can assert deterministic values in the
+            # "term_id_labels" portion of the response body; note that the correct behavior of the ontology_labels
+            # module is separately unit tested, and here we just want to verify the response building logic is correct.
+            ontology_term_label.side_effect = lambda ontology_term_id: f"{ontology_term_id}_label"
+            gene_term_label.side_effect = lambda gene_term_id: f"{gene_term_id}_label"
+
+            # this is the convention used by the FE to indicate no genes are selected
+            # see https://github.com/chanzuckerberg/single-cell-data-portal/pull/2729
+            genes = ["."]
```
atarashansky, line 280: as stated above, genes will be ["."] by convention when FE is querying with no genes selected.

PR #4415, opened 2023-03-24
fix(WMG): no genes selected should not return empty response (#4405)

atarashansky:
## Reason for Change

- This cherry picks fix #4405 into staging

## Notes for reviewer
 - This fix was tested in dev and confirmed by newly added unit tests in #4414


atarashansky:
hotfix already applied and tests can make it in the following deployment cycle.

PR #4427, opened 2023-03-24
fix: enable logging for processing jobs

Bento007:
## Reason for Change
No logs are emitted by the processing container. This makes debugging difficult.

## Changes

- Add functions to log when a functions started and stopped
- Enable logging in processing container.

## Testing steps
- Tested using rdev to make sure logs were emitted.


atarashansky (approved):
thanks for doing this! lgtm

PR #4437, opened 2023-03-27
chore(wmg): prevent majority of word-wrap by expanding right column of tooltip

atarashansky:
## Reason for Change

- #4213

## Changes

- shrink left column of tooltip to 30% width
- make left column text not wrap
- vertical alignment baseline so word-wrap is easier to read

## Testing steps

- Tested locally
- tested on the longest cell type name in the corpus (below screenshot)
image


ashin-czi (approved):
LGTM! TY!

Issue #4363, opened 2023-03-17
Remove Fallthrough Canonical/Version ID Lookup from non-version path dataset/collection endpoints


brianraymor:
@nayib-jose-gloria wrote:

Only do this after version endpoints have passed QA 

update all endpoints that currently do fallthrough lookup to only return responses on canonical IDs OR, for collections, revision IDs. We want to do this for every 'collection' and 'dataset' tagged endpoint in the swagger 

Return 403 Forbidden on attempted version_id fetches for initial unpublished collections, published collections, and datasets

PR #4441, opened 2023-03-27
feat: remove support for version_ids (unless unpublished revision IDs) in non-versioned Discover API endpoints

nayib-jose-gloria:
## Reason for Change

- #4363

## Changes

- remove support for version_ids (unless unpublished revision IDs) in all non-versioned Discover API endpoints
- added and updated tests
- some minor refactors in a few functions for simplicity / DRYness of code

## Testing steps
- unit tests
- functional tests


backend/curation/api/v1/curation/collections/common.py
```diff
@@ -288,37 +288,38 @@ def get_collection_level_processing_status(datasets: List[DatasetVersion]) -> st
     return return_status
 
 
-def get_inferred_collection_version_else_forbidden(collection_id: str) -> CollectionVersionWithDatasets:
+def get_inferred_collection_version(collection_id: str) -> CollectionVersionWithDatasets:
```
danieljhegeman, line 291: the name of this method should retain `_else_forbidden` suffix since it still throws `ForbiddenHTTPException`

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -171,10 +180,12 @@ def _test(token, is_super_curator: bool = False):
...
                 }
             )
-            unpublished_collection = self.generate_unpublished_collection()
             headers = {"Authorization": f"Bearer {token}"}
 
-            for id in [unpublished_collection.collection_id, unpublished_collection.version_id]:
+            with self.subTest(f"{token}, unpublished collection"):
+                unpublished_collection = self.generate_unpublished_collection()
+
+                id = unpublished_collection.collection_id
```
danieljhegeman, line 188: `id` shadows the built-in `id`. Prefer `_id` or `collection_id` as var name

backend/curation/api/v1/curation/collections/common.py
nayib-jose-gloria: It can also throw NotFoundHTTPException though

nayib-jose-gloria: unless we want to throw Forbidden for all use cases? But we've been using NotFound in other endpoints

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -934,19 +975,25 @@ def test_get_colletion_with_dataset_no_metadata(self):
     def _test_response(self, collection_version: CollectionVersion, status_code=200, auth=True) -> dict:
         version_id = collection_version.version_id
         collection_id = collection_version.collection_id
+        is_open_revision = (
+            collection_version.published_at is None
+            and collection_version.canonical_collection.originally_published_at is not None
```
danieljhegeman, line 980: why would you reuire the canonical entity to have a `originally_published_at` value of `None` in order to say it is an open revision? This seems incorrect. Am I missing a hack?

nayib-jose-gloria: it checks for `is not None`

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -934,19 +975,25 @@ def test_get_colletion_with_dataset_no_metadata(self):
...
+        )
         headers = self.make_owner_header() if auth else self.make_not_owner_header()
         res_1 = self.app.get(f"/curation/v1/collections/{version_id}", headers=headers)
-        self.assertEqual(status_code, res_1.status_code)
-        if status_code == 200:
-            self.assertEqual(collection_version.collection_id.id, res_1.json["collection_id"])
+        if is_open_revision:
+            self.assertEqual(status_code, res_1.status_code)
+            if status_code == 200:
+                self.assertEqual(collection_version.collection_id.id, res_1.json["collection_id"])
```
danieljhegeman, line 987: @nayib-jose-gloria I don't think this is correct. I must have missed this in the code. We should be returning the version id in this field for revisions.

danieljhegeman: From the user's perspective, a revision id is a `collection_id`.

danieljhegeman: it would simply be the same value. We had a discussion previously that implied this, I think. @brianraymor confirm?

backend/curation/api/v1/curation/collections/common.py
danieljhegeman: ah fair point. Ok, fine to remove if we aren't being consistent.

danieljhegeman (approved):
🎸

PR #4443, opened 2023-03-28
feat: add published_at and collection_version_id to Dataset [version] response object

danieljhegeman:
## Reason for Change

- #4324 

## Changes

- Add `published_at` and `collection_version_id` per discussion on the issue for this PR
- Consolidate shared logic
- minor cleanup of unrelated endpoints and tests

## Testing steps

- unit tested ✅ 
- rdev tested ✅

### DIFF
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -179,10 +179,6 @@ def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canoni
         ds["dataset_assets"] = assets
         ds["processing_status_detail"] = dataset_version.status.validation_message
         _published_at = dataset_version.canonical_dataset.published_at
-        if _published_at and _published_at < dataset_version.created_at:
-            ds["revised_at"] = dataset_version.created_at
-        else:
-            ds["revised_at"] = None
         ds["title"] = ds.pop("name", None)
         ds["explorer_url"] = generate_explorer_url(dataset_version, use_canonical_url)
         ds["tombstone"] = False  # TODO this will always be false. Remove in the future
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -212,10 +212,7 @@ def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canoni
     dataset = reshape_dataset_for_curation_api_common(dataset_version, use_canonical_url, preview)
     if not preview:
         dataset["published_at"] = dataset_version.canonical_dataset.published_at
-        print("\n\n\n\n\n")
-        print(dataset["published_at"])
         dataset["revised_at"] = dataset_version.canonical_dataset.revised_at
-        print(dataset["revised_at"])
     return dataset
 
 
@@ -328,16 +325,10 @@ def get_inferred_collection_version_else_forbidden(collection_id: str) -> Collec
     """
     validate_uuid_else_forbidden(collection_id)
     version = get_business_logic().get_published_collection_version(CollectionId(collection_id))
-    print("0000")
-    # print(f"\n\n{version.datasets[0].canonical_dataset.revised_at}\n")
     if version is None:
         version = get_business_logic().get_collection_version(CollectionVersionId(collection_id))
-        print("1111")
-        # print(f"\n\n{version.datasets[0].canonical_dataset.revised_at}\n")
     if version is None:
         version = get_business_logic().get_unpublished_collection_version_from_canonical(CollectionId(collection_id))
-        print("2222")
-        # print(f"\n\n{version.datasets[0].canonical_dataset.revised_at}\n")
     if version is None or version.canonical_collection.tombstoned is True:
         raise ForbiddenHTTPException()
     return version
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -212,8 +212,10 @@ def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canoni
     dataset = reshape_dataset_for_curation_api_common(dataset_version, use_canonical_url, preview)
     if not preview:
         dataset["published_at"] = dataset_version.canonical_dataset.published_at
-        if isinstance(dataset_version, PublishedDatasetVersion):
-            dataset["revised_at"] = dataset_version.published_at
+        print("\n\n\n\n\n")
+        print(dataset["published_at"])
+        dataset["revised_at"] = dataset_version.canonical_dataset.revised_at
+        print(dataset["revised_at"])
     return dataset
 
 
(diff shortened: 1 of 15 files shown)


backend/layers/persistence/persistence.py
```diff
@@ -508,36 +507,11 @@ def get_all_mapped_datasets_and_collections(self) -> Tuple[List[DatasetVersion],
             dataset_version_ids.extend(collection.datasets)
         return list(self._get_datasets(dataset_version_ids)), active_collections
 
-    def get_all_mapped_collection_versions_with_datasets(self) -> List[CollectionVersionWithDatasets]:
+    def get_all_mapped_collection_versions_with_datasets(self) -> Tuple[List[DatasetVersion], List[CollectionVersion]]:
```
nayib-jose-gloria, line 510: If this just calls a different function in this class, lets delete it

nayib-jose-gloria: and replace its usages with the call its invoking ofc

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -198,6 +208,20 @@ def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canoni
...
 
+def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canonical_url: bool, preview=False) -> dict:
+    dataset = reshape_dataset_for_curation_api_common(dataset_version, use_canonical_url, preview)
+    if not preview:
+        dataset["published_at"] = dataset_version.canonical_dataset.published_at
+        dataset["revised_at"] = dataset_version.canonical_dataset.revised_at
+    return dataset
+
+
+def reshape_dataset_for_curation_api_as_version(
```
nayib-jose-gloria, line 219: I don't really see the value of having this separate util function if it just calls another function with the same args

nayib-jose-gloria: If the only diff is the IF statement in reshape_dataset_for_curation_api, I'd rather just consolidate all of it into the previous singular  reshape_dataset_for_curation_api function and handle the conditional in there

nayib-jose-gloria: all of it == all 3 of reshape_dataset_for_curation_api_as_version, reshape_dataset_for_curation_api, and reshape_dataset_for_curation_api_common

nayib-jose-gloria:
> 

I believe you need to revise this logic https://github.com/chanzuckerberg/single-cell-data-portal/blob/0c2a4b80e4cc525961f79232a84a9b222179136e/backend/curation/api/v1/curation/collections/common.py#L183-L187

backend/layers/business/business.py
```diff
@@ -83,6 +86,32 @@ def __init__(
         self.uri_provider = uri_provider
         super().__init__()
 
+    @staticmethod
```
nayib-jose-gloria, line 89: nit: doesn't really need to be a static method, technically every method in this class could be, but we should stick to the established pattern

backend/layers/business/business.py
```diff
@@ -83,6 +86,32 @@ def __init__(
...
+    def _get_published_at_and_collection_version_id_else_not_found(
```
nayib-jose-gloria, line 90: also, might be worth renaming to be explicit that this is for fetching a dataset version's published_at

nayib-jose-gloria: since published_at and collection_version_id may, at first glance, suggest this is about collection version fields

nayib-jose-gloria:
Missing Curation API spec comments

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -198,6 +208,20 @@ def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canoni
     return ds
 
 
+def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canonical_url: bool, preview=False) -> dict:
+    dataset = reshape_dataset_for_curation_api_common(dataset_version, use_canonical_url, preview)
+    if not preview:
+        dataset["published_at"] = dataset_version.canonical_dataset.published_at
+        dataset["revised_at"] = dataset_version.canonical_dataset.revised_at
```
nayib-jose-gloria, line 215: I might be missing it, but are we setting the dataset_version.canonical_dataset revised_at anywhere? If we're not, I think this is why the test is failing with an unexpected None

danieljhegeman: You are correct!

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -848,7 +854,7 @@ def _test_responses_are_equal(test_ids, expected_id, auth_headers) -> dict:
             self.assertIsNotNone(resp_collection["published_at"])
             self.assertTrue(resp_collection["collection_url"].endswith(revision.version_id.id))
             self.assertEqual(revision.version_id.id, resp_collection["collection_version_id"])
-            self.assertIsNotNone(resp_collection["datasets"][0]["revised_at"])
+            self.assertIsNone(resp_collection["datasets"][0]["revised_at"])
```
danieljhegeman, line 857: this was incorrect service behavior before

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -208,20 +202,16 @@ def reshape_dataset_for_curation_api_common(
     return ds
 
 
-def reshape_dataset_for_curation_api(dataset_version: DatasetVersion, use_canonical_url: bool, preview=False) -> dict:
-    dataset = reshape_dataset_for_curation_api_common(dataset_version, use_canonical_url, preview)
+def reshape_dataset_for_curation_api_as_canonical(
```
nayib-jose-gloria, line 205: nit: suggestion but not a required change--I think it would be more consistent with the pattern we've been using in other functions in this class if we turned 'as_canonical' into an optional flag in `reshape_dataset_for_curation_api` (ala `as_version` in the `reshape_collection_for_curation_api` function) and added the below published_at & revised_at logic there, instead of the separate function

backend/layers/persistence/persistence.py
```diff
@@ -257,6 +258,25 @@ def _get_datasets(self, ids: List[DatasetVersionId]) -> List[DatasetVersion]:
                 datasets.append(self._row_to_dataset_version(version, canonical_dataset, version_artifacts))
         return datasets
 
+    @staticmethod
+    def _set_revised_at_field(
```
nayib-jose-gloria, line 262: this belongs in the helpers class

backend/layers/persistence/persistence.py
```diff
@@ -257,6 +258,25 @@ def _get_datasets(self, ids: List[DatasetVersionId]) -> List[DatasetVersion]:
...
+    ) -> None:
+        """
+        Sets the `revised_at` field on the CanonicalDataset object for each DatasetVersion object
+        """
+        for dataset_version in dataset_versions:
+            try:
+                version_published_at, collection_version_id = get_published_at_and_collection_version_id_else_not_found(
+                    dataset_version, collection_versions
+                )
+                if version_published_at > dataset_version.canonical_dataset.published_at:
```
nayib-jose-gloria, line 273: We should add None checks here, to avoid an Exception on dataset_version.canonical_dataset.published_at is None or version_published_at is None

backend/curation/api/v1/curation/collections/common.py
danieljhegeman: done

backend/layers/persistence/persistence.py
danieljhegeman: the function that returns `version_published_at` can't return `None` for it, and in the case where a Dataset has never been published before, i.e. `dataset_version.canonical_dataset.published_at` is `None`, then a `DatasetVersionNotFoundException` is thrown, and in this case it is caught because the calling function wishes to simply skip over such cases.

nayib-jose-gloria (approved):

Issue #4315, opened 2023-03-08
Gene Expression long cell type names are not readable


signechambers1:
Long ell type names are truncated and there is no way to read the full name. Ideally on hover you can read the whole name.

Addtl ask from Pablo that cell type names are findable using Ctrl-F.

rdev: https://cell-type-names-tooltip-frontend.rdev.single-cell.czi.technology/gene-expression

atarashansky:
Breadcrumbs from @seve, look how truncation is done in explorer.

ashin-czi:
@signechambers1 Here's what I came up with so far

https://user-images.githubusercontent.com/109984998/228116414-aea0e13c-ab54-4bf0-b75e-9e618101aa16.mov

Only for truncated cell type labels, a tooltip will appear when hovered revealing the full name to the left of the label

Also the full names should be findable using the browser's Ctrl+F/Cmd+F hotkey

This can be played with in rdev here: https://cell-type-names-tooltip-frontend.rdev.single-cell.czi.technology/gene-expression

signechambers1:
Thank you @ashin-czi ! The functionality looks great, I'm excited about this. A few questions:

1. when CTRL+F the string im searching for becomes visible in the cell type but cuts off another word. Is there a way to add a "..." after the search term?![Screenshot 2023-03-28 at 9.48.08 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/76f4f72d-957c-4bc3-a39d-f7db7686c887) 

2. The hover over looks like there is too much padding on the left and right for the centered text. Can we left align the text and reduce the left right padding? cc @karenliang0 if there is another small tweak you'd suggest instead.![Screenshot 2023-03-28 at 9.51.36 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/0ee971b5-bd33-44e2-8459-1fa096651587)

ashin-czi:
@signechambers1 

1. Unfortunately the highlighting is browser dependent so I can't change the behavior of the highlighting. @tihuan Any ideas off the top of your head for a workaround?
2. I fixed the padding issue

signechambers1:
Thanks! @ashin-czi don't worry about #1, let's go ahead and merge in. Thanks for the info!

ashin-czi:
One drawback with this approach is that since the full name is "hidden" only until you do a search, you will sometimes see something like this:

[...]

PR #4444, opened 2023-03-28
fix(wmg): Cell Type Name Label Tooltip

ashin-czi:
## Reason for Change

- #4315

## Changes

- add
    - Tooltip SDS component for cell type names
- remove
- modify

## Testing steps

- Select Tissue
- Hover over a truncated label
- A tooltip should appear
- Copy tooltip contents
- Ctrl+F to and paste full cell type name
- Full cell type name should be highlighted on page

Rdev: https://cell-type-names-tooltip-frontend.rdev.single-cell.czi.technology/gene-expression

Demo:

https://user-images.githubusercontent.com/109984998/228119256-5a363bec-f022-456c-89f7-12d6b385a10e.mov

### DIFF
--- a/frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
@@ -12,9 +12,11 @@ import InfoSVG from "./icons/info-sign-icon.svg";
 import {
   CellCountLabelStyle,
   CellTypeLabelStyle,
+  CellTypeLabelTooltipStyle,
   Container,
   FlexRow,
   FlexRowJustified,
+  HiddenCellTypeLabelStyle,
   StyledImage,
   TissueName,
   TissueWrapper,
@@ -26,6 +28,7 @@ import { EVENTS } from "src/common/analytics/events";
 import { EXCLUDE_IN_SCREENSHOT_CLASS_NAME } from "../../../GeneSearchBar/components/SaveExport";
 import { COMPARE_OPTION_ID_FOR_AGGREGATED } from "src/common/queries/wheresMyGene";
 import { InfoButtonWrapper } from "src/components/common/Filter/common/style";
+import { Tooltip } from "czifui";
 
 interface Props {
   cellTypes?: CellType[];
@@ -86,7 +89,8 @@ export default memo(function YAxisChart({
             return (
               <CellTypeButton
                 key={`${cellType}-cell-type-button`}
-                name={paddedName}
+                formattedName={paddedName}
+                name={name}
                 metadata={cellType}
                 tissueID={tissueID}
                 tissue={tissue}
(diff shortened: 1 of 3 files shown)


frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -125,8 +131,25 @@ const CellTypeButton = ({
...
+            title={
+              // Set tooltip content only if name is truncated
+              formattedName.includes("...") ? (
+                <CellTypeLabelTooltipStyle>{name}</CellTypeLabelTooltipStyle>
+              ) : null
+            }
+            sdsStyle="light"
+            arrow
+            placement="left"
+            enterNextDelay={700}
```
atarashansky, line 145: Out of curiosity, what is the default value?

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -125,8 +131,25 @@ const CellTypeButton = ({
...
+          >
+            {/* Must be wrapped in div and not fragment or else tooltip content won't render */}
+            <div>
+              <HiddenCellTypeLabelStyle>{name}</HiddenCellTypeLabelStyle>
+              <div data-test-id="cell-type-name">{formattedName}</div>
```
atarashansky, line 150: Can we put this `data-test-id` on `HiddenCellTypeLabelStyle`? I can imagine a situation where we use this ID to select a cell type with a specific name. In that case, we want to use the full (not truncated) name.

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/style.ts
```diff
@@ -53,6 +53,19 @@ export const CellTypeLabelStyle = styled.div`
   text-align: left;
 `;
 
+export const HiddenCellTypeLabelStyle = styled.div`
+  /* Overlay invisible, un-intractable element with full name of cell type for ctrl+f page search */
```
atarashansky, line 57: Really smart solution!

atarashansky (approved):
Looks good, very nice solution :) - approving preemptively. Just one change I'd like before merging (the `data-test-id` comment).

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
atarashansky: Alternatively, we could update `data-test-id` to be `cell-type-name-${name}`, but the e2e test selector for all cell types will need to be updated (e.g. like what [this line](https://github.com/chanzuckerberg/single-cell-data-portal/blob/25b4236e7d466f25a0a91cc4c024737fcc76dc7a/frontend/tests/features/wheresMyGene.test.ts#L14) is doing. So, something like `const CELL_TYPE_LABELS_ID = "[data-test-id^=cell-type-name-]";`

frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx
ashin-czi: If `enterNextDelay` isn't set then it just immediately spams the tooltip when you hover over the labels leaving tooltips overlapping each other until they disappear

The default for `enterNextDelay` is 0

Issue #2647, opened 2022-06-07
Curators can start a revision from the Collection page


jahilton:
## Need
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I am looking at a Collection, identify an update to make in the Title, Description, Links or Contacts (non-dataset information). To make the update, I need to copy the title, go to My Collections, then search for the title, then Start Revision.

## Approach

**Describe the solution you'd like**
I would like to be able to click a button on the Collection page to Start Revision. I could be redirected to the new Revision page, or a new tab could open to the Revision page - either would be OK.

## Design

[Start revision from Collection Detail page](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=4461%3A39456&t=gVeFYp4JnUWIrvke-1)
[Public collection, pending revision banner](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=4531%3A39402&t=gVeFYp4JnUWIrvke-1)
[Private revision, public collection banner](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=4532%3A39739&t=gVeFYp4JnUWIrvke-1)

brianraymor:
@metakuni @NoopDog - I'm assuming that this epic is complete, but the on-call failed to close its _Ready for Prod_ issues?

PR #4455, opened 2023-03-29
feat: Curator can start a revision from the Collection page (#2647)

NoopDog:
## Reason for Change

- #2647

## Backend Changes

- added a `dp/v1/my-collections/index` API
- modified the `portal-api.yml`, `portal_api.py` and `test_portal_api.py`
- this API:
    - is authenticated
    - returns a superset of fields returned by  `/dp/v1/collections/index` 
    - returns the same rows as `/dp/v1/collections` that is, all public collections and 1) for curators, all private collections they created (are the owner of), and 2 ) for super curators all private collections. 

### Notes for BE reviewer

There is a bit of code duplication now between the `/dp/v1/collections` and `/dp/v1/my-collections/index implementations`. I did not address this as the `/dp/v1/collections` implementation will be deleted once we are live with the updated "My Collections" experience.

## Front End Changes
WIP (not yet in this PR)

### DIFF
--- a/backend/portal/api/portal_api.py
+++ b/backend/portal/api/portal_api.py
@@ -443,10 +443,10 @@ def get_collection_index():
 
     return make_response(jsonify(response), 200)
 
-def get_user_writable_collections(token_info: dict):
 
+def get_user_writable_collections(token_info: dict):
     user_info = UserInfo(token_info)
-     # Get all public collections
+    # Get all public collections
     collections = get_business_logic().get_collections(CollectionQueryFilter(is_published=True))
 
     # Get all private collections the user has access to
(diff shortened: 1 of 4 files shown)


backend/portal/api/portal_api.py
```diff
@@ -433,6 +433,65 @@ def get_collection_index():
...
+            transformed_collection["publisher_metadata"] = _publisher_metadata_to_response(
+                collection.publisher_metadata
+            )
+
+        if collection.is_unpublished_version():
+            transformed_collection["id"] = collection.version_id.id
+        else:
+            transformed_collection["id"] = collection.collection_id.id
+
+        if not collection.is_published():
```
ebezzi, line 478: I think this code already exists somewhere. `_collection_to_response`?

ebezzi (commented):
Looks good overall. Before approval, I'd recommend:
1. Checking for duplicate code/logic. Most of the `transformed_collection` checks should already exist elsewhere in the code. If they need to be different, add a comment that states why.
2. If this deprecates `get_collections_list`, let's remove it, or at least mark it as deprecated.

Issue #4458, opened 2023-03-29
Update test Ids


charles-testco:
Currently we use the format data-test-id="menu" for test Ids.

We want to change this to:
data-testid="menu"

For several reasons:
1. It is Playwright default format
2. Will allow us to Playwright command out of the box: getByTestId without writing our own wrapper
3. Will fit with some Ids from external libraries

tihuan:
Done in https://github.com/chanzuckerberg/single-cell-data-portal/pull/4432

PR #4432, opened 2023-03-27
chore(wmg): add fmg e2e tests and refactor to use locators

atarashansky:
## Reason for Change

- #3550 

## Changes

- added FMG e2e tests to test the following cases
  - Marker genes display and the add-to-dotplot button works
  - Native cell displays no markers genes
  - Marker scores are all > 0.5
- big refactor to use locators instead of `$`/`$$` invocations.

## Testing steps

- the new e2e tests were tested locally.
- all tests pass locally

PR #4459, opened 2023-03-29
chore: update test id from `data-test-id` to `data-testid` to match default playwright configuration

charles-testco:
## Reason for Change

Currently we use the format data-test-id="menu" for test Ids.

We want to change this to:
data-testid="menu"

For several reasons:

- It is Playwright default format
- Will allow us to use Playwright command out of the box: getByTestId without writing our own wrapper
- Will fit with some Ids from external libraries

- #4458

## Changes

- modify
- data-test-id to data-testid

## Testing steps
Ensure all tests pass

### DIFF
--- a/.github/ISSUE_TEMPLATE/analytics-events-update.md
+++ b/.github/ISSUE_TEMPLATE/analytics-events-update.md
@@ -1,16 +1,15 @@
 ---
 name: Analytics events update
-about: Used to submit requests for updates to current events in Zenhub
+about: Used to submit requests for updates to current analytic events
 title: "Analytics (update): <fill in description> "
-labels: analytics, frontend
+labels: analytics, frontend, tech
 assignees: ""
 ---
 
 ### Overview
 
-**Tool**: `<Which tool are these events for, i.e. scExpression, explorer>`
-
-- Feature (Optional): `<Which feature in that tool are these events for, i.e. Filters>`
+**Tool**: `<Which tool are these events for, i.e. gene expression, explorer>`
+Feature: `<Which feature in that tool are these events for, i.e. filters>`
 
 **Description**: `<Reason for update>`
 
@@ -20,6 +19,13 @@ assignees: ""
 
 - Current: `<Description of how the event works currently>`
 - Requested change: `<Description of the update needed>`
+
+Checklist
+
+- [ ] Once event updates are deployed to Staging ping @ainfeld to check analytics
+- [ ] Verify events are showing up in the Staging space in Plausible correctly
+- [ ] Once event updates are deployed to Prod ping @ainfeld
+- [ ] Confirm Plausible events are still showing up in the Prod space in Plausible correctly
 
 #### People to cc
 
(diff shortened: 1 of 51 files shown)


atarashansky:
Hi - this PR can be closed. I have an open PR refactoring our unit tests where I set the default test attribute to be `data-test-id` in the playwright config: https://github.com/chanzuckerberg/single-cell-data-portal/pull/4432/files#diff-1ed7245e0c7c634928b52839600c3d493c9d29150d369e9ab20979e61e7f8f97R56

If you need this to unblock your work, then please add the linked line to your config.

tihuan:
I had set up `data-test-id` pattern way before Playwright went with `data-testid` as default (and still a little surprised that they went with that format, since that's not very HTML attribute looking format lol) 

But if switching to `data-testid` means one less customization in our system, I think it's worth doing it!

I was thinking the same thing Alec pointed out about us adding our own test id instead of relying on unofficial API third party libraries uses, but in this case it's unlikely existing libraries will change from `data-testid` to `data-test-id`, so I'm also fine relying on them if it means we get to write a little less code in our production code

tihuan:
> > I had set up `data-test-id` pattern way before Playwright went with `data-testid` as default (and still a little surprised that they went with that format, since that's not very HTML attribute looking format lol)
> > But if switching to `data-testid` means one less customization in our system, I think it's worth doing it!
> > I was thinking the same thing Alec pointed out about us adding our own test id instead of relying on unofficial API third party libraries uses, but in this case it's unlikely existing libraries will change from `data-testid` to `data-test-id`, so I'm also fine relying on them if it means we get to write a little less code in our production code
> 
> Sounds good to me. I'll go ahead and make the necessary changes to my branch to dodge merge conflicts.

Thanks so much, Alec 🎉 !!

ashin-czi:
frontend/src/views/WheresMyGene/components/HeatMap/components/YAxisChart/index.tsx

has a typo! 
https://github.com/chanzuckerberg/single-cell-data-portal/pull/4459/files#diff-e1359d72040323eb2beb360f986f33ac1914636781618ddf1aa744b92b02582cR98

charles-testco:
> I was thinking the same thing Alec pointed out about us adding our own test id instead of relying on unofficial API third party libraries uses, but in this case it's unlikely existing libraries will change from `data-testid` to `data-test-id`, so I'm also fine relying on them if it means we get to write a little less code in our production code

It is butter vs magarine war. data-test-id was the convention and cypress recommends/sticks. But I notice that majority or react components/libraries use data-testid and that probably is why Playwright went that road. An example is the "We would appreciate your feedback, please fill out a" alert render

atarashansky:
Closing this PR as these changes have been merged in #4432. Thanks for raising this, @charles-testco.

tihuan:
> > I was thinking the same thing Alec pointed out about us adding our own test id instead of relying on unofficial API third party libraries uses, but in this case it's unlikely existing libraries will change from `data-testid` to `data-test-id`, so I'm also fine relying on them if it means we get to write a little less code in our production code
> 
> It is butter vs magarine war. data-test-id was the convention and cypress recommends/sticks. But I notice that majority or react components/libraries use data-testid and that probably is why Playwright went that road. An example is the "We would appreciate your feedback, please fill out a" alert render

Love the analogy 😂 Yeah so interesting how the community evolves, but a small change for us so that's all good!

Thanks again for bringing this up, Charles!

Issue #4472, opened 2023-03-31
bug(cxg conversion): tiledb parameters out of date.


Bento007:
## Describe the bug
Invalid configuration parameters are passed to tildb.ctx.
https://github.com/chanzuckerberg/single-cell-data-portal/blob/8fc25a0eb52f8d7cc38071f9666327ba00c1c600/backend/layers/processing/h5ad_data_file.py#L61-L67

```
2023-03-31T17:58:32.319ZCopy[2023-03-31 17:58:32.319] [Process: 1] [error] [Context: 1] Config parameter "sm.num_reader_threads" has been removed; use config parameter "sm.compute_concurrency_level". | [2023-03-31 17:58:32.319] [Process: 1] [error] [Context: 1] Config parameter "sm.num_reader_threads" has been removed; use config parameter "sm.compute_concurrency_level".
-- | --
  | 2023-03-31T17:58:32.319ZCopy[2023-03-31 17:58:32.319] [Process: 1] [error] [Context: 1] Config parameter "sm.num_writer_threads" has been removed; use config parameter "sm.compute_concurrency_level". | [2023-03-31 17:58:32.319] [Process: 1] [error] [Context: 1] Config parameter "sm.num_writer_threads" has been removed; use config parameter "sm.compute_concurrency_level".
  | 2023-03-31T17:58:32.321ZCopy[2023-03-31 17:58:32.321] [Process: 1] [error] [Context: 1] Config parameter "sm.num_reader_threads" has been removed; use config parameter "sm.compute_concurrency_level". | [2023-03-31 17:58:32.321] [Process: 1] [error] [Context: 1] Config parameter "sm.num_reader_threads" has been removed; use config parameter "sm.compute_concurrency_level".
  | 2023-03-31T17:58:32.321ZCopy[2023-03-31 17:58:32.321] [Process: 1] [error] [Context: 1] Config parameter "sm.num_writer_threads" has been removed; use config parameter "sm.compute_concurrency_level". | [2023-03-31 17:58:32.321] [Process: 1] [error] [Context: 1] Config parameter "sm.num_writer_threads" has been removed; use config parameter "sm.compute_concurrency_level".
```
## To Reproduce

Processes an h5ad file with labels through the CXG processing container.

## Expected behavior

The errors should not be showing, and the configuration parameters should be updated.

## Environment

[...]

PR #4473, opened 2023-03-31
chore: remove deprecated tiledb configuration parameters

Bento007:
## Reason for Change

- #4472

## Changes

- Remove depreciated tiledb configuration parameters
- The tiledb default is the number of cores.  See ["sm.compute_concurrency_level"](https://docs.tiledb.com/main/how-to/configuration#configuration-parameters).

## Testing steps

- running on a datasets through cxg locally.


backend/layers/processing/h5ad_data_file.py
```diff
@@ -60,8 +60,6 @@ def to_cxg(self, output_cxg_directory, sparse_threshold, convert_anndata_colors_
         logging.info("Beginning writing to CXG.")
         ctx = tiledb.Ctx(
             {
-                "sm.num_reader_threads": 32,
-                "sm.num_writer_threads": 32,
```
atarashansky, line 64: Can we replace these with `sm.compute_concurrency_level`?

atarashansky (approved):

PR #4471, opened 2023-03-31
fix(wmg): Homepage not scrollable

ashin-czi:
## Reason for Change

- #TICKET_NUMBER
- If the reason for this PR's code changes are not clear in the issue, state value/impact

## Changes

- add
- remove
- modify

## Testing steps

- Either list QA steps or reasoning you feel QA is unnecessary
- Describe how you made sure to know that your changes worked. Should allow someone else to go verify your code without in depth knowledge.
- "Unit tested only", "tested in rdev by a, b, c, verifying feature worked by... ", "manually ran pipeline locally with these results: ..."

PR #4474, opened 2023-03-31
fix(wmg): Homepage not scrollable (#4471)

seve:
* fix(wmg): Homepage not scrollable - setting overflow-y to auto

back merging hotfix


ashin-czi (approved):
LGTM! TYSM!!

Issue #4460, opened 2023-03-29
Update CSV download headers to enable smooth loading into python and R


signechambers1:
Ambrose found some problems when uploading the csv download using the pandas package. We'd like this to be smooth and easy to understand for users so let's make the following changes:

1. Add descriptions to the comments so users know what the values are
2. Filter values should be in line with the comments (the organism value is currently not commented out and causing alignment issues)
3. All comments should be grouped together, no spaces

Example [here](https://docs.google.com/spreadsheets/d/1_H8TWAVUOYKLCx5B1vTBG8fbH4gP8CrEyCVQDRp45pk/edit#gid=0), screenshot below. Please test by loading into R and python.

![Screenshot 2023-03-28 at 9.51.36 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/e6a72067-2e31-44ed-96aa-f193fceb4ae6)

ashin-czi:
Going to wait for banner changes since there are some changes in the CSV output for that

niknak33:
Looks good from my side Andrew - thank you.

signechambers1:
LGTM! Thanks Andrew, this is great!

PR #4480, opened 2023-04-03
fix(wmg): Update CSV download headers to enable smooth loading into p…

ashin-czi:
…ython and R #4460

## Reason for Change

- #4460

## Changes

- add
- remove
- modify
    - Updating metadata header rows in CSV

## Testing steps

1. Select tissue/gene
2. Select any filters
3. Download CSV output
4. Make sure output has updated metadata rows

From ticket:

image

### DIFF
--- a/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveExport/csvUtils.ts
+++ b/frontend/src/views/WheresMyGene/components/GeneSearchBar/components/SaveExport/csvUtils.ts
@@ -16,6 +16,8 @@ import {
 } from "../../../HeatMap/utils";
 import { generateAndCopyShareUrl } from "../ShareButton/utils";
 
+const NO_SELECTION_STRING = "No selection";
+
 interface CsvMetadata {
   name: string;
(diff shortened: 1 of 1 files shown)


tihuan (approved):
LGTM! Just wanted to double check that when there's multiple selected values, the comment has `" "` surrounding it. Is that expected?

Screenshot 2023-04-06 at 7 16 10 AM

Also I didn't try to load the downloaded CSV in Python @ashin-czi did you check that? 

Thanks again!

tihuan:
> > LGTM! Just wanted to double check that when there's multiple selected values, the comment has `" "` surrounding it. Is that expected?
> > Also I didn't try to load the downloaded CSV in Python @ashin-czi did you check that?
> > Thanks again!
> 
> Yup the quotes are expected! The CSV library automatically surrounds the quote if there's a comma in the content
> 
> I haven't tried loading the CSV in Python, I might have to defer to someone else for that

Ah cool got it!

Sounds good! Can you send an example CSV file to Ambrose please? I think he was the one that raised the issue in the first place

Issue #4389, opened 2023-03-22
GET /v1/datasets has regressed the DOI value


brianraymor:
The datasets index notebook incorrectly returns:

`  'collection_doi': 'https://doi.org/10.1016/j.celrep.2018.09.006',`

Compare with the correct value from the the collection itself:

`'doi': '10.1016/j.celrep.2018.09.006'`

Please check that all other cases are correct too.

PR #4484, opened 2023-04-04
feat(discover API): return expected format for collection_doi in dataset index endpoint

nayib-jose-gloria:
## Reason for Change

- #4389

## Changes

- use common `extract_doi_from_links` function in Discover API datasets index endpoint, also used for collection DOIs
- update tests

## Testing steps
- unit tests


danieljhegeman (approved):
Thanks @nayib-jose-gloria ! 🚀

Issue #4488, opened 2023-04-05
Shard E2E tests in the CI


charles-testco:
Explore the use of sharding to speed up tests in CI

Context: #4487

Playwright Github Issue on native support for merging sharded test reports:

https://github.com/microsoft/playwright/issues/10437

tihuan:
Completed in CI/CD epic

PR #4489, opened 2023-04-05
feat: #4488 - Added sharding E2E tests in the CI

charles-testco:
## Reason for Change

- #4488 

- added project matrix strategy so each browser is executed on a separate runner so that they run in parallel


tihuan:
@charles-testco quick feedback -- in PR description we can drop prefix `ZEN` and just use `#NNNN` to directly link to the Github issue!

E.g., `#4488` instead of `#ZEN-4488`

Thank you 👍

charles-testco:
> The sharding is amazing! Thanks so much for adding it 🙏
> 
> And feels like we might be able to drop `e2e-all` and other related script commands, since it still uses the same amount of workers anyway?
> 
> Thanks again!

Yes, we can remove them

frontend/package.json
```diff
@@ -108,10 +108,12 @@
     "serve": "next start",
     "test": "playwright test",
     "e2e": "playwright test",
+    "e2e:ci": "playwright test --config playwright-ci.config.ts",
     "e2e:chrome": "playwright test --project chromium",
     "e2e:chrome:headed": "playwright test --headed --project chromium",
-    "e2e:ff": "playwright test --project firefox",
-    "e2e:ie": "playwright test --project edge",
+    "e2e:browser:chrome": "playwright test --workers 1 --project chromium",
```
tihuan, line 114: Feels like we could remove these three scripts, because we no longer have e2e-all?

Thank you!

tihuan: ah sorry I meant revert it back to:

```
"e2e:ff": "playwright test --project firefox",
    "e2e:ie": "playwright test --project edge",
```

since I don't think we need `--workers 1` in these scripts?

charles-testco:
I am closing this PR and raise a new one as suggested by some people to resolve the stale status:

e2e-test Expected — Waiting for status to be reported

Issue #4502, opened 2023-04-06
Make service- and happy-related infra changes for datasets bucket: dev & rdev


danieljhegeman:
In order for the pipeline to be able to write to the new datasets bucket, we have to make the necessary changes to permit the env variable containing the bucket name to be populated via happy env. Additionally, we need to add the rdev analogue of the datasets bucket so that rdev stacks can mimic the same write behavior.

Service code (processing pipeline) needs to be updated to write labeled `.h5ad` and `.rds` files to the datasets bucket with the format: `<bucket>/<dataset_version_id>.<type>`.

danieljhegeman:
Once the associated changes are in, we can try removing the backend role access to the datasets s3 bucket that is granted in the assets bucket in modules/corpora/backend/s3.tf

PR #4507, opened 2023-04-06
feat: add datasets bucket to ecs stack and batch modules

danieljhegeman:
## Reason for Change

- #4502 

## Changes

- include datasets bucket name as env var for ecs stack / batch modules

## Testing steps


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -56,6 +56,7 @@ locals {
 
   artifact_bucket            = try(local.secret["s3_buckets"]["artifact"]["name"], "")
   cellxgene_bucket           = try(local.secret["s3_buckets"]["cellxgene"]["name"], "")
+  datasets_bucket            = try(local.secret["s3_buckets"]["datasets"]["name"], "")
```
nayib-jose-gloria, line 59: just for my own clarification--where are all these secrets actually set? Is this secret already set?

backend/layers/processing/process_logic.py
```diff
@@ -49,41 +49,47 @@ def download_from_s3(self, bucket_name: str, object_key: str, local_filename: st
...
-        self.logger.info(f"Uploading [{dataset_id}/{file_name}] to S3 bucket: [{artifact_bucket}].")
         try:
-            s3_uri = self.upload(file_name, bucket_prefix, artifact_bucket)
-            self.logger.info(f"Updating database with  {artifact_type}.")
+            s3_uri = self.upload_artifact(file_name, key_prefix, artifact_bucket)
+            self.logger.info(f"Uploaded [{dataset_id}/{file_name}] to {s3_uri}")
             self.business_logic.add_dataset_artifact(dataset_id, artifact_type, s3_uri)
+            self.logger.info(f"Updated database with {artifact_type}.")
+            if datasets_bucket:
+                datasets_s3_uri = self.s3_provider.upload_file(
```
nayib-jose-gloria, line 89: `self.s3_provider.upload_file` does not return the s3_uri, you'll have to build it yourself for the log statement

nayib-jose-gloria: or update the method in s3_provider

backend/layers/processing/process_logic.py
```diff
@@ -49,41 +49,47 @@ def download_from_s3(self, bucket_name: str, object_key: str, local_filename: st
...
 
-    def upload(
+    def upload_artifact(
         self,
         file_name: str,
-        bucket_prefix: str,
+        key_prefix: str,
         artifact_bucket: str,
     ) -> str:
         # TODO: make sure that the files are correct
```
nayib-jose-gloria, line 61: nit: we can probably remove this old todo

backend/layers/processing/process_logic.py
nayib-jose-gloria: Also, I think whats leading to the test errors is that, despite the name, extra_args is NOT an optional argument for upload_file (neither in the mock nor the actual s3_provider). Let's update the mock + s3_provider to give it a default arg

danieljhegeman: Great catches. Hadn't had time to investigate fully last night. Resubmitting...

nayib-jose-gloria (commented):
This looks good! Have you tested in rdev to ensure no regressions and/or successful PUTs to the dataset assets bucket? I think we still need to update the bucket policy's Allow PutObjects role for the latter

.happy/terraform/modules/ecs-stack/main.tf
danieljhegeman: great question, I was confused on that initially too 😅 

this is set by the happy-env module in the single cell infra; I added the necessary changes [here](https://github.com/chanzuckerberg/single-cell-infra/pull/745/files#diff-d94f492fad3dd0bd4db31186d7558d6079078e53d0b9999af5b69a8947c620adR99-R102)

danieljhegeman:
> This looks good! Have you tested in rdev to ensure no regressions and/or successful PUTs to the dataset assets bucket? I think we still need to update the bucket policy's Allow PutObjects role for the latter

successfully tested in rdev ✅

tests/unit/processing/test_processing.py
```diff
@@ -121,11 +124,15 @@ def test_process_all(self, mock_cxg, mock_seurat, mock_h5ad):
             self.business_logic, self.uri_provider, self.s3_provider, self.downloader, self.schema_validator
         )
         for step_name in ["download-validate", "cxg", "seurat"]:
-            pm.process(dataset_version_id, step_name, dropbox_uri, "fake_bucket_name", "fake_cxg_bucket")
+            pm.process(
+                dataset_version_id, step_name, dropbox_uri, "fake_bucket_name", "datasets_bucket", "fake_cxg_bucket"
```
Bento007, line 128:
```suggestion
                dataset_version_id, step_name, dropbox_uri, "fake_bucket_name", "fake_datasets_bucket", "fake_cxg_bucket"
```

tests/unit/processing/test_processing.py
```diff
@@ -167,7 +174,9 @@ def test_process_all_download_validate_fail(self):
         )
 
         for step_name in ["download-validate"]:
-            pm.process(dataset_version_id, step_name, dropbox_uri, "fake_bucket_name", "fake_cxg_bucket")
+            pm.process(
+                dataset_version_id, step_name, dropbox_uri, "fake_bucket_name", "datasets_bucket", "fake_cxg_bucket"
```
Bento007, line 178:
```suggestion
                dataset_version_id, step_name, dropbox_uri, "fake_bucket_name", "fake_datasets_bucket", "fake_cxg_bucket"
```

backend/layers/processing/process_logic.py
```diff
@@ -49,41 +49,48 @@ def download_from_s3(self, bucket_name: str, object_key: str, local_filename: st
...
+        datasets_bucket: Optional[str] = None,  # If provided, dataset will be uploaded to this bucket for public access
     ):
         self.update_processing_status(dataset_id, processing_status_key, DatasetConversionStatus.UPLOADING)
-        self.logger.info(f"Uploading [{dataset_id}/{file_name}] to S3 bucket: [{artifact_bucket}].")
         try:
-            s3_uri = self.upload(file_name, bucket_prefix, artifact_bucket)
-            self.logger.info(f"Updating database with  {artifact_type}.")
+            s3_uri = self.upload_artifact(file_name, key_prefix, artifact_bucket)
+            self.logger.info(f"Uploaded [{dataset_id}/{file_name}] to {s3_uri}")
             self.business_logic.add_dataset_artifact(dataset_id, artifact_type, s3_uri)
```
Bento007, line 85: why don't we create an artifact for dataset uploaded to the "dataset_bucket"? It's something we should keep track of.

backend/layers/processing/process_logic.py
```diff
@@ -49,41 +49,48 @@ def download_from_s3(self, bucket_name: str, object_key: str, local_filename: st
...
+            self.logger.info(f"Updated database with {artifact_type}.")
+            if datasets_bucket:
```
Bento007, line 87: You could also trigger off an s3 event  in the artifact bucket to start an upload to the public bucket whenever a file is upload to the that matches your criteria. Your solution also works.

Bento007 (changes requested):
I have a few questions

backend/layers/processing/process_logic.py
danieljhegeman: Well, the question is, why do we _need_ an artifact row if the artifact uri can be easily determined by, presently, looking at the suffix on the existing artifact uri's, and soon, simply by appending the dataset version id as the key on the bucket?

backend/layers/processing/process_logic.py
danieljhegeman: I thought about this but it's unnecessary complication for the flow. I'd prefer to just insert the write here and allow it to be governed by the same error reporting that's already built in.

PR #4532, opened 2023-04-10
feat(discover API): return dataset assets as part of dataset api responses

nayib-jose-gloria:
## Reason for Change

- #4511 

## Changes

- deprecate dataset_assets from endpoints that return non-preview dataset response objects 
- add assets field to non-preview dataset response objects
- for assets field, construct + return cloudfront link to download H5AD and RDS assets
- deprecate /assets endpoint

## Testing steps
-  unit tests + rdev

## Notes for Reviewer
- For now, corpora config will point to cloudfront domain and we will return public dataset asset download links with the cloudfront distribution domain. Once infra is ready, we will change this config value to point to the vanity URL datasets.cellxgene.cziscience.com instead. This will not require a code change


backend/curation/api/v1/curation/collections/common.py
```diff
@@ -31,6 +31,27 @@ def get_collections_base_url():
...
+    asset_list = list()
+    for asset in dataset_version.artifacts:
+        if asset.type not in allowed_dataset_asset_types:
+            continue
+        filesize = get_business_logic().s3_provider.get_file_size(asset.uri)
+        if filesize is None:
+            filesize = -1
+        # Note: this key parsing assumes asset.uri from DB has `<bucket>/<uuid>/local.<file_type>` directory structure
+        key = urlparse(asset.uri).path.split("/")[-2]
+        url = f"{base_url}/{key}.{asset.type}"
```
danieljhegeman, line 45:
```suggestion
        key_uuid = urlparse(asset.uri).path.split("/")[-2]
        url = f"{base_url}/{key_uuid}.{asset.type}"
```

danieljhegeman (approved):
see suggestion! 🙏  Great work on refactoring tests

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -31,6 +31,27 @@ def get_collections_base_url():
...
 
 
+def extract_dataset_assets(dataset_version: DatasetVersion):
+    base_url = CorporaConfig().dataset_assets_base_url
+    asset_list = list()
+    for asset in dataset_version.artifacts:
+        if asset.type not in allowed_dataset_asset_types:
+            continue
+        filesize = get_business_logic().s3_provider.get_file_size(asset.uri)
+        if filesize is None:
```
Bento007, line 41: nit: 
```
filesize = get_business_logic().s3_provider.get_file_size(asset.uri) if not None else -1
```

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1567,7 +1597,39 @@ def test_get_datasets_nonexistent_collection_4xx(self):
             response = self.app.get(test_url, headers=headers)
             self.assertEqual(403, response.status_code)
 
-    def test_get_all_datasets_200(self):
+    @patch("backend.common.corpora_config.CorporaConfig.__getattr__")
+    def test_get_dataset_no_assets(self, mock_config: Mock):
+        def mock_config_fn(name):
```
Bento007, line 1602: suggestion: Reuse `mock_config_fn`? It could then be used in the patch decorator and parameterized as needed.

```
@patch("backend.common.corpora_config.CorporaConfig.__getattr__", side_effect=mock_config_fn())
```

Also I think this could be simplified to use a dict rather than a bunch of ifs and returns.
```
def mock_config_fn(name):
   mocked_attr = { 
       "curator_role_arn":"test_role_arn",
       ""submission_bucket": "cellxgene-dataset-submissions-test",
      ....
}
      return mocked_attr[name]
```

The mocked_attr dict could even be defined outside of the function to avoid invoking it every time.

Bento007 (approved):
Looks good. A few suggestions, nothing required.

Issue #4537, opened 2023-04-10
Delete dataset assets from public-access datasets bucket for tombstones


danieljhegeman:
- [ ] Add logic to delete corresponding `.h5ad` and `.rds` files in the public-access datasets bucket when a dataset is tombstoned (confirm: this should apply to all versions for the dataset).
- [ ] Check to see if tombstoned datasets were copied to the existing public-access bucket. This will need to be reassessed for staging and prod after those buckets are populated with the existing datasets. Develop script to delete such files.

PR #4541, opened 2023-04-11
feat: delete datasets in public-access s3 when tombstoning a Collection

danieljhegeman:
- fix error for submissions bucket info returned by ../s3-upload-credentials endpoint
- add datasets bucket env var to 'service' module

## Reason for Change

- Addresses first part of #4537 
- We will address cleaning up tombstoned datasets from public bucket by performing a mirror from prod->staging, and then staging->prod.

## Testing steps

- unit tests updated
- not yet rdev-tested

## Notes for Reviewer

This PR also needs to include invalidation of cloudfront caches for the datasets distribution. This work is in-progress and will be added soon.


backend/curation/api/v1/curation/collections/collection_id/s3_upload_credentials.py
```diff
@@ -29,6 +30,9 @@ def get(collection_id: str, token_info: dict):
         upload_key_prefix = f"super/{collection_id}/"
     else:
         upload_key_prefix = f"{user_info.user_id}/{collection_id}/"
+    rdev_prefix = os.environ.get("REMOTE_DEV_PREFIX", "").strip("/")
+    if rdev_prefix:
+        upload_key_prefix = f"{rdev_prefix}/{upload_key_prefix}"
```
danieljhegeman, line 35: see comment above. Remote dev prefix is part of the _key_, not the _bucket_.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -83,13 +83,21 @@ def test__delete_private_collection_by_collection_version_id(self):
...
+    def test__delete_tombstone_collection(self, mock_delete: Mock):
         tests = [("not_owner", 403), ("noauth", 401), ("owner", 403), ("super", 403)]
         for auth, expected_response in tests:
             with self.subTest(auth):
-                collection_id = self.generate_published_collection().collection_id
-                self.business_logic.tombstone_collection(collection_id)
-                self._test(collection_id, auth, expected_response)
+                collection = self.generate_published_collection()
+                self.business_logic.tombstone_collection(collection.collection_id)
+                if self.run_as_integration:
```
danieljhegeman, line 93: no need to complicate the db mock code to handle this, so we'll just run the arg check for integration tests only

tests/unit/processing/base_processing_test.py
```diff
@@ -22,6 +23,9 @@ def upload_file(self, src_file: str, bucket_name: str, dst_file: str, extra_args
     def upload_directory(self, src_dir: str, s3_uri: str):
         self.mock_s3_fs.append(s3_uri)
 
+    def delete_files(self, bucket_name: str, object_keys: List[str]):
+        pass
+
```
danieljhegeman, line 28: for completeness

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -83,13 +83,21 @@ def test__delete_private_collection_by_collection_version_id(self):
...
+                    dataset_version_ids = [d_v.version_id.id for d_v in collection.datasets]
+                    delete_args = set(mock_delete.call_args.args[1])
+                    expected_delete_args = set()
+                    for file_type in ("h5ad", "rds"):
+                        expected_delete_args.update([f"{d_v_id}.{file_type}" for d_v_id in dataset_version_ids])
```
danieljhegeman, line 98: I will be doing the dataset file structure migration before this hits prod followed by mirrors from prod to dev and staging, so this will work (i.e. every dataset file will, at that point, live at a `<dataset_version_id>.<file_type>` key)

backend/layers/thirdparty/s3_provider.py
```diff
@@ -48,6 +48,12 @@ def upload_file(self, src_file: str, bucket_name: str, dst_file: str, extra_args
             ExtraArgs=extra_args,
         )
 
+    def delete_files(self, bucket_name: str, object_keys: List[str]):
+        """
+        Deletes the objects `object_keys` from bucket `bucket_name`
+        """
+        self.client.delete_objects(Bucket=bucket_name, Delete={"Objects": [{"Key": key} for key in object_keys]})
```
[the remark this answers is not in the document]
Bento007: Also add unit tests for the above cases.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1475,5 +1475,32 @@ def test_dataset_published_at_with_replaced_dataset_ok(self):
         self.assertEqual(dataset_1.canonical_dataset.published_at, published_collection.published_at)
 
 
+class TestCollectionUtilities(BaseBusinessLogicTestCase):
+    def test__delete_datasets_from_public_access_bucket(self):
+        published_collection = self.initialize_published_collection()
+        new_version = self.business_logic.create_collection_version(published_collection.collection_id)
+
+        # We will replace the first dataset
```
Bento007, line 1483:
```suggestion
        # Update the first dataset
```

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1475,5 +1475,32 @@ def test_dataset_published_at_with_replaced_dataset_ok(self):
...
+        replaced_dataset_version_id, _ = self.business_logic.ingest_dataset(
+            new_version.version_id, "http://fake.url", None, dataset_id_to_replace
+        )
+
+        self.complete_dataset_processing_with_success(replaced_dataset_version_id)
+
+        dataset_version_ids = [d_v.version_id.id for d_v in published_collection.datasets] + [
+            replaced_dataset_version_id
+        ]
+        delete_keys = set(
```
Bento007, line 1495:
```suggestion
        actual_delete_keys = set(
```

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1475,5 +1475,32 @@ def test_dataset_published_at_with_replaced_dataset_ok(self):
         self.assertEqual(dataset_1.canonical_dataset.published_at, published_collection.published_at)
 
 
+class TestCollectionUtilities(BaseBusinessLogicTestCase):
+    def test__delete_datasets_from_public_access_bucket(self):
```
Bento007, line 1479: add a docstring. You are testing delete published datasets and unpublished datasets, add those details here.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1475,5 +1475,32 @@ def test_dataset_published_at_with_replaced_dataset_ok(self):
         self.assertEqual(dataset_1.canonical_dataset.published_at, published_collection.published_at)
 
 
+class TestCollectionUtilities(BaseBusinessLogicTestCase):
```
Bento007, line 1478: Add a test where multiple collection versions refers to the same dataset version.

Bento007 (changes requested):
A few edge cases to take into account

backend/curation/api/v1/curation/collections/collection_id/s3_upload_credentials.py
danieljhegeman: It was not noticed. It would cause an issue only in rdev, and only in certain scenarios. But it was wrong and should be fixed.

backend/layers/business/business.py
```diff
@@ -570,10 +571,29 @@ def delete_collection_version(self, version_id: CollectionVersionId) -> None:
...
         self.database_provider.delete_collection_version(version_id)
 
+    def delete_datasets_from_public_access_bucket(self, collection_id: CollectionId) -> List[str]:
+        """
+        Delete all associated publicly-accessible Datasets in s3
+        """
+        collection_versions = self.database_provider.get_all_versions_for_collection(collection_id)
+        datasets_bucket = os.getenv("DATASETS_BUCKET")
+        rdev_prefix = os.environ.get("REMOTE_DEV_PREFIX", "").strip("/")
+        object_keys = []
```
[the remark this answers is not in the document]
danieljhegeman: great point

tests/unit/backend/layers/business/test_business.py
danieljhegeman: that is encompassed in the existing test since one dataset is updated and the other is not.

backend/layers/thirdparty/s3_provider.py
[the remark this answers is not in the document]
danieljhegeman: > We don't want to silently fail to delete. If we do fail we need to roll back the and database changes made. t's is an edge case, but how difficult are these datasets to track down if this fails silently?

In some cases I would agree with you, but there are some important considerations: the user is _trying_ to delete these assets; the public dataset deletion is occurring before the updates to entities in the database, so we guarantee that we don't accidentally leave datasets behind -- the request only fails silently if the key doesn't exist, which does not constitute a problem in this scenario. If the actual attempt to delete fails, perhaps for network reasons, the request fails loudly and the Collection will not be deleted from the db. The user sees 500. The user will therefore be aware that the Collection still exists. Perhaps we should capture 500s for deletes and inform them that public assets may be missing.

Actually, in lieu of that, we can perform a HEAD request for all asset files for all `dataset.assets[]` objects. That way we can say whether or not the public asset is still available. If it's not, then the curator can get in touch with us.

cc @nayib-jose-gloria for thoughts on this issue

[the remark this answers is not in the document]
Bento007: I changed my mind we should make the request without `quite` and log the response. Knowing what request deleted a file could help with debugging and audits in the future.

tests/unit/backend/layers/business/test_business.py
Bento007: I would mention that in the docstring, that test case is obfuscated in the `initialize_published_collection`

backend/layers/business/business.py
```diff
@@ -570,10 +571,28 @@ def delete_collection_version(self, version_id: CollectionVersionId) -> None:
...
+        rdev_prefix = os.environ.get("REMOTE_DEV_PREFIX", "").strip("/")
+        object_keys = set()
+        for collection_version in collection_versions:
+            for d_v in collection_version.datasets:
+                for file_type in ("h5ad", "rds"):
+                    dataset_version_s3_object_key = f"{d_v.version_id.id}.{file_type}"
+                    if rdev_prefix:
+                        dataset_version_s3_object_key = f"{rdev_prefix}/{dataset_version_s3_object_key}"
+                    object_keys.add(dataset_version_s3_object_key)
+        self.s3_provider.delete_files(bucket, list(object_keys))
```
Bento007, line 588: do we need to convert it into a list?

backend/layers/thirdparty/s3_provider.py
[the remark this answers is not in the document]
nayib-jose-gloria: I think capturing 500s and returning a message to the user that tombstoning failed, but some public assets may have already been deleted, and suggesting they try again is a reasonable route. Perhaps we should introduce our own retry (unless boto3 already has that as a default?) here to account for transient network blips

backend/layers/thirdparty/s3_provider.py
```diff
@@ -48,6 +48,18 @@ def upload_file(self, src_file: str, bucket_name: str, dst_file: str, extra_args
             ExtraArgs=extra_args,
         )
 
+    def delete_files(self, bucket_name: str, object_keys: List[str]):
+        """
+        Deletes the objects `object_keys` from bucket `bucket_name`
+        """
+
+        def chunk_list(items: List[str], chunk_size: int = 1000):  # s3 client delete_objects accepts up to 1000 keys
```
Bento007, line 56: I don't think this is complicated enough to need sub function. Please inline.

backend/layers/thirdparty/s3_provider.py
```diff
@@ -48,6 +48,18 @@ def upload_file(self, src_file: str, bucket_name: str, dst_file: str, extra_args
...
+            for i in range(0, len(items), chunk_size):
+                yield items[i : i + chunk_size]
+
+        for keys_batch in chunk_list(object_keys):
```
Bento007, line 60: We should make the 1000 into a constant.

```suggestion
        AWS_S3_BULK_DELETE_LIMIT = 1000
        for i in range(0, len(object_keys), AWS_S3_BULK_DELETE_MAX):
            keys_batch = object_keys[ i: i+AWS_S3_BULK_DELETE_MAX]
```

Bento007 (changes requested):

backend/layers/thirdparty/s3_provider.py
```diff
@@ -48,6 +48,18 @@ def upload_file(self, src_file: str, bucket_name: str, dst_file: str, extra_args
...
+            self.client.delete_objects(Bucket=bucket_name, Delete={"Objects": [{"Key": key} for key in keys_batch]})
```
Bento007, line 61: log this response

backend/layers/thirdparty/s3_provider.py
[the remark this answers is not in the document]
nayib-jose-gloria: Agree w/ logging responses for debugging should it come up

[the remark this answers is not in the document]
brianraymor: > returning a message to the user that tombstoning failed, but some public assets may have already been deleted, and suggesting they try again is a reasonable route.

That sounds like ambiguity/complexity is being pushed to the user. The operation should either succeed or not?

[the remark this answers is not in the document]
nayib-jose-gloria: There's always the rare possibility that a network error causes some object(s) not to be deleted in a bulk delete request. We can enable dataset-assets to be a versioned bucket, which would allow us to restore deleted objects and rollback if we come across an error--but that seems like overengineering for this edge case IMO 

We can alternatively bulk delete and then alert engineers of an error if the boto3 delete_objects command fails on any asset + return the list of objects it failed to delete, like we do with upload errors. And users can see 500 and follow the same workflow as with upload errors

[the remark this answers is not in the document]
danieljhegeman: >That sounds like ambiguity/complexity is being pushed to the user. The operation should either succeed or not?

That is true, and in all likelihood we will never run into this problem. Really what we're dealing with here is the unlikely event that _some_ of the assets are deleted successfully and then a transient error causes the whole operation to return 500 (meaning the Collection entity does not actually get marked as tombstoned/deleted), at which point we'd be left with the Collection still in play and some of the assets gone, but not all. We don't have a way to perfectly transactionalize this operation since it is db + s3 without putting a lot of effort into it.

Plus, the ambiguity will only ever get put on the user _if_ they have tried to delete the Collection (which means they would surely just try to delete it again anyways). I don't want us to spend time solving perfectly the scenario where someone tries to delete something, it fails, and then they decide they don't want to delete it anymore. 😄

backend/layers/thirdparty/s3_provider.py
```diff
@@ -52,13 +57,14 @@ def delete_files(self, bucket_name: str, object_keys: List[str]):
...
-        def chunk_list(items: List[str], chunk_size: int = 1000):  # s3 client delete_objects accepts up to 1000 keys
-            for i in range(0, len(items), chunk_size):
-                yield items[i : i + chunk_size]
-
-        for keys_batch in chunk_list(object_keys):
-            self.client.delete_objects(Bucket=bucket_name, Delete={"Objects": [{"Key": key} for key in keys_batch]})
+        for i in range(0, len(object_keys), AWS_S3_MAX_ITEMS_PER_BATCH):
+            key_batch = object_keys[i:AWS_S3_MAX_ITEMS_PER_BATCH]
+            logger.info(f"Deleting assets from public-access bucket '{bucket_name}': {key_batch}")
+            self.client.delete_objects(
```
Bento007, line 63: We should also log the Deleted and Errors fields from this response.

Bento007: The value mentioned above are hard to find after deleting the files. Since deletion is a sensitive topic, I'd rather error on the side of caution and log more than we might need.

backend/layers/thirdparty/s3_provider.py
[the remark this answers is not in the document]
danieljhegeman: @Bento007 @nayib-jose-gloria  I actually think further optimization here is not worth the effort, and here's why: the curator only performs this action if they intend to erase the Collection. If the operation fails, i.e., if any of the deletes don't succeed, then the curator will be made aware of this in the UI because the request will fail. Therefore they've been made aware that the Collection still exists. They'll either try again, and this code will succeed or fail (try, try again), and then they'll contact us if it continues failing. When they contact us, our move will be to delete everything, which is extremely easy to do with a couple postgres queries for dataset version ids; it will not even be worth referencing the logs to figure out exactly what happened with the failed delete attempts. cc @brianraymor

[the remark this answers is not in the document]
danieljhegeman: So I submit that we leave it with the current level of logging. What really matters is having the s3 deletes coming _before_ the tombstoning of the Collection entity, which is what this code is doing.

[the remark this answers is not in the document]
nayib-jose-gloria: Actually, Dan, are you sure it won't fail silently for all errors? Per the [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/delete_objects.html), it seems like it returns the same object response shape regardless, which provides list of Errors. So I think we'd need to check here for `if resp['Errors']` and raise an exception if list has entries

nayib-jose-gloria (changes requested):
1 more note about error handling, otherwise looking good

backend/layers/thirdparty/s3_provider.py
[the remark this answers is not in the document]
danieljhegeman: It fails loudly if the request is malformed, if permissions are not correct, etc.

So if the key doesn't exist, it fails silently. But that would indicate a business logic failing... and our standard for interactions with s3 so far has not been to double check the key name for, say, writes.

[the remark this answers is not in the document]
danieljhegeman: I can't imagine a network failure would be swallowed by the boto client; that would surely be a loud failure as well.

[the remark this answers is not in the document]
nayib-jose-gloria: It 'fails loudly' by returning the error in the Errors list though, not by throwing an Exception. And if we don't check the Errors list for entries, aren't we failing silently as a result?

Bento007 (approved):

backend/layers/thirdparty/s3_provider.py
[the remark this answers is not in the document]
danieljhegeman: Fail loudly == throws Exception

Issue #4560, opened 2023-04-13
Add create collection button to collections list for curators


NoopDog:
When logged in, show a "Create Collection" button for users. 

- Selecting the button shows the create collection modal. 
- Completing the collection shows the unpublished collection detail.
- Canceling the create closed the modal. 
- Update create button to [match mocks](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=4626-40099&t=2oru9IONKhCWW9xK-0).

NoopDog:
Looks good to me on dev.

PR #4570, opened 2023-04-14
feat: Add create collection button to collections list for curators (#4560)

frano-m:
## Reason for Change

Curators are able to create a collection from the collections index.

- #4560

## Changes

- Added Create Collection button to collections index, for curators only.
- Updated Create Collection button from `blueprint` to `czifui`, and styles to match [mocks](https://www.figma.com/file/18a4QULVIfpFQhTn1FGh8a/Filtering?node-id=4626-40099&t=0WAYPzpZ8eh7QPpT-0).

## Testing steps

- Go to collections index, the "Create Collection" button should not be visible.
- Use the curator feature flag, the "Create Collection" button should not be visible.
- Log in, the "Create Collection" button should be visible.
- Click on the "Create Collection" button and add a new collection. Flow should be unchanged i.e. when a new private collection is created the page should route to the private collection.
- Cancelling the "Create Collection" modal should close the modal, and return the user to the collections index.

### DIFF
--- a/frontend/src/components/CreateCollectionModal/index.tsx
+++ b/frontend/src/components/CreateCollectionModal/index.tsx
@@ -24,7 +24,9 @@ const AsyncCTA = loadable(
 );
 
 const CreateCollectionButton = (props: Partial<IButtonProps>) => (
-  <StyledButton {...props}>Create Collection</StyledButton>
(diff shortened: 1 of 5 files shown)


NoopDog:
Thanks @ashin-czi !

Issue #4582, opened 2023-04-14
GET /datasets/versions documentation is unclear about published_at and revised_at


brianraymor:
See the [swagger](https://api.cellxgene.cziscience.com/curation/ui/#/dataset/backend.curation.api.v1.curation.datasets.dataset_id.versions.actions.get).

1. There's no discussion about `published_at` being the publication timestamp for a version of a dataset.
2. `revised_at` is not documented, but present in the response?

```
[{'assay': [{'label': "10x 3' v2", 'ontology_term_id': 'EFO:0009899'}],
  'assets': [{'filesize': 213763978,
              'filetype': 'H5AD',
              'url': 'https://datasets.cellxgene.dev.single-cell.czi.technology/652623db-3ed1-4748-a25a-76b0a4abde38.h5ad'},
             {'filesize': 181873820,
              'filetype': 'RDS',
              'url': 'https://datasets.cellxgene.dev.single-cell.czi.technology/652623db-3ed1-4748-a25a-76b0a4abde38.rds'}],
  'batch_condition': None,
  'cell_count': 24544,
  'cell_type': [{'label': 'basal cell of prostate epithelium',
                 'ontology_term_id': 'CL:0002341'},
                {'label': 'epithelial cell of urethra',
                 'ontology_term_id': 'CL:1000296'},
                {'label': 'luminal cell of prostate epithelium',
                 'ontology_term_id': 'CL:0002340'},
                {'label': 'neuroendocrine cell',
                 'ontology_term_id': 'CL:0000165'},
                {'label': 'secretory cell', 'ontology_term_id': 'CL:0000151'}],
  'collection_id': '913dcd71-45aa-4062-9529-504da18d4865',
  'collection_version_id': 'c625d29e-153f-4486-8a28-784672092c0d',
  'dataset_id': '8870053d-ce82-4244-bb23-b9473d7c033b',
  'dataset_version_id': '652623db-3ed1-4748-a25a-76b0a4abde38',
  'development_stage': [{'label': '29-year-old human stage',
                         'ontology_term_id': 'HsapDv:0000123'},
                        {'label': '31-year-old human stage',
                         'ontology_term_id': 'HsapDv:0000125'}],
  'disease': [{'label': 'normal', 'ontology_term_id': 'PATO:0000461'}],
  'donor_id': ['D17', 'D27', 'D35'],
[...]

Issue #4581, opened 2023-04-14
get_dataset_versions returns the original published_at timestamp for all versions of a dataset


brianraymor:
1. Create collection
1. Add dataset
1. Publish collection
1. Revise collection
1. Add dataset
1. Publish revision

The dataset timestamps returned in GET /collection are:

```

 'published_at': '2023-04-14T22:22:55.524100+00:00',
 'revised_at': '2023-04-14T22:43:05.154712+00:00',
```

Per [revised_at and published_at for DatasetVersions](https://github.com/chanzuckerberg/single-cell-data-portal/issues/4413#issuecomment-1485798718), the expectation is that GET /dataset/versions will return the publication timestamp for each version in `published_at` and `revised_at` will not be set.

But `GET /datasets/revision` returns the original `published_at` for both the original and updated version:

```

'2023-04-14T22:22:55.524100+00:00'
'2023-04-14T22:22:55.524100+00:00'

```

nayib-jose-gloria:
Issue is as_canonical flag is not being passed as False in the versions endpoint function. 

I think we should take the opportunity to attempt to consolidate as_canonical vs. use_canonical_url args, as I suspect there is overlap. And if there is a clear distinction in which we do need both args, we should rename them to reflect the distinction more

danieljhegeman:
Unassigning only because I have not had time to acquaint myself with the bug at all yet. Will re-assign once I've done so.

@nayib-jose-gloria if you do happen to have bandwidth to pick this up, that would be excellent

PR #4610, opened 2023-04-17
feat: fix response shape for curation API datasets/<dataset_id>/versions endpoint

nayib-jose-gloria:
## Reason for Change

- #4581
- #4582

## Changes
-  Passed arg that will 1) remove unexpected revised_at timestamp from dataset version history endpoint entries, and 2) populate published_at timestamp with the expected logic for the version history endpoint (logic for both already exists in common function, call was just missing the right arg)
- Updated curation API spec to clarify that dataset_versions are returned by  datasets/<dataset_id>/versions endpoint. In the Swagger, the different meaning of the published_at timestamp for dataset_versions objects (vs dataset objects) is documented, but the Open API spec had tagged this endpoint under 'datasets' instead, so the distinction wasn't clear.
- Added test case for bug scenario

## Testing steps
- unit tests 

## Notes for Reviewer
- I think we need to clean up the arg names / branching logic in the reshape functions (i.e. as_canonical vs. use_canonical_url sound too similar, but datasets in revisions returned in the canonical collection/dataset endpoints represent a scenario where you want as_canonical = True but use_canonical_url = False). We should make the distinction more clear. 
- We should also try to fix the fact we use as_version as the arg in some reshape functions, but as_canonical in others.
- Above is not in scope for the bug fix

### DIFF
--- a/backend/curation/api/v1/curation/datasets/dataset_id/versions/actions.py
+++ b/backend/curation/api/v1/curation/datasets/dataset_id/versions/actions.py
@@ -23,4 +23,6 @@ def get(dataset_id: str):
     # business function returns in chronological order; must return in reverse chronological
     dataset_versions.reverse()
 
-    return make_response(jsonify(reshape_datasets_for_curation_api(dataset_versions, use_canonical_url=False)), 200)
+    return make_response(
+        jsonify(reshape_datasets_for_curation_api(dataset_versions, use_canonical_url=False, as_version=True)), 200
+    )
--- a/tests/unit/backend/layers/api/test_curation_api.py
+++ b/tests/unit/backend/layers/api/test_curation_api.py
@@ -1783,13 +1783,13 @@ class TestGetDatasetIdVersions(BaseAPIPortalTest):
         dataset_id = collection.datasets[0].dataset_id
         dataset_version_id = collection.datasets[0].version_id
         published_revision = self.generate_revision(collection_id)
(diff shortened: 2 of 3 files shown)


backend/curation/api/v1/curation/datasets/dataset_id/versions/actions.py
```diff
@@ -23,4 +23,6 @@ def get(dataset_id: str):
     # business function returns in chronological order; must return in reverse chronological
     dataset_versions.reverse()
 
-    return make_response(jsonify(reshape_datasets_for_curation_api(dataset_versions, use_canonical_url=False)), 200)
+    return make_response(
+        jsonify(reshape_datasets_for_curation_api(dataset_versions, use_canonical_url=False, as_version=True)), 200
```
danieljhegeman, line 27: I guess I missed this when I made these changes a month ago.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1798,16 +1798,21 @@ def test_get_dataset_id_versions_ok(self):
         self.assertEqual(200, response.status_code)
         expected = defaultdict(list)
         for dataset in response.json:
+            self.assertIsNone(dataset.get("revised_at"))
```
danieljhegeman, line 1801: 👍

danieljhegeman (approved):
Thanks @nayib-jose-gloria !

PR #4622, opened 2023-04-18
feat: always return dataset_version_id dataset permalink

danieljhegeman:
Previously, all pre-redesign datasets lived under a key in the public-access datasets bucket that was different from their dataset version ids. Now, all past and future assets live under their dataset version id.

## Reason for Change

- #4621 

## Changes

- modify asset uri to reflect dataset version id, and not the uuid that is part of the corpora prod asset

## Testing steps

- unit tested

### DIFF
--- a/tests/unit/backend/layers/api/test_curation_api.py
+++ b/tests/unit/backend/layers/api/test_curation_api.py
@@ -24,7 +24,7 @@ from backend.layers.common.entities import (
 from backend.layers.thirdparty.crossref_provider import CrossrefDOINotFoundException
 from tests.unit.backend.layers.api.test_portal_api import generate_mock_publisher_metadata
 from tests.unit.backend.layers.common.base_api_test import BaseAPIPortalTest
-from tests.unit.backend.layers.common.base_test import DatasetStatusUpdate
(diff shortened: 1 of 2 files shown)


tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -632,12 +632,6 @@ def test__get_collection_verify_body_is_reshaped_correctly__OK(self, mock_config
         self.generate_dataset(
             collection_version=collection_version,
             metadata=dataset_metadata,
-            artifacts=[
-                DatasetArtifactUpdate(type="h5ad", uri="http://test_filename/1234-5678-9/local.h5ad"),
-                DatasetArtifactUpdate(type="rds", uri="http://test_filename/1234-5678-9/local.rds"),
-                DatasetArtifactUpdate(type="cxg", uri="http://test_filename/1234-5678-9/local.cxg"),
```
nayib-jose-gloria, line 638: minor: mind keeping the CXG update? I believe it's the only test where we confirm we filter out CXG artifacts, since the default generate_dataset doesn't include a CXG artifact.

danieljhegeman: ah, good point!

nayib-jose-gloria (approved):

PR #4636, opened 2023-04-19
test(e2e): Fix e2e tests increase timeout

tihuan:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps


tihuan:
Turned out that FF and Edge tests were failing because we need to increase timeout from 3 to 5 mins, so it has enough time to tear down context. Chrome seems to tear down faster, so we never needed to do this

PR #4640, opened 2023-04-19
test(e2e): only run chrome until GHA sharding is done

tihuan:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps


tihuan:
[This PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5492) shards e2e tests, but we still can't enable testing for FF and Edge yet, since we need 30 workers to keep Chrome e2e tests under 10 mins. By adding FF and Edge to the testing matrix, GHA would randomly cancel jobs, probably because GHA has the 250 max jobs limit 😢

PR #4655, opened 2023-04-20
fix(445): Only return processing status for unpublished revision or private collection.

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/445

## Changes

- add is_published flag to modify the dataset and collection api response. 
- modify tests

### DIFF
--- a/tests/unit/backend/layers/api/test_curation_api.py
+++ b/tests/unit/backend/layers/api/test_curation_api.py
@@ -588,7 +588,7 @@ class TestGetCollectionVersions(BaseAPIPortalTest):
 
 class TestGetCollectionID(BaseAPIPortalTest):
     @patch("backend.common.corpora_config.CorporaConfig.__getattr__", side_effect=mock_config_fn)
-    def test__get_collection_verify_body_is_reshaped_correctly__OK(self, mock_config: Mock):
+    def test__get_published_collection_verify_body_is_reshaped_correctly__OK(self, mock_config: Mock):
         # Setup
         # test fixtures
         dataset_metadata = copy.deepcopy(self.sample_dataset_metadata)
--- a/tests/unit/backend/layers/api/test_curation_api.py
+++ b/tests/unit/backend/layers/api/test_curation_api.py
@@ -475,8 +475,8 @@ class TestGetCollections(BaseAPIPortalTest):
         resp_collection = resp[0]
         self.check_fields(EntityColumns.link_cols, resp_collection["links"][0], "links")
         self.assertEqual(public_collection.datasets[0].dataset_id.id, resp_collection["datasets"][0]["dataset_id"])
-        self.expected_collection_columns.pop("processing_status")
-        self.expected_dataset_columns.pop("processing_status")
+        self.expected_collection_columns.remove("processing_status")
+        self.expected_dataset_columns.remove("processing_status")
         self.check_fields(self.expected_dataset_columns, resp_collection["datasets"][0], "datasets")
         self.check_fields(self.expected_collection_columns, resp_collection, "collection")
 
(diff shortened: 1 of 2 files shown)


tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -475,6 +475,7 @@ def test__verify_expected_public_collection_fields(self):
         resp_collection = resp[0]
         self.check_fields(EntityColumns.link_cols, resp_collection["links"][0], "links")
         self.assertEqual(public_collection.datasets[0].dataset_id.id, resp_collection["datasets"][0]["dataset_id"])
+        self.expected_collection_columns.remove("processing_status")
```
danieljhegeman, line 478: Regarding the 'expected_columns' assertion pattern more generally: this is fine for now but I'm thinking we should really be taking care of this in the connexion yml, as required fields, right?

danieljhegeman: I've noticed that we basically defaulted to not requiring any fields.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -679,7 +678,6 @@ def test__get_collection_verify_body_is_reshaped_correctly__OK(self, mock_config
                 "datasets": [expect_dataset],
                 "doi": None,
                 "links": links,
-                "processing_status": "PENDING",
```
danieljhegeman, line 682: why are you removing this? It seems to me that we shouldn't be removing any test lines where the processing status was `"PENDING"` or `"INITIALIZED"` because that by definition means the dataset wasn't published, and so the processing status should be reported.

danieljhegeman (commented):
Question about one item in the tests

tests/unit/backend/layers/api/test_curation_api.py
Bento007: That would be nice if we could read the yaml and know what fields should be returned, but that would be a significant refactor. If we have time we should go that route.

tests/unit/backend/layers/api/test_curation_api.py
Bento007: because this test only tests a published collection is returned correctly and published collections don't return a processing_status. `test_get_collection_dynamic_fields` tests for processing status in all of the other cases.

danieljhegeman: Yeah I guess what I mean is how did we previously have a _passing test_ that included `"processing_status": "PENDING"` in the `expected_body` when one of the asserts is `self.assertEqual(expected_body, res_body)`. 

Does that mean we were permitting publication of Collections with Datasets that did not have a processing statue of `SUCCESS`?? Or is there a nuance here in the "publication" of test Datasets that allows bypassing of that check, which then meant this test would pass? Because this test appears to have been for a _published_ Collection (version).

My understanding is that we were errantly returning this attribute for published Collections, and NOT that some published Collections had Datasets with statuses other than `SUCCESS`.

Bento007: In the old test case we only cared that a processing_status was in the response, not what the actually value is. This is a nuance of the test. In the wild the frontend would gate keep us from publishing anything with datasets that are erroneous or processing.

Bento007: remember that only the frontend permits publishing of a collection.

ebezzi (approved):

tests/unit/backend/layers/api/test_curation_api.py
danieljhegeman: Ok, thanks for clarifying

>remember that only the frontend permits publishing of a collection.

Right, this wasn't a concern.

Issue #4613, opened 2023-04-17
get_collection_versions returns the same published_at and revised_at timestamp for all versions of a collection


brianraymor:
1. Create collection
1. Add dataset
1. Publish collection
1. Revise collection
1. Add dataset
1. Publish revision

The collection timestamps returned in GET /collection are:

```

 'published_at': '2023-04-14T22:22:55.524100+00:00',
 'revised_at': '2023-04-14T22:43:05.154712+00:00',

```

Per [revised_at and published_at for DatasetVersions](https://github.com/chanzuckerberg/single-cell-data-portal/issues/4413#issuecomment-1485798718), the expectation is that GET /collections/versions will return the publication timestamp for each version in `published_at` and `revised_at` will not be set.

But `GET /collections/versions` returns the original `published_at` for both the original and updated version:

```

'2023-04-14T22:22:55.524100+00:00'
'2023-04-14T22:22:55.524100+00:00'

```

and the same `revised_at` for both the original and updated version:

```

'2023-04-14T22:43:05.154712+00:00'
'2023-04-14T22:43:05.154712+00:00'


```

PR #4661, opened 2023-04-21
fix: return correct published_at values for all Dataset version objects

danieljhegeman:
## Reason for Change

- #4613 

## Changes

- add
- remove
- modify

## Testing steps


backend/layers/business/business.py
```diff
@@ -143,33 +143,10 @@ def get_published_collection_version(self, collection_id: CollectionId) -> Optio
         """
         return self.database_provider.get_collection_mapped_version(collection_id)
 
-    def get_published_collection_version__discover_api(
+    def get_published_collection_version_with_datasets(
```
Bento007, line 146: I don't see enough of a difference between this and `get_collection_version` on line:165 to warrant a new function.

Bento007 (changes requested):
One thing to fix

Bento007 (approved):

Issue #4386, opened 2023-03-22
The Discover API must introduce the concept of collection and dataset versions


brianraymor:
The documentation does not explain the rationale for `dataset_id` and `dataset_version_id` or `collection_id` and `collection_version_id`. 

There are some definitions in [Surfacing Versions](https://docs.google.com/document/d/1wkbV67CI-jWgwjzDAtlpZ4qE3evZdHEaETaK0Y7RnSo/edit#heading=h.8yxeqpft8mig) that could be reused?

```
dataset_id
A unique identifier of a Dataset.

dataset_version_id	
A unique identifier of a Dataset version.
```

Could be simple like:

```
dataset_id

The unique identifier that always references the latest published version of this dataset.

dataset_version_id	

The unique identifier that always identifies a specific published version of this dataset, 
even after subsequent versions of this dataset have been published via revisions.
```

I might include an example of dataset_revision_id (perhaps using uuid(s) to be more realistic and not suggest an incremental set of numbers.)

dataset_id is 1.
dataset_version_id is 2.

The dataset is updated and republished.
dataset_id is 1.
dataset_version id is 3.

The dataset is updated again and republished.
dataset_id is 1.
dataset_version_id is 4.

It's still possible to access and download earlier versions such as dataset_version_id 3.

PR #4670, opened 2023-04-25
docs: introduce Collection versions and Dataset versions

danieljhegeman:
## Reason for Change

- #4386 

## Changes

- add clear description of how the concept of 'version' impinges on Collections/Datasets and interacting with them via the Discover API.

## Testing steps

- https://dan-docs-backend.rdev.single-cell.czi.technology/curation/ui/#/

### DIFF
--- a/backend/curation/api/curation-api.yml
+++ b/backend/curation/api/curation-api.yml
@@ -554,7 +554,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "#/components/schemas/collection"
+                  $ref: "#/components/schemas/collection_version"
         "403":
           $ref: "#/components/responses/403"
         "404":
--- a/backend/curation/api/curation-api.yml
+++ b/backend/curation/api/curation-api.yml
@@ -26,12 +26,12 @@ info:
     #### Version identifiers
     While the *Collection* identifier `collection_id` will always refer to the latest published instance of a Collection
     (like the "head" in version-control terminology), and must be used with `../collections/..` endpoints, the
-    *Collection version* identifier `collection_version_id` points always to one specific published instance (version)
+    *Collection Version* identifier `collection_version_id` points always to one specific published instance (version)
     of a Collection (like a commit hash) and must be used with `../collection_version/..`
-    endpoints. The same is true for *Datasets* and *Dataset versions*, though it is important to note that if a given Dataset
-    is not updated during the course of a revision of its parent Collection, then its *Dataset version* identifier
-    `dataset_version_id` will not change and will appear identically in the list of *Dataset versions* associated with the
-    Collection version from *before* the revision as well as with the Collection version from *after* the revision. Private Collections,
+    endpoints. The same is true for *Datasets* and *Dataset Versions*, though it is important to note that if a given Dataset
+    is not updated during the course of a revision of its parent Collection, then its *Dataset Version* identifier
+    `dataset_version_id` will not change and will appear identically in the list of *Dataset Versions* associated with the
+    Collection Version from *before* the revision as well as with the Collection Version from *after* the revision. Private Collections,
     private revisions of published Collections, and their constituent (private) Datasets are not available as *versions*
     until they are published, and therefore must be referenced via their `collection_id` and `dataset_id` identifiers
     using `../collections/..` and `../datasets/..` endpoints, respectively. As such, mutation actions cannot be
@@ -536,14 +536,14 @@ paths:
 
   /v1/collections/{collection_id}/versions:
     get:
-      summary: Fetch all versions of a Collection with corresponding Dataset versions
+      summary: Fetch all versions of a Collection with corresponding Dataset Versions
       description: >
-        Fetch full metadata for all versions of a Collection and their corresponding Dataset versions. The
-        `collection_id` must reference a *Collection*, **NOT** a *Collection version*. Collection versions are returned
+        Fetch full metadata for all versions of a Collection and their corresponding Dataset Versions. The
+        `collection_id` must reference a *Collection*, **NOT** a *Collection Version*. Collection Versions are returned
         in reverse-chronological order.
       operationId: backend.curation.api.v1.curation.collections.collection_id.versions.get
       tags:
-        - Collection version
+        - Collection Version
       parameters:
         - $ref: "#/components/parameters/path_collection_id"
       responses:
@@ -564,13 +564,13 @@ paths:
 
   /v1/collection_versions/{collection_version_id}:
     get:
-      summary: Fetch a Collection version with Dataset versions
+      summary: Fetch a Collection Version with Dataset Versions
       description: >
-        Fetch full metadata for a published Collection version and its Dataset versions. The `collection_version_id` 
-        must reference a *Collection version*, **NOT** a *Collection*.
+        Fetch full metadata for a published Collection Version and its Dataset Versions. The `collection_version_id`
+        must reference a *Collection Version*, **NOT** a *Collection*.
       operationId: backend.curation.api.v1.curation.collection_versions.collection_version_id.actions.get
       tags:
-        - Collection version
+        - Collection Version
       parameters:
         - $ref: "#/components/parameters/path_collection_version_id"
       responses:
@@ -610,10 +610,10 @@ paths:
     get:
       summary: List all versions for a Dataset
       tags:
-        - Dataset version
+        - Dataset Version
       description: |
-        List all versions for a given Dataset. The `dataset_id` must reference a *Dataset*, **NOT** a *Dataset version*.
-        Dataset versions are returned in reverse-chronological order.
+        List all versions for a given Dataset. The `dataset_id` must reference a *Dataset*, **NOT** a *Dataset Version*.
+        Dataset Versions are returned in reverse-chronological order.
       operationId: backend.curation.api.v1.curation.datasets.dataset_id.versions.actions.get
       parameters:
         - $ref: "#/components/parameters/path_dataset_id"
@@ -633,11 +633,11 @@ paths:
 
   /v1/dataset_versions/{dataset_version_id}:
     get:
-      summary: Get full metadata for a Dataset version
+      summary: Get full metadata for a Dataset Version
       tags:
-        - Dataset version
+        - Dataset Version
       description: |
-        Fetch Dataset metadata for a Dataset version. The `dataset_version_id` must reference a *Dataset version*,
+        Fetch Dataset metadata for a Dataset Version. The `dataset_version_id` must reference a *Dataset Version*,
         **NOT** a *Dataset*.
       operationId: backend.curation.api.v1.curation.dataset_versions.dataset_version_id.actions.get
       parameters:
@@ -869,7 +869,7 @@ components:
           $ref: "#/components/schemas/visibility"
       type: object
     collection_version:
-      description: Collection version metadata
+      description: Collection Version metadata
       properties:
         collection_id:
           $ref: "#/components/schemas/collection_id"
@@ -889,7 +889,7 @@ components:
           $ref: "#/components/schemas/curator_name"
         datasets:
           type: array
-          description: A list of Dataset metadata associated with the Collection version
+          description: A list of Dataset metadata associated with the Collection Version
           items:
             $ref: "#/components/schemas/dataset_version"
         description:
@@ -1138,7 +1138,7 @@ components:
           $ref: "#/components/schemas/collection_id"
         collection_version_id:
           description: |
-            The *Collection Version* in which this *Dataset version* was first published.
+            The *Collection Version* in which this *Dataset Version* was first published.
           type: string
           format: uuid
         dataset_id:
--- a/backend/curation/api/curation-api.yml
+++ b/backend/curation/api/curation-api.yml
@@ -37,22 +37,6 @@ info:
     using `../collections/..` and `../datasets/..` endpoints, respectively. As such, mutation actions cannot be
     performed using version identifiers.
 
-    #### Interpreting `published_at` and `revised_at` timestamps with versions
-    When retrieving data about *Collections* or *Datasets*, these fields convey the following:
-
-    * `published_at`: the timestamp of the first publication of the *Collection*/*Dataset*, i.e., the publication
-    of the first version.
-
-    * `revised_at`: the timestamp of the publication of the most recent revision of the *Collection*/*Dataset*
-
-    However, when retrieving data about *Collection versions* or *Dataset versions*:
-
-    * `published_at`: the timestamp of the publication of this particular version
-
-    Therefore, in the case of a Collection, the `published_at` timestamp for its latest *Collection version* will be
-    equal to the `revised_at` timestamp for the *Collection*. There is no `revised_at` attribute associated with an
-    entity version.
-
 servers:
   - description: Production environment
     url: https://api.cellxgene.cziscience.com/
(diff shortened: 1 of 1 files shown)


danieljhegeman:
@brianraymor I opted for a precise explanation over using examples, because I felt the added verbiage was going to be required either way. In the associated issue description you indicated you _might_ like an example; if this has changed to _definitely want_, let me know.

But I think it is clear as-is.

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,45 @@ info:
     ## Data Model
     <!-- Data model diagram created via Lucid: https://lucid.app/lucidchart/8d085832-34f4-4413-9c1d-44e77d06e326/edit?invitationId=inv_ed354532-7256-4e0a-9ad0-b622905e5198&page=Vfzkzdv~I1fL# -->
     <img src="/static/discover-api-data-model.png" width="50%"/>
+
+    ## Working with versions
+
+    #### Access to prior published version history
+    The CZ CELLxGENE Discover API employs the concept of *versions* for our two core entities, *Collections* and
+    *Datasets*. This is to permit access to prior published instances (versions) of these entities even as they are
+    changed over time via revisions (at present the Data Portal web application UI offers access to only the latest
```
nayib-jose-gloria, line 23: I would remove this reference to how Data Portal UI handles revisions--it's not related to curation API (and its also not true yet)

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,45 @@ info:
...
+
+    #### Version identifiers
+    While the *Collection* identifier `collection_id` will always refer to the latest published instance of a Collection
+    (like the "head" in version-control terminology), and must be used with `../collections/..` endpoints, the
+    *Collection version* identifier `collection_version_id` points always to one specific published instance (version)
+    of a Collection (like a commit hash) and must be used with `../collection_version/..`
+    endpoints. The same is true for *Datasets* and *Dataset versions*, though it is important to note that if a given Dataset
+    is not updated during the course of a revision of its parent Collection, then its *Dataset version* identifier
+    `dataset_version_id` will not change and will appear identically in the list of *Dataset versions* associated with the
+    Collection version from *before* the revision as well as with the Collection version from *after* the revision. Private Collections,
```
nayib-jose-gloria, line 34: This note reads a little confusing to me. What about something like 

'The same is true for Datasets and Dataset Versions. The same Dataset Version will persist across Collection Versions if the Dataset is unchanged. A new Dataset Version is created only when a Collection Version makes a change to that Dataset.'

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,45 @@ info:
...
+    private revisions of published Collections, and their constituent (private) Datasets are not available as *versions*
+    until they are published, and therefore must be referenced via their `collection_id` and `dataset_id` identifiers
```
nayib-jose-gloria, line 36: I think it would be clearer if we defined where you can use version IDs instead of where you can't. Something like

 'version identifiers are used exclusively to fetch prior published versions of Collections and Datasets, via the collection_version and dataset_version endpoints. You may also fetch the metadata for all the prior published versions of a Collection or Dataset using collection_id or dataset_id using the '/versions' endpoints'

backend/curation/api/curation-api.yml
danieljhegeman: Hold on. It is true. The Data Portal UI does not permit access to past versions.

danieljhegeman: Do you mean you want me to remove the parenthetical note? Or the entire sentence?

backend/curation/api/curation-api.yml
danieljhegeman: I like the first two sentences that you suggest. Trying to think of how to word the third one.

backend/curation/api/curation-api.yml
nayib-jose-gloria: Just remove the parenthetical note.

Also, you can put a collection version ID in the URL of the Data Portal UI and it'll render that prior published version. Albeit there is a ticket to redirect, its not implemented yet

Ex. 
https://api.cellxgene.dev.single-cell.czi.technology/curation/v1/collections/5e512a4a-4f42-4c1a-8220-cfb9df2e7421/versions

https://cellxgene.dev.single-cell.czi.technology/collections/5e512a4a-4f42-4c1a-8220-cfb9df2e7421

https://cellxgene.dev.single-cell.czi.technology/collections/eacc1698-85f7-4583-bc12-1668af7613f3

backend/curation/api/curation-api.yml
danieljhegeman: I'm aligning the tags strictly with which endpoints use which type of identifier

backend/curation/api/curation-api.yml
danieljhegeman: Interesting... ok

Got it, we'll skip this.

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    *Datasets*. This is to permit access to prior published instances (versions) of these entities even as new updates
+    for them are published.
+
+    #### Version identifiers
+    While the *Collection* identifier `collection_id` will always refer to the latest published instance of a Collection
+    (like the "head" in version-control terminology), and must be used with `Collection` endpoints, the
+    *Collection Version* identifier `collection_version_id` points always to one specific published instance (version)
+    of a Collection (like a commit hash) and must be used with `Collection Version` endpoints.
+    endpoints. The same is true for *Datasets* and *Dataset Versions*, though it is important to note that if a given Dataset
+    is not updated during the course of a revision of its parent Collection, then its *Dataset Version* identifier
```
nayib-jose-gloria, line 31: nit: I like this better, but I still think it's a bit clunky talking about the identifier changing or not changing instead of saying a Dataset Version is persisted or a new one is created. It's not just the identifier that changes if a Dataset has a new Version, so it feels incomplete to hone in on that and it's implied that the version ID will stay the same if it's the same Dataset Version.

danieljhegeman: I've tried to split the difference and refer to both 🙂

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    #### Access to prior published version history
+    The CZ CELLxGENE Discover API employs the concept of *Versions* for our two core entities, *Collections* and
+    *Datasets*. This is to permit access to prior published instances (versions) of these entities even as new updates
+    for them are published.
+
+    #### Version identifiers
+    While the *Collection* identifier `collection_id` will always refer to the latest published instance of a Collection
+    (like the "head" in version-control terminology), and must be used with `Collection` endpoints, the
+    *Collection Version* identifier `collection_version_id` points always to one specific published instance (version)
+    of a Collection (like a commit hash) and must be used with `Collection Version` endpoints. The same is true for
```
Bento007, line 29: I would link do an example of a commit hash or a definition for the uninitiated.

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
     ## Data Model
     <!-- Data model diagram created via Lucid: https://lucid.app/lucidchart/8d085832-34f4-4413-9c1d-44e77d06e326/edit?invitationId=inv_ed354532-7256-4e0a-9ad0-b622905e5198&page=Vfzkzdv~I1fL# -->
     <img src="/static/discover-api-data-model.png" width="50%"/>
+
+    ## Working with versions
+
+    #### Access to prior published version history
```
brianraymor, line 20: I don't think "prior" is required because it's implicit in "history"

Accessing published versions of collections and datasets

Bento007 (approved):

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    The CZ CELLxGENE Discover API employs the concept of *Versions* for our two core entities, *Collections* and
+    *Datasets*. This is to permit access to prior published instances (versions) of these entities even as new updates
+    for them are published.
+
```
brianraymor, line 24: The CZ CELLxGENE Discover API supports access to all published versions of Collections and Datasets. For example, if a Dataset was published and then later updated and published again, then both Dataset Versions can be accessed.

brianraymor: General comment. It's help to have line breaks for reviewers of markdown. Otherwise, it's hard to place the review comment.

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    #### Version identifiers
+    While the *Collection* identifier `collection_id` will always refer to the latest published instance of a Collection
+    (like the "head" in version-control terminology), and must be used with `Collection` endpoints, the
```
brianraymor, line 27: The *Collection* identifier `collection_id` refers to the most recent published version of a Collection.

brianraymor: **(like the "head" in version-control terminology)**

Are our consumers assumed to be familiar with version-control terminology? Why would this need to be introduced?

brianraymor, line 27: General comment - Why keep intermingling _instance_ and _version_?

backend/curation/api/curation-api.yml
brianraymor, line 29: ~like a commit hash~

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    *Collection Version* identifier `collection_version_id` points always to one specific published instance (version)
+    of a Collection (like a commit hash) and must be used with `Collection Version` endpoints. The same is true for
+    *Datasets* and *Dataset Versions*, though it is important to note that if a given Dataset
+    is not updated during the course of a revision of its parent Collection, then its Dataset Version (and identifier
+    `dataset_version_id`) will not change even though a new Collection Version has been published. Private Collections,
+    private revisions of published Collections, and their constituent (private) Datasets are not available as Versions
```
brianraymor, line 33: Datasets that are not updated during a revision are not versioned - their Dataset Version do not change ...

danieljhegeman: Good point. I'll specify _updated_ version of those Datasets.

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    until they are published, and therefore must be referenced via their `collection_id` and `dataset_id` identifiers
+    using `Collection` and `Dataset` endpoints. As such, mutation actions cannot be
```
brianraymor, line 35: _mutation actions_ - like X-Men?

danieljhegeman: This is the correct CRUD term I believe 😄

danieljhegeman: @Bento007 for opinion on this

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+
+    #### Access to prior published version history
+    The CZ CELLxGENE Discover API employs the concept of *Versions* for our two core entities, *Collections* and
+    *Datasets*. This is to permit access to prior published instances (versions) of these entities even as new updates
+    for them are published.
+
+    #### Version identifiers
+    While the *Collection* identifier `collection_id` will always refer to the latest published instance of a Collection
+    (like the "head" in version-control terminology), and must be used with `Collection` endpoints, the
+    *Collection Version* identifier `collection_version_id` points always to one specific published instance (version)
```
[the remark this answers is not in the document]
brianraymor: In general, you want to use similar language - "refer" was used above so there's no reason to switch to "points" in this case?

backend/curation/api/curation-api.yml
danieljhegeman: That's a bit of carryover from initially wanting to separate the explanation from the terminology, but now I'm doing that with `version` and `Version`, so I can replace `instance` with `version`.

danieljhegeman: No, but it is a useful analogy.

You want to get rid of it?

danieljhegeman: wrt the beginning of the sentence: I think "While..." makes sense since the focus is on explaining the _version_ identifiers, which happens in the second half of the sentence.

backend/curation/api/curation-api.yml
danieljhegeman: A published version could be the most recent version

A prior published version is a published version that is not the most recent version.

No?

backend/curation/api/curation-api.yml
brianraymor: ~Private Collections, private revisions of published Collections, and their constituent (private) Datasets are not available as Versions~

Versions were defined as published "instances" of collections and datasets, so I don't believe that this point needs to be stressed. Only curators are going to be familiar with private cases, so we don't need to confuse consumers with this detail?

backend/curation/api/curation-api.yml
danieljhegeman: You mean one line per sentence...? I'm not quite sure what you mean.

backend/curation/api/curation-api.yml
```diff
@@ -14,6 +14,27 @@ info:
...
+    of a Collection (like a commit hash) and must be used with `Collection Version` endpoints. The same is true for
+    *Datasets* and *Dataset Versions*, though it is important to note that if a given Dataset
+    is not updated during the course of a revision of its parent Collection, then its Dataset Version (and identifier
+    `dataset_version_id`) will not change even though a new Collection Version has been published. Private Collections,
+    private revisions of published Collections, and their constituent (private) Datasets are not available as Versions
+    until they are published, and therefore must be referenced via their `collection_id` and `dataset_id` identifiers
+    using `Collection` and `Dataset` endpoints. As such, mutation actions cannot be
+    performed using Version identifiers.
+
```
brianraymor, line 37: _mutation actions_ - like X-Men?

backend/curation/api/curation-api.yml
danieljhegeman: Changing to
> Private Collections,
    private revisions of published Collections, and any new or updated Dataset are not available as Versions
    until they are published, and

backend/curation/api/curation-api.yml
danieljhegeman: This is the correct CRUD term I believe 😄 cc @Bento007

backend/curation/api/curation-api.yml
danieljhegeman: >Only curators are going to be familiar with private cases, so we don't need to confuse consumers with this detail?

We include, by necessity, other pieces of curator-specific documentation. But, if you feel this is obviated by the explicit declaration of _versions_ being _published history_, then I can (and will) delete this.

backend/curation/api/curation-api.yml
brianraymor: Maximum number of characters per line ... 

See https://about.gitlab.com/blog/2016/10/11/wrapping-text/ or https://peps.python.org/pep-0008/#maximum-line-length ...

brianraymor: Basically, I should be able to comment line by line in a PR.

danieljhegeman: You can comment line by line. I was just following our PyCharm guide for line length, which is longer than is suggested by the links you put above

backend/curation/api/curation-api.yml
brianraymor: > No, but it is a useful analogy.
> You want to get rid of it?

Absolutely. It's not required to explain versions and assumes a shared background in version-control.

PR #4686, opened 2023-04-26
docs: add clarity around version identifier for Datasets

danieljhegeman:
## Reason for Change

Final touchup in the introduction for versions


Bento007 (approved):

backend/curation/api/curation-api.yml
```diff
@@ -25,9 +25,12 @@ info:
     While the *Collection* identifier `collection_id` will always refer to the latest published version of a Collection,
     and must be used with `Collection` endpoints, the *Collection Version* identifier `collection_version_id`
     refers to one specific published version of a Collection and must be used with `Collection Version` endpoints.
-    The same is true for *Datasets* and *Dataset Versions*, though it is important to note that if a given Dataset
-    is not updated during the course of a revision of its parent Collection, then its Dataset Version (and identifier
-    `dataset_version_id`) will not change even though a new Collection Version has been published.
+    The same is true for the *Dataset* identifier `dataset_id` and the *Dataset Version* identifier
+    `dataset_version_id`.
+
```
brianraymor, line 30: Rewrite:

The *Collection* identifier `collection_id` refers to the latest published version of a Collection and and is required by Collection endpoints.

The *Collection Version* identifier `collection_version_id` refers to one specific published version of a Collection and is required by the Collection Version endpoint.


The *Dataset* identifier `dataset_id` refers to the latest published version of a Dataset and and is required by Dataset endpoints.

The *Dataset Version* identifier `dataset_version_id` refers to one specific published version of a Dataset and is required by the Dataset Version endpoint.

---

It might also be useful to have Dataset Version and Collection Version link to their sections in the swagger for clarity (there's only one endpoint for each.)

backend/curation/api/curation-api.yml
```diff
@@ -25,9 +25,12 @@ info:
...
+    **Important Note:** if a given Dataset is not updated during the course of a
```
brianraymor, line 31: ~given~

_will not change **even though a new Collection Version has been published**_

A revision may be cancelled, which is why I rewrote in slack as:

Important Note: if a Dataset is not updated during a revision of its parent Collection, then its Dataset Version (and identifier `dataset_version_id`) do not change when the revision is published.

if could also be _if the revision is published_. 

If you want to stress that there's still a new Collection Version id, then:

Important Note: if a Dataset is NOT updated during a revision of its parent Collection and the revision is published, then its Dataset Version (and identifier `dataset_version_id`) do not change although there is a new Collection Version (and identifier `collection_version_id`).

brianraymor (changes requested):
See my comments.

backend/curation/api/curation-api.yml
danieljhegeman: addressed

backend/curation/api/curation-api.yml
danieljhegeman: addressed

danieljhegeman: >It might also be useful to have Dataset Version and Collection Version link to their sections in the swagger for clarity (there's only one endpoint for each.)

I've tried to figure out how to link to other internal sections of the swagger and I have been unsuccessful in that effort

brianraymor: How was this more detailed rewrite addressed?

danieljhegeman: >Rewrite:

The Collection identifier collection_id refers to the latest published version of a Collection and is required by Collection endpoints.

The Collection Version identifier collection_version_id refers to one specific published version of a Collection and is required by the Collection Version endpoint.

The Dataset identifier dataset_id refers to the latest published version of a Dataset and and is required by Dataset endpoints.

The Dataset Version identifier dataset_version_id refers to one specific published version of a Dataset and is required by the Dataset Version endpoint.

danieljhegeman: But I'm guessing you mean to replace the whole `#### Version identifiers` section

brianraymor: That would be correct.

brianraymor: We have not captured that the values of collection_id and dataset_id are immutable once the specific collection or dataset is created. 

Perhaps:

The Collection identifier collection_id is set during the creation of a Collection and never changes if the Collection is updated. It always refers to the latest published version of a Collection and is required by Collection endpoints. 

The Dataset identifier dataset_id is set during the creation of a Dataset and never changes if the Dataset is updated. It always refers to the latest published version of a Dataset and is required by Dataset endpoints. 

Thoughts?

brianraymor (changes requested):
https://github.com/chanzuckerberg/single-cell-data-portal/pull/4686/files#r1178480484 was resolved without discussion.

backend/curation/api/curation-api.yml
danieljhegeman: Generally good, but consider my slight edits:

```
The Collection identifier collection_id is set during the creation of a Collection; it does not change, even if the Collection is updated. It always refers to the latest published version of the Collection and is the required identifier to use with Collection endpoints.

The Dataset identifier dataset_id is set during the creation of a Dataset; it does not change, even if the Dataset is updated. It always refers to the latest published version of the Dataset and is the required identifier to use with Dataset endpoints.
```

danieljhegeman: I do think we should be explicit about the canonical identifier, though, so good catch that we hadn't added any verbiage to that effect.

brianraymor: Your edits are even more clarifying. LGTM.

backend/curation/api/curation-api.yml
```diff
@@ -21,13 +21,24 @@ info:
...
     if a Dataset was published and then later updated and published again, then both Dataset Versions can be accessed.
 
-    #### Version identifiers
-    While the *Collection* identifier `collection_id` will always refer to the latest published version of a Collection,
-    and must be used with `Collection` endpoints, the *Collection Version* identifier `collection_version_id`
-    refers to one specific published version of a Collection and must be used with `Collection Version` endpoints.
-    The same is true for *Datasets* and *Dataset Versions*, though it is important to note that if a given Dataset
-    is not updated during the course of a revision of its parent Collection, then its Dataset Version (and identifier
-    `dataset_version_id`) will not change even though a new Collection Version has been published.
+    #### Different types of identifiers
```
danieljhegeman, line 24: @brianraymor I've changed the heading from `Version identifiers` to this

brianraymor (approved):
Thanks for iterating.

PR #4692, opened 2023-04-27
chore: prod deploy; 4/27/23

nayib-jose-gloria:
Staging branch, minus changes that are dependent on prod infra changes that would require downtime. We will re-apply those changes to prod at a later date


.github/workflows/deploy-happy-stack.yml
```diff
@@ -70,7 +70,7 @@ jobs:
             DOMAIN_NAME=public-frontend.production.single-cell.czi.technology
             ALIAS=cellxgene.cziscience.com
           else
-            DOMAIN_NAME=public-frontend.${DEPLOYMENT_STAGE}.single-cell.czi.technology
+            DOMAIN_NAME=frontend.${DEPLOYMENT_STAGE}.single-cell.czi.technology
```
Bento007, line 73: We can keep that change that was previous on line 70, removing the public prefix.

Bento007: nevermind this is temporary and will not be merged back to main. ignore my above comment

Bento007 (approved):

Issue #4578, opened 2023-04-14
Dev spike for creating a well performing authenticated /user-datasets API endpoint to return all private datasets the user can see


NoopDog:
- [ ] Refine the requirement for this to determine precisely what datasets to return and how to represent datasets for a collection revision where the dataset has not changed but the collection has and the dataset now belongs to both the public collection and the private revision.

- [ ] Determine if the required data can be efficiently retrieved.

- [ ] Review result with BE team

API is
- authenticated
- not cached in AWS Cloud Front

Issue #4577, opened 2023-04-14
Add an authenticated /user-collections/index endpoint to return private collections and revisions visible to the curator


NoopDog:
#### NEED
To support filtering in the collections page, we need a list of collections that have all public collections and also all of the private collections and revisions the curator can see. (Where they have WRITE ACCESS).

- For creators, these are collections and revisions they created and where they are the "owner".
- For super curators, this is all private collections and revisions.

#### Approach

Add a /user-collections/index authenticated endpoint that returns the extra rows as specified above and also the additional collection fields:
- acces_type
- curator_name
- revision_of
- revising_in

NoopDog:
Complete

PR #4726, opened 2023-05-02
feat: user and dataset collections (#4577, #4578)

NoopDog:
## Reason for Change

- #TICKET_NUMBER
#4577,
#4578

## Changes

- adds API endpoints for /user-collections/index and /user-datasets/index

## Testing steps

- Could you describe how you made sure to know that your changes worked? Should allow someone else to go verify your code without in depth knowledge.

### DIFF
--- a/backend/portal/api/portal-api.yml
+++ b/backend/portal/api/portal-api.yml
@@ -1080,6 +1080,8 @@ components:
           type: string
         publisher_metadata:
           $ref: "#/components/schemas/publisher_metadata"
+        revising_in:
+          $ref: "#/components/schemas/revising_in"
         datasets:
           type: array
           items:
@@ -1113,7 +1115,13 @@ components:
           type: number
         published_year:
           type: number
-
+    revising_in:
+      type: string
+      description: >-
+        If the Collection is published and has an outstanding private Revision, and the user has write access, then
+        `revising_in` is set to the id of the private Revision. The value is `None` if the user is not authorized or no
+        private Revision exists.
+      nullable: true
     ontology_element:
       type: object
       properties:
--- a/backend/portal/api/portal-api.yml
+++ b/backend/portal/api/portal-api.yml
@@ -418,6 +418,54 @@ paths:
         "400":
           $ref: "#/components/responses/400"
 
+  /v1/user-collections/index:
+    get:
+      tags:
+        - collections
+      summary: >-
+        Returns all public, non tombstoned collections. This endpoint is meant to be used for displaying collections
+        in a compact view. It doesn't return all the fields, so it is not suitable for general usage.
+      security:
+        - cxguserCookie: []
+      operationId: backend.portal.api.portal_api.get_user_collection_index
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  required:
+                    - id
+                    - name
+                  properties:
+                    access_type:
+                      $ref: "#/components/schemas/access_type"
+                    curator_name:
+                      type: string
+                    id:
+                      $ref: "#/components/schemas/collection_id"
+                    name:
+                      type: string
+                    publisher_metadata:
+                      $ref: "#/components/schemas/publisher_metadata"
+                    published_at:
+                      type: number
+                      nullable: true
+                    revised_at:
+                      type: number
+                      nullable: true
+                    consortia:
+                      $ref: "#/components/schemas/consortia"
+                    visibility:
+                      $ref: "#/components/schemas/visibility"
+        "400":
+          $ref: "#/components/responses/400"
+        "401":
+          $ref: "#/components/responses/401"
+
   /v1/datasets/{dataset_id}:
     delete:
       tags:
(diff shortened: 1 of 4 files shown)


backend/portal/api/portal-api.yml
```diff
@@ -418,6 +418,54 @@ paths:
         "400":
           $ref: "#/components/responses/400"
 
+  /v1/user-collections/index:
+    get:
+      tags:
+        - collections
+      summary: >-
+        Returns all public, non tombstoned collections. This endpoint is meant to be used for displaying collections
+        in a compact view. It doesn't return all the fields, so it is not suitable for general usage.
```
danieljhegeman, line 427: This summary is copied/pasted from the above route. Let's make it accurate.

backend/portal/api/portal-api.yml
```diff
@@ -418,6 +418,54 @@ paths:
         "400":
           $ref: "#/components/responses/400"
 
+  /v1/user-collections/index:
```
danieljhegeman, line 421: is this route essentially returning the same stuff as [this route](https://github.com/chanzuckerberg/single-cell-data-portal/blob/6bf761664ad4e0db8157dd36392e8f64653d79b0/backend/portal/api/portal-api.yml#L16-L17) except for a narrower selection of Collections (i.e., only those for which the user is authorized)? If these routes serve the same general purpose, we should consider collapsing them so that you merely pass a query param to the existing route: `GET /v1/collections?owner=true` or similar.

danieljhegeman: The old route that has been there forever is pretty poorly-defined in terms of response object... 😞

backend/portal/api/portal-api.yml
danieljhegeman: Also, the summary should be a short sentence. Further detail should be provider in a `description:` block, as is being done with the datasets endpoint you added.

backend/portal/api/portal-api.yml
```diff
@@ -702,7 +750,88 @@ paths:
                       type: number
         "400":
           $ref: "#/components/responses/400"
-
+  /v1/user-datasets/index:
+    get:
+      tags:
+        - datasets
+      summary: >-
+        Returns all public and private, non tombstoned datasets where the user has WRITE access. This endpoint is meant to be used for displaying datasets
```
danieljhegeman, line 758:
```suggestion
        Returns all public and private non-tombstoned Datasets where the user has WRITE access. This endpoint is meant to be used for displaying Datasets
```

backend/portal/api/portal-api.yml
```diff
@@ -702,7 +750,88 @@ paths:
...
+        in a compact view. It doesn't return all the fields, so it is not suitable for general usage.
```
danieljhegeman, line 759: same thing; update summary

danieljhegeman: Although it appears actually you have added the `WRITE access` detail already... is the summary otherwise as you intend it?

backend/portal/api/portal-api.yml
```diff
@@ -702,7 +750,88 @@ paths:
...
+      security:
+        - cxguserCookieLenient: []
+        - {}
+      description: >-
+        Returns all datasets that belong to public and private, non tombstoned collections where the user has WRITE access. This endpoint is meant to be used 
+        for displaying datasets in a compact view. It doesn't return all the fields, so it is not suitable for general usage.
```
danieljhegeman, line 765: This is a good spot for detail. Remove the extra verbiage from the summary section and keep it here. Check to make sure it is accurate.

backend/portal/api/portal_api.py
```diff
@@ -434,6 +425,72 @@ def get_collection_index():
...
 
 
+def get_user_private_collections(token_info: Optional[dict] = None):
+    user_info = UserInfo(token_info)
+
+    # Get all private collections the user has write access to
+    if user_info.is_none():
+        all_user_private_collections = []
+    elif user_info.is_super_curator():
+        all_user_private_collections = get_business_logic().get_collections(CollectionQueryFilter(is_published=False))
```
danieljhegeman, line 435: this is a better var name than in the previous implementation above, before you modularized it 👍

backend/portal/api/portal_api.py
```diff
@@ -434,6 +425,72 @@ def get_collection_index():
...
+            transformed_collection["id"] = collection.collection_id.id
+
+        if not collection.is_published():
+            transformed_collection["revision_of"] = collection.collection_id.id
+        else:
+            transformed_collection["revision_of"] = None
+
+        if collection.is_published():
+            transformed_collection["published_at"] = collection.canonical_collection.originally_published_at
+            transformed_collection["revised_at"] = collection.published_at
```
danieljhegeman, line 484: if this Collection has only been published one time, then this value should probably be `None`. Though I don't know what the FE is expecting... if the FE accepts such a value, then it should be `None`.

danieljhegeman: Let's make sure the tests cover this code path

backend/portal/api/portal_api.py
```diff
@@ -680,8 +736,34 @@ def get_datasets_index():
...
 
+    datasets = get_business_logic().get_all_mapped_datasets()
+    response = enrich_dataset_response(datasets)
+
+    return make_response(jsonify(response), 200)
+
+
+def get_user_datasets_index(token_info: dict):
+    """
+    Returns a list of all datasets associated with public or private collections where
```
danieljhegeman, line 747:
```suggestion
    Returns a list of all Datasets associated with public Collections or with private Collections where
```

danieljhegeman: Pls confirm my change is the intended desription

danieljhegeman (changes requested):
See comments and suggestions. Most, but not all, are docs-related.

backend/portal/api/portal_api.py
NoopDog: TY!

backend/portal/api/portal_api.py
NoopDog: revised_at == published_at is the standard way to indicate that a collection has never been revised. This line exists (once ) elsewhere so so I think this is ok.

backend/portal/api/portal-api.yml
danieljhegeman: @NoopDog confirm

backend/portal/api/portal-api.yml
danieljhegeman: @NoopDog move detailed description to `description` block

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -780,6 +822,130 @@ def test__get_all_collections_for_index(self):
...
+        response = self.app.get(test_url.url, headers=headers)
+        self.assertEqual(200, response.status_code)
+        body = json.loads(response.data)
+
+        ids = [collection["id"] for collection in body]
+        self.assertIn(collection.collection_id.id, ids)
+        self.assertNotIn(private_collection.collection_id.id, ids)
+        self.assertNotIn(private_collection.version_id.id, ids)
+        self.assertNotIn(collection_to_tombstone.collection_id.id, ids)
+        self.assertNotIn(collection_to_tombstone.version_id.id, ids)
```
danieljhegeman, line 849: @NoopDog (regarding the behaviors you're testing here and above) this is already the behavior we expect: that users do not see their private Collections when on the `Collections` page; only when they are on the `My Collections` page do they see them. 

I guess we just lacked tests for this before?

backend/portal/api/portal-api.yml
```diff
@@ -418,6 +418,54 @@ paths:
...
+                      nullable: true
+                    revised_at:
+                      type: number
+                      nullable: true
+                    consortia:
+                      $ref: "#/components/schemas/consortia"
+                    visibility:
+                      $ref: "#/components/schemas/visibility"
+        "400":
+          $ref: "#/components/responses/400"
```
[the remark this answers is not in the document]
danieljhegeman: I think it only returns `401`

danieljhegeman:
@NoopDog I'm a little confused on the following point: the `user-collections` and `user-datasets` endpoints do not accept query params; they read the user info from the cxg cookie. In this case, those two endpoints should mirror each other in terms of error responses in the API definitions, which at the moment they do not. 

Also, didn't I see you demo a drop-down 'curator name' selector that super curators could use to limit the selection of a user's Collections/Datasets? Is that FE work part of this epic, and if so, what endpoint is it going to use? Unclear to me

tests/unit/backend/layers/api/test_portal_api.py
NoopDog: Correct.

backend/portal/api/portal-api.yml
NoopDog: Updated summary to move detail to description.

backend/portal/api/portal-api.yml
NoopDog: This is updated thanks.

backend/portal/api/portal-api.yml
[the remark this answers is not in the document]
NoopDog: Hmm,  maybe 400 should come off this and collections/index?

backend/portal/api/portal_api.py
```diff
@@ -680,8 +746,34 @@ def get_datasets_index():
...
+def get_user_datasets_index(token_info: dict):
+    """
+    Returns a list of all Datasets associated with public Collections or with private Collections where
+    the user has WRITE access and the collection is not a revision.
+    """
+    public_datasets = get_business_logic().get_all_mapped_datasets()
+
+    private_collections = get_user_private_collections(token_info)
+    # filter out collections that are revisions
+    non_revisions = [c for c in private_collections if c.is_unpublished_version() is False]
```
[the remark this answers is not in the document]
NoopDog: Updated, thanks.

PR #4758, opened 2023-05-05
fix(upload): increase resources to dataportal upload pipeline 

Bento007:
## Changes
- increase timeout. Now that we support larger datasets, they will take longer to run. Extend the timeout for processing cxgs.
- increase VCPUS. Now that we support larger datasets, give the upload pipeline more CPU to run faster.
- add logging. This will help gauge what parts of the cxg conversion pipeline take the longest.


.happy/terraform/modules/batch/main.tf
```diff
@@ -43,9 +43,9 @@ resource aws_batch_job_definition batch_job_def {
       "value": "${var.frontend_url}"
     }
   ],
-  "vcpus": 2,
+  "vcpus": 8,
```
ebezzi, line 46: What machines are we using in each environment? Ideally, since we allocate a static amount of memory, this should be proportional to the available vCPUs.

ebezzi (approved):

Issue #4575, opened 2023-04-14
Roll up metadata for parent cell types in addition to gene expression


signechambers1:
Using the compare feature, currently a particular cell type's compare options only include options associated with cells that were explicitly labeled as that cell type. We'd like to roll up metadata for parent cell types in addition to gene expression so that users are able to accurately see metadata for parent nodes.

We currently support sex, disease and self reported ethnicity in the compare functionality. Disease has the most options, with brain coming in as the tissue with the most diseases (7 diseases). There should not be any UX problem with showing 7 rows per cell type.

[Relevant slack thread](https://czi-sci.slack.com/archives/C0247APK621/p1681411030831259)

PR #4773, opened 2023-05-08
fix: Fix rollup implementation to include compare dimension

atarashansky:
## Reason for Change

- #4575 

## Changes

Modified the gene expression rollup code to also rollup the compare dimension cell counts correctly. Particularly, the change is in computation for set of ancestor cell types that should included rolled up numeric values from its descendant cell types.

## Testing steps

- Visual inspection of the dot plot and cell counts in [rdev](https://rdev-ps-fix-rollup2-frontend.rdev.single-cell.czi.technology/gene-expression) based on manual testing.
- Added unit tests for the top level rollup functionality which is public function
- Added unit tests for the private functions that the public rollup function calls

### DIFF
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -71,9 +71,8 @@ def query():
                     expression_summary=build_expression_summary(dot_plot_matrix_df, compare),
                     term_id_labels=dict(
                         genes=build_gene_id_label_mapping(criteria.gene_ontology_term_ids),
-                        cell_types=_aggregated_cell_counts_by_tissue_cell_type_with_metadata(
-                            cell_counts_cell_type_agg,
-                            compare,
+                        cell_types=build_ordered_cell_types_by_tissue(
+                            cell_counts_cell_type_agg, snapshot.cell_type_orderings, compare
                         ),
                     ),
                 )
@@ -623,7 +622,7 @@ def build_ordered_cell_types_by_tissue(
     joined = distinct_tissues_cell_types[indexer_bool_filter]
 
     for column in cell_type_orderings:
-        joined[column] = list(cell_type_orderings[column][indexer])
+        joined[column] = list(cell_type_orderings[column][indexer_filter])
 
     # Remove cell types without counts
     joined = joined[joined["n_total_cells"].notnull()]
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -72,11 +72,9 @@ def query():
                     term_id_labels=dict(
                         genes=build_gene_id_label_mapping(criteria.gene_ontology_term_ids),
                         cell_types=build_ordered_cell_types_by_tissue(
-                            cell_counts,
-                            cell_counts_cell_type_agg.T,
+                            cell_counts_cell_type_agg,
                             snapshot.cell_type_orderings,
                             compare,
-                            group_by_terms,
                         ),
                     ),
                 )
@@ -385,18 +383,14 @@ def build_ontology_term_id_label_mapping(ontology_term_ids: Iterable[str]) -> Li
 
 # getting only cell type metadata, no genes
 def build_ordered_cell_types_by_tissue(
-    cell_counts: DataFrame,
-    cell_counts_cell_type_agg_T: DataFrame,
+    cell_counts_cell_type_agg: DataFrame,
     cell_type_orderings: DataFrame,
     compare: str,
-    group_by_terms: List[str] = None,
 ) -> Dict[str, Dict[str, Dict[str, Any]]]:
-    if group_by_terms is None:
-        group_by_terms = DEFAULT_GROUP_BY_TERMS
 
-    distinct_tissues_cell_types: DataFrame = cell_counts.groupby(group_by_terms, as_index=False).first()[
-        group_by_terms + ["n_total_cells"]
-    ]
+    distinct_tissues_cell_types = cell_counts_cell_type_agg.reset_index().rename(
+        columns={"n_cells_cell_type": "n_total_cells"}
+    )
 
     # building order for cell types for FE to use
     cell_type_orderings["order"] = range(cell_type_orderings.shape[0])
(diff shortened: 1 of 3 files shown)


backend/wmg/api/v1.py
```diff
@@ -257,39 +254,103 @@ def get_dot_plot_data(
     return dot_plot_matrix_df, cell_counts_cell_type_agg
 
 
-def add_missing_combinations_to_dot_plot_matrix(dot_plot_matrix_df, cell_counts_cell_type_agg) -> DataFrame:
-    ### Add missing cell types to the expression dataframe so they can be rolled up ###
-
+def rollup(gene_expression_df, cell_counts_grouped_df) -> Tuple[DataFrame, DataFrame]:
```
atarashansky, line 257: The new organization is great! The docstrings look good, too.

tests/unit/backend/wmg/api/test_expression_rollup.py
```diff
@@ -54,14 +56,326 @@ def test__cell_types_in_same_lineage_are_colinear(self):
             a, b = cell_types
             assert are_cell_types_colinear(a, b) == expected
 
-    def test__missing_combinations_added_to_expression_dataframe_before_rollup(self):
+
+def _create_cell_counts_df_helper(cell_counts_rows: list[list], columns: list[str], index_cols: list[str]) -> DataFrame:
+    cell_counts_df = pd.DataFrame(cell_counts_rows, columns=columns)
+    cell_counts_df = cell_counts_df.set_index(index_cols)
```
atarashansky, line 62: Didn't know you could do this - so if you pass a dataframe a list of column names in `set_index`, it creates a multi-index automatically? What happens if there are duplicate combinations in the passed index columns?

tests/unit/backend/wmg/api/test_expression_rollup.py
```diff
@@ -54,14 +56,326 @@ def test__cell_types_in_same_lineage_are_colinear(self):
...
+                ["ENSG00000085265", "UBERON:0000955", "CL:0002605", "HANCESTRO:0008", 1, 1, 30, 1000],
+                ["ENSG00000085265", "UBERON:0000955", "CL:0002605", "multiethnic", 1, 1, 40, 1000],
+                ["ENSG00000085265", "UBERON:0000955", "CL:0002627", "HANCESTRO:0008", 1, 1, 20, 1000],
+                ["ENSG00000169429", "UBERON:0000955", "CL:0002627", "multiethnic", 1, 1, 30, 1000],
+            ],
+        }
+
+        return [
+            (
+                test_1["name"],
```
atarashansky, line 318: Are the test names being used anywhere (e.g. output in test logs)? If not, I'd move the names to comments above the different test cases.

tests/unit/backend/wmg/api/test_expression_rollup.py
```diff
@@ -54,14 +56,326 @@ def test__cell_types_in_same_lineage_are_colinear(self):
...
+                ["UBERON:0000955", "CL:0000644", 70],
+                ["UBERON:0000955", "CL:0002605", 80],
+                ["UBERON:0000955", "CL:0002627", 90],
+                ["UBERON:0002113", "CL:0000127", 540],
+                ["UBERON:0002113", "CL:0000644", 70],
+                ["UBERON:0002113", "CL:0002605", 80],
+                ["UBERON:0002113", "CL:0002627", 90],
+            ],
+            "input_gene_expression": [
+                ["ENSG00000085265", "UBERON:0000955", "CL:0000644", 1, 1, 70, 1000],
```
atarashansky, line 247: Let's say that we augment the input gene expression DF with a new numeric column that will be rolled up. In that case, these tests need not be updated since the rollup code is mostly agnostic to which numeric columns are being rolled up.

Is that correct? if not, I'm always hesitant to hardcode all the expected inputs / outputs (even though it is the most transparent) if future updates will require these test fixtures to be manually updated.

tests/unit/backend/wmg/api/test_expression_rollup.py
prathapsridharan: @atarashansky - It is not clear but _I think_ it might throw exception if you set `verify_integrity` argument to `True`: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.set_index.html

For this case, I wanted to create a dataframe containing the rows in `cell_counts_rows` with the multi-index. The assumption (because this is a test) is that rows can be properly multi-indexed in the way specified.

To make it clear, I can set `verify_integrity=True` to make the expectations about the rows clear.

tests/unit/backend/wmg/api/test_expression_rollup.py
prathapsridharan: @atarashansky - yes, the names are used in the test output logs. For example if you run

```
pytest tests/unit/backend/wmg/api/test_expression_rollup.py -v
```

You will see test names in the output:

```
tests/unit/backend/wmg/api/test_expression_rollup.py::TestHighLevelRollupFunction::test__rollup_0_no_compare_dim_all_tissues_have_all_cell_types PASSED                                                                                       [ 57%]
tests/unit/backend/wmg/api/test_expression_rollup.py::TestHighLevelRollupFunction::test__rollup_1_no_compare_dim_one_ancestor_cell_type_missing_in_one_tissue_but_exists_in_all_others PASSED                                                 [ 71%]
tests/unit/backend/wmg/api/test_expression_rollup.py::TestHighLevelRollupFunction::test__rollup_2_with_ethnicity_compare_dim_on_single_tissue PASSED  
```

tests/unit/backend/wmg/api/test_expression_rollup.py
prathapsridharan: @atarashansky - No these test fixtures need not be updated _if you don't want to check that the new column is rolled up_. Note that I am checking that `sum` and `nnz` are rolled up but you might very well deem that that is enough and and additional columns need not be checked and therefore the test fixture need not change.

However, note that if the `rollup` implementation itself changes where you specifically access a particular column by name (Let's say the new column is called `abc`), then `abc` is not in the test fixture and it would pass a dataframe into `rollup` that doesn't contain the column `abc` and therefore the `rollup` function will throw an exception - This is the expected behavior and demonstrates the value of the unit test.

prathapsridharan (approved):
LGTM (on behalf of @atarashansky) - Yes, that sounds suspicious but rest assured it is not. @atarashansky started the branch with code, I added to it, we both reviewed it, and I approved it

PR #4774, opened 2023-05-08
chore: updating robots.txt policy

sethfeingold:
adding a disallow rule to robots.txt file for https://api.cellxgene.cziscience.com/curation/ui/


sethfeingold:
@tihuan I've updated the robots.txt policy with a disallow for the `api.cellxgene.cziscience` subdomain, added a wildcard so any endpoint at that url will be applied

tihuan (approved):
Perfect thanks so much for adding this!

PR #4779, opened 2023-05-09
chore: update push-rdev to use only first 7 github sha characters for image tags

atarashansky:
## Reason for Change
- In elite programmer fashion, I always try to use the github UI to copy and paste the commit hash corresponding to the image tag only to be thwarted by the fact that github only displays the first 7 characters in the abbreviated hash, whereas the images are being tagged with the first 8 characters. 

## Changes

- Modify image tags to use first 7 characters of github SHAs


prathapsridharan (approved):
LGTM

Issue #4741, opened 2023-05-03
Versions of the explorer_url must not be returned in responses


brianraymor:
### Context

See [**single-cell-discover-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1683138788836969). 

`explorer_url`versions should either not be returned in the dataset versions response or the field should be set to its permanent url.

```

'id: 440fcef2-8bce-4964-9593-98d424e3c63f
version id: 3c37b8ae-241d-4be5-812b-08d75722b63e
explorer_url: https://cellxgene.dev.single-cell.czi.technology/e/3c37b8ae-241d-4be5-812b-08d75722b63e.cxg/'

'id: 440fcef2-8bce-4964-9593-98d424e3c63f
version id: bb301370-e85f-4576-9281-95f6a9e9952b
explorer_url: https://cellxgene.dev.single-cell.czi.technology/e/bb301370-e85f-4576-9281-95f6a9e9952b.cxg/'

```

PR #4838, opened 2023-05-15
feat: remove explorer_url from Dataset version object response

danieljhegeman:
## Reason for Change

- #4741 

## Changes

- add
- remove
- modify

## Testing steps

### DIFF
--- a/backend/curation/api/curation-api.yml
+++ b/backend/curation/api/curation-api.yml
@@ -974,7 +974,6 @@ components:
           $ref: "#/components/schemas/donor_id"
         explorer_url:
           $ref: "#/components/schemas/explorer_url"
-          required: true
         is_primary_data:
           $ref: "#/components/schemas/is_primary_data"
         mean_genes_per_cell:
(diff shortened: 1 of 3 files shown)


backend/curation/api/curation-api.yml
```diff
@@ -974,6 +974,7 @@ components:
           $ref: "#/components/schemas/donor_id"
         explorer_url:
           $ref: "#/components/schemas/explorer_url"
+          required: true
```
danieljhegeman, line 977: shouldn't we be using this feature? 🤔 cc @Bento007

Bento007: I found it best to only use `required` when sanitizing requests from the user. Using it on our responses force us to comply with the format, makes our code brittle, and results more 500 responses for our users. It's better to return a response that is missing some fields than throw a 500 at our users. Using the openapi to enforce our output via "required" is heavy handed and ultimately results in a worse experience for our users, because they either get all the info or none of the info based on a mistake outside of their control. We should be validating our responses in unit tests anyways, so using "required" for outputs _should_ not be required. 

If we were not using connexions which interprets and enforces the openapi spec I would say use required since the spec would be more for documentation purposes.

danieljhegeman: Ok, I buy that. Cool.

Bento007 (approved):

PR #4847, opened 2023-05-16
chore: staging<-main

Bento007:
- chore: update push-rdev to use only first 7 github sha characters for image tags (#4779)


Bento007:
chore: update push-rdev to use only first 7 github sha characters for image tags (https://github.com/chanzuckerberg/single-cell-data-portal/pull/4779)

Issue #4626, opened 2023-04-18
Improvements to Slack notifications for failed Datasets


danieljhegeman:
[see discussion](https://czi-sci.slack.com/archives/C0244PQK934/p1681836092350959?thread_ts=1681832314.264669&cid=C0244PQK934)

Two things: 

1. The Slack notification is currently including a link to the _published_ Collection whenever a Dataset upload to an ongoing _revision_ for that Collection fails. Instead, the Collection link included in the Slack notification should be a link to that revision Collection.

2. The Batch Job ID link in the notification is dead / None / not being included. This needs to be remedied. [See bottom of notification](https://czi-sci.slack.com/archives/C0244PQK934/p1681832314264669).

metakuni:
Assigning to @Bento007 since he's on-call this week.

PR #4857, opened 2023-05-17
fix: add more info to slack error messages for failed datasets. 

Bento007:
## Reason for Change

- #4626

## Changes

- Always return the unpublished collection or revision in the slack error message.
- include link to batch job. This is quality of life imporovement for debugging
- include link to step function. This is quality of life imporovement for debugging
- add function to parse the lambda event.
- add error handling to the error handler to ensure a slack notification is sent.
- add unit test for error handler functions.
- pass execution_arn into step function error handler. This is so the step function url can be built
- add formatting string for batch_job url and step function urls. This makes build the URLS easier and reusable.

## Testing steps

- add unit tests and increase test coverage.
- deployed to rdev and sent test messages to slack.


.happy/terraform/modules/sfn/main.tf
```diff
@@ -174,6 +174,12 @@ resource "aws_sfn_state_machine" "state_machine" {
         "Type": "Task",
         "InputPath": "$",
         "Resource": "${var.lambda_error_handler}",
+        "Parameters": {
+          "execution.$": "$$.Execution.Id",
+          "error.$": "$.error",
+          "dataset_id.$": "$.dataset_id",
+          "collection_id.$": "$.collection_id"
+        },
```
danieljhegeman, line 182: @Bento007 could you, for ease of review, add some clarity about that `$` and `$$` is doing here? I found [some documentation](https://states-language.net/) but I'm sorry to say it really only makes things like 10% more interpretable to me 😅

danieljhegeman, line 182: @Bento007 could you, for ease of review, add some clarity about that `$` and `$$` is doing here? I found [some documentation](https://states-language.net/) but I'm sorry to say it really only makes things like 10% more interpretable to me 😅

backend/common/utils/result_notification.py
```diff
@@ -24,7 +33,7 @@ def notify_slack(data: dict):
 def format_failed_batch_issue_slack_alert(data: dict) -> dict:
     aws_region = os.getenv("AWS_DEFAULT_REGION")
     job_id = os.getenv("AWS_BATCH_JOB_ID")
-    job_url = f"https://{aws_region}.console.aws.amazon.com/batch/v2/home?region={aws_region}#jobs/detail/{job_id}"
+    job_url = aws_batch_job_url_fmt_str.format(aws_region=aws_region, job_id=job_id)
```
danieljhegeman, line 36: I'm not sure I see the utility in creating str templates given they're only used once, but I guess it doesn't really hurt

backend/layers/processing/upload_failures/app.py
```diff
@@ -2,30 +2,69 @@
...
+    (
+        dataset_id,
+        collection_version_id,
+        error_step_name,
+        error_job_id,
+        error_aws_regions,
+        error_cause,
+        execution_arn,
+    ) = parse_event(event)
+    trigger_slack_notification(
```
danieljhegeman, line 26: worth putting this in a try/except clause so as to ensure the dataset processing status is still updated even if the slack notification generates an unhandled error?

backend/layers/processing/upload_failures/app.py
```diff
@@ -74,24 +121,46 @@ def get_failure_slack_notification_message(dataset_id):
...
+
+
+def cleanup_artifacts(dataset_id: str, error_step_name: str) -> None:
+    """Clean up artifacts depending on error; default to full clean-up if error step is unknown"""
+    object_key = os.path.join(os.environ.get("REMOTE_DEV_PREFIX", ""), dataset_id).strip("/")
+    if not error_step_name or error_step_name == "download-validate":
+        delete_many_from_s3(os.environ["ARTIFACT_BUCKET"], object_key)
+    if not error_step_name or error_step_name == "cxg":
+        cellxgene_bucket = os.getenv("CELLXGENE_BUCKET", default=f"hosted-cellxgene-{os.environ['DEPLOYMENT_STAGE']}")
+        delete_many_from_s3(cellxgene_bucket, object_key)
```
danieljhegeman, line 166: if we're going to clean up artifacts for instances of failed dataset processing, shouldn't we delete _all_ artifacts _every_ time?

danieljhegeman: I would think we'd either want to delete all associated assets or none.

danieljhegeman: note this also means cleaning up the assets from the public-access datasets bucket as well

danieljhegeman: Ah, I'm seeing that this is appearing as added code because you wrapped it in a method; this functionality was all as-is prior to your changes. Hm. I should have added the removal of assets from the public-access datasets bucket myself. Can I get you to introduce that fix? 🙏

tests/unit/processing/test_handle_error.py
```diff
@@ -0,0 +1,360 @@
...
+        execution_arn,
+    ) = parse_event(event)
+
+    assert dataset_id == "123"
+    assert collection_version_id == "456"
+    assert error_step_name is None
+    assert error_job_id is None
+    assert error_aws_regions is None
+    assert error_cause == ""
+    assert execution_arn is None
```
danieljhegeman, line 115: isn't convention to use the unittest framework to make these assertions via `self.assertIsNone`, etc.

is there a reason you decided to structure these tests differently rather than create a class and inherit from `unittest.TestCase`?

danieljhegeman, line 115: isn't our convention to use the unittest framework to make these assertions via `self.assertIsNone`, etc.

is there a reason you decided to structure these tests differently rather than create a class and inherit from `unittest.TestCase`?

tests/unit/processing/test_handle_error.py
```diff
@@ -0,0 +1,360 @@
...
+        error_cause,
+        execution_arn,
+    ) = parse_event(event)
+
+    assert dataset_id == "123"
+    assert collection_version_id == "456"
+    assert error_step_name == "Step1"
+    assert error_job_id == "789"
+    assert error_aws_regions == "us-east-1"
+    assert error_cause is error_cause
```
danieljhegeman, line 92: 🤔

tests/unit/processing/test_handle_error.py
```diff
@@ -0,0 +1,360 @@
...
+    assert dataset_id == "123"
+    assert collection_version_id == "456"
+    assert error_step_name is None
+    assert error_job_id is None
+    assert error_aws_regions is None
+    assert error_cause == "invalid JSON"
+    assert execution_arn is None
+
+
+def mock_get_dataset_version(dataset_id, collection_id):
```
danieljhegeman, line 140: `dataset_id` is unused

.happy/terraform/modules/sfn/main.tf
Bento007: Fair enough, this took me a while to grep as well. All of the data in a step function is JSON and can be accessed using a query language similar to [JSONPath](https://goessner.net/articles/JsonPath/). I didn't see anywhere that says they officially use this query language, but that's what it looks like.

The `$$` lets you access the [context object](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html) associated with the step function.

The `$` lets you access [input and output values](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-inputpath-params.html) from other steps.

The `.$` suffix is required for the key value if you are getting a value using JSONPath.  From AWS: 
> For key-value pairs where the value is selected using a path, the key name must end in .$.

backend/common/utils/result_notification.py
Bento007: it was mainly to reducing cognitive load in that function. The long format string caused `black` to reformat it up in a way that was more confusing to understand.

tests/unit/processing/test_handle_error.py
Bento007: good catch!

tests/unit/processing/test_handle_error.py
Bento007: We've been moving to use pytest when possible for a few reason, but I think this biggest reason is it has better support for reporting. Since this was a new set of tests, there was no reason to keep it as a `unittest.TestCase`

Bento007: I've also found that pytest has trouble running unittests with subtests. Using pytest test format avoid this problem.

backend/layers/processing/upload_failures/app.py
Bento007: in a separate PR

Bento007: https://github.com/chanzuckerberg/single-cell-data-portal/pull/4944

.happy/terraform/modules/sfn/main.tf
danieljhegeman: Thanks. This clears things up; I was missing the last bit about the req for the key to have the `.$` suffix.

.happy/terraform/modules/sfn/main.tf
```diff
@@ -174,6 +174,12 @@ resource "aws_sfn_state_machine" "state_machine" {
         "Type": "Task",
         "InputPath": "$",
         "Resource": "${var.lambda_error_handler}",
+        "Parameters": {
+          "execution.$": "$$.Execution.Id",
```
danieljhegeman, line 178: I would change the name of the key to `"execution_id.$"` since it is the id of the execution object/block

tests/unit/processing/test_handle_error.py
danieljhegeman: I see. And pytest does not provide any proprietary assertion statements. Got it.

danieljhegeman: >We've been moving to use pytest when possible for a few reason 

in summary what were these? I think I missed this discussion / decision

backend/layers/processing/logger.py
```diff
@@ -22,3 +23,14 @@ def wrapper(*arg, **kw):
...
+
+class LogSuppressed(suppress):
+    def __init__(self, *args, message="Suppressed Exception"):
+        super().__init__(*args)
+        self.message = message
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if exc_type is not None:
+            logging.error(self.message, exc_info=(exc_type, exc_value, traceback))
+        return super().__exit__(exc_type, exc_value, traceback)
```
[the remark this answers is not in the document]
Bento007: This is an established patter for context manager and all context managers are is a simplified try-except-finally block.

tests/unit/processing/test_handle_error.py
Bento007: simplified test structure, better reporting, already being used in other single-cell projects.

backend/layers/processing/logger.py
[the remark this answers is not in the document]
danieljhegeman: It seems to me like ln 36 would cause the super to throw an exception which would not get caught

But I'm missing something...?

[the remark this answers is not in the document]
danieljhegeman: Ah... I [see](https://docs.python.org/3/reference/datamodel.html#object.__exit__:~:text=If%20an%20exception%20is%20supplied%2C%20and%20the%20method%20wishes%20to%20suppress%20the%20exception%20(i.e.%2C%20prevent%20it%20from%20being%20propagated)%2C%20it%20should%20return%20a%20true%20value.%20Otherwise%2C%20the%20exception%20will%20be%20processed%20normally%20upon%20exit%20from%20this%20method.)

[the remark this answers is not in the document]
danieljhegeman: that is really not at all intuitive 

Although it still seems like the super might raise that exception since you're passing it....??

danieljhegeman (approved):
useful exercise. Thanks @Bento007 for getting this done, too.

Issue #4863, opened 2023-05-18
feat(WMGv2): Create V2 version of `\query` endpoint


prathapsridharan:
Here are the execution steps in order:

1. **Version the query API** - Make a copy of [v1.py](https://github.com/chanzuckerberg/single-cell-data-portal/blob/aa092f81fdb944ad13f9007bbb65c917fa626f80/backend/wmg/api/v1.py) and [test_v1.py](https://github.com/chanzuckerberg/single-cell-data-portal/blob/aa092f81fdb944ad13f9007bbb65c917fa626f80/tests/unit/backend/wmg/api/test_v1.py) and name them as `v2.py` and `test_v2.py`

_**The reason for this is to mitigate against the case of making API code changes that are backward incompatible with the current frontend code running in production. For example, if the API changes require changes to the structure of the data sent to the frontend that the current production version of frontend is not prepared to handle, then it is safer to simply version the API**_

2. `tissue_ontology_term_ids` ([here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/aa092f81fdb944ad13f9007bbb65c917fa626f80/backend/wmg/data/query.py#L14)) should be updated to allow no items (`min_items=0`)
3. Modify `v2.py` to query the `expression_summary` and `expression_summary_default` cubes without specifying tissues.
4. Modify `wmg-api.yml` to include `v2/query` and check that `v2/query` executes code in `v2.py`
5. Refactor unit test files `test_v2.py` and `test_query.py` to include changes made in `v2.py`
6. Modify the data structures returned gene expression calculator functions and rollup functions (gene expression and cell counts) appropriately to serve the requirements of the frontend.

At the end of these 6 steps, the **V2 Query API** _should be functional but may not be optimized for performance in production_. It may not be performance optimized the backend cube changes may have not been made. Furthermore, other optimizations besides cube changes may be needed and is as yet undetermined.

prathapsridharan:
@seve - It might be helpful to put in this ticket what the request payload and response format are V1 and V2 of the `query` api as it might be helpful during implementation. I will put what I know in comments below _but if you could put down the expected V2 response format that would be good_.

prathapsridharan:
# V1 Request Payload

```
{
  "filter": {
    "dataset_ids": [],
    "development_stage_ontology_term_ids": [],
    "disease_ontology_term_ids": [],
    "gene_ontology_term_ids": [
      "ENSG00000000003"
    ],
[...]

PR #4864, opened 2023-05-18
refactor: Create a strawman V2 version of the WMG API

prathapsridharan:
## Reason for Change

- #4863
- This ticket requires multiple commits. This commit **addresses steps 1 to 3** as described in the ticket. _**Specifically, this commit versions the WMG API (make a `v2` version) instead of making changes in the existing API version (`v1`) so as to prevent backward incompatible changes in the API (ex: structure of payload sent back to the frontend) from breaking the current production version of frontend rendering code.**_

## Changes
- add v2 version of the WMG query API and accompanying unit test file
- add v2 version of the `WMGQueryCriteria` class called `WMGQueryCriteriaV2` to isolate the WMG V2 requirement of querying without specifying `tissue_ontology_term_id`

## Testing steps

- Currently has no tests since this is effectively unreachable code. Moreover it is effectively prototype code meant to unblock the frontend developer. Tests will come in subsequent commits after the frontend developer is unblocked by this commit.

## Notes for Reviewer
This is done be creating an exact copy of the files: `v1.py` and `test_v1.py` and renaming them as `v2.py` and `test_v2.py` respectively. The aim of this commit is to create an isolated fork of the API code in which to develop the new functionality specified by the WMG V2 requirements. _**The reason for this is to mitigate against the case of making API code changes that are backward incompatible with the current frontend code running in production. For example, if the API changes require changes to the structure of the data sent to the frontend that the current production version of frontend is not prepared to handle, then it is safer to simply version the API**_

This V2 version contiains the functionality for the client to make queries without specifying `tissue_ontology_term_id` in the request body. However, the implementation is not entirely functional at this commit. This is not a problem because the incomplete implementation is hidden from the main line of code execution and therefore poses no risk due in the production environment.

### DIFF
--- a/backend/wmg/api/v2.py
+++ b/backend/wmg/api/v2.py
@@ -1,0 +1,625 @@
+import itertools
+from collections import defaultdict
+from typing import Any, Dict, Iterable, List, Tuple
+
+import connexion
+import numpy as np
+import pandas as pd
+from flask import jsonify
+from pandas import DataFrame
+from server_timing import Timing as ServerTiming
+
+from backend.wmg.data.ontology_labels import gene_term_label, ontology_term_label
+from backend.wmg.data.query import (
+    MarkerGeneQueryCriteria,
+    WmgFiltersQueryCriteria,
+    WmgQuery,
+    WmgQueryCriteria,
+    retrieve_top_n_markers,
+)
+from backend.wmg.data.rollup import rollup_across_cell_type_descendants
+from backend.wmg.data.schemas.cube_schema import expression_summary_non_indexed_dims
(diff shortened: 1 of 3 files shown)


prathapsridharan:
@atarashansky - How could I update the [`wmg-api.yml`](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/wmg/api/wmg-api.yml) to include a path to the `v2` API file?

Basically, I want to unblock @seve so he can just pass in `wmg/v2/query` in the url and hit the `v2` code.

atarashansky (approved):

Issue #2099, opened 2022-03-10
Deletion of published collections in the Curation UX must be removed


brianraymor:
Currently, curators are presented with a `Delete Collection` button in the collection detail page for all published collections that they're authorized to delete. This UX must be completely removed to avoid accidents, track deletion occurrences and justifications, and to validate that the redirect is in place for the permanent links to the collection and its CXG(s). 

Curators are still allowed to `Cancel a revision` in a private revision of a published collection.

Bento007:
The current user types I'm aware of are: 
- visitor:
  - unauthenticated
  - permissions: read collections
- curator:
  - authenticated
  - permissions: read, write, delete, and modify own collections
- super curator:
  - authenticated
  - permissions: read, write, delete and modify all collections

Are we adding an additional user type, or is this a modification of the `curator` user type?

brianraymor:
This is a modification to the `curator` user type.

- curator:
  - authenticated
  - permissions: read, write, ~delete~, and modify own collections

PR #4865, opened 2023-05-19
feat: Deletion of published collections in the Curation UX must be removed (#2099)

frano-m:
## Reason for Change

- #2099 

## Changes

- Removed "Delete" button and associated "Delete" modal from Collection detail page.

## Testing steps

- A curator should not be able to delete a published collection.
- A curator should still be able to delete a private collection, or delete a collection in revision.

## Notes for Reviewer

- Tests are unaffected by the removal of the delete button.


frano-m:
Thank you so much @ashin-czi!

Issue #4867, opened 2023-05-19
BUG: Gene rollup exception causes 500 error at query end point


prathapsridharan:
# Steps to reproduce

1. Pick `organism: homo sapiens`
2. Pick 2 `Tissues: [spinal cord, liver]`
3. Pick 1 `Gene: MS4A1`

This happens when the genes and tissues picked on the frontend are such that one of the tissues has NO expression of any of the genes picked. As the simple example above suggests, if you pick 2 tissues and 1 gene, and that single gene is not expressed in one of the tissues, then that tissue has no gene expression at all within the query constraints.

### Request body
```
{"filter":{"dataset_ids":[],"development_stage_ontology_term_ids":[],"disease_ontology_term_ids":[],"gene_ontology_term_ids":["ENSG00000156738"],"organism_ontology_term_id":"NCBITaxon:9606","self_reported_ethnicity_ontology_term_ids":[],"sex_ontology_term_ids":[],"tissue_ontology_term_ids":["UBERON:0002240","UBERON:0002107"]},"is_rollup":true}
```

### Error stack trace from 

https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Fdp$252Fprod$252Fprodstack$252Fbackend/log-events/fargate$252Fweb$252F0c40d564d5504d808640c6941fba2d35$3Fstart$3D-1800000$26filterPattern$3D$2522ERROR$2522

Search for *"ERROR"* in the cloudwatch link above

```
{
    "levelname": "ERROR",
    "asctime": "2023-05-19T17:33:17.019Z",
    "name": "corpora-api-prod",
    "message": "InternalServerError",
    "lineno": 182,
    "pathname": "/single-cell-data-portal/backend/api_server/app.py",
    "request_id": "838b1559-bcfc-490c-8cf6-1c839f123fc5",
[...]

PR #4869, opened 2023-05-19
fix: Fix rollup bug when selected tissue has no gene expressions

prathapsridharan:
## Reason for Change

- #4867 

## Changes

When the query API gets a list of tissues and genes, and it is the case that one of the tissues in the query arguments does not express any of the genes in the query argument, the rollup function would throw a KeyError when looking up the "gene expression free" tissue in one of the data structures.

The fix is to prevent this look up by filtering these "expression free tissues".

## Testing steps

- Follow reproduction steps in the ticket on rdev
- Added unit tests to verify the fix at the function level

### DIFF
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -355,13 +355,12 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
         missing_combinations = available_combinations.difference(available_combinations_per_gene)
         for combo in missing_combinations:
             entry = {dim: combo[i] for i, dim in enumerate(group_by_terms)}
-            tissue_has_gene_expression = n_cells_tissue_dict.get(entry["tissue_ontology_term_id"], False)
 
             # If a tissue, T1, DOES NOT have ANY of the QUERIED GENES expressed, then entirely
             # omit all combos that contain tissue T1 from the candidate list of combos
             # that should be considered for rollup. Otherwise, include combos containing T1
             # in the candidate list of combos for rollup.
-            if tissue_has_gene_expression:
+            if entry["tissue_ontology_term_id"] in n_cells_tissue_dict:
                 entry.update({col: 0 for col in numeric_columns})
                 entry["n_cells_tissue"] = n_cells_tissue_dict[entry["tissue_ontology_term_id"]]
                 entry["gene_ontology_term_id"] = gene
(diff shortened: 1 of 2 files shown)


backend/wmg/api/v1.py
```diff
@@ -353,10 +353,17 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
         missing_combinations = available_combinations.difference(available_combinations_per_gene)
         for combo in missing_combinations:
             entry = {dim: combo[i] for i, dim in enumerate(group_by_terms)}
-            entry.update({col: 0 for col in numeric_columns})
-            entry["n_cells_tissue"] = n_cells_tissue_dict[entry["tissue_ontology_term_id"]]
```
atarashansky, line 357: Thanks for catching this, Prathap!! I think the simplest fix would be to replace this line with:

```python
entry["n_cells_tissue"] = n_cells_tissue_dict.get(entry["tissue_ontology_term_id"],0)
```

What do you think?

tests/unit/backend/wmg/api/test_expression_rollup.py
```diff
@@ -330,10 +410,24 @@ def _rollup_testcases():
             ),
             (
                 test_3["name"],
-                _cell_counts_df_with_ethnicity_compare_dim(test_3["input_cell_counts"]),
-                _cell_counts_df_with_ethnicity_compare_dim(test_3["expected_rolled_up_cell_counts"]),
-                _gene_expression_df_with_ethnicity_compare_dim(test_3["input_gene_expression"]),
-                _gene_expression_df_with_ethnicity_compare_dim(test_3["expected_rolled_up_gene_expression"]),
+                _cell_counts_df_without_compare_dim(test_3["input_cell_counts"]),
```
atarashansky, line 413: The new test cases, look great, thanks so much!

backend/wmg/api/v1.py
prathapsridharan: @atarashansky - Even better! I will make this fix

prathapsridharan: @atarashansky - On second thought, I see a couple of disadvantages here. We get the advantage of conciseness with your suggestion but we also get a couple of disadvantages (as I see it):

1. _Loss of clarity_ - It is not clear that the combos involving the tissue will be removed. In fact, a bug in a function further up the callstack that manipulates the dataframe returned from this function could mess things up. The current approach limits that possibility by not including these zero entries in dataframe.
2. _Greater memory footprint that might negatively affect performance_ - the memory footprint of the dataframe that is returned is higher by including the combos that will be eliminated anyway. The extra memory is 

```
|genes with expression values| x |tissues_with_no_gene_expression| x |compare_dim|
```

What do you think?

atarashansky: I'll leave it up to you. More clarity is indeed better.  Also, for what it's worth, I think the extra memory used by the 0's that would be added in this edge case is negligible.

atarashansky (approved):

PR #4876, opened 2023-05-23
feat(cell-guide): add canonical marker genes table and e2e tests

atarashansky:
## Reason for Change

- Adding the canonical marker genes table and E2E tests

## Testing steps

- E2E tests pass
- Tested locally

### DIFF
--- a/backend/wmg/api/v1.py
+++ b/backend/wmg/api/v1.py
@@ -355,10 +355,16 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
         missing_combinations = available_combinations.difference(available_combinations_per_gene)
         for combo in missing_combinations:
             entry = {dim: combo[i] for i, dim in enumerate(group_by_terms)}
-            entry.update({col: 0 for col in numeric_columns})
-            entry["n_cells_tissue"] = n_cells_tissue_dict[entry["tissue_ontology_term_id"]]
-            entry["gene_ontology_term_id"] = gene
-            entries_to_add.append(entry)
+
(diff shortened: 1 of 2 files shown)


frontend/src/views/CellCards/common/fixtures/allCellTypeMarkerGenes.json
atarashansky, line 1: Updates to fixtures that I had added to the prototype branch and forgot to port into the boilerplate PR.

frontend/src/views/CellCards/components/CellCard/components/common/HelpTooltip/index.tsx
```diff
@@ -0,0 +1,27 @@
+import { Tooltip } from "czifui";
```
atarashansky, line 1: This tooltip will be used to display extra information for the various data (marker genes, ontology view, enriched genes, etc)

frontend/src/views/CellCards/components/CellCard/components/common/Table/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  StyledHeadCell,
+  StyledRow,
+  StyledCell,
+  TableWrapper,
+} from "./style";
+
+interface TableProps<T> {
+  columns: Array<Extract<keyof T, string>>;
+  rows: T[];
+  columnIdToName?: Record<Extract<keyof T, string>, string>;
```
atarashansky, line 14: This is an optional property so users can set the column names to complex strings that cannot be encoded as properties in the row objects (for example "% cells" cannot be a property name).

If unset, then the property names in each `row` are capitalized and set as column names.

PR #4877, opened 2023-05-23
feat(cell-guide): add enriched genes table and e2e tests

atarashansky:
## Reason for Change

- add enriched genes table and e2e tests

## Testing steps

- e2e tests pass
- tested locally

### DIFF
--- a/frontend/src/views/CellCards/components/CellCard/components/EnrichedGenesTable/index.tsx
+++ b/frontend/src/views/CellCards/components/CellCard/components/EnrichedGenesTable/index.tsx
@@ -80,6 +80,9 @@ const EnrichedGenesTable = ({ cellTypeId }: Props) => {
 
   useEffect(() => {
     return () => {
+      // (alec) when the component unmounts, reset the organism to its initial state.
+      // not all cell types may have homo sapiens as a valid option so we need to invoke the conditional
+      // logic above for setting the initial organism.
       setSelectedOrganism("");
     };
   }, []);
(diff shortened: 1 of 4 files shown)


frontend/src/views/CellCards/components/CellCard/components/EnrichedGenesTable/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+      });
+    }
+    return rows;
+  }, [genes, selectedOrganism, cellTypeId]);
+
+  useEffect(() => {
+    return () => {
+      setSelectedOrganism("");
+    };
+  }, []);
```
seve, line 85: This is a cleanup function, right? Since this is a bit fringe, might want to comment the what and why

frontend/src/views/CellCards/components/CellCard/components/EnrichedGenesTable/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+              <DropdownSelect
+                handleChange={handleChange}
+                options={uniqueOrganisms}
+                selectedOption={selectedOrganism}
+                testId={CELL_CARD_ENRICHED_GENES_TABLE_DROPDOWN}
+              />
+              <Link
+                url={`${ROUTES.WHERE_IS_MY_GENE}?genes=${genesForShareUrl}`}
+                label="Open in Gene Expression"
+              />
```
seve, line 131: I guess we don't need to, but do we want to use a share url version?

seve: Is this going to f up our analytics?

seve: We can probably append an additional `?internal-share=true` flag that doesn't trigger analytics

Issue #4729, opened 2023-05-02
Documentation outline does not show up when you click in from Gene Expression


signechambers1:
The documentation outline does not show up when you click in from Gene Expression. This inhibits users from knowing there are additional pages that show the methodology and data processing. I would like the outline to expand to show where the user is on the page by default.

[Link to documentation](https://cellxgene.cziscience.com/docs/04__Analyze%20Public%20Data/4_2__Gene%20Expression%20Documentation/4_2_1__Get%20Started)

**Current view:**
![Screenshot 2023-05-02 at 11.01.48 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/991c97ea-46dd-462d-866a-918b0402898e)

**Desired view:**
![Screenshot 2023-05-02 at 11.01.57 AM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/336f5ecc-590b-4714-bfed-2b66f4cc3c33)

joyceyan:
should be fixed in this PR https://github.com/chanzuckerberg/single-cell-data-portal/pull/4888

not entirely sure when it will deploy to prod

PR #4888, opened 2023-05-25
fix: documentation outline when you enter from an external source

joyceyan:
## Reason for Change

ticket:
https://github.com/chanzuckerberg/single-cell-data-portal/issues/4729

## Changes

- sets the `isExpanded` state to 1 if `directory.files.includes(activeFile)` using the react hook `useMemo`
- after doing some reading + consulting chatGPT, i believe a react hook is necessary because the values `directory.files` and `activeFile` may not be set when the component is first rendered. the `useMemo` hook will ensure that the state is set if the data is updated asynchronously
- i think `useMemo` is preferred over `useEffect` since this component doesn't rely on any external dependencies
- recursively checks the directory as well as any child subdirectories to see if the activeFile is in there, and sets the expansion state as necessary
- i also included some aesthetic changes to declare `ExpandedValue` as an enum, so that it's more clear what the different numerical values mean. this should not affect any functionality (ex: `if isExpanded % 2` should still work)

## Testing steps

- tbh i don't fully understand the product well enough to get to this page from gene expression (lol), but i was able to repro the bug by just opening up a new tab with the following URL and seeing that the navigation outline was expanded: https://localhost:3000/docs/05__Annotate%20and%20Analyze%20Your%20Data/5_0__Get%20Started

https://github.com/chanzuckerberg/single-cell-data-portal/assets/5653616/1c28b02e-76af-4066-824a-84fc25feb99a

## Notes for Reviewer
let me know if there's a better way to do this!

### DIFF
--- a/frontend/src/pages/docs/[...slug].tsx
+++ b/frontend/src/pages/docs/[...slug].tsx
@@ -15,7 +15,7 @@ import Head from "next/head";
 import Image, { ImageProps } from "next/image";
 import NextLink from "next/link";
 import pathTool from "path";
-import { Fragment, memo, useState } from "react";
+import { Fragment, memo, useState, useMemo } from "react";
 import rehypeSlug from "rehype-slug";
 import { noop } from "src/common/constants/utils";
 import { OFF_WHITE, PINK } from "src/common/theme";
@@ -179,8 +179,13 @@ const DirectoryListItem = ({
   directory: Directory;
   activeFile: string;
 }) => {
+  const initialState = useMemo(() => {
+    return directory.files.includes(activeFile) ? 1 : 0;
+  }, [directory.files, activeFile]);
+
   // 0 = default collapse, 1 = default expand, 2 = user collapse, 3 = user expand
-  const [isExpanded, setIsExpanded] = useState<0 | 1 | 2 | 3>(0);
+  const [isExpanded, setIsExpanded] = useState<0 | 1 | 2 | 3>(initialState);
+
   return (
     <Fragment>
       <li onClick={() => setIsExpanded(isExpanded % 2 == 0 ? 3 : 2)}>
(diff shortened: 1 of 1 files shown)


seve:
> todo: should i add a video to this?
Prob not but they're always nice to have. If you don't know how to make one quickly just lmk and I'll show you how to do it

frontend/src/pages/docs/[...slug].tsx
```diff
@@ -179,8 +179,13 @@ const DirectoryListItem = ({
   directory: Directory;
   activeFile: string;
 }) => {
+  const initialState = useMemo(() => {
+    return directory.files.includes(activeFile) ? 1 : 0;
+  }, [directory.files, activeFile]);
+
```
Comment (seve) [inline frontend/src/pages/docs/[...slug].tsx:185]: This is only looking at the current directory and its children, and opening if necessary.  There are also cases where the current directory contains additional directories, and inside of one of those directories is the active file. See: https://localhost:3000/docs/04__Analyze%20Public%20Data/4_2__Gene%20Expression%20Documentation/4_2_2__Cell%20Type%20and%20Gene%20Ordering

seve (changes requested):

frontend/src/pages/docs/[...slug].tsx
Comment (joyceyan) [reply frontend/src/pages/docs/[...slug].tsx:185]: just to make sure i'm understanding this correctly: so i had thought this line took care of that https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/frontend/src/pages/docs/%5B...slug%5D.tsx#L226 but looking further, it looks like this will expand the "Gene Expression Documentation" directory, but not the "Analyze Public Data" directory, resulting in the navigation being collapsed since the "grandparent" directory is not expanded, even if the "parent" directory is?

seve (approved):

atarashansky:
@joyceyan 
I think this PR might have introduced a bug in the doc site. You can no longer expand collapsible sections by clicking on them.

PR #4896, opened 2023-05-26
chore(cell-guide): update canonical marker gene tooltip to point to latest ASCT+B version

atarashansky:
## Reason for Change

- The tooltip was pointing to ASCT+B v1.2 but we are using v1.3.

## Testing steps

- Tested locally


prathapsridharan (approved):
lgtm

PR #4897, opened 2023-05-26
chore(cell-guide): update dag tooltip to remove descendent cells for leaf nodes

atarashansky:
## Reason for Change

- Update the DAG tooltip to remove descendant # cells indicator for leaf nodes
- Leaf nodes have no descendants so this information isn't meaningful.

## Testing steps

- Tested locally


prathapsridharan (approved):
lgtm

PR #4900, opened 2023-05-27
refactor: Reorganize code and unit tests

prathapsridharan:
Reorganized WMG API code and unit tests minimally to enable writing unit tests for WMG V2 API in isolation from the unit tests associated with WMG V1 API.

## Reason for Change

- #4863 
- This PR addresses part of **step 5** in the above ticket so that unit tests for WMG V2 API can be written in isolation from the unit tests for WMG V1 API. Previously, the functions making up the WMG V1 API code was not partitioned into modules such that commonalities between WMG V1 and WMG V2 could be extracted and tested in isolation.

This refactor extracts the common function between V1 and V2 and covers them with tests that previously existed all over the unit test codebase. After this refactor is merged, only what is changed between V1 and V2 need to be unit tested.

## Changes

- pull out common functions between WMG V1 and V2 API into modules
- pull out the test that covered these common functions into their own modules
- have the test codebase hierarchy mirror the actual codebase hierarchy so it is easy to see the correspondence between unit tests and the code under test

## Testing steps

- Unit tests

## Notes for Reviewer

Note that what is refactored was already covered by unit tests. Therefore, the refactor is safe. The functions that are not covered by unit tests and left where they are. The intent here is to do the minimal refactor necessary to evolve the V2 API (code and tests) independently of the WMG V1 API


seve (commented):
LGTM on logistics and general refactoring, would wait for Alec on more detailed python review

tests/unit/backend/wmg/test_query.py
```diff
@@ -128,240 +100,6 @@ def test__query_expression_summary_default_cube__returns_correct_results(self):
             expected = [553379.0, 261191.0]
```
atarashansky, line 100: Why not move `test__query_expression_summary_default_cube__returns_correct_results` into `test_expression_dotplot`?

atarashansky (approved):
LGTM! Just one small nit about a unit test that I think should be moved to the test_expression_dotplot.py module (the expression_summary_default unit test)

tests/unit/backend/wmg/test_query.py
prathapsridharan: @atarashansky - My thought was `test_query.py` tests methods in the `query.py` module which holds the `WMGQuery` class. Therefore I thought this test calling `WMGQuery.expression_summary_default()` should belong here.

What I hope we can do, is that the `unit` folder (holding the unit tests), reflects the directory hierarchy of `backend/wmg`. I would have moved `test_query.py` into the `test/unit/backend/wmg/data` folder but I wanted to do minimal change to accomplish the isolation that I need so as not to create too much churn in the codebase.

I will go ahead and merge this as is. And if you have a different idea of how to go about refactoring/reorganizing tests and code, let's discuss.

PR #4915, opened 2023-05-31
feat: Unit tests and security for WMG V2 API

prathapsridharan:
## Reason for Change

- #4863 
- This PR addresses step 5 in the ticket above - specifically creating unit test coverage for the WMG V2 API where the `query` endpoint expects request bodies without `tissue_ontology_term_ids` key.

## Changes

- add unit tests
- Make api more secure by making strict the schema of the request body on `HTTP POST`. Specifically, the openApi spec is modified such that properties not specified in the request body schema are rejected. This prevents potential injection attacks that could lead to a denial of service or worse things.

## Testing steps

- Unit tests

### DIFF
--- a/backend/wmg/data/query.py
+++ b/backend/wmg/data/query.py
@@ -127,7 +127,9 @@ class WmgQuery:
     # TODO: refactor for readability: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues
     #  /chanzuckerberg/single-cell-data-portal/2133
     @staticmethod
-    def _query(cube: Array, criteria: Union[WmgQueryCriteria, FmgQueryCriteria], indexed_dims: List[str]) -> DataFrame:
+    def _query(
+        cube: Array, criteria: Union[WmgQueryCriteria, WmgQueryCriteriaV2, FmgQueryCriteria], indexed_dims: List[str]
+    ) -> DataFrame:
         query_cond = ""
         attrs = {}
         for attr_name, vals in criteria.dict(exclude=set(indexed_dims)).items():
--- a/tests/unit/backend/wmg/api/test_v2.py
+++ b/tests/unit/backend/wmg/api/test_v2.py
@@ -91,14 +91,13 @@ def generate_expected_expression_summary_dictionary(genes, tissues, cell_types, 
     return result
 
 
-def generate_test_inputs_and_expected_outputs(genes, tissues, organism, dim_size, me, expected_count, compare_dim=None):
+def generate_test_inputs_and_expected_outputs(genes, organism, dim_size, me, expected_count, compare_dim=None):
     """
     Generates test inputs and expected outputs for the /wmg/v2/query endpoint.
 
     Arguments
     ---------
     genes: list of gene ontology term IDs
-    tissues: list of tissue ontology term IDs
     organism: organism ontology term ID
     dim_size: size of each dimension of the test cube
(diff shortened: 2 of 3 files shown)


backend/wmg/data/query.py
```diff
@@ -127,7 +127,9 @@ def cell_counts(self, criteria: Union[WmgQueryCriteria, FmgQueryCriteria]) -> Da
     # TODO: refactor for readability: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues
     #  /chanzuckerberg/single-cell-data-portal/2133
     @staticmethod
-    def _query(cube: Array, criteria: Union[WmgQueryCriteria, FmgQueryCriteria], indexed_dims: List[str]) -> DataFrame:
+    def _query(
```
joyceyan, line 130: if it wouldn't require too much refactoring, it might be nice to add `*` as an argument here, which will require callers to name the arguments that are passed in.

https://stackoverflow.com/questions/2965271/how-can-we-force-naming-of-parameters-when-calling-a-function

tests/unit/backend/wmg/api/test_v2.py
```diff
@@ -235,55 +232,10 @@ def test__query_single_primary_dims__returns_200_and_correct_response(
             gene_term_label.side_effect = lambda gene_term_id: f"{gene_term_id}_label"
 
             genes = ["gene_ontology_term_id_0"]
-            tissues = ["tissue_ontology_term_id_0"]
```
joyceyan, line 238: just for my own understanding, why are these being removed?

tests/unit/backend/wmg/api/test_v2.py
```diff
@@ -91,14 +91,13 @@ def generate_expected_expression_summary_dictionary(genes, tissues, cell_types,
     return result
 
 
-def generate_test_inputs_and_expected_outputs(genes, tissues, organism, dim_size, me, expected_count, compare_dim=None):
```
joyceyan, line 94: same here, would be nice to add `*`, though for functions used only in tests, it's really only beneficial for yourself so that you're forced to change all the things within your tests that call into it 😅

joyceyan (approved):

backend/wmg/data/query.py
prathapsridharan: @joyceyan - I think that is good idea. I presume you mean something like this:

```
def _query(*, cube: Array, criteria: Union[WmgQueryCriteria, FmgQueryCriteria], indexed_dims: List[str]) -> DataFrame
```

which forces all callers to use keyword arguments?

Do you mind if I do this in another commit? Because I think we also need to think about:

- which functions should be called with _keyword-only_ arguments - that is, the _argument names are part of the explicit interface of the function_ and
- which functions should be called by _positional_ arguments (using a `/` to enforce that arguments preceding it should _not be named by the caller and only called positionally_) - that is, the _argument names are NOT part of the explicit interface of the function_

joyceyan: yep the code snippet you posted is what i'm imagining. putting it in another PR or commit is fine - that's why i left the comment but still approved the PR 😄 i think any sort of query-related thing we would probably want to enforce keyword only arguments, and i'd also say that generally speaking any function with > 1 argument should probably enforce keyword only arguments (though obviously not a priority to refactor existing code). i can't really think of any reason why positional arguments would be preferred over keyword arguments, but there might be some math-y use cases here that i haven't seen before?

PR #4944, opened 2023-06-02
fix: delete from public dataset bucket when submission fails during upload

Bento007:
## Reason for Change

- need to delete datasets from the public bucket if a dataset fails in the submission pipeline.
- [see previous conversation](https://github.com/chanzuckerberg/single-cell-data-portal/pull/4857#discussion_r1213981729)
## Changes

- add dataset public bucket as an environment variable to the submission failure handler lambda.
- generalize delete_many_from_s3 to work with all of the buckets.
- modify cleanup_artifacts to include deleting from the dataset public bucket if upload fails.

## Testing steps

- expanded unit tests

## Notes for Reviewer
related infra changes https://github.com/chanzuckerberg/single-cell-infra/pull/802


backend/layers/processing/upload_failures/app.py
```diff
@@ -161,9 +161,11 @@ def cleanup_artifacts(dataset_id: str, error_step_name: str) -> None:
...
-            delete_many_from_s3(os.environ["ARTIFACT_BUCKET"], object_key)
+            delete_many_from_s3(os.environ["ARTIFACT_BUCKET"], object_key + "/")
+            delete_many_from_s3(os.environ["DATASET_BUCKET"], object_key + ".")
+
         if not error_step_name or error_step_name == "cxg":
             cellxgene_bucket = os.getenv(
                 "CELLXGENE_BUCKET", default=f"hosted-cellxgene-{os.environ['DEPLOYMENT_STAGE']}"
             )
-            delete_many_from_s3(cellxgene_bucket, object_key)
+            delete_many_from_s3(cellxgene_bucket, object_key + ".cxg/")
```
danieljhegeman, line 171: @Bento007  Shouldn't we also be deleting from the artifact and dataset buckets when the error step is "cxg"?

If the processing pipeline errors, we should delete everything. So, since download-validate is first, we only have to worry about the buckets that get files uploaded to them during that step. But if cxg fails, we should delete from buckets that get files during download-validate OR cxg.

Bento007: That makes sense.

backend/common/utils/aws.py
```diff
@@ -93,13 +93,15 @@ def _debug(self, message):
...
+s3 = boto3.resource("s3", endpoint_url=os.getenv("BOTO_ENDPOINT_URL") or None)
+
+
+def delete_many_from_s3(bucket_name: str, prefix: str) -> None:
     """
     This deletes everything with a specific prefix from the given bucket
 
     """
-    if not dataset_id:
+    if not prefix:  # protect from deleting everything
```
danieljhegeman, line 104: 🧠

danieljhegeman (approved):
Looks good. Thanks for implementing across-the-board delete on any failure.

PR #4950, opened 2023-06-05
fix: update ontology file location in tissue_mapper

ebezzi:
## Reason for Change

Fixes an issue where the tissue mapper is unable to download the UBERON ontology file.

## Changes

- Updates the uberon.owl file to a different location (which is the redirect target for the old URI).

## Testing steps

- Unit tests should cover the update.


atarashansky (approved):

prathapsridharan (approved):
LGTM

Issue #4730, opened 2023-05-02
Compare cell types across publications and filter by publication


signechambers1:
**What is the problem?**
Our users would like to be able to compare cell types across publications in order to see where expression patterns are consistent and where they diverge. Understanding these patterns will help users to interpret results, filter by publications to narrow results if they want to, and download collections or datasets they are interested in.

This is a top feature request from our users, see user feedback [here](https://airtable.com/app8fNSQ8ieIiHLOv/tblqoYEr9Ee7rYrut/viwFPtZPfH7r2hIzk/rec5NZa4Orn1eKbay?blocks=hide).

**How does this align to single cell strategy?**
This functionality furthers researchers ability to identify patterns of expression across cell types, which is core to CELLxGENE Discover’s main goal.

**[Figma Prototype](https://www.figma.com/proto/nUbwlxcSPAGoRGWyYaIkyv/Q2-2023-Gene-Expression-Improvements?page-id=226%3A52176&node-id=226-58146&viewport=601%2C309%2C0.21&scaling=scale-down&starting-point-node-id=226%3A58146&show-proto-sidebar=1)**
**[Figma Designs](https://www.figma.com/file/nUbwlxcSPAGoRGWyYaIkyv/Q2-2023-Gene-Expression-Improvements?type=design&node-id=226-52176&t=jEufF2mpVSuHem46-0)**

**Definition of a publication**
* Each collection is associated with a publication (they have a 1:1 relationship)
* Publications are formatted as: First author last name. (year) Journal
* Example: Carr et al. (2023) Nature
* For collections without a publication DOI, we will use the last name of the collection contact, and no publilcation. Example: Carr et al. (2023) No publication

**User stories**
1. Users can filter by publication
- [x] User can multiselect from a publication filter and results shown in the dot plot will filter to the selected publications
- [x] User can clear publication selections
- [x] Filter works consistently with other filters in the left side bar
- [x] Analytics work: New event tracking the filter usage & values, update share & download to save state of filter (WMG_download_complete, WMG_share_clicked, WMG_share_loaded)

2. User can Group By a publication
- [x] User can select Publication as an option in the Group By option and dot plot will update showing rows for each publication for each cell type that is visible
- [x] Loading indicator is present if loading time is >1 sec
Iteration likely needed:
- [ ] _Iterate with product and design on performance and speed options_
- [ ] _Open question: Iterate with product and design to resolve UX complications where there are many publications per cell type_
[...]

PR #4965, opened 2023-06-06
feat: adding a new secondary filter to filter by publication

carolinechoi:
## Adding a secondary filter called "Publication"

- #4730
- Adding a new secondary filter called "Publication" to Gene Expressions page, such that users can filter by publication.

## Changes

- add
- remove
- modify

## Testing steps

- Two directions of testing: (1) Datasets -> Publications; if a dataset is selected, it filters out the publication options based on the publication that the dataset is attached to. (2) Publications -> Datasets; if a publication is selected, the only options in the dataset filter are the selected publication's corresponding datasets.
- The dot plot is filtered out automatically based on the dataset selection due to the intersection of selected dataset IDs and the selected publication's dataset IDs. 
- Tested locally. 

## Notes for Reviewer
- Used the 'explorer_url' to detach the stable dataset IDs, such that I could work on this without touching the DP API's structure. Solution suggested by Emanuele.


frontend/src/views/WheresMyGene/components/Filters/index.tsx
```diff
@@ -75,6 +78,12 @@ interface FilterOption {
   id: string;
 }
 
+interface publicationFilterOption {
+  name: string;
+  label: string;
+  id: string | string[];
```
seve, line 84: Since we're not doing "No publications" we can probably drop the `string[]`

frontend/src/views/WheresMyGene/components/Filters/index.tsx
```diff
@@ -324,6 +400,23 @@ export default memo(function Filters({
           DropdownMenuProps={DropdownMenuProps}
           InputDropdownProps={InputDropdownProps}
         />
+
+        {/* CAROLINE U ARE HERE  */}
```
seve, line 404:
```suggestion
```

seve (changes requested):

.happy/config.json
seve, line 1: i would drop this file from your PR since you aren't actually changing anything other than formatting

package.json
seve, line 1: looks like you ran `npm install prettier` in the wrong directory

seve (changes requested):

package.json
carolinechoi: poop

frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -443,6 +449,40 @@ export function useFilterDimensions(version: 1 | 2 = 1): {
   data: FilterDimensions;
   isLoading: boolean;
 } {
+  const { mode, status } = useViewMode();
```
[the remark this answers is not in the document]
carolinechoi: Should I just set it to the default mode/status then? I need something to pass through useFetchCollectionRows on line 455.

seve (approved):

PR #4968, opened 2023-06-06
update to api

carolinechoi:
## Adding field to portal API

- #4730 (?)
- Currently the id field returns the dataset _version_ ids, not the actual dataset ids, which complicates the code for the publication filter. This code adds a new field that returns the actual dataset id. 

## Changes

- add field to portal API file's dataset_to_response function

## Testing steps
??


seve:
Yikes

Issue #3037, opened 2022-08-04
Deletion of published datasets in the Curation UX must be removed


brianraymor:
Currently, curators are presented with a `Delete Dataset` button in the private revision of a published collection. To avoid 
accidents, track deletion occurrences and justifications, and to validate that the redirect is in place for the permanent links to the CXG(s), the following changes must be made:

1. If a dataset is _New_, the `Delete Dataset` menu item is available.
2. If a dataset is _in the public collection_, then the `Delete Dataset` menu item is disabled or hidden.
3. If a dataset is _in the public collection_ and _Updated_, then the `Delete Dataset` menu item is disabled or hidden.

PR #4980, opened 2023-06-08
feat: Deletion of published datasets in the Curation UX must be removed (#3037)

frano-m:
## Reason for Change

- #3037 

## Changes

- Curators are no longer able to delete a dataset in the private revision of a published collection i.e. if a dataset has been previously published, then it may not be deleted. For clarification see [slack thread](https://clevercanary.slack.com/archives/C02CF7MQPGV/p1685602748854589?thread_ts=1685602175.729339&cid=C02CF7MQPGV).

## Testing steps

- Curator may delete a dataset from a private collection.
- Curator may NOT delete a dataset from a published revision collection.
- Curator may add a new dataset to a published revision collection, and is able to delete the newly added dataset.

### DIFF
--- a/frontend/src/common/entities.ts
+++ b/frontend/src/common/entities.ts
@@ -147,6 +147,7 @@ export interface Dataset {
   created_at: number;
   original_id?: string;
   published?: boolean;
+  published_at?: number;
   tombstone?: boolean;
   updated?: boolean;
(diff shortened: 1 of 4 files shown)


atarashansky (approved):
lgtm!

PR #4986, opened 2023-06-10
chore: Pin package version for WMG api and pipeline

prathapsridharan:
## Reason for Change

The changes in this PR address the following problems:

1. Versions for WMG api and pipeline are unpinned and sometimes incompatible. This is very risky because transparent changes to the version of the underlying dependencies (on account of packages not being pinned in a consistent manner), can create vicious bugs in local and production environments.
2. The current local development environment is memory intensive and slow on account of running many docker containers.

## Changes

1. Pin python packages for WMG api and pipeline to mitigate bugs in all environments (particularly production environment) due to transparent version upgrade of dependencies
2. The above pinning also enables installing packages on the local machine such that the test suite can be run on a developers local machine without having to invoke docker
3. Add detailed instructions on how to create a python virtual environment that contains all the dependencies necessary to run the test suite on a developer's local machine without docker 

## Testing steps

- Ran unit tests locally


DEV_ENV.md
```diff
@@ -1,12 +1,11 @@
 # Local Development Environment
 
-## Development quickstart
+## Install Docker
 
 1. [install docker](https://docs.docker.com/get-docker/). If brew is installed run `brew install docker`. If you have a Mac, [install Docker Desktop](https://www.docker.com/products/docker-desktop) and open it so it's running on your machine. Note: If you have a Mac M1 (arm64 CPU), follow instructions in [`docker-compose.yml`](docker-compose.yml) under `oidc` service, for manually building the oidc-server-mock images for the M1 architecture.
```
joyceyan, line 5: can you add a line about setting up docker with the CZI enterprise account? something like: "Go to https://hub.docker.com/ and sign in with SSO, which should add yourself to the CZI org automatically. Verify that you see yourself here: https://hub.docker.com/orgs/chanzuckerberg/members"

DEV_ENV_WITHOUT_DOCKER.md
```diff
@@ -0,0 +1,30 @@
...
+   - `pip install --global-option=build_ext --global-option="-I$(brew --prefix graphviz)/include/" --global-option="-L$(brew --prefix graphviz)/lib/" pygraphviz==1.11`
+
+1. Install base packages - `pip install -r requirements.txt`
+1. Install packages for WMG api - `pip install -r backend/api_server/requirements.txt`
+1. Install packages for WMG pipeline - `pip install -r backend/wmg/pipeline/requirements.txt`
+
+### Run unit tests
+
+1. Run unit tests for WMG api - `pytest -v tests/unit/backend/wmg`
+1. Run unit tests for WMG pipeline - `pytest -v tests/unit/wmg_processing`
```
joyceyan, line 23: might be good to add an example of testing a specific file. ex: `DEPLOYMENT_STAGE=test python3 -m unittest tests/unit/wmg_processing/test_load_corpus.py`

DEV_ENV.md
ebezzi: This is a public repo, so I don't think this is a good idea.

joyceyan: :homer-disappears:

atarashansky (changes requested):
Can you make an rdev stack off this commit? I think it's prudent to be extra careful with any dependency changes, even though these specific ones seem innocuous.

DEV_ENV_WITHOUT_DOCKER.md
```diff
@@ -0,0 +1,30 @@
...
+   - `brew install graphviz`
+   - `pip install --global-option=build_ext --global-option="-I$(brew --prefix graphviz)/include/" --global-option="-L$(brew --prefix graphviz)/lib/" pygraphviz==1.11`
+
+1. Install base packages - `pip install -r requirements.txt`
+1. Install packages for WMG api - `pip install -r backend/api_server/requirements.txt`
+1. Install packages for WMG pipeline - `pip install -r backend/wmg/pipeline/requirements.txt`
+
+### Run unit tests
+
+1. Run unit tests for WMG api - `pytest -v tests/unit/backend/wmg`
```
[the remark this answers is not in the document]
joyceyan: resolved this by creating a new virtual env with python 3.10 instead of python 3.9

prathapsridharan:
> Can you make an rdev stack off this commit? I think it's prudent to be extra careful with any dependency changes, even though these specific ones seem innocuous.

@atarashansky - Here is a new PR since I needed to change the branch name to contain the suffix `rdev`: https://github.com/chanzuckerberg/single-cell-data-portal/pull/4992

prathapsridharan:
Closing this PR in favor of a new PR so the `rdev` can be created. I had to rename the remote branch to contain the suffix `rdev` which meant I had to create a new PR. Here is the new PR:

https://github.com/chanzuckerberg/single-cell-data-portal/pull/4992

PR #4992, opened 2023-06-12
chore: Pin package version for WMG api and pipeline

prathapsridharan:
## Reason for Change

The changes in this PR address the following problems:

1. Versions for WMG api and pipeline are unpinned and sometimes incompatible. This is very risky because transparent changes to the version of the underlying dependencies (on account of packages not being pinned in a consistent manner), can create vicious bugs in local and production environments.
2. The current local development environment is memory intensive and slow on account of running many docker containers.

## Changes

1. Pin python packages for WMG api and pipeline to mitigate bugs in all environments (particularly production environment) due to transparent version upgrade of dependencies
2. The above pinning also enables installing packages on the local machine such that the test suite can be run on a developers local machine without having to invoke docker
3. Add detailed instructions on how to create a python virtual environment that contains all the dependencies necessary to run the test suite on a developer's local machine without docker 

## Testing steps

- Ran unit tests locally
- manual test on rdev frontend - https://rdev-ps-pin-wmg-pkg-frontend.rdev.single-cell.czi.technology/gene-expression
- manual test on rdev backend - https://rdev-ps-pin-wmg-pkg-backend.rdev.single-cell.czi.technology/


joyceyan (approved):
i was able to recreate a new conda env + run unit tests locally with the commands in `DEV_ENV_WITHOUT_DOCKER`

atarashansky (approved):

Issue #4908, opened 2023-05-30
Bug(curation) Flash of "No Results" when refreshing the collections page / No loading indicator


NoopDog:
#### Description

When refreshing the collections page, in the new curator experience, the page flashes "No Results"  then does not give any loading indicator while the page is actually loading.

https://images.zenhubusercontent.com/228681195/d8ce7f9f-7337-46dc-9124-d9398a7104b8/screen_recording_2023_05_26_at_11_00_56_am.mov

Expecting no flash of "No Results" and an indicator that the page is loading

#### Definition of Done

1. Left-hand filters do not show as a skeleton or disabled during load (current state for collections/datasets)
1. Loading indicator displays for the non-auth view only during loading.
1. Flash of no results is not displayed during loading for both auth and non-auth views.

PR #5005, opened 2023-06-13
fix: Bug(curation) Flash of "No Results" when refreshing the collections page / No loading indicator (#4908)

frano-m:
## Reason for Change

- #4908 

## Changes

- The Collections and Datasets index has been modified to display only the filter sidebar and list upon successful execution of the collections and datasets query, resolving the issue of unwanted and premature display of "No Results".
- A loading indicator has been implemented specifically for curator mode to provide a visual representation of the pending collections and datasets query.

## Testing steps

- A curator should see on page refresh, *a loading indicator*, followed by the display of the side bar filters and list.
- A non-curator should see on page refresh, *no loading indicator*, followed by the display of the side bar filters and list.
- Both curator and non-curator should not see a flash of "No Results" during loading.

### DIFF
--- a/frontend/src/common/queries/filter.ts
+++ b/frontend/src/common/queries/filter.ts
@@ -56,6 +56,7 @@ const QUERY_ID_DATASETS = "datasetIndex";
 export interface FetchCategoriesRows<T extends Categories> {
   isError: boolean;
   isLoading: boolean;
+  isSuccess: boolean;
   rows: T[];
 }
 
@@ -169,6 +170,7 @@ export function useFetchCollectionRows(
     data: datasets,
     isError: datasetsError,
     isLoading: datasetsLoading,
+    isSuccess: datasetsSuccess,
   } = useFetchDatasets(mode, status);
 
   // Fetch collections.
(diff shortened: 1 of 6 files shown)


frontend/src/common/queries/filter.ts
```diff
@@ -191,6 +194,7 @@ export function useFetchCollectionRows(
   return {
     isError: datasetsError || collectionsError,
     isLoading: datasetsLoading || collectionsLoading,
+    isSuccess: datasetsSuccess && collectionsSuccess,
```
frano-m, line 197: Success when both the collections and datasets query

frontend/src/views/Datasets/style.ts
```diff
@@ -0,0 +1,8 @@
+import styled from "@emotion/styled";
+import { View } from "src/views/globalStyle";
+
+export const DatasetsView = styled(View)`
```
frano-m, line 4: Updated DatasetsView to match CollectionsView.

frano-m:
Thank you again @ashin-czi!

Issue #4179, opened 2023-02-10
Remove 'X' icon from Datasets Download Dialog


karenliang0:
To align the design of dialogs across our tools, we should remove the 'x' / close icon button from the Datasets Download Dialog on Data Portal. 

For the updated collection details dialog, there is already no 'x' icon. The downloads dialog that will be built for Gene Expression to allow users to download svg, png, or csv will also have no 'x' icon. 

The thinking here is that there is already a cancel button and a user can also click anywhere on the page outside the Dialog to close the dialog as well. For accessibility, screen readers have an easier time locating a cancel button than the 'x' icon on a dialog as well.

PR #5015, opened 2023-06-14
feat: Remove 'X' icon from Datasets Download Dialog (#4179)

frano-m:
## Reason for Change

- #4179

## Changes

- Removed `x` button from datasets download dialog (both datasets index and detail pages).


frano-m:
@ashin-czi Many thanks for the review 🙇🏽‍♀️ !

PR #5031, opened 2023-06-15
feat(wmg): faster queries

atarashansky:
## Reason for Change

- #5146 
- #5294 

## Changes

- Invoke `load_snapshot` at server start time
- Replace `load_snapshot` usage with `get_cached_snapshot` in `v2.py`. This new function only returns the current cached snapshot without checking if a new one should be loaded.

## Testing steps

- Updated unit tests
- Tested in [rdev](https://fast-frontend.rdev.single-cell.czi.technology/gene-expression)


prathapsridharan (approved):
Synchronous code review with Alec. LGTM

PR #5047, opened 2023-06-15
fix: Adding option for "No Publication" to the secondary filter 

carolinechoi:
## Need to add an option for "No Publication" for selection in the publication filter 

- #4730 

## Changes

- add code for aggregating no publications' dataset ids 

## Testing steps
?? ¯\\_(ツ)_/¯ (seve fixed his arm) (the lil guy thanks you seve)


seve (approved):
Approval contingent on creating a const for no publication id/name

Issue #5050, opened 2023-06-16
[WMG] Conditionally run WMG E2E tests based on API URL


tihuan:
Problem:
Currently WMG E2E tests don't run in PRs and only run in deployed environments (dev/staging/prod)
This means that we frequently push bugs and broken E2E tests to `main`

Solution:
Detect which API URL the FE connects to and determine whether to run WMG E2E tests or not

tihuan:
this is for PR tests to run WMG tests, so moving to complete now

PR #5051, opened 2023-06-16
test(WMG): 5050 Conditionally run WMG e2e tests based on API URL

tihuan:
## Reason for Change

- #5050 

## Changes

- add
- remove
- modify

## Testing steps

### DIFF
--- a/frontend/playwright-globalSetup.ts
+++ b/frontend/playwright-globalSetup.ts
@@ -1,24 +1,33 @@
 import { chromium } from "@playwright/test";
-import { SKIP_LOGIN, TEST_URL } from "tests/common/constants";
+import { SKIP_LOGIN } from "tests/common/constants";
 import { COMMON_PLAYWRIGHT_CONTEXT } from "tests/common/context";
 import featureFlags from "tests/common/featureFlags";
-import { login } from "tests/utils/helpers";
+import { goToPage, login } from "tests/utils/helpers";
 
 module.exports = async () => {
+  const browser = await chromium.launch();
+
+  const browserContext = await browser.newContext({
+    ...COMMON_PLAYWRIGHT_CONTEXT,
+    storageState: featureFlags,
+  });
+
+  const page = await browserContext.newPage();
+
+  await goToPage(undefined, page);
+
+  const apiUrl = await page.evaluate(() => {
+    return document.body.getAttribute("data-api-url");
+  });
+
+  process.env.API_URL = apiUrl || "";
+
   try {
     // Skip login for tests that don't require it
     if (SKIP_LOGIN) {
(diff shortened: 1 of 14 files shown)


frontend/playwright-globalSetup.ts
```diff
@@ -2,23 +2,32 @@ import { chromium } from "@playwright/test";
...
+  const browserContext = await browser.newContext({
+    ...COMMON_PLAYWRIGHT_CONTEXT,
+    storageState: featureFlags,
+  });
+
+  const page = await browserContext.newPage();
+
+  await goToPage(undefined, page);
+
+  const apiUrl = await page.evaluate(() => {
```
tihuan, line 19: during global setup phase of Playwright, we hit the homepage to detect which API URL the FE app connects to and save that as env variable for subsequent tests to access via helper function `conditionallyRunTests()`

frontend/tests/utils/wmgUtils.ts
```diff
@@ -1,10 +1,25 @@
...
 
+const { skip } = test;
+
+const ENVS_TO_RUN_TESTS = [
+  "api.cellxgene.dev.single-cell.czi.technology",
+  "api.cellxgene.staging.single-cell.czi.technology",
+  "api.cellxgene.cziscience.com",
+];
+
+export function conditionallyRunTests() {
```
tihuan, line 16: this is the function that will use the env variable and check if the API URL is one of the URLs in `ENVS_TO_RUN_TESTS` to determine whether to run the tests or not

frontend/tests/utils/helpers.ts
```diff
@@ -57,11 +57,13 @@ const TEST_ENVS_DEV_STAGING = ["dev", "staging"];
...
 export const isDevStaging = TEST_ENVS_DEV_STAGING.includes(TEST_ENV);
 
+const GO_TO_PAGE_TIMEOUT_MS = 30 * 1000;
+
 export async function goToPage(
   url: string = TEST_URL,
   page: Page
 ): Promise<void> {
-  await page.goto(url);
+  await page.goto(url, { timeout: GO_TO_PAGE_TIMEOUT_MS });
```
tihuan, line 66: default 30 seconds is not enough in GHA

atarashansky (approved):
Very nifty. Thanks for this!

tihuan:
Thanks for the quick review, @atarashansky 🎉 !

tihuan:
Ah the test is failing for real, because we've updated the WMG cubes, and since the test selects the first dataset, which is now `22 integrated samples`, it doesn't narrow down the options. 

I'll update the test to select disease `acute kidney failure` on Tuesday, so our test is more predictable and not dependent on datasets order, which changes more frequently 👌

tihuan:
Fixed the test 👍

PR #5056, opened 2023-06-20
fix: publication filter crossfiltering

atarashansky:
## Reason for Change

- Publication filter was not crossfiltering into other filters besides dataset
- Datasets/publications for human included mouse datasets and vice-versa

## Changes

- Updated backend to ensure that organism was being factored into determining the valid filters
- Updated the frontend so that publication filters crossfiltered into other secondary filters

## Testing steps

- RDEV


backend/wmg/api/v1.py
```diff
@@ -140,13 +140,12 @@ def find_dimension_id_from_compare(compare: str) -> str:
...
     criteria = criteria.dict()
     for key in criteria:
-        if key != "organism_ontology_term_id":
-            if isinstance(criteria[key], list):
-                if len(criteria[key]) > 0:
-                    return False
-            else:
-                if criteria[key] != "":
-                    return False
+        if isinstance(criteria[key], list):
```
prathapsridharan, line 143: @atarashansky - Should this change be copied into `v2.py` as well?

prathapsridharan (changes requested):
@atarashansky -

carolinechoi (approved):
looks good 2 me

Issue #4800, opened 2023-05-09
Collection Migration Batch Job


nayib-jose-gloria:
Create Batch Job process that:

- Installs latest cellxgene-schema from PyPI
- Creates revision IF public collection
- Pulls raw.h5ad from S3 for all of the Collection’s datasets
- Runs 'migrate' command against raw.h5ad files
- At minimum, all datasets (even those unaffected by scripted changes) need to have their metadata updated to bump the schema version
- Triggers Processing Step Function for each migrated dataset

PR #5057, opened 2023-06-20
feat: add collection migrate + dataset migrate functions for schema migration

nayib-jose-gloria:
## Reason for Change

- #4800
- #5040

## Changes

- collection migrate + dataset migrate functions for schema migration
- refactor tests + migrate class set-up

## Testing steps
- unit tests

### DIFF
--- a/backend/schema_migration/migrate.py
+++ b/backend/schema_migration/migrate.py
@@ -1,17 +1,33 @@
+import json
 import logging
 import os
 import sys
 from typing import Dict, List
 
+import cellxgene_schema
+
 from backend.layers.business.business import BusinessLogic
 from backend.layers.business.entities import CollectionQueryFilter
+from backend.layers.common.entities import DatasetArtifactType
 from backend.layers.persistence.persistence import DatabaseProvider
 from backend.layers.processing import logger
 from backend.layers.thirdparty.s3_provider import S3Provider
 from backend.layers.thirdparty.uri_provider import UriProvider
-from backend.portal.api.providers import get_business_logic
 
 logger.configure_logging(level=logging.INFO)
+
+
+database_provider = DatabaseProvider()
+s3_provider = S3Provider()
+uri_provider = UriProvider()
+
+business_logic = BusinessLogic(
+    database_provider,
+    None,  # Not required
+    None,  # Not required
+    s3_provider,
+    uri_provider,
+)
 
 
 def gather_collections() -> Dict[str, Dict[str, List[str]]]:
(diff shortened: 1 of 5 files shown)


backend/schema_migration/migrate.py
```diff
@@ -1,88 +1,132 @@
...
+        raw_h5ad_uri = [
+            artifact.uri
+            for artifact in self.business_logic.get_dataset_artifacts(dataset_version_id)
+            if artifact.type == DatasetArtifactType.RAW_H5AD
+        ]
+        bucket_name, object_key = self.business_logic.s3_provider.parse_s3_uri(raw_h5ad_uri)
+        self.business_logic.s3_provider.download_file(bucket_name, object_key, "raw.h5ad")
+        cellxgene_schema.migrate("raw.h5ad", "migrated.h5ad", collection_id, dataset_id)
+        # TODO: define where this upload bucket actually should be
+        upload_bucket_name = os.environ["UPLOAD_BUCKET"]
```
Bento007, line 76: We decided we would upload to the main bucket. I think we should go ahead and overwrite the raw.h5ad always because if a user re-uploads the same dataset in the same versions, we want to overwrite all of the files previously stored under that version_id.

backend/schema_migration/migrate.py
```diff
@@ -1,88 +1,132 @@
...
+        dst_uri = f"{dataset_version_id}/migrated.h5ad"
+        self.business_logic.s3_provider.upload_file("migrated.h5ad", upload_bucket_name, dst_uri)
+        url = f"s3://{upload_bucket_name}/{dst_uri}"
+        new_dataset_version_id, _ = self.business_logic.ingest_dataset(collection_id, url, None, dataset_version_id)
```
Bento007, line 80: skip the step function trigger step, and return the parameters needs to trigger the step function in the response.

backend/schema_migration/migrate.py
```diff
@@ -1,88 +1,132 @@
...
+            {
+                "collection_id": private_collection_id,
+                "dataset_id": dataset["dataset_id"],
+                "dataset_version_id": dataset["dataset_version_id"],
+            }
+            for dataset in datasets
+        ]
+
+    def corpus_migrate(self) -> bool:
+        self.gather_collections()
```
Bento007, line 99: you can return this response directly

backend/schema_migration/migrate.py
```diff
@@ -1,88 +1,132 @@
...
+            "revision": {"collection_version_id": ["dataset_id", "dataset_id", ...], ...}
+            "private": {"collection_id": ["dataset_id", "dataset_id", ...], ...},
+        }
+        """
 
-if __name__ == "__main__":
-    database_provider = DatabaseProvider()
-    s3_provider = S3Provider()
-    uri_provider = UriProvider()
+        CollectionQueryFilter()
```
Bento007, line 37: remove this

tests/unit/schema_migration/test_dataset_migrate.py
```diff
@@ -0,0 +1,25 @@
...
+        schema_migrate.business_logic.s3_provider.parse_s3_uri.return_value = ("fake-bucket", "object_key.h5ad")
+        schema_migrate.business_logic.get_dataset_artifacts.return_value = [
+            DatasetArtifact(id=None, type="raw_h5ad", uri="s3://fake-bucket/object_key.h5ad")
+        ]
+        dataset_version_id = DatasetVersionId().id
+        schema_migrate.business_logic.ingest_dataset.return_value = (
+            dataset_version_id,
+            private.datasets[0].dataset_id.id,
+        )
+        with mock.patch("backend.schema_migration.migrate.cellxgene_schema"):
```
Bento007, line 21: Assert that this is called

Bento007 (changes requested):
Add things we discussed and a few things to fix

PR #5063, opened 2023-06-20
feat(schema-migration): move schema_migration into processing container

Bento007:
## Reason for Change

- simplifies the needed infra changes.

## Changes

- move schema_migration into the processing container.

## Testing steps

- moved unit tests into the processing test suit.


nayib-jose-gloria (approved):
thanks! lets go for it

nayib-jose-gloria (commented):
One note--we should remove schema_migration from the docker compose yaml as well

Issue #4801, opened 2023-05-09
Error Handling Job -- Gather error reports from other steps


nayib-jose-gloria:
notify ops channel on any failures during migrated dataset processing

Idea: Aggregate any failures in a separate Batch Job rather than separate notifications. May require an update to processing pipeline (or to how we invoke it) to avoid hundreds of individual failure notices in the channel in case of a bad script causing validation failures across the corpus migration

PR #5094, opened 2023-06-23
feat(schema migrations): add error collecting for report

Bento007:
## Reason for Change

- #4801 
- The report step needs to know the status of previous step. Since we are using batch job and recycling the processing container, it's difficult to pass in large number of parameters. To get around this we can upload the results of each step that requires reporting to s3 and have the report step use those files.

## Changes

- add report that read from a directory in s3 and processes each file

## Testing steps
- add unit tests


nayib-jose-gloria (commented):
I like this proposed flow! Do you intend to address the report generation TODO in this PR or a separate one? If separate, this looks good to merge

backend/layers/thirdparty/s3_provider.py
```diff
@@ -100,3 +100,8 @@ def upload_directory(self, src_dir: str, s3_uri: str):
             command,
             check=True,
         )
+
+    def list_directory(self, bucket_name: str, src_dir: str) -> list[str]:
```
nayib-jose-gloria, line 104: either return the list here directly or fix return type to generator

Bento007:
regarding the todo, we don't have a format establish and need the slack secret. I think it can be done as a separate issue.

ebezzi (approved):

PR #5112, opened 2023-06-27
fix: fix the double-selection bug on publication filter

carolinechoi:
## Fixing the double selection of keys with the same name in filters

- n/a

## Changes

- add to default dropdown props


frontend/src/views/WheresMyGene/components/Filters/index.tsx
```diff
@@ -71,10 +72,15 @@ const filterOptions = createFilterOptions({
     `${option.label} ${option.collection_label}`,
 });
 
+function isOptionEqualToValue(option: FilterOption, value: FilterOption) {
+  return option.id === value.id;
```
tihuan, line 76: just curious, by default it's matching `name`, how did that cause double selection? Thanks!

tihuan (approved):
LGTM! Thanks for the fix, @carolinechoi 😄 🙏

PR #5122, opened 2023-06-30
feat: SFN for schema migration

ebezzi:
Infra&code changes for the schema migration step function.


.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -0,0 +1,321 @@
...
+                    "Name": "STEP_NAME",
+                    "Value": "collection_publish"
+                  },
+                  {
+                    "Name": "MIGRATE",
+                    "Value": "True"
+                  },
+                  {
+                    "Name": "COLLECTION_ID",
+                    "Value.$": "$[0].Input.collection_id"
```
ebezzi, line 258: Here, the previous Map steps returns an array of datasets. I am just grabbing the first element since they're all from the same collection.

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -0,0 +1,321 @@
...
+                  },
+                  {
+                    "Name": "CAN_PUBLISH",
+                    "Value": "true"
```
ebezzi, line 262: This was previously coming from the input, but as far as I can see, nothing ever sets this variable. I hardcoded it as true for now.

backend/layers/processing/requirements.txt
```diff
@@ -1,4 +1,4 @@
-cellxgene-schema
+git+https://github.com/chanzuckerberg/single-cell-curation/#subdirectory=cellxgene_schema_cli
```
ebezzi, line 1: This one needs to be reverted at some point, but it might be a wise idea to keep this (at least) until we test on the whole corpus.

backend/layers/processing/schema_migration.py
```diff
@@ -24,88 +26,93 @@ class SchemaMigrate:
...
-                datasets.append({"dataset_id": dataset.dataset_id.id, "dataset_version_id": dataset.version_id.id})
-            return datasets
+        response = []
 
         collections = itertools.chain(unpublished_collections, published_collections)
         # evaluate unpublished collections first, so that published versions are skipped if there is an active revision
         has_revision = []  # list of collections to skip if published with an active revision
         for collection in collections:
+
+            # TODO <testing code>
```
ebezzi, line 52: This snippet was introduced to test in rdev. This is because pre-seeded collections can't be migrated (they don't belong to the environment). It should be removed before merging.

Issue #5136, opened 2023-07-06
Duplicate asset artifacts being stored in DatasetArtifact table


danieljhegeman:
See [**#single-cell-discover-api**](https://czi-sci.slack.com/archives/C02UGUV5R3J/p1688595007007299). 

[Originally reported by Bruce with census error](https://github.com/chanzuckerberg/cellxgene-census/issues/606)

The `DatasetArtifact` table has a few uris that are duplicated:

`select count(uri), uri from "DatasetArtifact" group by uri having count(uri) > 1 order by count(uri) desc;`

The corresponding `DatasetVersion.artifacts` array contains all of the duplicates.

A simple fix for now would be de-duping when formulating the API response. Likely arising from insufficient cleanup of artifacts during a failed processing attempt, but not known for sure.

atolopko-czi:
Might be worth considering adding relevant scheme constraints to database tables.

danieljhegeman:
@atolopko-czi yeah that's a better fix, along with handling any such errors that would then come out of the processing pipeline. Quick fix for now.

brianraymor:
> Might be worth considering adding relevant schema constraints to database tables.

@atolopko-czi @danieljhegeman 

Should there be a separate tracking issue after this band-aid is applied?

danieljhegeman:
@brianraymor ☝️

PR #5140, opened 2023-07-06
fix: filter out errant duplicate assets on Discover API responses

danieljhegeman:
## Reason for Change

- #5136

## Testing steps

- tested against dev with adhoc deploy ✅

### DIFF
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -35,8 +35,12 @@ def get_collections_base_url():
 def extract_dataset_assets(dataset_version: DatasetVersion):
     base_url = CorporaConfig().dataset_assets_base_url
     asset_list = list()
+    included_asset_types = set()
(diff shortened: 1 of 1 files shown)


Bento007 (changes requested):
Where is the test for this?

danieljhegeman:
@Bento007 https://czi-sci.slack.com/archives/C02UGUV5R3J/p1688679605578739?thread_ts=1688595007.007299&cid=C02UGUV5R3J

Bento007:
Given the above thread, I recommend pulling out the filter into it's own function that takes in the list of assets and filters them. This will allow us to not only test it works, but also removing the code will be easier.

maniarathi:
+1 to what @Bento007 said. This PR needs tests. Data integrity issues are not ok and we should be unit testing failure cases to ensure it never happens again.

danieljhegeman:
closing in favor of #5163

PR #5163, opened 2023-07-06
fix: prevent duplicate assets from being returned by Discover API

danieljhegeman:
This temporary fix addresses a more fundamental bug in the pipeline logic which is resulting in occasional instances of duplicate asset rows in the DatasetArtifact table.

## Reason for Change

- #5136 

## Testing steps

- unit tested


Bento007 (approved):

danieljhegeman:
Wrong base branch 🤦‍♂️ 

Fixed.

PR #5184, opened 2023-07-10
feat: add sqsum array to expression summary default cube [WIP]

joyceyan:
## Reason for Change

first component of this ticket: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5093

this PR adds the `sqsum` array to the expression summary default cube. up next:
- actually read from it in `calculate_markers`
- deprecate `expression_summary_fmg`

this PR also includes some aesthetic code changes, such as:
1. ensuring we have the return type annotated on every function i touched
2. adding the `*` decorator to all functions i touched, which requires callers to name the params. this helps ensure that you don't end up touching the wrong parameter just because you ordered something wrong
3. created Result class objects rather than returning tuples. i think having a typed result object is a lot easier to read when you're returning more than two things. similarly, this helps prevent the caller of the function from setting one of the result values to be the wrong thing if you just ordered your variables wrong

## Testing steps

re-generated realistic test snapshots


backend/wmg/pipeline/summary_cubes/expression_summary_fmg/job.py
```diff
@@ -28,7 +28,7 @@ def _load(
     cube_sqsum: np.ndarray,
     cube_nnz: np.ndarray,
     cube_nnz_thr: np.ndarray,
-) -> (list, dict):
```
joyceyan, line 31: we're not actually returning anything here, so i updated this to `None`

backend/wmg/pipeline/summary_cubes/expression_summary/job.py
```diff
@@ -83,7 +92,15 @@ def create_expression_summary_cube(corpus_path: str, default=False):
         gene_ontology_term_ids = extract_var_data(corpus_path, ctx)
 
         # transform
-        cube_index, cube_sum, cube_nnz = transform(corpus_path, gene_ontology_term_ids, cube_dims)
-        _load(uri, gene_ontology_term_ids, cube_index, cube_sum, cube_nnz, default=default)
+        result = transform(corpus_path, gene_ontology_term_ids, cube_dims)
```
atarashansky, line 95: I think the args here need to be keyword arguments. The pipeline failed with an error indicating as such:
https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/%2Fdp%2Frdev%2Fstack-name%2Fwmg-processing/log-events/wmg-rdev-stack-name-wmg-processing%2Fdefault%2F1bd4a7129c9d4943a9881f7974f6389b

atarashansky (changes requested):

joyceyan:
closing to open a new PR with rdev stack

Issue #5164, opened 2023-07-07
Introduced a facade for writes that will handle the s3 location to which the wmg pipeline writes the snapshot and metadata information


prathapsridharan:
This is an interface that currently implements the current way of writing the snapshot and `latest_snapshot_identifier` file to `s3`. This step enables us to change where the snapshot is written in s3 in the new design.

The starting point to introduce the facade is [around here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/eb71ce726adc151478aa4243e62414c706225739/backend/wmg/pipeline/cube_pipeline.py#L67)

prathapsridharan:
This is an interface that currently implements the current way of writing the snapshot and `latest_snapshot_identifier` file to `s3`. This step enables us to change where the snapshot is written in s3 in the new design

Issue #4680, opened 2023-04-26
Decouple WMG application lifecycle from data lifecyle


prathapsridharan:
This [tech spec](https://docs.google.com/document/d/1A1pLYUus1uDQXfDeuT_JiFLZwfsthhOL-g3Mq-2Bvjg/edit?pli=1#heading=h.jcf27gbiblyg) lays out the problem statement, the goals, and the design of the feature encapsulated in this epic.

The problem statement and goal below is copied from the tech spec:

The motivation for this work is to improve the evolvability and operability of the WMG application. In other words, we want to be able to make changes to the application more easily and reduce the likelihood of production outages. There may be multiple solutions to improve the evolvability and operability of the system. Each solution may involve multiple steps or phases. The first step in any solution is to decouple the data lifecycle from the application lifecycle. This document specifies a method to logically decouple the data generating component (WMG pipeline) from the data consuming component (WMG API) by introducing a data version pinning scheme such that:

- The WMG pipeline component can generate new snapshots of the data that may alter the data schema or semantics without worrying about breaking the WMG API component.

- The WMG API component can rollback to consume another data version if it encounters trouble when reading the current data version.

metakuni:
@joyceyan / @prathapsridharan  , this epic is listed as a blocker for the CI/CD work but it looks like all work has been completed. Can it be closed?

PR #5187, opened 2023-07-11
refactor: Introduce a facade api for the write path of wmg pipeline

prathapsridharan:
## Reason for Change

The purpose of this ticket is to address ticket #5164 for the epic: #4680. In this change list, we create a facade interface for the write path of wmg pipeline so that subsequent implementation changes to how the write path can be done behind this interface. Introducing the facade makes possible iterative and stable changes to the write path.

- #5164 

## Changes

- add a facade api for the write path of wmg pipeline so that implementation can be changed behind the facade interface
- add a step to write `latest_snapshot_run` file that contains the relative path (relative to the s3 bucket) of the folder containing the latest cube data artifact

## Testing steps

1. Kick off run of the cube generation on `staging` or `dev`
2. Copy the s3 data artifacts of the run in step (1) to an `rdev` environment and manually test the web application

## Notes for Reviewer

In this PR, no major behavior changes are introduced. Specifically,

- The data artifacts are still written at a folder at the root of the s3 bucket and `latest_snapshot_identifier` still contains the name of this folder
- We introduce a small change to write a `latest_snapshot_run` file that contains the relative path (relative to the s3 bucket) of the folder containing the latest cube data artifact


backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+    Specifically:
+
+    1. Snapshot artifacts are stored in s3://<environment_bucket_name>/<snapshot_id>/
+       A. Note that, as of now, <snapshot_id> is a integer representing the time elapsed
+          in seconds since epoch
+       B. If the snapshot generation pipeline is run in a "remote dev environment", then the snapshot
+          is stored in s3://<environment_bucket_name>/<remote_stack_name>/<snapshot_id>/
+       C. If validation checks on the snapshot fails, the data artifact will be stored in
+          s3://<environment_bucket_name>/<snapshot_id>_validation_failed_snapshot/
+          (or s3://<environment_bucket_name>/<remote_stack_name>/<snapshot_id>_validation_failed_snapshot/)
```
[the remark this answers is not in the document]
atolopko-czi: Oh the failed snapshot going to be originally uploaded with the "validation_failed" infix, right? As opposed to renaming an S3 "dir".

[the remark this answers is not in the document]
prathapsridharan: @atolopko-czi - That's right. Moreover, this PR doesn't change any existing behavior (_yet_), except for adding another file called `latest_snapshot_run`. This `latest_snapshot_run` file won't be used by the reader but more meant as an easy way for operators to check where the latest run is located given that the new design will store snapshots in `s3://<bucket_name/snapshots/<data_schema_version>/` directory and it would be useful to have a `s3://<bucket_name>/latest_snapshot_run` at the top level that contains the path to the latest snapshot run (ex: `s3://<bucket_name/snapshots/v2/1688370972`)

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
 
 stack_name = os.environ.get("REMOTE_DEV_PREFIX")
 wmg_bucket_name = os.environ.get("WMG_BUCKET")
-artifact_bucket_name = os.environ.get("ARTIFACT_BUCKET")
```
joyceyan, line 16: was this previously just not being used at all? might be nice to also remove it from wherever we set the environment variables. not a big deal though

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+    sync_command = ["aws", "s3", "sync", snapshot_source_path, snapshot_s3_dest_path]
+
+    subprocess.run(sync_command)
+
+    _write_snapshot_metadata(snapshot_schema_version, snapshot_id, is_snapshot_validation_successful)
+
+    return snapshot_s3_dest_path
+
+
+###################################### PRIVATE FUNCTIONS #################################
```
joyceyan, line 88: just as a general comment, i found this a bit harder to follow / review with so much of the logic wrapped in small private functions since i need to jump all over the place to figure out what's actually happening

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+    data_root_path = _get_wmg_s3_root_dir_path()
+    _get_wmg_s3_data_schema_dir_path(snapshot_schema_version)
+    file_name = "latest_snapshot_identifier"
+
+    # TODO(prathap) - change to: return f"{data_root_path}/{data_schema_dir_path}/{file_name}"
+    # when working on: https://github.com/chanzuckerberg/single-cell-data-portal/issues/5166
+    return f"{data_root_path}/{file_name}"
+
+
+def _get_latest_snapshot_run_file_s3_key() -> str:
```
joyceyan, line 104: this function is only used in one place, so i think it would be fine to remove it and move the logic into `_write_latest_snapshot_run_file`

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+    """
+    Return s3 key of latest_snapshot_run file
+    """
+    data_root_path = _get_wmg_s3_root_dir_path()
+    file_name = "latest_snapshot_run"
+
+    return f"{data_root_path}/{file_name}"
+
+
+def _get_wmg_s3_bucket_uri():
```
joyceyan, line 114: i think it makes more sense to move this up to the top, something like:
```
wmg_bucket_name = os.environ.get("WMG_BUCKET")
wmg_bucket_uri = f"s3://{wmg_bucket_name}"
```
and just reference `wmg_bucket_uri` as a static variable throughout. mostly just for readability purposes but i think there are also some (very minimal) performance gains to removing the function call and referencing the same variable

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+    """
+    Get s3 bucket uri
+    """
+    return f"s3://{wmg_bucket_name}"
+
+
+def _get_wmg_s3_root_dir_path():
```
joyceyan, line 121: same here, would be fine to just have this be a static variable declaration at the top
```
stack_name = os.environ.get("REMOTE_DEV_PREFIX")
wmg_s3_root_dir_path = stack_name.strip("/") if stack_name else ""
```

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+def _get_wmg_snapshot_s3_path(
+    snapshot_schema_version: str, snapshot_id: str, is_snapshot_validation_successful: bool
+) -> str:
+    """
+    Get the path of the directory in which the data artifact is located
+    """
+    snapshot_dir = snapshot_id if is_snapshot_validation_successful else f"{snapshot_id}_validation_failed_snapshot"
+
+    data_root_path = _get_wmg_s3_root_dir_path()
+    _get_wmg_s3_data_schema_dir_path(snapshot_schema_version)
```
joyceyan, line 144: what is this call suposed to do? i think this needs to be `snapshot_dir = _get_wmg_s3_data_schema_dir_path(snapshot_schema_version)` since we're otherwise not using the return value?

i think it would also be fine to just hard-code `f"snapshots/{snapshot_schema_version}`

backend/wmg/data/load_cube.py
```diff
@@ -13,37 +13,186 @@
...
+
+def _make_snapshot_active(snapshot_schema_version, snapshot_id):
     """
-    Remove all data stored under the oldest timestamp
+    Update the latest snapshot reference file to point to the newly generated snapshot and remove older snapshots
+    """
+    _write_latest_snapshot_identifier_file(snapshot_schema_version, snapshot_id)
+
+    data_root_path = _get_wmg_s3_root_dir_path()
+    _get_wmg_s3_data_schema_dir_path(snapshot_schema_version)
```
joyceyan, line 169: what is this call supposed to do?

backend/wmg/data/load_cube.py
prathapsridharan: the left hand variable was removed by the linter. will fix

backend/wmg/data/load_cube.py
prathapsridharan: will fix with a refactor here by consolidating some code

backend/wmg/data/load_cube.py
prathapsridharan: agreed. will fix

backend/wmg/data/load_cube.py
prathapsridharan: agreed. will fix

backend/wmg/data/load_cube.py
prathapsridharan: agreed. will fix. same for `_get_latest_snapshot_identifier_file_s3_key` as well

joyceyan (approved):

PR #5188, opened 2023-07-11
feat: pull latest ontology mappings from s3 on webhook trigger

nayib-jose-gloria:
## Reason for Change

- #4796

## Changes

- add github action that, on webhook trigger (to be created in single-cell-curation upon new ontology reference file release), pulls latest ontology mappings for backend + frontend in s3 and commits to repo + pushes to main
- Changed frontend constants to pull data from json fixtures rather than hardcoded into constants.ts file (but retained constant name)

## Testing steps

- Tested in branch (see [AUTO: pull latest ontology files after schema migration](https://github.com/chanzuckerberg/single-cell-data-portal/commit/1e2361375ce9375d3fe287255406e5e7423d9e89) and the [action](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/5521559298) that created it)


.github/workflows/pull-latest-ontology-mappings.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          pip install awscli==1.29.1
+      - name: setup git
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+      - name: Pull Ontology Mappings from S3 and write to data-portal
+        run: |
+          aws s3 sync s3://cellxgene-schema-ref-files-prod/ontology-mappings/backend backend/common/utils/ontology_mappings/fixtures
+          aws s3 sync s3://cellxgene-schema-ref-files-prod/ontology-mappings/frontend frontend/src/components/common/Filter/descendant_mappings
```
Bento007, line 36: nit: for readability can you put these in the same order as you are committing them

Issue #5102, opened 2023-06-26
Tissue Descriptions


niknak33:
Based on internal feedback and some external feedback, the tissue page would benefit from having tissue descriptions

User Stories:

As a user, I would like to see a description of the tissue page I am viewing.

PR #5211, opened 2023-07-12
feat(cell-guide): add tissue descriptions to TissueCards

atarashansky:
## Reason for Change

- #5102 
- Descriptions are parsed from the `latest` UBERON OBO file.

## Testing
 - added e2e test
 - tests pass
 - tested locally


brianraymor:
> Descriptions are parsed from the latest UBERON OBO file.

Why can't the pinned version be used?

frontend/src/views/CellGuide/components/TissueCard/index.tsx
```diff
@@ -36,6 +48,14 @@ export default function TissueCard(): JSX.Element {
     return () => window.removeEventListener("resize", onResize);
   }, []);
 
+  const [descriptionUberon, setDescriptionUberon] = useState<string>("");
```
tihuan, line 51: Feels like we might not need this state and can just use rawDescriptionUberon directly?

tihuan (approved):
FE code looks great! Thanks for adding this quickly, Alec!

Would you be able to add a 150 char version of tissue description as part of the SEO JSON file too? 

Please see Brian Raymor's question before merging :pray: 

Thank you!

atarashansky:
> FE code looks great! Thanks for adding this quickly, Alec!
> 
> Would you be able to add a 150 char version of tissue description as part of the SEO JSON file too?
> 
> Please see Brian Raymor's question before merging 🙏
> 
> Thank you!

You'd think I would have learned by now to use pinned versions! Will update this now. Thanks for reminding me again

frontend/src/views/CellGuide/components/TissueCard/index.tsx
tihuan: @atarashansky PTAL thanks!

tihuan: Actually never mind, for SEO I'm going to fetch the description at build time, so I'll just remove this completely! Thanks!

Issue #5141, opened 2023-07-06
Update data schema and expression summary cube to include "publications"


carolinechoi:
Adding "publication" citation to the expression summary cube and updating the data schema to reflect this change. Also adding "publication" as part of the /wmg/v1/filters functionality in order to update the frontend code.  

- [x] update schema in backend.wmg.data.schemas.cube_schema
- [x] update cell_count.py in backend.wmg.pipeline.summary_cubes.cell_count

PR #5215, opened 2023-07-13
feat: add publications to the cube (BE)

carolinechoi:
## Adding in "publications" as an option for the cube 
- #5141 

## Changes 🤠

- added utility function to cell_count
- added publication citation attribute
- modded build_in_mem_cube function to add in publication range


seve (approved):
wait for more eyes

backend/wmg/pipeline/summary_cubes/cell_count.py
```diff
@@ -61,6 +66,32 @@ def load(corpus_path: str, df: pd.DataFrame) -> str:
     return uri
 
 
+def return_dataset_dict_w_publications():
+    # cchoi: modifying the dataframe to contain publication coll IDs
```
atarashansky, line 70: remove this line as its no longer accurate (or update with accurate description)

backend/wmg/pipeline/summary_cubes/expression_summary/load.py
```diff
@@ -57,8 +59,16 @@ def build_in_mem_cube(
...
+        dataset_index = 0
+
         for i, k in enumerate(other_cube_attrs):
-            vals[k][idx : idx + n_vals] = attr_values[i]
+            if k == "dataset_id":
+                dataset_index = i
+            if k != "publication_citation":
+                vals[k][idx : idx + n_vals] = attr_values[i]
+
+        dataset_dict = return_dataset_dict_w_publications()
```
atarashansky, line 70: put this outside the for loop - we don't want it to run every iteration.

backend/wmg/pipeline/summary_cubes/expression_summary/load.py
```diff
@@ -57,8 +59,16 @@ def build_in_mem_cube(
...
+        vals["publication_citation"][idx : idx + n_vals] = dataset_dict(attr_values[dataset_index])
```
atarashansky, line 71: `dataset_dict[attr_values[dataset_index]]`

Use `datastic_dict[]` instead of `dataset_dict()`

atarashansky (changes requested):
Looks good - a few minor mistakes. Fix these and we can merge :)

backend/wmg/pipeline/summary_cubes/expression_summary/load.py
carolinechoi: Oh I thought line 68 needed to be run for each column name. And then everything from line 70 down runs outside of the for loop?

atarashansky: I mean outside the outer for loop

atarashansky: `dataset_dict = return_dataset_dict_w_publications()` We don't want this line to run for every row of the cube index

atarashansky (approved):

PR #5216, opened 2023-07-13
fix(wmgv2): analytics changes

seve:
Addressing the following needs:

>  WMG_HEATMAP_LOADED
current behavior: still only fires the first time a heat map is rendered
desired behavior: should now fire whenever a new heat map is rendered even if there is already currently a heat map loaded (i.e. add another gene, use a filter, add a cell type)
WMG_SELECT_CELL_TYPE
payload name should be “cell_type” and only contain the single cell type that was just selected
WMG_DOWNLOAD_COMPLETE
has the “tissues_expanded” payload field, but it doesn’t track the tissues that were expanded at the time of download


frontend/src/views/WheresMyGeneV2/components/GeneSearchBar/components/SaveExport/index.tsx
```diff
@@ -734,7 +734,7 @@ function download_({
         tissues_expanded: expandedTissues,
         tissue_filter: selectedFilters.tissues,
         genes: selectedGenes,
-        cell_types_selected: filteredCellTypes,
+        cell_types_selected: [...filteredCellTypes],
```
tihuan, line 737: why do we need to create a new array here? Thanks!

seve: transforming a set to list

tihuan: ah cool! Thank you!

Issue #5214, opened 2023-07-13
tech(wmg): retrieve only the required attributes and dimensions per WMG query


atarashansky:
Currently, we are retrieving all attributes and dimensions per query:
 - https://github.com/chanzuckerberg/single-cell-data-portal/blob/0be546c02366cc4aa3da989e37e3c623a052271e/backend/wmg/data/query.py#L156

This is bad because we need to allocate 500Mb per attribute and 1 Gb per dimension, meaning we are using way more memory than is required since we only need a few dimensions and attributes.

To retrieve only the attributes and dimensions required for downstream handling, please set the `attrs` and `dims` parameters in `query(...)`.

When `compare` is not set in the request body, we need to set `dims` to `["tissue_ontology_term_id", "gene_ontology_term_id"]` and `attrs` to `["cell_type_ontology_term_id"]`. When `compare` is set, we set `attrs` to `["cell_type_ontology_term_id", f"{compare}"]`.

PR #5221, opened 2023-07-13
feat: retrieve only required attrs and dims [wip]

joyceyan:
## Reason for Change

https://github.com/chanzuckerberg/single-cell-data-portal/issues/5214

## Changes

- passes in the `dims` and `attrs` as specified
- adds `compare` as a param to the query criteria objects for `WmgQueryCriteria` and `WmgQueryCriteriaV2`
- not entirely sure what the expected behavior should be for the `FmgQueryCriteria`, though maybe this doesn't matter that much since that will be deprecated hopefully soon

## Testing steps

- make sure existing tests pass
- TODO: write test coverage to make sure the `compare` param is being passed down correctly


backend/wmg/data/query.py
```diff
@@ -153,7 +155,17 @@ def _query(
 
         tiledb_dims_query = tuple(tiledb_dims_query)
 
-        query_result_df = cube.query(cond=query_cond or None, use_arrow=True).df[tiledb_dims_query]
+        attributes = ["cell_type_ontology_term_id"]
+
+        if not isinstance(criteria, FmgQueryCriteria) and criteria.compare_dimension is not None:
```
joyceyan, line 160: the `compare` param isn't in the fmg queries, but i'm not 100% sure if it's correct to query just the `cell_type_ontology_term_id` attribute or if we should just not be passing in this param and querying all attributes

backend/wmg/data/query.py
```diff
@@ -153,7 +155,25 @@ def _query(
 
         tiledb_dims_query = tuple(tiledb_dims_query)
 
-        query_result_df = cube.query(cond=query_cond or None, use_arrow=True).df[tiledb_dims_query]
+        if isinstance(criteria, FmgQueryCriteria):
+            # fmg queries retain existing behavior of pulling all dims and attrs
+            query_result_df = cube.query(cond=query_cond or None, use_arrow=True).df[tiledb_dims_query]
```
joyceyan, line 160: not entirely sure if this is right, but when passing in the "no compare param" logic, all of the tests fail with this:
```
E   tiledb.cc.TileDBError: Selected dimension does not exist: 'gene_ontology_term_id'
```

so i'm guessing that we want to retain the existing behavior for FMG queries, and only do the optimization for WMG

atarashansky (approved):

Issue #4468, opened 2023-03-30
Data Portal - update FE dependencies


atolopko-czi:
- [ ] Update and pin FE dependencies
- [ ] Deploy and sanity check that nothing is broken
- [ ] Update [how-to-upgrade-dependencies.md](https://github.com/chanzuckerberg/single-cell/blob/main/how-we-work/how-to-upgrade-dependencies.md)

metakuni:
Labeled as `dp` and `data-viz` since the FE code is still tightly coupled. Handle with care!

metakuni:
^ Removed the `dp` label since FE dependencies for both Gene Expression and Portal are one in the same and Data Viz has already scheduled this work in an upcoming sprint. 🙇

PR #5222, opened 2023-07-13
chore: 4468 FE update deps

tihuan:
## Reason for Change

- #4468 

## Changes
1. Updating dependencies up to minor versions only (major versions would need separate PRs if we choose to do them, since they require more time and research)
2. Couldn't update next-mdx-remote beyond the current `4.1.0`, since it throws `_jsx not a function` error ([Issue](https://github.com/hashicorp/next-mdx-remote/issues/350))

## Testing steps

1. The app should still work as expected


tihuan:
Hmm need to troubleshoot mdx-remote a little - will continue tomorrow!

seve (approved):

Issue #4810, opened 2023-05-09
Integrate schema_version parameter with Discover API index endpoints


nayib-jose-gloria:
get latest published dataset version with version matching 'schema_version' parameter if passed into discover API dataset index endpoint

PR #5223, opened 2023-07-13
feat: add schema_version query param to curation API dataset index endpoints

nayib-jose-gloria:
## Reason for Change

- #4810

## Changes

- add optional schema_version query param to curation API datasets index endpoint in OpenAPI spec, with examples and description
- Added 'schema_version' to Collection Version Table in DB
     - To be populated on _publish_, based on schema_version of datasets to publish. If dataset versions do not match within the Collection, an exception will be thrown and block publish. 
     - Revision Collection Versions will not copy the previous version's schema_version. It will be set to None for ALL private collection versions. 
- Added 'schema_version' optional parameter support in curation/v1/datasets endpoint. 
        - No schema_version input == existing route to fetch all latest published dataset versions
        - Parses schema_version input to determine if input is matching for MAJOR, MINOR or PATCH version and if input is valid. 
        - Returns 400 if not valid (i.e. doesn't follow semver rules, uses non-digits since we don't support RCs or prerel versions). 
        - For MAJOR and MINOR versions, transforms schema_version input to include wildcards for db querying (i.e. '3' -> '3._._' to match all 3.X.X, '3.1' -> '3.1._' to match all '3.1.X'). For PATCH, passes input as-is. 
- Added business layer function that 1) calls persistence function to fetch all collection versions matching a schema_version, 2) filters for the most recently published collection version for every canonical collection version with at least one published version, 3) fetches dataset details for each of those collection versions 
- Added persistence layer function to query DB for all collection versions with a matching schema_version. Will query using LIKE statement if wildcards exist (i.e. major or minor schema version passed in), or == statement for full schema versions including patch

## Testing steps
- Testing in rdev
- Unit Tests In-Progress

## Notes for Reviewer
- [Swagger](https://rdev-schema-version-backend.rdev.single-cell.czi.technology/curation/ui/#) (Curation)
- Will Backfill DB with schema version ('3.0.0' for every currently published collection version) upon merge to each env. schema_version field is nullable, so it will not affect any existing endpoints behavior, and special coordination with on-call isn't required.


backend/curation/api/curation-api.yml
```diff
@@ -1457,6 +1459,14 @@ components:
...
       schema:
         $ref: "#/components/schemas/dataset_version_id"
+    schema_version:
+      description: Filter to fetch latest version of dataset with the given CELLxGENE schema version
+      in: query
+      name: schema_version
+      required: false
+      schema:
+        type: string
+        example: "3.0.0"
```
brianraymor, line 1469: What's the behavior if the schema version is unavailable?

5.0.0 or 1.1.0 ?

nayib-jose-gloria: No schema match == No results.

But that's flexible. I'm thinking the downstream consumers that will want to use this will have the necessary context to pass a valid schema version. And it won't cause breaking changes if we move past a particular version and consumers don't update, since we'll have the previous versions stored to serve them.

brianraymor (commented):
Besides my inlined question, I'm curious if we've addressed all the potential cases in the Discover UX (Data Portal API) since there are a multiplicity of API(s) in play. And there's also the separate REST API used by the Explorer to access collections/datasets.

backend/curation/api/curation-api.yml
brianraymor: That's fine. I think it should be documented, because the other alternative is to return an error status code.

backend/curation/api/v1/curation/datasets/actions.py
```diff
@@ -1,16 +1,28 @@
...
-
     all_datasets_with_collection_name_and_doi = []
+    if not schema_version:
+        collections_with_datasets = get_business_logic().get_all_mapped_collection_versions_with_datasets()
+    else:
+        version_parts = schema_version.split(".")
+        if len(version_parts) > 3 or not all(part.isdigit() for part in version_parts):
+            raise InvalidParametersHTTPException(detail="Invalid Schema Version Input")
+        while len(version_parts) < 3:
+            # wildcard match for exactly 1 character
```
Bento007, line 20: Add a test for this to make sure it escapes correctly, and the wild card filter works.

backend/layers/business/business.py
```diff
@@ -147,6 +147,39 @@ def get_published_collection_version(self, collection_id: CollectionId) -> Optio
...
+        Returns a list with the latest published collection version that matches the given schema_version, for each
+        canonical collection
+        """
+        has_wildcards = "_" in schema_version
+        collection_versions = self.database_provider.get_collection_versions_by_schema(schema_version, has_wildcards)
+
+        # for each published canonical collection, map its most recently published collection version
+        collections = dict()
+        for collection_version in collection_versions:
+            if collection_version.published_at is None:
```
Bento007, line 163: you can filter these out in the database query.

backend/layers/business/business.py
```diff
@@ -147,6 +147,39 @@ def get_published_collection_version(self, collection_id: CollectionId) -> Optio
...
+                continue
+            canonical_collection_id = collection_version.collection_id.id
+            if canonical_collection_id not in collections:
+                collections[canonical_collection_id] = collection_version
+            else:
+                current_collection_version = collections[canonical_collection_id]
```
Bento007, line 169: `current` is a confusing prefix since your are iterating over the collections. It makes me think it's the current collection_version in this iteration. Maybe use `last`, or don't rename it and just use `collections[canonical_collection_id]` bellow.

backend/layers/persistence/persistence.py
```diff
@@ -464,7 +475,7 @@ def save_collection_publisher_metadata(
     def add_collection_version(self, collection_id: CollectionId) -> CollectionVersionId:
         """
         Adds a collection version to an existing canonical collection. The new version copies all data from
-         the previous version except version_id and datetime-based fields (i.e. created_at, published_at)
+        the previous version except version_id, schema_version and datetime-based fields (i.e. created_at, published_at)
```
Bento007, line 478:
```suggestion
        the previous version except version_id, schema_version, and datetime-based fields (i.e. created_at, published_at)
```

Bento007 (commented):
Needs more unit tests

backend/layers/business/business.py
nayib-jose-gloria: I know, though I got tripped up determining whether this is technically business logic. We generally seem to do timestamp processing in the business layer, though I can be convinced to switch it back

Bento007: Either way is fine by me.

Bento007 (approved):

Issue #3039, opened 2022-08-04
Deletion of published datasets in the Discover API must be restricted


brianraymor:
### Context

Authorization to delete published datasets MUST be limited to CZI engineering to avoid accidents, track deletion occurrences and justifications, and to validate that the redirect is in place for the permanent links to the collection and its CXG(s). 

### Current Experience

In a private revision of a public collection, a curator MAY use [`DELETE
​/v1​/collections​/{collection_id}​/datasets​/{dataset_id}`](https://api.cellxgene.cziscience.com/curation/ui/#/Collection/backend.curation.api.v1.curation.collections.collection_id.datasets.dataset_id.actions.delete) to delete a published dataset. The collection can then be published in the curation UX.

### Updated Experience

For all curators:
1. If a dataset is _New_ either in an unpublished collection or private revision of a public collection, `DELETE` is allowed.
2. If a dataset is _in the public collection_, then `DELETE` MUST fail.
3. If a dataset is _in the public collection_ and  _Updated_, then `DELETE` MUST fail.

For CZI engineering:
1. If a dataset is _New_, `DELETE` is allowed.
2. If a dataset is _in the public collection_, `DELETE` is allowed.
3. If a dataset is _Updated_, `DELETE` is allowed.

brianraymor:
@danieljhegeman - let's review together when you're ready to start on this one.

danieljhegeman:
@brianraymor sounds good, will come to you when ready 👌

danieljhegeman:
Going to introduce `CXG_ADMIN` env variable which will be set for situations when admin-override functionality is desired, to be used primarily in backend business logic; pair with context manager for unit tests to permit usage of CXG_ADMIN env var only for subsequences of logic (because normal, non-admin functionality is generally the global behavior that is desired for unit tests).

danieljhegeman:
Confirmed working in dev -- deletes of new Datasets are allowed during a revision, but deletes of a previously-published Dataset yields a 405, even when that Dataset has been updated with a new Dataset.

As an aside/separate issue, I discovered staging is failing when trying to generate an auth token.

PR #5194, opened 2023-07-12
chore: create 'tombstone Collection' utility

danieljhegeman:
## Reason for Change

- #5186 

## Changes

- add logic to the `tombstone_collection` function skeleton in `tombstones.py`
- force `dev`, `staging`, or `prod` value for `--deployment` in general usage of the cxg_admin script.
- use Secrets to populate happy env variable(s) necessary for Business / Persistence logic

## Testing steps

- tested manually, adhoc, tombstoning Collections in dev and staging. Confirmed all entities tombstoned and public assets deleted from public datasets bucket.

## Notes for Reviewer

- try it out! In dev 😅

PR #5226, opened 2023-07-13
feat: restrict deletion of published data

danieljhegeman:
## Reason for Change

- #3039 

## Changes

- add
  - `is_cxg_admin` env variable that can be used to bypass guardrails in the business logic. This way, SCB engineers can use business_logic.py in the cxg_admin scripts to perform business logic that is otherwise-illegal when attempted via the API (i.e., by users). Accordingly, a context manager is added to test infrastructure for ease-of-use in tests.
- remove
  - ability to delete published datasets via the Curation API, whether updated or not
- modify 
  - move mock of CorporaConfig with `mock_audience` return value to narrowest class necessary
  - use dict with mocked CorporaConfig for other tests to permit easy modifications

## Testing steps

- unit tested


tests/unit/backend/layers/common/base_api_test.py
```diff
@@ -24,6 +24,10 @@ def setUp(self):
         )
         self.mock_config.start()
 
+        # TODO: this can be improved, but the current authorization method requires it
+        self.mock = patch("backend.common.corpora_config.CorporaAuthConfig.__getattr__", return_value="mock_audience")
+        self.mock.start()
+
```
danieljhegeman, line 30: Moved here, to the narrowest class necessary. Only required for auth tests, which inherit from this class.

tests/unit/backend/layers/common/base_api_test.py
```diff
@@ -60,10 +64,6 @@ class BaseAPIPortalTest(BaseAuthAPITest, BaseTest):
     def setUp(self):
         super().setUp()
 
-        # TODO: this can be improved, but the current authorization method requires it
-        self.mock = patch("backend.common.corpora_config.CorporaAuthConfig.__getattr__", return_value="mock_audience")
-        self.mock.start()
-
```
danieljhegeman, line 66: moved to BaseAuthAPITest (above)

tests/unit/backend/layers/common/base_test.py
```diff
@@ -80,11 +90,8 @@ def setUp(self):
 
         # Mock CorporaConfig
         # TODO: deduplicate with base_api
-        def mock_config_fn(name):
-            if name == "upload_max_file_size_gb":
-                return 30
```
danieljhegeman, line 85: Replace with a dict (mutable, can add `is_cxg_admin` value without overriding the other config values and the result returned from the mock `side_effect`)

tests/unit/backend/layers/common/base_test.py
```diff
@@ -222,7 +236,7 @@ def generate_revision(self, collection_id: CollectionId) -> CollectionVersionWit
     def generate_dataset(
         self,
         owner: str = "test_user_id",
-        collection_version: Optional[CollectionVersion] = None,
+        collection_version: Optional[typing.Union[CollectionVersion, CollectionVersionWithDatasets]] = None,
```
danieljhegeman, line 239: More flexible for test scripting

backend/common/corpora_config.py
```diff
@@ -31,6 +31,7 @@ def get_defaults_template(self):
             "submission_bucket": os.getenv("DATASET_SUBMISSIONS_BUCKET", "cellxgene-dataset-submissions-test"),
             "collections_base_url": collections_base_url,
             "dataset_assets_base_url": dataset_assets_base_url,
+            "is_cxg_admin": None,
```
ebezzi, line 34: Why not `False`?

backend/layers/business/business.py
```diff
@@ -413,6 +414,10 @@ def remove_dataset_version(
         Removes a dataset version from an existing collection version
         """
         self._assert_collection_version_unpublished(collection_version_id)
+        dv = self.database_provider.get_dataset_version(dataset_version_id)
+        if not CorporaConfig().is_cxg_admin and dv.canonical_dataset.published_at:
+            raise CollectionUpdateException from None
```
ebezzi, line 419: `is_cxg_admin` doesn't belong to the business logic, since it's user based. You should add this as a parameter and move this call to the API layer.

backend/common/corpora_config.py
danieljhegeman: Fair point 😄

backend/layers/business/business.py
danieljhegeman: It determines rules for the business logic; it is involved in API calls.

ebezzi: Who will even set `is_cxg_admin`?

danieljhegeman: Sorry, I should have included the link to the previous issue / PR #5194

danieljhegeman: My thinking on the parameter vs env var: because the entire suite of interactions with the entire service code will either be _entirely_ cxg admin or _entirely_ not cxg admin, it makes more sense to have the logic served by an env var. 

A parameter such as 'get_tombstoned' doesn't make sense as an env var because it could (and does) change during a set of actions. The service, when running in deployed environments, will always have a value of `False` for `is_cxg_admin`.

danieljhegeman: Also the business API should not expose the ability to perform cxg admin functions to the API layer, because cxg admin ops will never occur via the API.

ebezzi: In this case, I'd just delete the method and move it to cxg_admin, or make it private. I think the idea of using an environment variable here is incredibly confusing.

danieljhegeman: The method is still used for deleting unpublished data.

danieljhegeman: Looking forward: as cxg admin scripts expand in scope, we run the risk of duplicating logic (risk of bugs) if we have to have separate functions for cxg admin work.

tests/unit/backend/layers/common/base_test.py
```diff
@@ -80,11 +90,8 @@ def setUp(self):
 
         # Mock CorporaConfig
         # TODO: deduplicate with base_api
-        def mock_config_fn(name):
-            if name == "upload_max_file_size_gb":
-                return 30
 
-        mock_config = patch("backend.common.corpora_config.CorporaConfig.__getattr__", side_effect=mock_config_fn)
+        mock_config = patch("backend.common.corpora_config.CorporaConfig.__getattr__", side_effect=self.mock_config_fn)
         mock_config.start()
```
Bento007, line 95: Add tear down for the mock

Bento007 (changes requested):

backend/layers/business/business.py
```diff
@@ -413,6 +414,11 @@ def remove_dataset_version(
         Removes a dataset version from an existing collection version
         """
         self._assert_collection_version_unpublished(collection_version_id)
+        if not CorporaConfig().is_cxg_admin:
+            dv = self.database_provider.get_dataset_version(dataset_version_id)
+            if dv.canonical_dataset.published_at:
+                raise CollectionUpdateException from None
```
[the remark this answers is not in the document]
danieljhegeman: This is covered by the `405` test I added.

tests/unit/backend/layers/common/base_api_test.py
```diff
@@ -24,6 +24,10 @@ def setUp(self):
         )
         self.mock_config.start()
 
+        # TODO: this can be improved, but the current authorization method requires it
+        self.mock = patch("backend.common.corpora_config.CorporaAuthConfig.__getattr__", return_value="mock_audience")
+        self.mock.start()
```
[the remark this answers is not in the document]
danieljhegeman: Interesting. This wasn't there before; I just moved it to a different class. But I'll add, sure.

backend/layers/business/business.py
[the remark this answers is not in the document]
danieljhegeman: Oh you mean a test in `test_business.py` 🆗

tests/unit/backend/layers/common/base_api_test.py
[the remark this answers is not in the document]
danieljhegeman: I'm actually just not going to touch this since I won't be messing with mock stuff at all with the new approach in the new PR (coming imminently). Let me know if you think this is an actual problem for the existing tests, as it was.

danieljhegeman:
Closing in favor of #5261

PR #5233, opened 2023-07-14
fix(wmg): fix find marker genes in wmg v2

atarashansky:
## Reason for Change

- FMG is broken in WMG v2

## Changes

- Hardcode useMarkerGenes to use version=2 (later, all versioning should be removed from the hooks module)

## Testing steps


joyceyan (approved):
approving cause this will work but i assume we want to have v2 be default not v1

atarashansky:
> approving cause this will work but i assume we want to have v2 be default not v1

Yeah, we will need to rip out the version parameter  anyway since we aren't going back to v1.

Issue #5145, opened 2023-07-06
Modify "Publication" filter to match new data schema


carolinechoi:
- [x] Rewrite code to point to `/wmg/v1/filters`
- [x] Remove unnecessary collection calls from useManyCollections()
- [ ] Revert skipping E2E tests PR

PR #5271, opened 2023-07-18
feat(wmg): added publication citation to v2

carolinechoi:
## Adding Publication col to the v2 filters function

- #5145 
🙉🙈


backend/wmg/api/wmg-api-v2.yml
```diff
@@ -667,6 +667,8 @@ paths:
                         $ref: "#/components/schemas/wmg_ontology_term_id_label_list"
                       tissue_terms:
                         $ref: "#/components/schemas/wmg_ontology_term_id_label_list"
+                      publication_citation:
+                        $ref: "#/components/schemas/wmg_ontology_term_id_label_list"
```
atarashansky, line 671: Can you replace this with:
```
publication_citation:
    type: array
    items:
      type: string
```

atarashansky (changes requested):

PR #5280, opened 2023-07-19
Revert "temp: increase staging resources for debugging"

atarashansky:
Reverts chanzuckerberg/single-cell-data-portal#5279


.happy/terraform/envs/dev/main.tf
```diff
@@ -19,6 +19,6 @@ module stack {
   backend_memory               = 8192
```
Bento007, line 19: we could make this 8192 * backend_cpus

PR #5282, opened 2023-07-19
feat(schema-migration): filter out datasets that have already been upgraded

Bento007:
## Reason for Change

- #5272 

## Changes

- add filter for datasets with update schema version
- add check during publish_and_cleanup to skip if the collection version is published.

## Testing steps

- Updated unit tests to reflect change.


backend/layers/processing/schema_migration.py
```diff
@@ -125,6 +126,8 @@ def collection_migrate(self, collection_id: str, collection_version_id: str, can
                     "dataset_version_id": dataset.version_id.id,
                 }
                 for dataset in version.datasets
+                if dataset.metadata.schema_version != current_schema_version
```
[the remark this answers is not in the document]
Bento007: resolved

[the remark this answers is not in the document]
Bento007: Also this was a great catch. Thank you for the great review.

backend/layers/processing/schema_migration.py
```diff
@@ -102,16 +102,28 @@ def dataset_migrate(
...
+            # check if there are datasets to migrate.
+            return {
+                "can_publish": str(False),  # skip publishing, because the collection is already published and no
+                # revision is created, or the collection is private or a revision.
+                "collection_version_id": collection_version_id,
+                "datasets": [],
+                "no_datasets": str(True),
+            }
 
         if can_publish:
```
nayib-jose-gloria, line 119: edge case to consider: if a collection was originally a can_publish==True during run 1, but for whatever reason did not finish publishing perhaps 1 dataset of N failed to process or timed out), it would become a can_publish==False during run 2 because its in an open revision now. However, we can publish revisions that we created, just not those created before run 1. As an optimization, we should consider addressing this by either looking at timestamps or perhaps even looking for the machine account user under curator_name

nayib-jose-gloria: Up to you whether this should be a separate issue or you want to tackle it here

nayib-jose-gloria (approved):

nayib-jose-gloria (approved):
1 further optimization to consider, either as part of this PR or a separate issue

backend/layers/processing/schema_migration.py
Bento007: Or we check if published in `publish_and_cleanup`. I'll add it to the PR and ask for another review.

nayib-jose-gloria: Will that account for the collections that were can_publish==False in run 1 and should stay can_publish==False in all subsequent runs? Just making sure we don't accidentally publish any private or revisions from before migration that are not ready

backend/layers/processing/schema_migration.py
```diff
@@ -109,8 +109,7 @@ def collection_migrate(self, collection_id: str, collection_version_id: str, can
         if not any(dataset.metadata.schema_version != current_schema_version for dataset in version.datasets):
             # check if there are datasets to migrate.
             return {
-                "can_publish": str(False),  # skip publishing, because the collection is already published and no
-                # revision is created, or the collection is private or a revision.
+                "can_publish": str(can_publish),
```
nayib-jose-gloria, line 112: I don't think this will work to address the publish concern. can_publish will be set to False in gather_collections if the collection is in a revision or private. So, this will always be False for any publishable collection created outside of this run of this function

backend/layers/processing/schema_migration.py
Bento007: > looking at timestamps

The time stamps sounds too variable to be reliable.

> perhaps even looking for the machine account user under curator_name

This is not reliable since most of the dataset will eventually be migrated by the machine account. Also are we setting the curator_name to the machine account when we migrate?

I think the best solution is to do it manually for now.

PR #5283, opened 2023-07-19
chore: update num workers for dev, staging, and prod

atarashansky:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

### DIFF
--- a/.happy/terraform/envs/dev/main.tf
+++ b/.happy/terraform/envs/dev/main.tf
@@ -19,6 +19,6 @@ module stack {
   backend_memory               = 8192
   frontend_memory              = 4096
   backend_instance_count       = 1
-
+  backend_workers              = 1
   wait_for_steady_state = var.wait_for_steady_state
 }
--- a/.happy/terraform/envs/prod/main.tf
+++ b/.happy/terraform/envs/prod/main.tf
@@ -16,11 +16,11 @@ module stack {
   batch_container_memory_limit = 300000
   wmg_batch_container_memory_limit = 248000
   wmg_desired_vcpus                = 128
-  backend_memory               = 61440
+  backend_memory               = 30 * 1024
   frontend_memory              = 4096
   backend_instance_count       = 6
-  backend_cpus                 = 8
-  backend_workers              = 17
+  backend_cpus                 = 4
+  backend_workers              = 5 # Rule of thumb is num CPUs + 1
 
   wait_for_steady_state        = var.wait_for_steady_state
 }
--- a/.happy/terraform/envs/prod/main.tf
+++ b/.happy/terraform/envs/prod/main.tf
@@ -20,6 +20,7 @@ module stack {
   frontend_memory              = 4096
   backend_instance_count       = 6
   backend_cpus                 = 8
+  backend_workers              = 17
 
   wait_for_steady_state        = var.wait_for_steady_state
 }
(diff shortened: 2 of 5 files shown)


Bento007 (commented):
Why not simply increase the memory to the machines?

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -24,7 +24,7 @@ locals {
   frontend_cmd                 = []
   # TODO: Assess whether this is safe for Portal API as well. Trying 1 worker in rdev portal backend containers, to minimize use of memory by TileDB (allocates multi-GB per process)
   # Note: keep-alive timeout should always be greater than the idle timeout of the load balancer (60 seconds)
-  backend_workers              = var.backend_cpus * 2 + 1 # Number of backend workers to run. Using the flask formula 2*cpu+1
+  backend_workers              = var.backend_workers
```
Bento007, line 27: It would be cleaner to make this `var.backend_cpus + 1`. Then we don't need to add the `backend_worker` var, less hardcode variables, and consistently assigned number of workers.

.happy/terraform/envs/prod/main.tf
```diff
@@ -16,10 +16,11 @@ module stack {
   batch_container_memory_limit = 300000
   wmg_batch_container_memory_limit = 248000
   wmg_desired_vcpus                = 128
-  backend_memory               = 61440
+  backend_memory               = 30 * 1024
   frontend_memory              = 4096
   backend_instance_count       = 6
-  backend_cpus                 = 8
+  backend_cpus                 = 4
```
Bento007, line 22: aren't we doing 8/30?

Bento007 (commented):
Eliminating the number of magic number would improve readability for future developers. I'd also suggest adding comments as to why we have those numbers. In particular the amount of memory, and the number of CPUs.

.happy/terraform/envs/dev/main.tf
```diff
@@ -19,6 +19,6 @@ module stack {
   backend_memory               = 8192
```
Bento007, line 19: we should make this `8192 * backend_cpus`

.happy/terraform/modules/ecs-stack/main.tf
atarashansky: I'd like to fix dev and staging to 1 worker, though.

.happy/terraform/envs/prod/main.tf
atarashansky: 8/30 is invalid. 4/30 is valid

.happy/terraform/envs/dev/main.tf
atarashansky: I don't think this is valid for all values of backend_cpus. Check here for valid configs:
https://aws.amazon.com/fargate/pricing/

Bento007: Ah, that makes more sense, why we need to have these asymmetric configurations.

.happy/terraform/modules/ecs-stack/main.tf
Bento007: In that case set the default of backend_workers to 1 and add a comment as to why it is 1.

atarashansky: Gotcha

Issue #5024, opened 2023-06-14
Use independent requirements.txt files for each image


atolopko-czi:
We currently have a schema whereby Docker images install requirements from two requirements.txt files, a "base" file and an image-specific file:

```
ADD requirements.txt requirements-base.txt
ADD backend/api_server/requirements.txt requirements-api.txt
RUN python3 -m pip install -r requirements-base.txt -r requirements-api.txt
```

The makes it more difficult to infer exactly which python package versions are installed since you must 1) first know that this is being done in the Dockerfile, 2) examine both files (which may even specify redundant packages but at different versions).

Also, defining common dependencies across images is not necessarily a good thing, since:
* Different teams may "own" different images, but a single change to the base requirements impacts all images
* We can end up with a base requirement that isn't actually needed by an image, causing image bloat (this is currently happening).
* We sometimes need different versions for the same package in different images. One known case is the TileDB reader-stays-ahead-of-writer case, between the WMG FE/BE image and the WMG pipeline image.

To address these concerns, the team agreed at the 2023-06-14 staff meeting that each Docker image will maintain its own, independent requirements.txt file that lists _all_ of it requirements in one place. Related, we agreed that dev-time-only requirements (testing packages, linters, etc.) should be placed a requirements-dev.txt file, to differentiate between prod run-time requirements and dev requirements.

metakuni:
Removing the `dp` label since Prathap said that he'd be addressing all requirements.txt files. 🙇

Issue #5075, opened 2023-06-21
Gene Expression - update BE dependencies


metakuni:
- [ ] Update and pin BE dependencies
- [ ] Deploy and sanity check that nothing is broken
- [ ] Update [how-to-upgrade-dependencies.md](https://github.com/chanzuckerberg/single-cell/blob/main/how-we-work/how-to-upgrade-dependencies.md)

Consider addressing #5024 first.

PR #5291, opened 2023-07-19
chore: Make single python requirements file for each dockerfile

prathapsridharan:
## Reason for Change

- #5024 

## Changes

- add appropriately named requirements file for each dockerfile at the root of repository
- modify the docker files to **only** include these new requirements file and install them
- remove deprecated `requirements.txt` files
- update README documentation to reflect changes

## Testing steps

- Run `make local-sync` and check that all docker images are reconstructed
- `pip install -r <requirement-files>` in a _local python virtual environment_ and unit tests
- Let CI pipeline on GHA reconstruct the docker images and run the pipeline successfully


Dockerfile.upload_success
```diff
@@ -3,7 +3,8 @@ FROM public.ecr.aws/lambda/python:3.8
 COPY backend/layers/processing/upload_success .
 COPY backend/layers ./backend/layers
 
-RUN pip3 install -r requirements.txt
+ADD requirements-upload-success.txt equirements-upload-success.txt
+RUN pip3 install -r equirements-upload-success.txt
```
atolopko-czi, line 7:
```suggestion
ADD requirements-upload-success.txt requirements-upload-success.txt
RUN pip3 install -r requirements-upload-success.txt
```
Tested?

Dockerfile.data_cube_creation
```diff
@@ -11,9 +11,8 @@ ENV PIP_NO_CACHE_DIR=1
 
 WORKDIR /
 
-ADD backend/wmg/pipeline/requirements.txt requirements.txt
-ADD requirements.txt requirements-base.txt 
-RUN pip3 install -r requirements-base.txt -r requirements.txt 
+ADD requirements-wmg-pipeline.txt requirements-wmg-pipeline.txt
+RUN pip3 install -r requirements-wmg-pipeline.txt
```
atolopko-czi, line 15: Out-of-scope, but it would be nice to align the names of Dockerfiles, requirements files, and python module names. If it doesn't cause downstream pain, could rename this to `Dockerfile.wmg-pipeline`.

Dockerfile.processing_image
atolopko-czi, line 1: Again, if Dockerfile renames aren't too painful, could rename to just `Dockerfile.processing`.

atolopko-czi (changes requested):
One small, but critical typo.

A couple _optional_ filerename suggestions.

Finally, is this a good time to pin all package versions, since we're touching all of the reqs.txt files?

prathapsridharan:
> One small, but critical typo.
> 
> A couple _optional_ filerename suggestions.
> 
> Finally, is this a good time to pin all package versions, since we're touching all of the reqs.txt files?

@atolopko-czi - Pinning will be done #5075 for gene expression at least. I think it is better to separate pinning packages from refactoring dockerfiles so that things are tested in smaller units and regressions can be isolated more easily

Dockerfile.data_cube_creation
prathapsridharan: @atolopko-czi - will do

Dockerfile.processing_image
prathapsridharan: @atolopko-czi - will do

Dockerfile.upload_success
prathapsridharan: @atolopko-czi - I realized I didn't test it properly locally. I added a mods to the `.dockerignore` and tested docker container creation locally! Thanks for catching this.

atolopko-czi (approved):
LGTM as is, and understand some file renames may still occur.

Bento007 (approved):

Issue #5144, opened 2023-07-06
Add Group By functionality for "publications" (FE)


carolinechoi:
- [x] Add "Publication" option to the Group By dropdown
- [x] Confirm that loading indicator is present if loading time is >1 sec
- [ ] Confirm that rows are sorted by size of "Publication" grouping
- [ ] If >10 publications per cell type, add control to 'Show X more' which allows users to see the full list of publications for that cell type.
- [ ] Also add control to 'Show Fewer' to collapse full list 

[Designs](https://www.figma.com/file/nUbwlxcSPAGoRGWyYaIkyv/Q2-2023-Gene-Expression-Improvements?type=design&node-id=434%3A161197&mode=design&t=E1xeQE03C5srUsQt-1)

PR #5300, opened 2023-07-20
feat: adding compare by publication

carolinechoi:
## GroupBy by Publication

- #5144

## Changes

- add option to the v2.py logic
- add FE option in dropdown


tihuan (approved):

tihuan:
Before merging, can you spin up an rdev for Signe to test if the performance is acceptable? Thank you!

Issue #4955, opened 2023-06-05
Analytics (update): Update Group By to account for Publication option


signechambers1:
### Overview

**Tool**: Gene Expression
Feature: Group by

**Description**: Users will now be able to select Publication as an option, we want to know how many people utilize this new feature.

### Updates needed
~It's possible no updates are actually needed, but please confirm the below will happen ~

`WMG_OPTION_SELECT_GROUP_BY`
- Confirm this event will fire when the `publication` dimension is selected 
- Confirm `Publication` value will be surfaced in the `group_by_option` payload

`WMG_DOWNLOAD_COMPLETE`, `WMG_SHARE_CLICKED`, `WMG_SHARE_LOADED`
- Confirm `Publication` value will be surfaced in the `group_by_option` payload

Checklist

- [ ] Once event updates are deployed to Staging ping @ainfeld to check analytics
- [ ] Verify events are showing up in the Staging space in Plausible correctly
- [ ] Once event updates are deployed to Prod ping @ainfeld
- [ ] Confirm Plausible events are still showing up in the Prod space in Plausible correctly

#### People to cc

**PM**: @signechambers1 
**DS**: @ainfeld

Privacy policy: https://cellxgene.cziscience.com/privacy

PR #5318, opened 2023-07-24
fix: analytics for publications

carolinechoi:
## Adding back in the analytics for Publications

- #4955

## Changes

- adding in filter analytics
- verified groupBy analytics
- modified option in dropdown (UI)


prathapsridharan (approved):
lgtm

PR #5329, opened 2023-07-26
refactor: Add SDS utility tokens

tihuan:
## Reason for Change

1. Currently utilizing SDS design tokens to style components is pretty cumbersome, given the amount of boilerplate. 
2. Thanks to @frano-m for pioneering a new pattern - design token utility function, it cuts down the boilerplate to 0
3. This PR extends the pattern being used in a few files to centralize it in `theme.ts`, so the whole app can import the same utility functions
4. This PR also refactors the whole app to use the new utility functions
5. Next step: Work with SDS to introduce design token utility functions, so we can just import from `@czi-sds/components` once it's done

## Changes

1. Add SDS Token utility functions in `theme.ts`
2. Replace the whole app to use the new utility functions

## Testing steps

1. The whole app should retain the same style as before
2. RDEV: https://thuang-styled-util-frontend.rdev.single-cell.czi.technology


frontend/src/components/common/Filter/components/FilterContent/components/FilterMenu/components/FilterMenuItems/style.ts
```diff
@@ -1,16 +1,16 @@
...
 }
 
-const semibold = (props: CommonThemeProps) => getFontWeights(props)?.semibold; // font-weight 600.
-
 export const ListItem = styled("li")<Props>`
   .MuiListItemText-root {
     span:first-of-type {
-      font-weight: ${({ selected }) => (selected ? semibold : undefined)};
+      font-weight: ${(props) =>
+        props.selected ? fontWeightSemibold(props) : undefined};
```
tihuan, line 13: we need to explicitly call `fontWeightSemibold()` with `props` here, since `font-weight:` is already calling an anonymous function `${(props) => ... }`

frontend/src/components/common/Filter/components/FilterContent/components/common/style.ts
```diff
@@ -73,7 +71,7 @@ export const listItemButtonSelectedCss = (props: CommonThemeProps) => {
 
       .MuiListItemText-root {
         span:first-of-type {
-          font-weight: ${semibold(props)};
+          font-weight: ${fontWeightSemibold(props)};
```
tihuan, line 74: we need to explicitly call `fontWeightSemibold(props)` here, because it's using `css from "@emotion/react"`

atarashansky (approved):
LGTM!

PR #5335, opened 2023-07-26
feat(schema-migration): upgrade schema migration resources + skip failed/initialized datasets during migration

nayib-jose-gloria:
## Changes

- split the schema-migration batch job into job that uses swap and one that doesn't. This allows us to spin up smaller jobs and make use of the compute resources allocated.
- parameterize job_queue in the upload sfn. This allow us to use the schema-migration job_queue for the migrations so we don't block the upload job queue when migrating
- fix checks for latest schema version and missing datasets that was causing failures.

## Testing steps
- test in rdev by running schema-migration


Bento007 (approved):

backend/layers/processing/schema_migration.py
```diff
@@ -149,7 +149,7 @@ def collection_migrate(self, collection_id: str, collection_version_id: str, can
                     "dataset_version_id": dataset.version_id.id,
                 }
                 for dataset in version.datasets
-                if dataset.processing_status == "SUCCESS" and dataset.metadata.schema_version != current_schema_version
+                if dataset.status == "SUCCESS" and dataset.metadata.schema_version != current_schema_version
```
nayib-jose-gloria: I think this should be `dataset.status.processing_status` actually

nayib-jose-gloria: nvm, see that you caught it later--wrong diff!

Issue #5000, opened 2023-06-12
CXG(s) for deleted datasets are not deleted


brianraymor:
### Context

See [**data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1686604455914939). 

This is another regression. 

1. If a collection is deleted by a submitter, then each CXG for the datasets in the collections must be deleted.
2. If a dataset is deleted by a submitter, then its CXG must be deleted.

danieljhegeman:
@brianraymor reading the Slack thread, Jenny has asked if we can redirect users to the homepage when a user visits a deleted cxg url `../e/<deleted_dataset_id>.cxg/`

Should that ask be included as a requirement for this ticket?

brianraymor:
> Edit: [redirect regression](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5004) describes this, so it will be handled as part of that ticket within this epic.

Correct. The original toast and redirect was regressed.

danieljhegeman:
@brianraymor I think I mentioned this in Slack already, but public-access Dataset versions were also not being deleted _in cases where the Dataset versions were never published._ I have included logic to this.

danieljhegeman:
Tested scenario where a subset of the expected assets are not actually in s3 through some sort of glitch; the deletes do not choke in this case 👍

Issue #4508, opened 2023-04-07
Clean up and prevent orphaned, unreachable entities


danieljhegeman:
[Some recent discussion](https://docs.google.com/document/d/19IZbojtc7eofV75NnL5C6fCgBJB13KIswWNOn12xAa0/edit?disco=AAAAuWWp4rw)

At present, the Data Portal does not delete entity rows in tables except for CollectionVersions when they are abandoned in the course of deleting a never-before-published Collection or a revision of a published Collection. When this happens, the `Collection` and any `Dataset` / `DatasetVersion` rows remain. The general motivation of this ticket is to manage database entities so that the database reflects only:

1) (previously) published entities, and
2) entities which are currently not _but may yet be_ published

This will be in keeping with an append-only approach to storing the data from the perspective of published versions of Collections and their constituent Datasets.

metakuni:
@danieljhegeman should we add this to #4693?

Bento007:
This should be performed after a collection is published. Here is a code snippet I was working on to find orphaned datasets associated with a canonical collection id. ```python 
    def get_orphaned_dataset_versions(self, collection_id: CollectionId) -> List[DatasetVersion]:
        """
        collection_id: CollectionId
        Returns a list of dataset versions that are not associated with a publishefd collection version.
        """
        with self._manage_session() as session:
            # Must be a published collection
            if (session.query(CollectionTable).filter(id=collection_id.id)
                    .filter(CollectionTable.originally_published_at != None).one_or_none()) is None:
                raise CollectionIsNotPublishedException(f"Collection {collection_id} not found")

            # Get all dataset versions associated with a collection
            all_dataset_version_ids = set( session.query(DatasetVersionTable.id)
                                             .filter(DatasetVersionTable.collection_id == collection_id.id)
            )

            # Get all dataset versions that have been published in a collection version
            published_dataset_version_ids = (
                session.query(CollectionVersionTable.datasets)
                .filter(CollectionVersionTable.datasets != None)
                .filter(CollectionVersionTable.published_at != None ).all()
            )
            published_dataset_version_ids = set(itertools.chain.from_iterable(published_dataset_version_ids))

            # find orphaned dataset versions
            orphaned_dataset_version_ids = all_dataset_version_ids - published_dataset_version_ids

            return [self.get_dataset_version(DatasetVersionId(dv_id[0])) for dv_id in orphaned_dataset_version_ids]
```

[...]

PR #5350, opened 2023-07-27
feat: delete cxgs for unpublished entities

danieljhegeman:
This commit introduces cleanup deletion for when unpublished Collections and Datasets are deleted, both for public-access Datasets and CXG assets. There is no change for tombstoning behavior, which was already in line with Product expectations. Includes moderate refactor of s3 deletion logic in Business provider.

## Reason for Change

- #5000
- #4508  
- Prevent users from being able to Explore unpublished Datasets that have since been deleted. Also prevent downloads of unpublished Dataset assets.

## Changes

- add logic to delete cxgs and public-access Datasets from their corresponding s3 buckets for previously-unpublished data. No changes needed for tombstoning data because 1) this was already happening for the tombstoning process for public-access Datasets, and 2) access to cxgs via Explorer link is gated by `Dataset.tombstone`.
- all deletion for unpublished Datasets occurs _on Collection publish_ rather than _on ingestion_ so as to avoid conflicts with the auto schema migration paradigm, which may require a rollback.

## Testing steps

- unit tests ✅ 
- tested in rdev ✅


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -353,13 +355,15 @@ data "aws_iam_policy_document" "lambda_step_function_execution_policy" {
...
+      "arn:aws:s3:::${local.datasets_bucket}",
+      "arn:aws:s3:::${local.datasets_bucket}/*"
     ]
   }
 }
 
 resource "aws_iam_policy" "lambda_step_function_execution_policy" {
   name = "lambda-step-function-execution-policy-${local.custom_stack_name}"
-  policy = data.aws_iam_policy_document.lambda_step_function_execution_policy.json
+  policy = data.aws_iam_policy_document.lambda_step_function_execution_policy_document.json
```
danieljhegeman, line 366: unrelated; improve clarity

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -219,6 +219,7 @@ module upload_success_lambda {
   lambda_execution_role      = local.lambda_execution_role
   subnets                    = local.subnets
   security_groups            = local.security_groups
+  datasets_bucket            = local.datasets_bucket
```
danieljhegeman, line 222: this was missing

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -256,6 +257,7 @@ module dataset_submissions_lambda {
   deployment_stage           = local.deployment_stage
   artifact_bucket            = local.artifact_bucket
   cellxgene_bucket           = local.cellxgene_bucket
+  datasets_bucket            = local.datasets_bucket
```
danieljhegeman, line 260: this was missing

.happy/terraform/modules/lambda/main.tf
```diff
@@ -10,10 +10,10 @@ resource aws_lambda_function lambda_job_def {
     variables = {
       ARTIFACT_BUCKET            = var.artifact_bucket,
       CELLXGENE_BUCKET           = var.cellxgene_bucket,
-      DATASETS_BUCKET            = var.datasets_bucket
+      DATASETS_BUCKET            = var.datasets_bucket,
```
danieljhegeman, line 13: missing as part of boilerplate

backend/layers/business/business.py
```diff
@@ -609,14 +613,30 @@ def create_collection_version(self, collection_id: CollectionId) -> CollectionVe
...
-    def delete_dataset_versions_from_bucket(self, dataset_version_ids: List[str], bucket: str) -> List[str]:
+        unpublished_versions_of_published_datasets, unpublished_datasets, versions_to_delete_from_s3 = [], [], []
+        for dv in collection_version.datasets:
+            unpublished_versions = self.get_unpublished_dataset_versions(dv.canonical_dataset.dataset_id)
+            versions_to_delete_from_s3.extend(unpublished_versions)  # All unpublished s3 assets should be deleted
+            if dv.canonical_dataset.published_at:
+                unpublished_versions_of_published_datasets.extend(unpublished_versions)
+            else:
+                unpublished_datasets.append(dv.canonical_dataset)
+        self.delete_dataset_version_assets(versions_to_delete_from_s3)
```
Bento007, line 629: Add some more spaces in this function for readability. It'll make it easier to know if a comment refers to the line above or bellow.

backend/layers/business/business.py
```diff
@@ -625,30 +645,51 @@ def delete_dataset_versions_from_bucket(self, dataset_version_ids: List[str], bu
...
-        dataset_version_ids_to_delete = []
-        for collection_version in collection_versions:
-            dataset_version_ids_to_delete.extend([dv.version_id.id for dv in collection_version.datasets])
-        return self.delete_dataset_versions_from_bucket(dataset_version_ids_to_delete, bucket)
+        dataset_versions = self.database_provider.get_all_dataset_versions_for_collection(collection_id)
+        return self.delete_dataset_versions_from_public_bucket([dv.version_id.id for dv in dataset_versions])
+
+    def get_unpublished_dataset_versions(self, dataset_id: DatasetId) -> List[DatasetVersion]:
+        """
+        Get all unpublished versions for a Dataset, as determined by the DatasetVersion argument, currently in the
```
Bento007, line 660: What `DatasetVersion` argument?

backend/layers/business/business.py
```diff
@@ -667,21 +708,35 @@ def publish_collection_version(self, version_id: CollectionVersionId) -> None:
             raise CollectionPublishException("Cannot publish a collection with datasets of different schema versions")
         schema_version = next(iter(schema_versions))
 
+        date_of_last_publish = datetime.min
```
Bento007, line 711: This could be the default value for `get_all_dataset_versions_for_collection(from_date=datetime.min)`

backend/layers/business/business.py
```diff
@@ -763,6 +818,23 @@ def get_prior_published_dataset_version(self, dataset_version_id: DatasetVersion
...
+                self.s3_provider.delete_files(bucket, keys)
+            if prefix:
+                self.s3_provider.delete_recursive(bucket, prefix)
+        except S3DeleteException as e:
+            raise CollectionDeleteException("Attempt to delete public Datasets failed") from e
+
+    def delete_artifacts(self, artifacts: List[DatasetArtifact]) -> None:
+        for artifact in artifacts:
+            logger.info(f"djh artifact is {artifact.uri} {artifact.type}")
+            matches_dict = re.match(r"^s3://(?P<bucket>[^/]+)/((?P<prefix>.*/$)|(?P<key>.*))", artifact.uri).groupdict()
```
Bento007, line 833: to avoid all of this regex logic you could factor out this into a comment function 
https://github.com/chanzuckerberg/single-cell-data-portal/blob/ef9c9baf50fe74737d7e9506e9885cbfc8b60131/backend/layers/processing/process_logic.py#L117C5-L124C30

backend/layers/business/business_interface.py
```diff
@@ -143,6 +160,9 @@ def get_latest_published_collection_versions_by_schema(
     ) -> List[CollectionVersionWithPublishedDatasets]:
         pass
 
+    def _delete_from_bucket(self, bucket: str, keys: List[str], prefix: str) -> None:
```
Bento007, line 163: you don't need to include private functions in the interface definition

backend/layers/business/exceptions.py
```diff
@@ -101,6 +101,14 @@ class DatasetUpdateException(BusinessException):
     pass
 
 
+class DatasetIsPublishedException(DatasetUpdateException):
+    pass
+
+
+class DatasetIsNotPublishedException(DatasetUpdateException):
```
Bento007, line 108: Is there a better way of say Not Published without using not. Maybe `DatasetIsPrivateException`. The Not could get lost when reading the name.

backend/layers/persistence/persistence.py
```diff
@@ -507,6 +507,17 @@ def add_collection_version(self, collection_id: CollectionId) -> CollectionVersi
             session.add(new_version)
             return CollectionVersionId(new_version_id)
 
+    def delete_collection(self, collection_id: CollectionId) -> None:
```
Bento007, line 510: change to ```suggestion
    def delete_unpublished_collection(self, collection_id: CollectionId) -> None:
```
Otherwise don't mention the unpublished part since it add business logic to the persistence layer. Not even sure if you should be checking for published here since that is also business layer logic.

**Review by Bento007 — COMMENTED** [review]
Still reviewing but here is what I have so far.

File: backend/layers/thirdparty/s3_provider_mock.py [hunk 12]
```diff
@@ -11,24 +11,37 @@ class MockS3Provider(S3ProviderInterface):
...
+    #     for key in self.mock_s3_fs:
+    #         if key.startswith(url_prefix):
+    #             del
+    #     if resp and isinstance(resp, list):
+    #         if deleted := resp[0].get("Deleted"):
+    #             logger.info(f"Deleted: {deleted}")
+    #         if errors := resp[0].get("Errors"):
+    #             logger.info(f"Errors: {errors}")
+    #             raise S3DeleteException(errors)
+
```
danieljhegeman, line 44: This will be rectified 😎

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -87,16 +88,20 @@ def delete(token_info: dict, collection_id: str, dataset_id: str, delete_publish
...
-            collection_version.version_id, dataset_version.version_id, delete_published_flag
-        )
-    except CollectionUpdateException:
-        raise MethodNotAllowedException(detail="Cannot delete a public Dataset") from None
+        try:
+            business_logic.remove_dataset_version(
+                collection_version.version_id, dataset_version.version_id, delete_published=True
+            )
+        except DatasetIsNotPublishedException:
+            raise InvalidParametersHTTPException(
```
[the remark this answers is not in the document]
danieljhegeman: My initial thinking was if someone tries to perform an action that indicates they have a misunderstanding of the state of the data, we want them to know that, rather than returning 2xx. They might've been trying to delete the wrong Dataset.

Sort of similar to how we force users to use the appropriate `/datasets/` vs `/dataset_versions/` route for retrieving Datasets even though we could have just used fall-through under a single route.

But I could be convinced to not fail here. Perhaps @brianraymor would like to weigh in.

backend/layers/business/business.py
danieljhegeman: That actually already is the default value. Slightly annoying I agree but the calling logic dictates that we pass the `date_of_last_publish` no matter what; in some cases it will be `None`, which overrides the default, so then we handle that case in the method itself.

backend/layers/business/business.py
danieljhegeman: You mean to avoid having the regex inlined here? As far as I can tell this specific regex isn't used/needed anywhere else, so factoring it out seems a bit premature to me. If it were several lines I think I would agree but otherwise I'm partial to keep as-is.

Removing the logger comments leftover from dev work... 😅 🤦‍♂️

backend/layers/business/business_interface.py
danieljhegeman: Don't need to... or shouldn't? 🤔 

Perhaps the latter so that the places where business logic is instantiated as BusinessInterface will not prompt the developer to call that private method... removing it now

backend/layers/business/exceptions.py
danieljhegeman: what about a `NotPublishedException`? Can be used for both types of entities.

backend/layers/persistence/persistence.py
danieljhegeman: Good points. I added the check for published/not published because I was following the precedent set by Emanuele's original code for `delete_collection_version` which is immediately below.

But yeah I will update the name.

backend/layers/business/business.py
danieljhegeman: Addressed with business layer unit test `test_delete_artifacts`. Let me know if that suffices.

backend/layers/business/exceptions.py
Bento007: I'm trying to avoid adding negative logic if possible.

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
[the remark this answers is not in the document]
Bento007: for deletion, i'm happy to play it more conservative.

backend/portal/api/portal_api.py
```diff
@@ -520,16 +519,10 @@
...
-        try:
-            get_business_logic().delete_collection_version(resource_id)
-        except CollectionIsPublishedException:
-            raise ForbiddenHTTPException() from None
-    elif isinstance(resource_id, CollectionId):
-        try:
-            get_business_logic().tombstone_collection(resource_id)
-        except CollectionDeleteException() as e:
-            raise ServerErrorHTTPException(e.errors) from e
+    try:
```
Bento007, line 522: These changes and everything related could have been a separate PR to make reviewing easier.

backend/layers/thirdparty/s3_provider_interface.py
```diff
@@ -14,6 +14,9 @@ def upload_file(self, src_file: str, bucket_name: str, dst_file: str, extra_args
     def delete_files(self, bucket_name: str, object_keys: List[str]):
         pass
 
+    def delete_recursive(self, bucket_name: str, prefix: str) -> None:
```
Bento007, line 17: document what this does.

backend/layers/thirdparty/s3_provider.py
```diff
@@ -80,6 +81,15 @@ def delete_files(self, bucket_name: str, object_keys: List[str]) -> None:
                 logger.info(f"Errors: {errors}")
                 raise S3DeleteException(errors)
 
+    def delete_recursive(self, bucket_name: str, prefix: str) -> None:
```
Bento007, line 84: You don't need a new function. Use`list_directory` and `delete_files` to accomplish this, or use those two function in `delete_recursive`

Bento007 (changes requested):
reuse code for deleting in s3_provider. Also enforce that prefix cannot be an empty string when listing out objects.

backend/layers/persistence/persistence_interface.py
```diff
@@ -105,15 +108,38 @@ def add_collection_version(self, collection_id: CollectionId) -> CollectionVersi
...
+        """
+        Delete an unpublished DatasetTable row (and its dependent DatasetVersionTable and DatasetArtifactTable rows)
+        """
+
+    def delete_dataset_versions(self, dataset_versions: List[Union[DatasetVersionId, DatasetVersion]]) -> None:
+        """
+        Deletes DatasetVersionTable rows.
+        """
+
+    def _delete_dataset_version_and_artifact_rows(
```
Bento007, line 131: you don't need to put private functions here

backend/layers/persistence/persistence.py
```diff
@@ -605,6 +654,22 @@ def get_dataset_version(self, dataset_version_id: DatasetVersionId, get_tombston
                     return None
             return self._hydrate_dataset_version(dataset_version)
 
+    def get_all_dataset_versions_for_collection(
+        self, collection_id: CollectionId, from_date: datetime = datetime.min
```
Bento007, line 658: is `datetime.min` immutable? If it's not then change it to:

 `from_date: Optional[datetime] = None`

and add this

`from_date = from_date if from_date else datetime.min`

This is to avoid weird bugs related to modifying a mutable attribute.

backend/layers/persistence/persistence.py
```diff
@@ -807,12 +872,16 @@ def delete_dataset_from_collection_version(
         self, collection_version_id: CollectionVersionId, dataset_version_id: DatasetVersionId
     ) -> None:
         """
-        Removes a mapping between a collection version and a dataset version
+        Removes a mapping between a collection version and a dataset version.
+
+        :param collection_version_id: the CollectionVersion or the CollectionVersionId
+        :param dataset_version_id: the DatasetVersionId
+        :param delete_dv_row: boolean flag - when True, delete DatasetVersion row (and dependent DatasetArtifact rows)
```
Bento007, line 879: not used

backend/layers/persistence/persistence.py
```diff
@@ -507,6 +507,17 @@ def add_collection_version(self, collection_id: CollectionId) -> CollectionVersi
             session.add(new_version)
             return CollectionVersionId(new_version_id)
 
+    def delete_collection(self, collection_id: CollectionId) -> None:
+        """
+        Delete an unpublished Collection
```
Bento007, line 512: what happens to all of the rows and table associated with the collection?

Bento007: Since dataset_version depends on a dataset, which depend on a collection_version, which depend on a collection, deleting a collection should be an atomic database operation. If there isn't a ticket filed for this already, please create and add one for better database session management.

Bento007: Another option it to add a session parameter to these database operations and use the existing session if it's passed in. This might be a heavy lift, I'll try an prototype it.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -905,6 +917,139 @@ def test_delete_published_dataset_in_revision__success(self):
         revision = self.business_logic.get_collection_version(revision.version_id)
         self.assertNotIn(dataset_version_id_to_delete.id, [dv.version_id.id for dv in revision.datasets])
 
+    def test_deletion_of_dataset_assets_and_rows(self):
+        with self.subTest("Updated Dataset in private Collection not deleted until Collection is deleted"):
```
Bento007, line 921: where possible avoid using subtest, pytest doesn't work will with it, which is the test engine we are using.

Bento007: Also given how large these tests are, breaking them up would be helpful

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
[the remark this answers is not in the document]
danieljhegeman: great 🤝

backend/layers/persistence/persistence.py
danieljhegeman: datetime.min is a constant attribute: the earliest possible `datetime` value. I don't think this changes the extent to which this is bug-prone but I'll change the logic because I agree, better to avoid `not` statements where possible.

danieljhegeman: done ✔️

backend/layers/persistence/persistence.py
danieljhegeman: We should leave these sorts of reorg questions for the Redesign 1.1 work. This is what we've been discussing wrt fixing the persistence/business layer interactions.

You can make a ticket if you like and add it to Redesign 1.1.

backend/layers/thirdparty/s3_provider.py
danieljhegeman: The prefix delete is going to be way faster; if you have to iterate over the dozens of individual s3 keys associated with a single cxg file, it'll increase network time costs a lot.

backend/layers/business/exceptions.py
metakuni: Besides choosing an easily understandable name, we should also be adding a docstring for each exception to make it clear when they should be raised. 🙏

tests/unit/backend/layers/business/test_business.py
danieljhegeman: done

backend/portal/api/portal_api.py
danieljhegeman: Very small change, though you're right in the sense that I should have labeled it as tangetially related

backend/layers/thirdparty/s3_provider_mock.py
danieljhegeman: ✅

backend/layers/persistence/persistence.py
Bento007: Add a todo to this function to make that change in the future. It's important we don't forget to make this change.

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/actions.py
```diff
@@ -87,16 +88,20 @@ def delete(token_info: dict, collection_id: str, dataset_id: str, delete_publish
     if dataset_version.version_id not in [v.version_id for v in collection_version.datasets]:
         raise ForbiddenHTTPException(f"Dataset {dataset_id} does not belong to a collection")
 
-    delete_published_flag = False
     if user_info.is_cxg_admin() and delete_published == "true" and not collection_version.published_at:
```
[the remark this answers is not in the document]
danieljhegeman: Right you are. I'd been misled by some stackoverflow posts. Changed!

backend/layers/business/business.py
```diff
@@ -822,13 +824,14 @@ def _delete_from_bucket(self, bucket: str, keys: List[str] = None, prefix: str =
             if keys:
                 self.s3_provider.delete_files(bucket, keys)
             if prefix:
-                self.s3_provider.delete_recursive(bucket, prefix)
+                object_keys = list(self.s3_provider.list_directory(bucket, prefix))
```
Bento007, line 827: nit: I would suggest pushing these to functions into another function in s3_provider. Maybe call it `delete_prefixed_files(bucket, prefix)`. It seems like an operation that could be useful to other areas of the code.

Bento007 (approved):
LGTM, one suggestion

backend/layers/business/business.py
danieljhegeman: incorporated 👍

Issue #4812, opened 2023-05-09
Create script to clean-up erroneous datasets from s3


nayib-jose-gloria:
Admin Script to run locally or AWS Lambda/Batch if not securely/safely able to implement such that we can run locally

Bento007:
- How do we know they are erroneous?
- The script shall delete all migrate.h5ad files
- Mark the collections as migrating 

## special cases

### migrating existing revision with replaced datasets.
The migration would have overwritten the updated datasets. To get to the previous state, the old datasets would need to be reuploaded manually

### migrating a private collection
The migration would have replaced the uploaded private datasets. They would need to be reuploaded to revert

Bento007:
The above cases are not a problem because the previous version is never replace. A new dataset version is created and the files are uploaded to the new dataset version. This does create a problem of orphaned datasets living in the data corpus, but that will be handled as part of [this](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/4508) work

Issue #4811, opened 2023-05-09
Create ‘Rollback' Job


nayib-jose-gloria:
- Run manually, only can be triggered by powerusers
- Points canonical Datasets / Collections in database to previous Dataset Version or Collection Version
- Delete database rows related to erroneous dataset versions that may be created by accident
- Necessary to block them from appearing in versions endpoints
- Can accept no input (will rollback the latest schema minor version for every collection where applicable), or list of collection IDs to only rollback specific collections

A lambda should be doable, but may need to do batch job if long-running. Or cxg admin script in repo

danieljhegeman:
@Bento007 @nayib-jose-gloria does the rollback feature for schema migration require the ability to rollback to a previous _unpublished_ Dataset version? Or only to the previous _published_ version?

Issue #4828, opened 2023-05-10
Create an 'untombstone' utility


metakuni:
## Motivation
Mistakes happen. If a collection is accidentally deleted, it should be easy for _any_ on-call engineer to be able to resurrect it. 

Thanks to the magic of immutable data, this is all we have to do: ([from @ebezzi in Slack](https://czi-sci.slack.com/archives/C024HCSH9PT/p1683758051542559?thread_ts=1683757407.222789&cid=C024HCSH9PT))
```
update persistence_schema."Collection" set tombstone=false where id='<collection ID>';
```

See also @Bento007's comment below re: recovery of S3 assets.

For quicker recovery after accidental deletions, it would be handy to have a script that can resurrect deleted data.

## Definition of Done
- [ ] Implement a script to make it easier for an on-call engineer to undo a deletion
- [ ] Add tests for the above
- [ ] Add instructions for use in [on-call docs](https://docs.google.com/document/d/1n5cngEIz-Lqk9737zz3makXGTMrEKT5kN4lsofXPRso/edit#heading=h.88nz21uu8s19)

## Tasks

Detail the specific tasks that can be used to accomplish the desired changes.
If detailed steps cannot be provided at this time, please file a [Tech Proposal](https://docs.google.com/document/d/1o2vuvl-kXwRJN1nBoPzJS_MAQgDGYnjmPZWa4qRDi-I/edit#heading=h.7dvzhm7gqc3v) instead.

- [ ]
- [ ]

danieljhegeman:
It is not sufficient to simply make the necessary updates in the entity rows in postgres because the tombstoning process also removes the asset files from the public-access bucket `dataset-assets-public-<env>`; files will need to be restored to that bucket, by the correct naming convention, from the private assets bucket `corpora-data-<env>`.

For a given Dataset Version `abc.h5ad` that lives at `corpora-data-prod/abc/local.h5ad` in the private bucket, this will need to be copied to `dataset-assets-public-prod/abc.h5ad` because that file will have been deleted during the tombstoning action.

Bento007:
Here is a script to restore a delete file from a versioned bucket:
```python
import boto3

s3 = boto3.client("s3")
dataset_ids = []  # TODO add datasets that need to be restored here

for dataset_id in dataset_ids:

    response = s3.list_object_versions(
        Bucket="dataset-assets-public-prod",
        Prefix=dataset_id)
[...]

PR #5355, opened 2023-07-28
feat(schema-migration): Add ability to restore a dataset to the previous version

Bento007:
## Reason for Change

- #4812
- #4811
- When things go wrong during schema migration we need a way to bring things back to a know state. Part of this work requires restoring a collection and its datasets to a state prior to the migration. 

## Changes
- add cxg_admin script to roll back all datasets from a report generated by the schema_migration pipeline
- modify the shape of the schema-migration report to make it easier to parse in the cxg_admin command.
- add `restore_previous_dataset_version` to modify a dataset_version associated with a collection_version to the previous dataset_version if it exists.
- add additional exceptions related to `set_previous_dataset_version`
- modify `replace_dataset_in_collection_version` to optionally include a `new_dataset_version_id` to replace `old_dataset_version_id`.
- add `get_previous_dataset_version_id` to the persistence layer to query the database for the previously created `dataset_version` associated with a `dataset_id`.
- Set the name of the upload sfn name to "migrate_{dataset_version_id}_{timestamp}"

## Testing steps

- Added unit test


danieljhegeman:
Nice. I will likely make use of some of the functionality here for #4828

nayib-jose-gloria (commented):
I think we'll want an equivalent 'restore' function for published Collections as well. Since, in theory, we might want to rollback a published mistake in the migration run. But that can be in a separate PR to avoid blocking some of the other changes in here

PR #5357, opened 2023-07-31
fix: WMGv2 React bugs

tihuan:
## Reason for Change

1.  Fix WMGv2 bugs: React cannot render Parent component while a child component is being updated. Two places as noted inline comments
2. Refactor heatmap to simplify code

## Testing steps

1. Should work as well as dev/staging


frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -181,7 +182,8 @@ function selectOrganism(
     ...state,
     selectedGenes: [],
     selectedOrganismId: action.payload,
-    selectedTissues: [],
+    selectedTissues: EMPTY_ARRAY,
+    selectedFilters: EMPTY_FILTERS,
```
tihuan, line 186: we need to reset filters when selecting a new organism too. This way we don't need to manually dispatch an event to reset the state in `Heatmap`

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -181,7 +182,8 @@ function selectOrganism(
     ...state,
     selectedGenes: [],
     selectedOrganismId: action.payload,
-    selectedTissues: [],
+    selectedTissues: EMPTY_ARRAY,
```
tihuan, line 185: use `EMPTY_ARRAY` instead of a new array `[]`

frontend/src/views/WheresMyGene/components/HeatMap/components/XAxisChart/index.tsx
```diff
@@ -74,10 +74,12 @@ function GeneButton({
     currentFont
   );
 
-  if (dispatch && length && length > currentMaxLabelHeight) {
```
tihuan, line 77: side effect like state change and dispatching events need to be inside a `useEffect()`, otherwise it will cause React error "cannot update component while a child component is being updated`

frontend/src/views/WheresMyGeneV2/components/HeatMap/components/YAxisChart/index.tsx
```diff
@@ -42,7 +42,6 @@ interface Props {
   generateMarkerGenes: (cellType: CellType, tissueID: string) => void;
   handleExpandCollapse: (tissueID: string, tissueName: Tissue) => void;
   expandedTissues: Set<Tissue>;
-  selectedOrganismId: string;
```
tihuan, line 45: not used

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -36,7 +36,7 @@ import {
 } from "src/views/WheresMyGene/common/types";
 import YAxisChart from "./components/YAxisChart";
 import { useTrackHeatMapLoaded } from "./hooks/useTrackHeatMapLoaded";
-import { memoize } from "lodash";
```
tihuan, line 39: `useMemo` works just fine I believe

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -82,14 +82,14 @@ interface Props {
...
   geneSortBy: SORT_BY;
-  selectedOrganismId: string;
   echartsRendererMode: "svg" | "canvas";
   setAllChartProps: Dispatch<
     SetStateAction<{
       [tissue: string]: ChartProps;
     }>
   >;
   allChartProps: { [tissue: string]: ChartProps };
+  tissuesByName: { [name: string]: OntologyTerm };
```
tihuan, line 92: use `tissuesByName` from parent, so we have single source of truth

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -110,19 +110,20 @@ export default memo(function HeatMap({
...
   setAllChartProps,
+  tissuesByName,
   setTissuesByName,
   expandedTissues,
   setExpandedTissues,
 }: Props): JSX.Element {
   const {
     xAxisHeight,
-    selectedFilters: { tissues: filteredTissues },
+    selectedFilters: { tissues: filteredTissueIds },
```
tihuan, line 123: make variable name explicit by suffixing with `Ids`. We actually have a share link breaking change that previously the share link tissue expects tissue name, but now it seems to expect tissue ids. Not sure if this is an intended change or a bug. Signe is out OOO, so we'll need to check when she comes back!

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -131,27 +132,27 @@ export default memo(function HeatMap({
   const { data } = usePrimaryFilterDimensions(2); //temp explicit version
 
   // Get tissueName to ID map for use in find marker genes
-  const tissuesByName = useMemo(() => {
```
tihuan, line 134: no longer needed, since we use the parent's `tissuesByName`

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -131,27 +132,27 @@ export default memo(function HeatMap({
...
-    let result: { [name: string]: OntologyTerm } = {};
-
-    if (!data) return result;
+  useEffect(() => {
```
tihuan, line 135: change from `useMemo` to `useEffect` to update the parent's `tissuesByName`

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -208,30 +209,27 @@ export default memo(function HeatMap({
...
-          filteredTissues.length === 0
-        ) {
-          acc.add(id + id);
-        }
-        return acc;
-      }, new Set<string>()),
-    [tissuesByName, filteredTissues]
-  );
+  const initialDisplayedCellTypeIds = useMemo(() => {
+    return Object.values(tissuesByName).reduce((acc, { id }) => {
```
tihuan, line 213: use `Object.values` instead of `Object.entries`, since we don't use the keys anyway

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -208,30 +209,27 @@ export default memo(function HeatMap({
...
   // set of tissue names that are visible and set of cell types that are visible
   // presence is visible
   const [displayedCellTypes, setDisplayedCellTypes] = useState<Set<string>>(
-    initialDisplayedCellTypes
+    initialDisplayedCellTypeIds
   );
 
   useEffect(() => {
-    setDisplayedCellTypes(initialDisplayedCellTypes);
-    setExpandedTissues(new Set<string>());
```
tihuan, line 233: no longer needs this, since we add `key={selectedOrganismId}` to the Heatmap, so it will automatically clear all states when the `key` changes

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -297,19 +296,21 @@ export default memo(function HeatMap({
   // update displayedCellTypes and expandedTissues
   useEffect(() => {
     if (filteredCellTypes.length === 0) {
-      setDisplayedCellTypes(initialDisplayedCellTypes);
```
tihuan, line 300: no longer needs this, since we add `key={selectedOrganismId}` to the Heatmap, so it will automatically clear all states when the `key` changes

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -356,6 +356,39 @@ export default memo(function HeatMap({
     selectedCellTypes: filteredCellTypes,
   });
 
+  /**
+   * All tissue cell types to render in YAxisCharts
+   */
+  const allTissueCellTypes = useMemo(() => {
```
tihuan, line 362: Keep the code DRY and extract calculation to its useMemo, so we don't need to use `memoize`

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -404,105 +437,76 @@ export default memo(function HeatMap({
...
                       handleExpandCollapse={handleExpandCollapse}
                     />
                   )
                 );
-              })}
+              }
+            )}
           </YAxisWrapper>
           <ChartWrapper ref={chartWrapperRef} top={xAxisHeight}>
-            {Object.values(tissuesByName)
```
tihuan, line 439: keep the code DRY and reuse `allTissueCellTypes`

frontend/src/views/WheresMyGeneV2/components/Main/index.tsx
```diff
@@ -399,6 +399,10 @@ export default function WheresMyGene(): JSX.Element {
           <ScreenTint isDownloading={downloadStatus} />
 
           <HeatMap
+            /**
+             * (thuang): Use the selected organism ID to reset the heatmap state
+             */
+            key={selectedOrganismId}
```
tihuan, line 405: add `key` to reset Heatmap use states

atarashansky (approved):
LGTM! Thanks for catching these bugs!

seve:
Introduced bug descibed in #5491

PR #5360, opened 2023-07-31
fix(schema-migration): double the timeout for seurat conversion step

Bento007:
## Reason for Change

- During a schema migration in dev seurat conversion jobs were timing out. This was likely cause by a number of jobs being ahead of the seurat job in the batch job queue. The lead to the job timing out before being processed.

## Changes

- modify the time for the seurat conversion job from 36000 -> 72000 seconds

### DIFF
--- a/.happy/terraform/modules/sfn/main.tf
+++ b/.happy/terraform/modules/sfn/main.tf
@@ -160,7 +160,7 @@ resource "aws_sfn_state_machine" "state_machine" {
                     ]
                   }
                 },
-                "TimeoutSeconds": 36000
+                "TimeoutSeconds": 72000
               }
             }
           }


.happy/terraform/modules/sfn/main.tf
```diff
@@ -160,7 +160,7 @@ resource "aws_sfn_state_machine" "state_machine" {
                     ]
                   }
                 },
-                "TimeoutSeconds": 36000
+                "TimeoutSeconds": 72000
```
nayib-jose-gloria, line 163: Why not increase to 360000 to match cxg conversion? The job is a multi-day job so I worry even 20 hours could stall in the queue. 

Also, we should up DownloadValidate as well, there was 1 instance of it timing out on 36000 as well

Bento007: I don't think we want a single upload job to take 10 days. We should instead allocate more machines to get the job done sooner.

Bento007: I will increase the timeout for download validate to match.

Bento007: We could also reduce the cxg processing time to 5 days.

nayib-jose-gloria: The cxg timeout is already about 4 days (360000 seconds == 100 hours == a little over 4 days) 

An upload job could still take multiple days because of the cxg processing timeout, so I don't think it affects much to have other jobs match. Though I agree we should look to increase resources; I'm just not sure what the 'right' amount would be to offset a 20 hour timeout, and I'd rather avoid timeouts in prod at all

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -389,7 +379,7 @@ resource aws_sfn_state_machine sfn_schema_migration {
             },
             "ItemsPath": "$.datasets",
             "Next": "CollectionPublish",
-            "MaxConcurrency": 40,
+            "MaxConcurrency": 32,
```
nayib-jose-gloria, line 382: I see there's a MaxConcurrency [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5360/files#diff-f4d94af2de6b76dcfe8833d56b096305952f2409ac0c0f1a2b6b0a7a20a67fe6R401) as well set to 40, I assume we should decrease this as well?

Bento007: That is for the collectionSpan which the machines will be able to support 40+ jobs.

nayib-jose-gloria (approved):

PR #5367, opened 2023-08-01
feat(schema-migration): improvements to schema migrations error reporting

Bento007:
## Reason for Change

- resolves https://github.com/chanzuckerberg/single-cell/issues/518
- Filtering out errors from the schema migration that are not helpful.

## Changes

- add `self.local_path` to `SchemaMigrate` to help with testing
- Send the response of `SchemaMigrate.collectionMigrate` to `SchemaMigrate.publish_and_cleanup`. This is use to filter out datasets that were not processed.
- `publish_and_cleanup` filters out dataset that were not processed
- The whole processing status is reported if there is an error with the `datasets.status._processing_status`
- Adding helper function `SchemaMigrate._retrieve_sfn_response` to simplify retrieving extra data used for processing a step.
- Add ability to limit a migration to a number of collections.

## Testing steps
- Add unit test for new functionality.


backend/layers/processing/schema_migration.py
```diff
@@ -188,11 +199,14 @@ def publish_and_cleanup(self, collection_version_id: str, can_publish: bool) ->
                         "rollback": False,
                     }
                 )
-            elif dataset.status.processing_status != DatasetProcessingStatus.SUCCESS:
+            elif dataset.status.processing_status not in [
+                DatasetProcessingStatus.SUCCESS,
+                DatasetProcessingStatus.INITIALIZED,
+            ]:
```
nayib-jose-gloria, line 205: I think we can keep the old check now that we only look at datasets the migration processed; since we filter for datasets that were already successfully processed in collection_migrate, INITIALIZED would be a bad sign

nayib-jose-gloria (approved):

PR #5375, opened 2023-08-02
feat: add sqsum array to expression summary default cube

joyceyan:
## Reason for Change

first component of this ticket: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5093

this PR adds the `sqsum` array to the expression summary default cube. up next:
- actually read from it in `calculate_markers`
- deprecate `expression_summary_fmg`

this PR also includes some aesthetic code changes, such as:
1. ensuring we have the return type annotated on every function i touched
2. adding the `*` decorator to all functions i touched, which requires callers to name the params. this helps ensure that you don't end up touching the wrong parameter just because you ordered something wrong
3. created Result class objects rather than returning tuples. i think having a typed result object is a lot easier to read when you're returning more than two things. similarly, this helps prevent the caller of the function from setting one of the result values to be the wrong thing if you just ordered your variables wrong

## Testing steps

in order to make sure this change works as expected, we need to regenerate our test fixtures. this also means we need to create a new snapshot with the `sqsum` array. i plan to accomplish this by:
1. creating an rdev stack against a commit on this PR
```
happy create $STACK_NAME --tag sha-$COMMIT_TAG --create-tag=false --skip-check-tag
```
2. running the WMG pipeline against that rdev stack
```
aws batch submit-job --job-name wmg-rdev --job-queue wmg-rdev --job-definition wmg-rdev-$STACK_NAME-wmg-processing --profile single-cell-dev
```
3. wait for pipeline to complete... ⏳ about 7 hours
4. download snapshot locally into `../snapshot` (relative to directory for `single-cell-data-portal`). this can be found somewhere in S3
5. generate realistic test fixtures from snapshot
```
python3 scripts/generate_realistic_test_snapshot.py ../snapshot tests/unit/backend/wmg/fixtures/realistic-test-snapshot
```
6. profit
image


tests/unit/wmg_processing/test_data_artifact_correctness.py
```diff
@@ -32,29 +27,6 @@ def find_dim_option_values_pandas(criteria: Dict, snapshot: WmgSnapshot, dimensi
 
 
 class DataArtifactCorrectness(unittest.TestCase):
-    # creates the expression summary default cube by group-by and summing the expression summary cube
-    # and checks to makes sure that the result is the same as the existing fixture (expression_summary_default)
-    def test__expression_summary_default_cube_correctness(self):
```
joyceyan, line 37: this test is no longer relevant. it asserts that taking the expression summary cube and group-by/summing only the default logical dimensions should result in the same thing as the expression summary default cube. however, since we're adding the sqsum array to only the default cube and not the expression summary cube, this is no longer expected to be true.

tests/unit/backend/wmg/test_query.py
```diff
@@ -96,8 +96,8 @@ def test__query_expression_summary_default_cube__returns_correct_results(self):
         with load_realistic_test_snapshot(TEST_SNAPSHOT) as snapshot:
             q = WmgQuery(snapshot)
             query_result = q.expression_summary_default(criteria)
-            query_sum = list(query_result[["sum", "nnz"]].sum())
-            expected = [804599.0, 370972.0]
+            query_sum = list(query_result[["sum", "nnz", "sqsum"]].sum())
+            expected = [804599.0, 370972.0, 1822108.0]
```
joyceyan, line 100: this is the only test that i altered. there are existing tests that verify that the schema of the test snapshot matches the declared schema of the default cube. not sure what other tests would be valuable to add, since we now have coverage to:
- ensure the default cube has the `sqsum` array
- assertion on expected value of the sum of the `sqsum` array

atarashansky (approved):
LGTM!

PR #5376, opened 2023-08-02
chore(wmg): add example dev notebook for running pipeline locally

atarashansky:
## Reason for Change

- Currently, the WMG pipeline dev cycle is way too long using rdev. Every change requires another deployment, and the pipeline currently cannot run against the small rdev database due to Auth0 walling off the curation endpoints. There are a number of solutions to enable that, but the easiest bypass is to be able to run the pipeline locally to test pipeline changes.

## Changes

- added an example notebook that goes through how to download some datasets locally and run the pipeline against them


joyceyan (approved):
can you add a comment about how the AWS S3 upload failure is expected and how to load the snapshot locally?

prathapsridharan (approved):
lgtm

PR #5377, opened 2023-08-02
fix: Make WMG frontend call v2 version of the /markers endpoint

prathapsridharan:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell-data-portal/pull/5377
- Make the WMG frontend call `/v2/markers` instead of `/v1/markers`

## Changes

- modify call sites on the frontend to call hit `/v2/markers`

## Testing steps

- Manually tested by starting the frontend server locally and checking that adding marker genes calls `/v2/markers` and returns `http 200 OK`


dsadgat (approved):
lgtm

tihuan (approved):

Issue #5378, opened 2023-08-02
Fix failure to generate WMG snapshot for large datasets that throw OOM exception


prathapsridharan:
Here is the stacktrace from the log stream in `staging`. The failure happens in production too and this failurewas replicated in the `staging`:

```
INFO:backend.wmg.pipeline.integrated_corpus.job:Processing dataset 1 of 294
INFO:backend.wmg.pipeline.integrated_corpus.job:h5ad_file_path='datasets/6f7fd0f1-a2ed-4ff1-80d3-33dde731cbc3/local.h5ad'
INFO:backend.wmg.pipeline.integrated_corpus.extract:Extracting datasets/6f7fd0f1-a2ed-4ff1-80d3-33dde731cbc3/local.h5ad...
INFO:backend.wmg.pipeline.integrated_corpus.transform:Applying filters: assay, and lowly-covered cells
ERROR:__main__:Pipeline failed
Traceback (most recent call last):
  File "/backend/wmg/pipeline/cube_pipeline.py", line 88, in main
    corpus_path, stats = load_data_and_create_cube("datasets")
  File "/backend/wmg/pipeline/cube_pipeline.py", line 51, in load_data_and_create_cube
    integrated_corpus.run(path_to_h5ad_datasets, corpus_path, extract_data)
  File "/backend/wmg/pipeline/integrated_corpus/__init__.py", line 15, in run
    build_integrated_corpus(dataset_directory, corpus_path)
  File "/backend/wmg/data/utils.py", line 19, in wrap_func
    result = func(*args, **kwargs)
  File "/backend/wmg/pipeline/integrated_corpus/job.py", line 55, in build_integrated_corpus
    dataset_id, gene_ids = process_h5ad_for_corpus(h5ad_file_path, corpus_path)
  File "/backend/wmg/pipeline/integrated_corpus/job.py", line 98, in process_h5ad_for_corpus
    apply_pre_concatenation_filters(anndata_object)
  File "/backend/wmg/pipeline/integrated_corpus/transform.py", line 31, in apply_pre_concatenation_filters
    scanpy.pp.filter_cells(anndata_object, min_genes=min_genes)
  File "/usr/local/lib/python3.10/site-packages/scanpy/preprocessing/_simple.py", line 133, in filter_cells
    filter_cells(adata.X, min_counts, min_genes, max_counts, max_genes)
[...]

PR #5383, opened 2023-08-03
fix: Fix the OOM error due to processing large datasets in wmg pipeline

prathapsridharan:
## Reason for Change

- #5378 

## Changes
- Replace the call to `scanpy.pp.filter_cells()` with a memory efficient implementation to filter cells

## Testing steps

- Manually tested on dataset_id `b6adf6d3-9f0b-402f-96b8-42d6114bc34b.h5ad` on `staging`

The new implementation has been tested in a jupyter notebook and confirmed to be able to load the above large anndata object in memory and filter the underlying matrix by removing cells that have gene counts below the threshold without throwing OOM error


prathapsridharan:
> Not sure I like the patching approach - I would prefer to reimplement `scanpy.pp.filter_cells`. Any reduction in surface area of our dependency on a bulky package like scanpy is a win.
> 
> If we do go the patching route, we need to undo the patch at the end of the function call by storing the original function definition in a variable and re-assigning it.

@atarashansky - I will add a reimplementation of `scanpy.pp.filter_cells` in the next commit

atolopko-czi (changes requested):
One problem with the patching approach is that we will not get future scanpy improvements to these methods. At a minimum, we need a plan to return to a future optimized scanpy implementation. This means at least filing an issue with anndata and/or scanpy, and ideally providing PR(s).  Also, have we reach out to the scipy/anndata community to see if there might be a more pure way to fix this?  E.g. I do see the `inplace` option, which potentially means we could take the computed set of filtered cells and do the necessary AnnData.X surgery ourselves; not sure if this is viable, just a thought...

prathapsridharan:
> One problem with the patching approach is that we will not get future scanpy improvements to these methods. At a minimum, we need a plan to return to a future optimized scanpy implementation. This means at least filing an issue with anndata and/or scanpy, and ideally providing PR(s). Also, have we reach out to the scipy/anndata community to see if there might be a more pure way to fix this? E.g. I do see the `inplace` option, which potentially means we could take the computed set of filtered cells and do the necessary AnnData.X surgery ourselves; not sure if this is viable, just a thought...

@atolopko-czi @atarashansky  - `inplace=True` by default and we are using the default option. There are multiple problems here that I have captured after doing some profiling with `top`:

https://github.com/chanzuckerberg/single-cell-data-portal/issues/5378#issuecomment-1664631933

atolopko-czi:
> @atolopko-czi @atarashansky - `inplace=True` by default and we are using the default option. There are multiple problems here that I have captured after doing some profiling with `top`:

I was thinking that we could actually use `inplace=False`, which appears to only return the matrix cells that pass the zero filtering operation, and not doing anything else to modify X. With that, we could rebuild the AnnData's X "ourselves", possibly in a memory efficient way. This assumes the returned data is a filtered view into the underlying X, otherwise we're clearly going to hit the memory problem.  Again, not sure if this is viable, but it was the first idea that came to mind.

prathapsridharan:
> > @atolopko-czi @atarashansky - `inplace=True` by default and we are using the default option. There are multiple problems here that I have captured after doing some profiling with `top`:
> 
> I was thinking that we could actually use `inplace=False`, which appears to only return the matrix cells that pass the zero filtering operation, and not doing anything else to modify X. With that, we could rebuild the AnnData's X "ourselves", possibly in a memory efficient way. This assumes the returned data is a filtered view into the underlying X, otherwise we're clearly going to hit the memory problem. Again, not sure if this is viable, but it was the first idea that came to mind.
> 
> Since the ultimate operation we want is `scipy.sparse.csr_matrix.eliminate_zeros()`, can we just call that ourselves directly on X?

@atolopko-czi - Perhaps I am misunderstanding. Before even hitting the condition `if not inplace:`, we would need to finish [this computation](https://github.com/scverse/scanpy/blob/89804c26bcbdde4449dfe98b3193a9d28c2a0e2f/scanpy/preprocessing/_simple.py#L132) (which is where it is failing due to OOM):

```
cell_subset, number = filter_cells(adata.X, min_counts, min_genes, max_counts, max_genes)
```

prathapsridharan (commented):
@atolopko-czi @atarashansky - In any case, any memory efficient modification can be done in the duplicated implementation of `filter_cells` in `utils.py` that I have submitted. Once we have determined what part needs to be made more efficient, we can look to separate out that part into a separate stage (call it stage `B`) and perform the prior stage (call it stage `A`) by calling `scanpy` and using the arguments to produce the output of stage `A` that will then go into our custom stage `B`

atolopko-czi:
> @atolopko-czi - Perhaps I am misunderstanding. Before even hitting the condition `if not inplace:`, we would need to finish [this computation](https://github.com/scverse/scanpy/blob/89804c26bcbdde4449dfe98b3193a9d28c2a0e2f/scanpy/preprocessing/_simple.py#L132) (which is where it is failing due to OOM):

No, I misinterpreted the code path earlier. But I'm still wondering if the obs filtering could be performed without copying so much of the scanpy code.  Goal is it to make any added, optimized code easier to read and maintain, and it seems like our min genes use case could be represented more succinctly. It would make the "surface area" of our optimization smaller.

We would keep the 2 lines of code from the optimized `_filter_compressed_matrix_gt_zero()` method and then use the result to subset the adata on the obs axis. Ultimately, the `AnnData._init_as_actual()` method is being called in the original implementation, by way of `_inplace_subset_obs()`, and it's doing the heavy lifting of the obs subsetting. But maybe we can achieve a similar outcome with AnnData.obs filtering via Pandas operations? 

Would still include a comment indicating that `scanpy.pp.filter_cells()` was explicitly avoided to address excessive memory usage. And I would also add a unit test for the custom filtering logic, since we "own" the code now. Thoughts?

backend/wmg/pipeline/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    # memory efficient implementation for getting count of non-zeros per row.
+    # For CSR we use the `indptr` structure to compute number of non-zeros per row
+    # For CSC we use the `indices` structure to compute number of non-zeros per row
+    # And finally for ndarray, we simply filter the matrix and sum the resulting boolean matrix.
+    if isinstance(adata.X, csr_matrix):
+        number_per_cell = np.diff(adata.X.indptr)
+    elif isinstance(adata.X, csc_matrix):
+        _, number_per_cell = np.unique(adata.X.indices, return_counts=True)
```
atarashansky, line 62: This is unlikely, but there could be cases where a cell has no reads associated with it. `number_per_cell` would be the wrong length in this case.

Can we update this portion to:

, number_per_cell = np.unique(adata.X.indices, return_counts=True)
```suggestion
cell_indices, counts = np.unique(adata.X.indices, return_counts=True)
number_per_cell = np.zeros(adata.shape[0],dtype='int')
number_per_cell[cell_indices] = counts
```

atarashansky (changes requested):

backend/wmg/pipeline/utils.py
prathapsridharan: @atarashansky - great catch! Will fix.

I am assuming the `indices` and `data` arrays will be constructed in such a way to tell the difference between implicit zeros (ie no value) vs explicit zero (a value that just happens to be zero)?

atarashansky: That's a good point. Unless `array.eliminate_zeros()` was called, there could be explicit zeros in a sparse matrix - though I don't know if that's the case for sparse data that was loaded from disk in AnnData.

atarashansky (approved):

backend/wmg/pipeline/utils.py
```diff
@@ -0,0 +1,86 @@
...
+        # index in the row corresponds to a row index in the sparse matrix and then filling
+        # this new array with counts non-zero columns for a corresponding row index
+        #
+        # Example: 4X4 sparse matrix stored in column orientation where ROW 3 has no values.
+        # That is, ROW 3 is not stored in the `adata.X.indices` array:
+        # column_oriented_sparse_matrix = np.array([[1, 0, 0, 2], [0, 4, 1, 0], [0, 0, 0, 0], [0, 0, 5, 0]])
+        number_per_cell = np.zeros(adata.shape[0], dtype="int")
+        number_per_cell[cell_indices] = counts
+    elif isinstance(adata.X, np.ndarray):
+        number_per_cell = (adata.X > 0).sum(axis=1).A1
```
atarashansky, line 76:
```suggestion
        number_per_cell = (adata.X > 0).sum(axis=1).flatten()
```
Apparently `.A1` is not a valid option.

atarashansky (changes requested):

backend/wmg/pipeline/utils.py
prathapsridharan: will fix!

atarashansky (approved):

PR #5414, opened 2023-08-07
feat: write to versioned snapshot paths when constructing the marker genes cube [wip]

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5166

## Changes

- if `snapshot_schema_version` is provided when loading the cube, then we'll upload the snapshot artifacts to `s3://<bucket_name>/snapshots/<schema_version>/`
- introduces a default param of `read_versioned_snapshots` into `load_snapshot`. this is set to `True` in the `calculate_markers` pipeline, to ensure that we're writing to the new location. however, it's defaulted to `False` so that all of our API's that call this function continue reading from the old location at the root directory

## Testing steps

- run WMG pipeline in dev after merging
- verify that the metadata got written to the new path in S3


backend/wmg/data/load_cube.py
```diff
@@ -35,28 +35,32 @@ def upload_artifacts_to_s3(
     information about the data schema version, the snapshot id, and the location
     of the snapshot.
 
-    TODO(prathap): This part of the docstring will be modified when work starts on ticket:
-    https://github.com/chanzuckerberg/single-cell-data-portal/issues/5166
+    The destination_path is determined by the following logic:
```
prathapsridharan, line 38: suggest `<destination_path>` (in angle brackets) since it is referred below

backend/wmg/data/load_cube.py
```diff
@@ -35,28 +35,32 @@ def upload_artifacts_to_s3(
...
+    - If `snapshot_schema_version` is set, then the destination path is
+        s3://<environment_bucket_name>/snapshots/<data_schema_version>/
+    - Otherwise, the destination path is the root directory: s3://<environment_bucket_name>/
```
prathapsridharan, line 41: I don't believe there is any `if-else` logic here. There is no option to omit `snapshot_schema_version`. I think this comment should just read:

```
The <destination_path> for the snapshot is `s3://<environment_bucket_name>/snapshots/<data_schema_version>/`

prathapsridharan (approved):
@joyceyan - The code looks good! I just have some comments about changing the docstrings a little bit for clarity.

prathapsridharan:
@joyceyan - Could you please move `WMG_DATA_SCHEMA_VERSION` from the `__init__.py` file under the `schemas` directory to a new file, call it `data_schema_config.py` under the `schemas` dir?

Reason being that it is uncommon to put constants in `__init__.py`. It is primarily used for module loading or creating a public interface to a package. 

You could argue that `WMG_DATA_SCHEMA_VERSION` belongs in the public interface of the `schemas` package but from someone reasoning about the code and wanting to update the schema version when they update the schema, it would be somewhat weird to not find this constant in a "config" or "constants" file.

PR #5422, opened 2023-08-08
fix(wmg): do not subset anndata and save filter state instead

atarashansky:
## Reason for Change

- #5378 

## Changes

Compute a boolean index of the cells (rows) to be filtered from the anndata sparse matrix and save it as an attribute in the `obs` array. Then perform the filtering when constructing the cubes. This allows getting through anndata processing (which is highly memory inefficient) and do the filtering in the memory efficient tiledb side of things.

## Testing steps

- Verified in a jupyter notebook loading a 53GB dataset that it was able to get through the portion where failure was happening due to filtering on the anndata object. Since we are skipping filtering in anndata, we get through this step with negligible memory overhead (~ 0.64 MiB). See below:

Screenshot 2023-08-08 at 5 44 20 PM


backend/wmg/pipeline/summary_cubes/cell_count.py
```diff
@@ -27,17 +27,26 @@ def extract(corpus_path: str) -> pd.DataFrame:
...
     integrated corpus obs arrays on relevant features
     """
+
+    # filter out observations in the 'filter_cells' attribute.
+    # It is important to filter out rejected observations BEFORE
+    # performing the `groupby` operation because the `groupby` operation
+    # would lose information about the `filter_cells` attribute because
+    # `filter_cells` is not one of the columns used in the `groupby` list
+    cells_to_keep_bool_index = np.logical_not(obs["filter_cells"])
+    obs_to_keep = obs[cells_to_keep_bool_index]
```
atarashansky, line 46:
```suggestion
    obs_to_keep = obs[~obs["filter_cells"]]
```
This is a style thing, so not a blocker, but the above is more concise - `cells_to_keep_bool_index` is not being reused anywhere so no need for the intermediate variable.

backend/wmg/pipeline/utils.py
```diff
@@ -72,22 +85,38 @@ def anndata_filter_cells_by_gene_counts_inplace(adata: AnnData, min_genes: int)
...
+        logger.info("Performing in place zeroing of AnnData.X rows to be filtered out")
+        rows_to_zero = np.where(lowly_covered_cells_to_filter_out)[0]
+        if isinstance(adata.X, csr_matrix):
+            for row in rows_to_zero:
+                adata.X.data[adata.X.indptr[row] : adata.X.indptr[row + 1]] = 0
+            adata.X.eliminate_zeros()
+        elif isinstance(adata.X, csc_matrix):
+            adata.X.data[np.in1d(adata.X.indices, rows_to_zero)] = 0
+            adata.X.eliminate_zeros()
+        else:
```
atarashansky, line 118: Just for safety, we should probably have `elif isinstance(adata.X, np.ndarray)` and then `else` raise invalid matrix type exception

backend/wmg/pipeline/summary_cubes/cell_count.py
prathapsridharan: will fix

backend/wmg/pipeline/utils.py
prathapsridharan: will fix

prathapsridharan (approved):
LGTM. @atarashansky started the PR. I contributed some. But since he is originator of the PR he can't be a reviewer. We talked about the PR and we both agree that it can be merged. Therefore, I am approving on our behalf.

PR #5428, opened 2023-08-09
feat: add Serializable isolation for select db operations

### LINKED ISSUE #5424: Race condition in `add_dataset_artifact`
# Context

See failure in [**single-cell-dataset-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1691539202724109?thread_ts=1690849743.567859&cid=C024HCSH9PT).

`add_dataset_artifact` in the persistence layer uses a write-after-read pattern which can cause a race condition, since Seurat and CXG are written in parallel. We can enforce serializable isolation to prevent this from happening. This will also require to retry the query if it fails due to concurrency, but even without the retry, it's still better than a silent failure.

ebezzi:
## Reason for Change

Fixes #5424 

## Changes

- Enforces a stricter transaction isolation level for some database operations. This should hopefully mitigate a race condition where the two pipelines (RDS and CXG) can update the same database record concurrently if they complete at the same time.
- TODO: this will only cause one of the transactions to fail. This will at least mark the dataset as failed, and not yield a silent failure like today. A better alternative is to add retries [as recommended by the PostgreSQL docs](https://www.postgresql.org/docs/current/transaction-iso.html). This can be done as part of this PR or as a follow up.

## Testing steps

- Added a unit test that simulates the concurrency via threading. I tried the unit test it main and it doesn't pass (as it should)
- For what it's worth, I tested this in rdev to make sure that the change doesn't affect the happy path.

### DIFF
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -5,13 +5,13 @@ import uuid
 from contextlib import contextmanager
 from datetime import datetime
 from typing import Any, Iterable, List, Optional, Tuple
-from tenacity import retry
-from tenacity.stop import stop_after_attempt
-from tenacity.wait import wait_fixed
 
 from sqlalchemy import create_engine
 from sqlalchemy.exc import ProgrammingError, SQLAlchemyError
 from sqlalchemy.orm import sessionmaker
+from tenacity import retry
+from tenacity.stop import stop_after_attempt
+from tenacity.wait import wait_fixed
 
 from backend.common.corpora_config import CorporaDbConfig
 from backend.layers.business.exceptions import CollectionIsPublishedException
--- a/backend/layers/persistence/persistence.py
+++ b/backend/layers/persistence/persistence.py
@@ -5,6 +5,9 @@ import uuid
 from contextlib import contextmanager
 from datetime import datetime
 from typing import Any, Iterable, List, Optional, Tuple
+from tenacity import retry
+from tenacity.stop import stop_after_attempt
+from tenacity.wait import wait_fixed
 
 from sqlalchemy import create_engine
 from sqlalchemy.exc import ProgrammingError, SQLAlchemyError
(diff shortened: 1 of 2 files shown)


backend/layers/persistence/persistence.py
```diff
@@ -92,6 +92,12 @@ def _manage_session(self, **kwargs):
             if session is not None:
                 session.close()
 
+    @contextmanager
+    def _get_serializable_session(self, **kwargs):
+        with self._manage_session(**kwargs) as session:
+            session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
```
Bento007, line 98: What does "SERIALIZABLE" do

atolopko-czi: Effectively, it prevents the race condition of two parallel db sessions clobbering one another's writes due each reading the same data, computing a result with that data,  and then writing back the result, where the losing process hasn't incorporated the winning process's update. The winning process's write gets clobbered.

https://www.postgresql.org/docs/current/transaction-iso.html#XACT-SERIALIZABLE

Postgres is actually quite smart about this mode and doesn't actually run the 2 txns in a serial manner--instead they run in parallel, but psql figures out if a "serializable anomaly" has occurred and errors out the offending process.

atolopko-czi (commented):
Looks reasonable so far!

It _is_ possible to write tests for this with threads, though it's tricky due to the explicit coordination you need to implement between them. I suspect CoPilot will make easy work of it. :)

ebezzi:
> Looks reasonable so far!
> 
> It _is_ possible to write tests for this with threads, though it's tricky due to the explicit coordination you need to implement between them. I suspect CoPilot will make easy work of it. :)

I have an idea for testing this in a "simple" way but I will have to timebox it since it's not guaranteed to work. If I don't get it by today I'd go ahead and merge this. I also added retries, PTAL.

atolopko-czi (approved):
Bring on the isolation!

backend/layers/persistence/persistence.py
danieljhegeman: >Serializable isolation typically relies upon locking of rows or ranges of rows in order to achieve this effect and can increase the chance of deadlocks and degrade performance

"increase the chance of deadlocks" -- I don't see how/why this would be the case... 🤷‍♂️

Issue #5168, opened 2023-07-07
Make WmgQuery accept optional indexed_dims parameter to enable the wmg pipeline to use WmgQuery when creating marker genes cube


prathapsridharan:
This allows `WmgQuery` object to [specify the dimensions to use to query the cubes](https://github.com/chanzuckerberg/single-cell-data-portal/blob/eb71ce726adc151478aa4243e62414c706225739/backend/wmg/data/query.py#L77).

The use of this technique is an expedient to get around the fact that `WmgQuery` is being used both by the wmg api and wmg pipeline. [The wmg pipeline uses `WmgQuery` to query `expression_summary_fmg` and `cell_counts` cube to create the `marker_genes` cube](https://github.com/chanzuckerberg/single-cell-data-portal/blob/eb71ce726adc151478aa4243e62414c706225739/backend/wmg/pipeline/summary_cubes/calculate_markers.py#L107)

This method is the fastest way to enable data schema version pinning for wmg api.

_A longer and more robust implementation involves the following steps_ (**which we will not attempt here but is mentioned here for context in the future**):

- Make wmg pipeline write to s3 the metadata about the snapshot that includes _latest_snapshot_ and _indexed dimensions of the cubes_
- Make WmgSnapshot object expose the data schema version and the index dimensions of each cube through fields and methods
- Make WmgQuery methods query the WmgSnapshot object it holds a reference to get the indexed dimensions of each cube

Issue #5361, opened 2023-07-31
Deprecate WMG V1 API and associated code and artifacts


prathapsridharan:
1. Delete the `v1.py` module
2. Move all associated tests to run with `v2` wmg api
3. Delete any other code or artifacts associated with `v1` that cannot be ported to `v2` or is not useful in the context of `v2`

prathapsridharan:
This ticket will be addressed in multiple PRs

PR #5429, opened 2023-08-10
refactor(wmg): Decouple wmg api and wmg pipeline when calling WMGQuery()

prathapsridharan:
## Reason for Change

- #5168 

## Changes

Modify `class WMGQuery` such that it can be instantiated with a reference to an object encapsulating the schema of the cube it is meant to query. This allows the WMG API and WMG Pipeline to query cubes at different schema versions

## Testing steps

- Unit Tests pass


backend/wmg/api/v2.py
```diff
@@ -51,7 +57,12 @@ def query():
             explicit_snapshot_id_to_load=WMG_API_FORCE_LOAD_SNAPSHOT_ID,
         )
 
-        q = WmgQuery(snapshot)
+        cube_query_params = WmgCubeQueryParams(
```
joyceyan, line 60: do we need to make the equivalent changes to the v1 API?

backend/wmg/data/query.py
```diff
@@ -77,9 +77,31 @@ class MarkerGeneQueryCriteria(BaseModel):
...
+    def _transform_cube_index_name(self, index_name: str, pluralize: bool = True) -> str:
+        if index_name != "organism_ontology_term_id" and pluralize:
+            return index_name + "s"
+
+        return index_name
+
+
 class WmgQuery:
-    def __init__(self, snapshot: WmgSnapshot) -> None:
+    def __init__(self, snapshot: WmgSnapshot, cube_query_params: WmgCubeQueryParams) -> None:
```
joyceyan, line 102: just for my own understanding, under what circumstances would we not want to set the `cube_query_params` to be
```
WmgCubeQueryParams(
        cube_query_valid_attrs=READER_WMG_CUBE_QUERY_VALID_ATTRIBUTES,
        cube_query_valid_dims=READER_WMG_CUBE_QUERY_VALID_DIMENSIONS,
    )
    ```
?

joyceyan (approved):

backend/wmg/api/v2.py
prathapsridharan: `V1` will be deprecated. I have already removed unit tests for `V1`. There is another ticket to deprecate V1: https://github.com/chanzuckerberg/single-cell-data-portal/issues/5361

backend/wmg/data/query.py
prathapsridharan: See the `calculate_markers.py` file in this PR for an example of when `WmgCubeQueryParams` is instantiated with `WRITER_*` instead of `READER_*`. That is the wmg pipeline uses the same query interface to construct cubes it has just finished building in order to build the `marker_genes_cube`.

However, the writer (pipeline) uses the latest schema to query the cube, whereas the reader (wmg api) might at a schema version that is `<=` the latest schema and therefore might have different `attrs` and `dimensions` to parse through as encoded in the `READER_*` constants

PR #5432, opened 2023-08-10
feat(cellguide): Replace Source Data Scroll with Pagination

ashin-czi:
## Reason for Change

- #5408

## Changes

- add
- remove
- modify
    - Remove scrolling for Source Data, replace with pagination similar to Marker Genes table

## Testing steps

1. Scroll down to Source Data
2. Paginate through Source Data pages for various cell types


atarashansky (approved):
LGTM!

PR #5477, opened 2023-08-14
feat(cellguide): build ontologyTree.json

atarashansky:
## Reason for Change

- This PR adds the logic for generating the ontologyTree.json fixture, which is a nested graph representation of the ontology tree with cell counts populated in each node.

## Testing steps

- Unit test added and passed
- The tree output by this module was programmatically compared to the one used by the existing app in production and no differences were observed


backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -0,0 +1,188 @@
...
+        cell_counts_df_rollup: pandas.Series
+            A series with index "cell_type_ontology_term_id" and values "n_cells".
+            "n_cells" contains the sum of "n_cells" for the cell type and its descendants.
+        """
+
+        cell_counts_df = (
+            cell_counts_df.groupby("cell_type_ontology_term_id").sum(numeric_only=True)[["n_cells"]].reset_index()
+        )
+
+        to_attach = pd.DataFrame()
```
joyceyan, line 75: might be nice to just add a code comment in english here that explains what we're attaching. if my understanding is correct, we're just attaching all the cell types that exist in "all cell type ids" that is not in our `cell_counts_df` and setting the number of cells to 0

backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -0,0 +1,188 @@
...
+            This series is the cell counts dataframe with the following modification:
+            1. The cell type ontology term ids that are not in the input cell counts dataframe are added with 0 counts.
+
+        cell_counts_df_rollup: pandas.Series
+            A series with index "cell_type_ontology_term_id" and values "n_cells".
+            "n_cells" contains the sum of "n_cells" for the cell type and its descendants.
+        """
+
+        cell_counts_df = (
+            cell_counts_df.groupby("cell_type_ontology_term_id").sum(numeric_only=True)[["n_cells"]].reset_index()
```
joyceyan, line 72: just for my own understanding, what does `reset_index` here do?

backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -0,0 +1,188 @@
...
+        The suffix is the number of times the node had been visited prior.
+
+        Arguments
+        ---------
+        node: pronto.Term
+            A node of the cell ontology. This should be the root node of the ontology at the top-level.
+        traverse_node_counter: dict, optional, default=None
+            A dictionary that maps cell type ontology term ids to the number of times they have been visited.
+        all_unique_nodes: set, optional, default=None
+            A set of all unique cell type ontology term ids that have been visited.
```
joyceyan, line 139: just to clarify, this is actually a set of all unique nodes + times they've been visited right?

backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns
+        -------
+        ontology_graph: dict
+            A nested dictionary representation of the cell ontology tree.
+        """
+        if traverse_node_counter is None:
```
joyceyan, line 146: this is kind of a nit, but i think it would be a bit better to remove this block here, instead initialize the call in `OntologyTreeBuilder` to set `traverse_node_counter = {}` and `all_unique_nodes = set()`, and require the parameters to be set rather than having a default of `None`

backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -0,0 +1,188 @@
...
+            A list of all cell type ontology term ids.
+        """
+        id_to_name = {}
+        for term in self.ontology.terms():
+            if term.id.startswith("CL:"):
+                id_to_name[term.id] = term.name
+        all_cell_type_ids = list(id_to_name.keys())
+        return id_to_name, all_cell_type_ids
+
+    def _traverse_ontology_with_counting(self, node, traverse_node_counter=None, all_unique_nodes=None):
```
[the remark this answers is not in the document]
joyceyan: nice - this is probably only the second time i've ever seen a tree traversal in production code :)

backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -0,0 +1,188 @@
...
+                None,
+                None,
+            )
+
+        children = []
+        for child in subclasses:
+            subtree, _, _ = self._traverse_ontology_with_counting(child, traverse_node_counter, all_unique_nodes)
+            children.append(subtree)
+
+        return (
```
joyceyan, line 176: you don't necessarily need to update this here, but in general i think it's a better pattern to have a typed return object instead of an unstructured tuple. so instead of returning a tuple here, you might have something like:

```
class TraverseOntologyResult:
  subtree: dict[str, Any]
  traverse_node_counter: dict[str, int]
  all_unique_nodes: set[str]

def _traverse_ontology_with_counting(self, node, traverse_node_counter=None, all_unique_nodes=None) -> TraverseOntologyResult:
  <all of your code>
```

that way the caller knows explicitly what to expect, and maybe more importantly, if you ever add more stuff in the return value here in the future, it becomes much more obvious that the caller needs to update things

backend/cellguide/pipeline/ontology_tree/tree_builder.py
joyceyan, line 126: also an actual code comment on this line: you can add an actual type annotation to the parameter, like:
```
def _traverse_ontology_with_counting(self, node, traverse_node_counter: Optional[dict[str, Any]], all_unique_nodes: Optional[set[str]]):
```
which avoids the need to document the type in the code comments and is more useful since you can use the type annotations to enforce the caller to pass in the right type

joyceyan (approved):

backend/cellguide/pipeline/ontology_tree/tree_builder.py
atarashansky: `groupby(...).sum(...)` converts the group-by dimensions into a MultiIndex (where each index value is a tuple of the group values). In the case where the group-by is one dimension, then it just converts the groupby dimension into an Index (each index value is a cell type in this case). 

`reset_index` converts the Index/MultiIndex back into the column/columns.

backend/cellguide/pipeline/ontology_tree/tree_builder.py
atarashansky: traverse_node_counter is a dictionary with cell type term ids as keys and number of times they've been visited as values.

all_unique_nodes is a set of cell type term ids suffixed by the number of times that cell type had been visited prior to that particular instance of it being reached during the traversal

backend/cellguide/pipeline/ontology_tree/tree_builder.py
atarashansky: That's what it was, but I think it's an anti-pattern. `traverse_node_counter` is an argument only because it's required to be passed to recursive calls of the traversal function. As long as the function documents what it returns, then there's no need to initialize it outside the function call.

backend/cellguide/pipeline/ontology_tree/tree_builder.py
atarashansky: Great suggestion, I'll update this in the PR and ping you to take a look before merging.

joyceyan: looks great!

PR #5499, opened 2023-08-15
chore(wmg+cellguide): use pinned ontology in single-cell-curation

atarashansky:
## Reason for Change

- WMG and CellGuide CL ontology versions need to be synced with the data portal schema.

## Testing steps
 - tests pass


backend/wmg/data/utils.py
```diff
@@ -207,3 +208,21 @@ def to_dict(a, b):
...
+    Parameters:
+    name (str): The name of the ontology (e.g. cl-basic.obo).
+
+    Returns:
+    str: The URL of the pinned ontology.
+    """
+
+    response = requests.get(CL_PINNED_CONFIG_URL)
+    decoded_yaml = yaml.safe_load(response.content.decode())
+    cl_url = list(decoded_yaml["CL"]["urls"].values())[0]
```
nayib-jose-gloria, line 226: there's no guarantee the latest value will always be the first in the list. I'd grab `key = decoded_yaml["CL"]["latest"]` and then `cl_url = decoded_yaml["CL"]["URLs"][key]`, then do the splitting

nayib-jose-gloria (approved):

PR #5504, opened 2023-08-15
chore(cellguide): add types to tree builder pipeline

atarashansky:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

### DIFF
--- a/backend/cellguide/pipeline/ontology_tree/tree_builder.py
+++ b/backend/cellguide/pipeline/ontology_tree/tree_builder.py
@@ -9,6 +9,11 @@ import pandas as pd
 from pronto import Ontology, Term
 
 from backend.cellguide.pipeline.constants import UBERON_BASIC_PERMANENT_URL_PRONTO
+from backend.cellguide.pipeline.ontology_tree.types import (
+    OntologyTree,
+    OntologyTreeStatePerCellType,
+    OntologyTreeStatePerTissue,
+)
 from backend.common.utils.rollup import rollup_across_cell_type_descendants
 from backend.wmg.data.constants import CL_BASIC_PERMANENT_URL_PRONTO
 
@@ -38,7 +43,7 @@ HEMATOPOIETIC_CELL_TYPE_ID = "CL:0000988"
 
 @dataclass
 class TraverseOntologyResult:
-    subtree: Dict[str, Any]
+    subtree: OntologyTree
     traverse_node_counter: Dict[str, int]
     all_unique_nodes: set[str]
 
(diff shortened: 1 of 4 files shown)


backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -247,8 +252,11 @@ def _traverse_ontology_with_counting(
             all_unique_nodes=all_unique_nodes,
         )
 
+    def get_ontology_tree(self) -> OntologyTree:
+        return self.ontology_graph
+
     ### Get the ontology tree state
-    def get_ontology_tree_state_per_celltype(self) -> Dict[str, Any]:
+    def get_ontology_tree_state_per_celltype(self) -> Dict[str, OntologyTreeStatePerCellType]:
```
joyceyan, line 259: this comment is applicable to all of the functions where the return type changed, but i believe you need to create an actual instance of the return type. so in this function, you'd need to have this be
```
                all_states_per_cell_type[end_node] = OntologyTreeStatePerCellType(
                    "isExpandedNodes": isExpandedNodes,
                    "notShownWhenExpandedNodes": notShownWhenExpanded,
                )
```

joyceyan (changes requested):

backend/cellguide/pipeline/ontology_tree/tree_builder.py
```diff
@@ -719,7 +722,7 @@ def _build_parents_per_node(self, node, all_parents=None) -> Dict[str, list[str]
             self._build_parents_per_node(child, all_parents)
         return all_parents
 
-    def _get_deepcopy_of_ontology_graph(self) -> Dict[str, Any]:
+    def _get_deepcopy_of_ontology_graph(self) -> OntologyTree:
```
joyceyan, line 725: for a python dataclass i believe you would want to update this to be
```
return copy.deepcopy(self.ontology_graph)
```
https://docs.python.org/3/library/copy.html

joyceyan (changes requested):
i don't think i can leave an inline comment unless it's something being changed by the PR, but a few more changes:
- update code comments to reflect type changes
- in places where we generate a dummy node i think we want that to be of type `OntologyTree` right?
- related to ^ might be useful to have something in `utils.py` that returns a dummy node so you avoid some code duplication

backend/common/utils/dataclass.py
```diff
@@ -0,0 +1,38 @@
+from dataclasses import asdict, is_dataclass
+
+
+def convert_dataclass_to_dict(data, remove_nones=True):
```
joyceyan, line 4: i think for this i would prefer if we either don't set a default or default it to `False`. if you consider what a generic "convert dataclass to dict" util would do, it feels more intuitive to not alter the data at all

backend/common/utils/dataclass.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    if is_dataclass(data):
+        data = asdict(data)
+
+    elif isinstance(data, dict):
+        for key, value in data.items():
+            if is_dataclass(value):
+                data[key] = asdict(value)
+
+    return _remove_none_values(data)
```
joyceyan, line 22: also the `remove_nones` param is not actually used here lol

joyceyan (changes requested):

backend/common/utils/dataclass.py
atarashansky: Whoopsie

backend/common/utils/dataclass.py
atarashansky: Nevermind, I just did a little refactoring and added a cellguide-specific wrapper for the convert dataclass to dict.

joyceyan (approved):

PR #5518, opened 2023-08-15
fix(schema-migration): Pass collection_version to publish_collection_version

Bento007:
## Reason for Change

- The wrong type was being passed into `publish_collection_version`

## Changes

- Pass CollectionVersion into publish_collection_version
- Add unit test to verify

## Testing steps
- Unit tested
- ran schema migration on rdev https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/v2/executions/details/arn:aws:states:us-west-2:699936264352:execution:dp-rdev-memory-schema-migration-sfn:7221f0fc-e2b3-37da-db2f-748e7963e83d

## Notes for Reviewer

- Implementing `mypy` would avoid problems like this.


backend/layers/processing/schema_migration.py
```diff
@@ -224,7 +224,7 @@ def publish_and_cleanup(self, collection_version_id: str, can_publish: bool) ->
         if errors:
             self._store_sfn_response("report", collection_version_id, errors)
         elif can_publish:
-            self.business_logic.publish_collection_version(collection_version_id)
+            self.business_logic.publish_collection_version(collection_version)
```
Bento007, line 227: This value is set [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d70ce570d359090e3b1aa84c94892ed06d1c68c8/backend/layers/processing/schema_migration.py#L185)

nayib-jose-gloria (approved):

PR #5530, opened 2023-08-17
feat(cellguide): computational marker genes

atarashansky:
## Reason for Change

- #5464 

## Changes

- add
- remove
- modify

## Testing steps


backend/cellguide/pipeline/canonical_marker_genes/canonical_markers.py
```diff
@@ -42,24 +42,8 @@ def __init__(self, *, wmg_tissues: list[str], wmg_human_genes: list[str]):
 
         self.wmg_human_genes = wmg_human_genes
 
-        file_path = self._get_symbol_to_gene_descriptions_file_path()
```
atarashansky, line 45: I refactored this fixture out into the utility module.

backend/cellguide/pipeline/computational_marker_genes/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+            marker_genes[key] += marker_genes_per_tissue[key]
+        else:
+            marker_genes[key] = marker_genes_per_tissue[key]
+
+    return marker_genes
+
+
+def output_marker_genes(marker_genes: dict, output_directory: str):
+    """
+    This function outputs the marker genes to a specified directory. Each cell type will be output to its own JSON file.
```
atarashansky, line 61: for FE reasons, the computational marker gene data needed to be split out into a bunch of different JSONs, one per cell type.

cell_guide_local_pipeline/cell_guide_data_pipeline.ipynb
atarashansky, line 1: Deleting the notebook as it is no longer needed once your source data PR gets in.

atarashansky: (YAY!)

scripts/generate_realistic_test_snapshot.py
```diff
@@ -26,7 +26,7 @@
 python3 scripts/generate_realistic_test_snapshot.py <PATH_TO_LOCAL_SNAPSHOT> tests/unit/backend/wmg/fixtures/realistic-test-snapshot
 """
 
-test_tissue = "UBERON:0002048"
+test_tissues = ["UBERON:0002048", "UBERON:0002113"]  # lung and kidney
```
atarashansky, line 29: I added another tissue so that I can test marker genes per tissue

backend/cellguide/pipeline/computational_marker_genes/computational_markers.py
atarashansky, line 1: Start here! @joyceyan

backend/cellguide/pipeline/computational_marker_genes/utils.py
```diff
@@ -0,0 +1,189 @@
...
+    sumsq2 - Sum of squares of the second set of data
+    n2 - Number of elements in the second set of data
+
+    Returns
+    -------
+    pvals - The p-values of the t-test
+    effects - The effect sizes of the t-test
+    """
+
+    with np.errstate(divide="ignore", invalid="ignore"):
```
joyceyan, line 87: do you think it would make sense to use a standard library for computing this instead of writing the logic ourselves?

https://www.geeksforgeeks.org/how-to-conduct-a-two-sample-t-test-in-python/#

i don't have a strong opinion on this particular use case since it's already written and appears to match this
image

but maybe something to keep in mind for the future?

backend/cellguide/pipeline/computational_marker_genes/utils.py
```diff
@@ -0,0 +1,189 @@
...
+    pvals - The p-values from the statistical test
+    effects - The effect sizes from the statistical test
+    percentile - The percentile to use for thresholding (default is 0.05)
+
+    Returns
+    -------
+    A dictionary mapping marker genes to their statistics.
+    """
+
+    is_colinear = np.array([are_cell_types_colinear(cell_type, cell_type_target) for cell_type in cell_types_context])
```
joyceyan, line 136: just for my own understanding, why do we set colinear cell types to null?

backend/cellguide/pipeline/computational_marker_genes/utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+    pvals[:, np.all(np.isnan(pvals), axis=0)] = 1
+    effects[:, np.all(np.isnan(effects), axis=0)] = 0
+
+    # aggregate
+    effects = nanpercentile_2d(effects, percentile * 100, 0)
+
+    effects[effects == 0] = np.nan
+
+    # pvals = np.array([stats.combine_pvalues(x[np.invert(np.isnan(x))] + 1e-300)[-1] for x in pvals.T])
```
joyceyan, line 148: should we delete this line instead of leaving it as a comment?

backend/cellguide/pipeline/computational_marker_genes/utils.py
```diff
@@ -0,0 +1,189 @@
...
+    arr - 2D array to calculate percentile of
+    percentile - percentile to calculate, as a number between 0 and 100
+    axis - axis along which to calculate percentile
+
+    Returns
+    -------
+    The specified percentile of the 2D array along the specified axis.
+    """
+    if axis == 0:
+        result = np.empty(arr.shape[1])
```
joyceyan, line 28: is there any reason why this isn't just
```
        result = np.empty(arr.shape[axis])
        for i in prange(arr.shape[axis]):
            arr_row = arr[i, :]
            result[i] = nanpercentile(arr_row, percentile)
        return result
```

backend/cellguide/pipeline/computational_marker_genes/types.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+@dataclass
+class ComputationalMarkerGenes:
+    me: float
+    pc: float
+    marker_score: float
+    symbol: str
+    name: str
+    groupby_dims: Dict[str, str]
```
joyceyan, line 12: nit: in the future you can just use the primitive `dict[str, str]` here. previously, the benefit of using `Dict` from the `typing` package was that it would allow you to type the key/value pairs. as of some version of python recently (i forget exactly, but within the last few years?) the primitive `dict` (and `list`) support typing of the key/value pairs. the benefit of primitive types is that it's one fewer thing for you to import, but there's also no real harm in importing from the `typing` package

joyceyan (approved):

backend/cellguide/pipeline/computational_marker_genes/utils.py
atarashansky: I don't think the standard library can do the t-test elementwise and with array broadcasting, but i could be wrong

backend/cellguide/pipeline/computational_marker_genes/utils.py
atarashansky: I'll add this as a comment - we don't want a parent node to be compared to its child because they share expressions, since we are rolling up expressions across descendants

PR #5539, opened 2023-08-17
feat: collection resurrection

danieljhegeman:
## Reason for Change
Collection: Resurrection 👽 🪦 ⏪ 
- #4828 

## Changes

- add cxg admin script to allow the "untombstoning", aka **resurrection**, of a tombstoned Collection. 

## Testing steps

✅ unit tested
🚧 rdev testing in progress...


danieljhegeman:
Whoops. I rebased earlier when I should have merged, hence the absurd number of commits it suggests are contributing to this diff 😂 . No harm though.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1398,18 +1398,34 @@ def test_delete_collection_version_ok(self):
         self.assertEqual(version.version_id, published_collection.version_id)
         self.assertNotEqual(version.version_id, new_version.version_id)
 
-    def test_tombstone_collection_ok(self):
+    def test_tombstone_then_resurrect_collection_ok(self):
```
danieljhegeman, line 1401: this test replacement improves upon the original `test_tombstone_collection_ok` and also includes resurrection test coverage.

backend/layers/business/business.py
```diff
@@ -692,6 +693,20 @@ def tombstone_collection(self, collection_id: CollectionId) -> None:
         self.delete_all_dataset_versions_from_public_bucket_for_collection(collection_id)
         self.database_provider.tombstone_collection(collection_id)
 
+    def resurrect_collection(self, collection_id: CollectionId) -> None:
+        """
+        Resurrect a tombstoned Collection (untombstone Collection and un-delete public s3 assets). For admin use only.
+        """
+        collection = self.get_canonical_collection(collection_id)
+        if not collection.tombstoned:
+            raise CollectionIsNotTombstonedException()
```
Bento007, line 702: avoid negative logic, and add a test for this error case.
```suggestion
            raise CollectionIsAliveException()
```

backend/layers/business/business.py
```diff
@@ -692,6 +693,20 @@ def tombstone_collection(self, collection_id: CollectionId) -> None:
         self.delete_all_dataset_versions_from_public_bucket_for_collection(collection_id)
         self.database_provider.tombstone_collection(collection_id)
 
+    def resurrect_collection(self, collection_id: CollectionId) -> None:
+        """
+        Resurrect a tombstoned Collection (untombstone Collection and un-delete public s3 assets). For admin use only.
```
Bento007, line 698: Initially reading this it was easy to assume this would only restore a single collection version and it's datasets. The code looks like it restores all previous version of the collection and previous versions of the datasets. Include that in the description or change the code.

backend/layers/business/business.py
```diff
@@ -692,6 +693,20 @@ def tombstone_collection(self, collection_id: CollectionId) -> None:
...
+        """
+        collection = self.get_canonical_collection(collection_id)
+        if not collection.tombstoned:
+            raise CollectionIsNotTombstonedException()
+        all_dataset_versions = self.database_provider.get_all_dataset_versions_for_collection(collection_id)
+        for dv in all_dataset_versions:
+            for ext in (DatasetArtifactType.H5AD, DatasetArtifactType.RDS):
+                object_key = f"{dv.version_id}.{ext}"
+                self.s3_provider.resurrect_object(os.getenv("DATASETS_BUCKET"), object_key)
```
Bento007, line 707: nit: consider using restore vs resurrected

backend/layers/business/business.py
```diff
@@ -692,6 +693,20 @@ def tombstone_collection(self, collection_id: CollectionId) -> None:
...
+        self.database_provider.untombstone_collection(collection_id)
```
Bento007, line 708: I would stick with untombstone or the resurrect notation. I can see this becoming a confusing topic differentiating between tombstone, deleted, untombstone, and resurrected,

backend/layers/persistence/persistence.py
```diff
@@ -461,6 +461,22 @@ def tombstone_collection(self, collection_id: CollectionId) -> None:
             for dataset in datasets:
                 dataset.tombstone = True
 
+    def untombstone_collection(self, collection_id: CollectionId) -> None:
+        """
+        Untombstones a canonical collection.
```
Bento007, line 466: and all of it's datasets and dataset versions.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1398,18 +1398,34 @@ def test_delete_collection_version_ok(self):
...
         """
-        A collection can be marked as tombstoned using 'tombstone_collection'
+        A Collection can be tombstoned and then resurrected/untombstoned
```
Bento007, line 1403: resurrected or untombstoned. I prefer **restore** but resurrect also works

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1398,18 +1398,34 @@ def test_delete_collection_version_ok(self):
...
-        # The collection version canonical collection has tombstoned marked as True
-        collection_version = self.business_logic.get_collection_version(
-            published_collection.version_id, get_tombstoned=True
-        )
-        self.assertTrue(collection_version.canonical_collection.tombstoned)
+        # The collection version canonical collection has tombstoned attribute marked as True
+        collection = self.business_logic.get_canonical_collection(published_collection.collection_id)
+        assert collection.tombstoned is True
+        # Verify public Dataset asset files are gone
+        assert all([not self.s3_provider.uri_exists(uri) for uri in public_dataset_asset_s3_uris])
```
Bento007, line 1420: All of the assert above should have been tested in other tests. They should be removed. The reason it to minimize the scope of the tests to only the specific function.

tests/unit/scripts/test_tombstone.py
```diff
@@ -16,9 +16,25 @@ def test__tombstone_collection(self):
...
+        context.obj = {"business_logic": self.business_logic}
+        tombstone_collection(context, c_v.collection_id.id)
+
+        c_v = self.business_logic.get_collection_version(c_v.version_id, get_tombstoned=True)
+        self.assertTrue(c_v.canonical_collection.tombstoned)
+
+    def test__resurrect_collection(self):
+        c_v = self.generate_published_collection()
+        self.assertFalse(c_v.canonical_collection.tombstoned)
+        [self.assertFalse(d.canonical_dataset.tombstoned) for d in c_v.datasets]
```
Bento007, line 29: This and the above assertions are not needed. This is tested else were. This goes for the `test__tombstone_collection` as well.

tests/unit/scripts/test_tombstone.py
```diff
@@ -16,9 +16,25 @@ def test__tombstone_collection(self):
...
         [self.assertFalse(d.canonical_dataset.tombstoned) for d in c_v.datasets]
 
+        context = MagicMock(spec=Context)
+        context.obj = {"business_logic": self.business_logic}
+        tombstone_collection(context, c_v.collection_id.id)
+
+        c_v = self.business_logic.get_collection_version(c_v.version_id, get_tombstoned=True)
+        self.assertTrue(c_v.canonical_collection.tombstoned)
+
+    def test__resurrect_collection(self):
```
Bento007, line 26: There is a lot of duplicated setup shared between these tests. Factor it out into a pytest fixture.

Bento007 (changes requested):
back to you

backend/layers/business/business.py
danieljhegeman: How about `CollectionIsPublicException` 😄  as much as I like the commitment to 🧟 💀 vibes...

backend/layers/business/business.py
danieljhegeman: Sure. I was thinking that "resurrect" is the _business_ term while "tombstone" is the _persistence_ term. But I can use "resurrect" everywhere.

tests/unit/backend/layers/business/test_business.py
danieljhegeman: I prefer "resurrect" because it is more directly associated with "tombstone" as being its opposite. I think "restore" makes more sense in the context of restoring deleted s3 assets.

danieljhegeman:
@Bento007 just realized I'm missing a portion of logic. When resurrecting, we need to check for individually-tombstoned Datasets prior to the tombstoning of the Collection. If any exist, those need to remain tombstoned. I'll make the change.

tests/unit/backend/layers/business/test_business.py
danieljhegeman: changed as requested ✔️

tests/unit/scripts/test_tombstone.py
danieljhegeman: changed as requested

tests/unit/scripts/test_tombstone.py
danieljhegeman: I can't get this to work nicely with the class-contained tests.

danieljhegeman:
```
    def _callTestMethod(self, method):
>       method()
E       TypeError: TestTombstone.test__resurrect_collection() missing 1 required positional argument: 'context'```

Bento007 (approved):

Bento007 (approved):

PR #5541, opened 2023-08-18
chore: use functest@example.com for functional testing

ebezzi:
## Reason for Change

- Tentatively fixes the TestApiKey functional test flakyness


Bento007:
Which environments have been updated?
- [ ] dev
- [ ] staging
- [ ] rdev
- [ ] prod

Bento007 (approved):

tests/functional/backend/common.py
```diff
@@ -41,7 +41,7 @@ def setUpClass(cls):
             allowed_methods={"DELETE", "GET", "HEAD", "PUT" "POST"},
         )
         cls.session.mount("https://", HTTPAdapter(max_retries=retry_config))
-        token = cls.get_auth_token(cls.config.test_account_username, cls.config.test_account_password)
+        token = cls.get_auth_token(cls.config.functest_account_username, cls.config.functest_account_password)
```
atolopko-czi, line 44: Where does this get configured and where does an engineer learn of its existence as config param? I see `test_account_username` appears in many places, including mentioned in a frontend/tests/README.md and `setup_data_data.sh`, so would have expected similar updates elsewhere.

ebezzi: It's a username/password type of secret, so it has to be added manually to SSM. The auth0 user associated with it can be easily tracked down in auth0 once you know the username (which is again part of SSM). I think this is self-documented.

PR #5542, opened 2023-08-18
feat(cellguide): implement pipeline for the source collections data artifact

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5466

## Changes

- generates the source collections data artifact
- i added a new citation formatter to format responses from the collections API

## Testing steps

- added a test for generating source collections data
- i also added a test for formatting citations from the collections API

### DIFF
--- a/backend/cellguide/pipeline/canonical_marker_genes/utils.py
+++ b/backend/cellguide/pipeline/canonical_marker_genes/utils.py
@@ -85,11 +85,10 @@ def format_citation_dp(message: dict) -> str:
     else:
         author_str = f"{first_author['name']} et al."
 
-    journal = message["journal"] if message["journal"] else ""
-
+    journal = " " + message["journal"] if message["journal"] else ""
     year = f"{message['published_year']}"
 
-    return f"{author_str} ({year}) {journal}"
+    return f"{author_str} ({year}){journal}"
 
 
 def format_citation_crossref(message: dict) -> str:
@@ -113,7 +112,7 @@ def format_citation_crossref(message: dict) -> str:
     else:
         author_str = f"{first_author['name']} et al."
 
-    journal = message["container-title"][0] if len(message["container-title"]) else ""
+    journal = " " + message["container-title"][0] if len(message["container-title"]) else ""
     year = message["created"]["date-parts"][0][0]
 
-    return f"{author_str} ({year}) {journal}"
+    return f"{author_str} ({year}){journal}"
--- a/backend/cellguide/pipeline/canonical_marker_genes/utils.py
+++ b/backend/cellguide/pipeline/canonical_marker_genes/utils.py
@@ -71,7 +71,7 @@ def format_citation_dp(message: dict) -> str:
     Parameters
     ----------
     message : dict
-        The message containing publication details from the /collections API.
+        The message containing publisher_metadata from the /collections API.
 
     Returns
     -------
--- a/backend/cellguide/pipeline/canonical_marker_genes/utils.py
+++ b/backend/cellguide/pipeline/canonical_marker_genes/utils.py
@@ -52,11 +52,11 @@ def get_title_and_citation_from_doi(doi: str) -> str:
         # Get the title and citation count from the data
         try:
             title = data["message"]["title"][0]
-            citation = format_citation(data["message"])
+            citation = format_citation_crossref(data["message"])
         except Exception:
             try:
                 title = data["message"]["items"][0]["title"][0]
-                citation = format_citation(data["message"]["items"][0])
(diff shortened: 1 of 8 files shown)


backend/cellguide/pipeline/canonical_marker_genes/utils.py
atarashansky, line 1: Woah, didn't know there were so many variations for the different pub metadata!

atarashansky: Oh, I just realized you're overloading the same function to format citations from both crossref and the data portal.

atarashansky: I think it would be better to split these out into `format_citation_crossref` and `format_citation_dp` to be explicit about what kind of publisher metadata each function expects.

backend/cellguide/pipeline/source_collections/source_collections_generator.py
```diff
@@ -0,0 +1,77 @@
...
+            collection_id = dataset["collection_id"]
+
+            # If we don't have any source data on this collection yet, create a new SourceCollectionsData object
+            if collection_id not in collections_to_source_data:
+                collection = collections_dict.get(collection_id)
+                source_data = SourceCollectionsData(
+                    collection_name=collection["name"],
+                    collection_url=collection["collection_url"],
+                    publication_url=collection["doi"],
+                    publication_title=format_citation(collection["publisher_metadata"])
```
atarashansky, line 46: See above comment about splitting `format_citation` into two functions for crossref and data portal publisher metadata, respectively:

> I think it would be better to split these out into format_citation_crossref and format_citation_dp to be explicit about what kind of publisher metadata each function expects.

backend/cellguide/pipeline/source_collections/source_collections_generator.py
```diff
@@ -0,0 +1,77 @@
...
+            # dataset to the existing object. If not, we'll want to create a new list for each of these fields.
+            source_data = collections_to_source_data[collection_id]
+
+            for tissue in dataset["tissue"]:
+                if tissue["ontology_term_id"] not in [t["ontology_term_id"] for t in source_data.tissue]:
+                    source_data.tissue.append(tissue)
+
+            for disease in dataset["disease"]:
+                if disease["ontology_term_id"] not in [d["ontology_term_id"] for d in source_data.disease]:
+                    source_data.disease.append(disease)
```
atarashansky, line 66: out of curiosity, have you observed duplicate diseases/tissues/organisms in the source data or is this just a guardrail?

tests/unit/cellguide_pipeline/test_canonical_marker_genes.py
```diff
@@ -79,3 +79,20 @@ def test__format_citation(self):
         }
         result = format_citation(message)
         self.assertEqual(result, "Doe, John et al. (2022) Test Journal")
+
+        message = {
```
atarashansky, line 83: nice :)

atarashansky (approved):
Approving now - I'll leave it up to you if you want to split format_citation into two functions or not. It's just a nit from me.

tests/unit/cellguide_pipeline/test_source_collections_generator.py
```diff
@@ -0,0 +1,30 @@
...
+from tests.test_utils.dict_compare import compare_dicts
+from tests.unit.backend.wmg.fixtures.test_snapshot import (
+    load_realistic_test_snapshot,
+)
+
+TEST_SNAPSHOT = "realistic-test-snapshot"
+
+
+class TestSourceCollectionsGenerator(unittest.TestCase):
+    def test__csource_collections_generator(self):
```
atarashansky, line 16: typo: test__source
```suggestion
    def test__source_collections_generator(self):
```

backend/cellguide/pipeline/canonical_marker_genes/utils.py
joyceyan: ohhh okay this makes so much more sense. i was wondering why the response data was so different 🤔

backend/cellguide/pipeline/source_collections/source_collections_generator.py
joyceyan: i have not - this is just a slightly rewritten guardrail from what was originally in the python notebook

tests/unit/cellguide_pipeline/test_source_collections_generator.py
joyceyan: lol ty

PR #5561, opened 2023-08-20
feat(cellguide): add script to regenerate pipeline test fixtures

atarashansky:
## Reason for Change

- #5502 

To run this script:
`python -m scripts.generate_cellguide_pipeline_test_fixtures`

## Changes

- added a script to regenerate all cellguide pipeline test fixtures
- added a function to `test_snapshot.py` that returned the test snapshot directly instead of using a context manager. this was needed to patch the load snapshot function in the cellguide pipeline to load the test snapshot instead.

## Testing steps

- I ran the script and observed no diffs in the generated test fixtures.


backend/cellguide/pipeline/canonical_marker_genes/canonical_markers.py
```diff
@@ -34,7 +34,10 @@ def __init__(self, *, wmg_tissues: list[str], wmg_human_genes: list[str]):
         """
 
         logger.info("Fetching ASCTB data...")
-        self.asctb_data = requests.get(ASCTB_MASTER_SHEET_URL).json()
+        session = setup_retry_session()
```
atarashansky, line 37: setup_retry_session is more robust than requests.get because it retries if connection is interrupted.

scripts/generate_cellguide_pipeline_test_fixtures.py
atarashansky, line 1: This script generates test fixtures for cellguide pipeline.

scripts/generate_cellguide_pipeline_test_fixtures.py
```diff
@@ -0,0 +1,127 @@
...
+
+TEST_SNAPSHOT = "realistic-test-snapshot"
+
+
+""" ################################# DANGER #######################################
+
+Run this file if and only if you are confident there are no bugs in the CellGuide
+pipeline.
+
+Any and all unit test assertion errors prior to running this script must be expected
```
atarashansky, line 48: this is important to flag

scripts/generate_cellguide_pipeline_test_fixtures.py
```diff
@@ -0,0 +1,127 @@
...
+due to intended changes in the pipeline.
+
+------------------------------------------------------------------------------------
+
+This module generates the CellGuide data using the test snapshot stored in {TEST_SNAPSHOT}.
+Requires an internet connection.
+"""
+
+
+def custom_load_snapshot_into_cube_dir(cube_dir: str, **kwargs):
```
atarashansky, line 58: this replaces "load_snapshot" for the purposes of this script

scripts/generate_cellguide_pipeline_test_fixtures.py
```diff
@@ -0,0 +1,127 @@
...
+    return load_realistic_test_snapshot_obj(TEST_SNAPSHOT, cube_dir)
+
+
+def custom_output_computational_marker_genes(
```
atarashansky, line 62: this is so that we can output computational marker genes into one test fixture instead of how the pipeline does it (one file per cell type)

tests/unit/backend/wmg/fixtures/test_snapshot.py
```diff
@@ -218,6 +218,50 @@ def load_realistic_test_snapshot(snapshot_name: str) -> WmgSnapshot:
             )
 
 
+def load_realistic_test_snapshot_obj(snapshot_name: str, cube_dir: str) -> WmgSnapshot:
```
atarashansky, line 221: this function is needed to be able to patch "load_snapshot" since `load_snapshot` returns WmgSnapshot whereas `load_realistic_test_snapshot` returns a context manager.

tests/unit/cellguide_pipeline/constants.py
atarashansky, line 1: Some constants that set the fixture filenames and fixtures basepath

scripts/generate_cellguide_pipeline_test_fixtures.py
```diff
@@ -0,0 +1,127 @@
...
+Run this file if and only if you are confident there are no bugs in the CellGuide
+pipeline.
+
+Any and all unit test assertion errors prior to running this script must be expected
+due to intended changes in the pipeline.
+
+------------------------------------------------------------------------------------
+
+This module generates the CellGuide data using the test snapshot stored in {TEST_SNAPSHOT}.
+Requires an internet connection.
```
joyceyan, line 54: can you add the command for running the script to this comment block?

joyceyan (approved):

Issue #5559, opened 2023-08-18
[Gene Expression] Heatmap Tooltip CellType Shows Tissue Id instead of CellType Name


tihuan:
Bug:
In prod - Heatmap Tooltip CellType Shows Tissue Id instead of CellType Name

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcXdiIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--55b22751b03b0f4da4ce9ba0c5749287eaf013a3/Screenshot%202023-08-18%20at%209.33.04%20AM.png)

DoD:
1. Fix this bug
2. Add a test to catch this bug

tihuan:
@niknak33 @signechambers1  please help add a priority tag thank you!

niknak33:
I put it down as a P0, mainly because it is a data quality issue. There is no workaround to the cell type, as the data in the field corresponds to a tissue ID. @signechambers1 let me know if you feel different.

tihuan:
Awesome thanks, Nik! @seve I know you have adding WMGv2 new tests this week, but are you able to send out a fix for this today 🙏 ? Thanks so much!

seve:
Should the tooltip still say `Cell Type` for organisms? I'm going to fix the bug and push for merging this because of the priority, but we may want to open a seperate ticket to address.

tihuan:
> Should the tooltip still say `Cell Type` for organisms? I'm going to fix the bug and push for merging this because of the priority, but we may want to open a seperate ticket to address.
> 
> CC: @hthomas-czi @niknak33

@seve Just to double check: you meant the tooltip should say "Tissue" instead of "Cell Type" for tissue rows, right?

PR #5567, opened 2023-08-21
fix(wmg): add correct tooltip cell type for tissue rows and add test coverage

seve:
## Reason for Change

- #5559

## Changes

- add
  -  Adds logic fork for tissue tooltips, uses `celltype.name` instead of celltype.cellTypeName` if the `name` contains `"UBERON"`
  -  ~Adds test to ensure UBERON isn't present in the tooltip text~

### DIFF
--- a/frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
@@ -86,7 +86,7 @@ const TOOLTIP_THROTTLE_MS = 100;
 let handleDotHoverAnalytic: NodeJS.Timeout;
 
 export default memo(function Chart({
-  cellTypes,
+  cellTypes: dataRows,
   selectedGeneData = EMPTY_ARRAY,
   setIsLoading,
   tissue,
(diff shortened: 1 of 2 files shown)


seve:
I originally added a test to `tests/features/wheresMyGene/cellTooltip.test.ts`, but that doesn't test the heatmap tooltip, rather the truncation functionality. We have no test coverage on the heatmap tooltip AFAIK.

frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
```diff
@@ -306,9 +306,13 @@ export default memo(function Chart({
       ],
     };
 
+    const cellTypeName = cellType.cellTypeName.startsWith("UBERON")
```
tihuan, line 309: to avoid confusion, should we rename `cellType` to `row` or something better, and add notes that row can be either a tissue or a cellType, so it's clearer why `cellType.cellTypeName` can be a tissueId?

Thank you!

tihuan:
> I originally added a test to `tests/features/wheresMyGene/cellTooltip.test.ts`, but that doesn't test the heatmap tooltip, rather the truncation functionality. We have no test coverage on the heatmap tooltip AFAIK.

Ah got it! I think it's worth adding a test for this in a separate PR. What do you think? Thanks for the quick fix too 🙏 ✨

seve:
> Ah got it! I think it's worth adding a test for this in a separate PR. What do you think? Thanks for the quick fix too 🙏 ✨

I'll write up a ticket!

tihuan (approved):
Readability is so much better! Thanks for taking the time to get it done 🚀 !

Issue #5491, opened 2023-08-14
Cell type filter selections do not clear


signechambers1:
When a user selects a cell type (B cell for example), a blue tag shows up under the cell type search bar. However when you click the x to clear the filter, the tag disappears but the results do not refresh and B cell continues to be shown.

Desired behavior is that the application shows all cell types when a cell type filter is removed. Or, if there are multiple cell types, the cell type the user clears should no longer show in the application.

seve:
cell types clear if we reduce the cell type filter to something > 0.

This is stinky https://github.com/chanzuckerberg/single-cell-data-portal/pull/5357/files#r1279803002

cc: @tihuan

seve:
cell types clear if we reduce the cell type filter to something > 0.

This is stinky https://github.com/chanzuckerberg/single-cell-data-portal/pull/5357/files#r1279803002

cc: @tihuan

tihuan:
Ooh yeah my bad! I thought that condition was just for resetting the state when organism changes 🤦 

Adding these back seems to work, but probably further testing is needed and adding tests too

```ts
if (filteredCellTypes.length === 0) {
  setDisplayedCellTypes(initialDisplayedCellTypeIds);
  setExpandedTissues(new Set<string>());
  return;
}
```

Thank you!

seve:
> Adding these back seems to work, but probably further testing is needed and adding tests too

Since this is main functionality and not an edge case, this functionality will be covered by testing described in #5447

tihuan:
> > Adding these back seems to work, but probably further testing is needed and adding tests too
> 
> Since this is main functionality and not an edge case, this functionality will be covered by testing described in #5447

Sounds good! I added this as a task item in #5447 👍 !

PR #5568, opened 2023-08-21
fix: correctly handle setting filtered cell types to 0

seve:
## Reason for Change

- #5491 

## Changes

- add
  Re add logic to reset state when resetting filtered cell type state

pair coded with @prathapsridharan

### DIFF
--- a/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
+++ b/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
@@ -11,7 +11,11 @@ import {
   useState,
 } from "react";
 import SearchIcon from "@mui/icons-material/Search";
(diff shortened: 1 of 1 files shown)


frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -296,6 +300,8 @@ export default memo(function HeatMap({
   // update displayedCellTypes and expandedTissues
```
tihuan, line 300:
```suggestion
  // Reset `displayedCellTypes` and `expandedTissues` when the user clears `filteredCellTypes`
```

tihuan (approved):
LGTM! Just one comment suggestion 👍 Thank you!

prathapsridharan (approved):
LGTM! Thanks for the pair session @seve

Issue #5555, opened 2023-08-18
Hotfix: Schema Migration Batch Job Publish did not Publish 3.1 Revisions


nayib-jose-gloria:
3.1 Corpus Schema Migration succeeds until reaching the 'collection_publish' step. At this point, a wrong input is passed into our business layer publish function (string of Collection Version ID rather than the expected CollectionVersionId dataclass type). This causes datasets to successfully migrate in a revision, but ultimately fail to be published. 

Impact: No user-facing impact, as nothing was published. However, Migration did not officially complete, and all published collections in prod were left with an open revision containing 3.1 migration changes. We could have published manually to resolve, but we chose instead (confirmed by EM/PM) to rollback the migration revisions and re-attempt migration with a hotfix in order to have a clean end-to-end run of the schema migration automation in prod, as this was our first attempt to do so and our last scheduled opportunity until schema 4.0.0 in the Fall. Therefore, public collections were reverted back to their state before migration (i.e. no revisions open if they did not have one prior). 

A Hotfix has been requested in order to attempt a new migration earlier than the next production deployment date. 

Fix was completed in ticket: https://github.com/chanzuckerberg/single-cell-data-portal/pull/5518

This issue is tracking the hotfix.

PR #5569, opened 2023-08-21
fix(schema-migration): migrated.h5ad file cleanup

nayib-jose-gloria:
## Reason for Change

- #5555 
- Issue with cleanup of temp migrated.h5ad files created during schema-migration; wrong IDs were used for comparison to determine whether cleanup + error reporting occurred. As a result, files were not cleaned up. 

## Changes

- compare canonical dataset IDs to determine whether a dataset in a collection was processed by schema migration (rather than version ID, which creates a new dataset version ID when you do dataset replacement ingestion but keeps the stable canonical dataset ID) 

## Testing steps

- rdev migration

### DIFF
--- a/backend/layers/processing/schema_migration.py
+++ b/backend/layers/processing/schema_migration.py
@@ -188,15 +188,17 @@ class SchemaMigrate(ProcessingLogic):
 
         # Get the datasets that were processed
         extra_info = self._retrieve_sfn_response("publish_and_cleanup", collection_version.collection_id.id)
-        processed_datasets = [d["dataset_version_id"] for d in extra_info["datasets"]]
+        processed_datasets = {d["dataset_id"]: d["dataset_version_id"] for d in extra_info["datasets"]}
 
         # Process datasets errors
         for dataset in collection_version.datasets:
+            dataset_id = dataset.dataset_id.id
             dataset_version_id = dataset.version_id.id
-            if dataset_version_id not in processed_datasets:
+            if dataset_id not in processed_datasets:
                 # skip datasets that were not processed
                 continue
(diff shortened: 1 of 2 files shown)


backend/layers/processing/schema_migration.py
```diff
@@ -188,15 +188,17 @@ def publish_and_cleanup(self, collection_version_id: str, can_publish: bool) ->
...
         for dataset in collection_version.datasets:
+            dataset_id = dataset.dataset_id.id
             dataset_version_id = dataset.version_id.id
-            if dataset_version_id not in processed_datasets:
+            if dataset_id not in processed_datasets:
                 # skip datasets that were not processed
                 continue
-            key_prefix = self.get_key_prefix(dataset_version_id)
+            # filepath to clean-up uses dataset_version_id from the replaced version; accessing with dataset_id as key
+            key_prefix = self.get_key_prefix(processed_datasets[dataset_id])
```
Bento007, line 201: Something is still wrong about this. This `dataset_version_id` is for the unmigrated dataset. We do want to  delete the /migrated.h5ad from it.

Bento007 (changes requested):

backend/layers/processing/schema_migration.py
nayib-jose-gloria: Isn't that where the migrated.h5ad is uploaded? The unmigrated dataset's directory? At the time we run migrate, we don't know the new dataset's version ID yet.

Bento007: You're right, the syntax is confusing. Can we break out `processed_datasets[dataset_id]` in a variable name `old_dataset_version_id`

Bento007 (approved):

PR #5577, opened 2023-08-21
fix(schema-migration): add more logging

Bento007:
## Reason for Change

Add more logging to make debugging easier in the event a report is not created or an error occurs during publish. The function to delete artifact was moved before publish to make sure artifact are deleted even if publish fails.

## Changes

- add logging for the error message for each dataset during publish
- Move delete datasets artifact before publish. This is to ensure the migrated files get cleanup.

### DIFF
--- a/backend/layers/processing/schema_migration.py
+++ b/backend/layers/processing/schema_migration.py
@@ -194,44 +194,52 @@ class SchemaMigrate(ProcessingLogic):
         for dataset in collection_version.datasets:
             dataset_id = dataset.dataset_id.id
             dataset_version_id = dataset.version_id.id
-            if dataset_id not in processed_datasets:
-                # skip datasets that were not processed
-                continue
             # filepath to clean-up uses dataset_version_id from the replaced version; accessing with dataset_id as key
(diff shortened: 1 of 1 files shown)


backend/layers/processing/schema_migration.py
```diff
@@ -194,44 +194,52 @@ def publish_and_cleanup(self, collection_version_id: str, can_publish: bool) ->
...
+            _log_extras = {
+                "dataset_id": dataset_id,
+                "dataset_version_id": dataset_version_id,
+                "previous_dataset_version_id": previous_dataset_version_id,
+            }
+            if dataset_id not in processed_datasets:
+                self.logger.info("skipping dataset", extra=_log_extras)
+                continue
+            self.logger.info("checking dataset", extra=_log_extras)
+            key_prefix = self.get_key_prefix(dataset_version_id)
```
Bento007, line 208:
```suggestion
            key_prefix = self.get_key_prefix(previous_dataset_version_id)
```

nayib-jose-gloria (approved):

PR #5582, opened 2023-08-22
feat(cellguide): upload pipeline output to S3

atarashansky:
- Refactor the pipeline so that it can upload the cellguide data in the proper format to S3.
- Instead of storing the data for a particular component (like marker genes) in a single file, we are storing the response for each cell type in a separate JSON file.
  - With the exception of the ontology tree, the cell metadata, and the tissue metadata.
 - This PR also refactors the subpipelines to adopt a consistent code structure.

## Testing steps

- No functional code changed (e.g. outputs remain the same) and tests pass.

## Notes for Reviewer
 - Some of the test fixtures were regenerated, but the actual contents are not different - It's just that some of the list values are in different orders. If you run `compare_dicts` on the before/after JSONs that got updated, it returns True.


prathapsridharan (approved):
LGTM

PR #5606, opened 2023-08-24
feat: set up ChatGPT description pipeline

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/zh/14

## Changes

- sets up the ChatGPT description generator pipeline outside of integration with the actual OpenAI
- adds a method to `S3Provider.does_object_exist` to easily determine if an object path has something or not
- introduces a 3rd party provider layer, `OpenAIProvider`, which will handle setting up the API key + actually hitting the API. this will be fleshed out in a separate PR

## Testing steps

n/a

## Notes for Reviewer
see inline comments


backend/cellguide/pipeline/gpt_descriptions/gpt_description_generator.py
```diff
@@ -0,0 +1,52 @@
...
+    📁 gpt_descriptions
+    ﹂CL_0000000.json
+    ﹂CL_0000001.json
+    ﹂...
+
+    Returns a mapping of cell type id to GPT description
+    """
+
+    # TODO: figure out how to actually configure this environment variable
+    bucket_name = os.getenv("CELL_GUIDE_DATA_BUCKET_NAME")
```
joyceyan, line 26: what's the recommended way to store bucket names? as environment variables or somewhere else?

backend/layers/thirdparty/openai_provider.py
atarashansky, line 1: Can we move this into `cellguide/pipeline/providers/`? `backend/layers` is used for the data portal backend + database business logic.

backend/layers/thirdparty/s3_provider.py
atarashansky, line 1: Let's make our own s3 provider in cellguide/pipeline/providers/

backend/cellguide/pipeline/gpt_descriptions/gpt_description_generator.py
atarashansky: Check the PR I'm working on - there's a CellGuideConfig object I added which can retrieve the bucket stored in an AWS secret. in fact, I would recommend building your PR off my branch (`cellguide-fe`).

backend/cellguide/pipeline/gpt_descriptions/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+from backend.cellguide.pipeline.gpt_descriptions.gpt_description_generator import generate_all_gpt_descriptions
+from backend.cellguide.pipeline.ontology_tree import get_ontology_tree_builder
+from backend.wmg.api.wmg_api_config import WMG_API_SNAPSHOT_SCHEMA_VERSION
+from backend.wmg.data.snapshot import load_snapshot
+
+logging.basicConfig(level=logging.INFO)
+
+
+def run():
```
atarashansky, line 11: Can you follow the pattern used in the other __init__ files (a `run` function that calls a `get_*` function). The `get_*` function is just a one-line wrapper around your generate function. For example, [source collections](https://github.com/chanzuckerberg/single-cell-data-portal/blob/19aafe00de0787274131fbb78f08e465f1461b47/backend/cellguide/pipeline/source_collections/__init__.py#L21). this pattern ensures consistency so the behavior of the functions are clearly delineated. For example, `run` gets the data and outputs the data. `get` gets the data.

atarashansky: This pattern was adopted to make it easier to disentangle data generation from data output while keeping imports less confusing (particularly useful for [the script](https://github.com/chanzuckerberg/single-cell-data-portal/blob/19aafe00de0787274131fbb78f08e465f1461b47/scripts/generate_cellguide_pipeline_test_fixtures.py#L65) used to generate test fixtures).

atarashansky (changes requested):

backend/layers/thirdparty/s3_provider.py
atarashansky: We might not even need a provider. We could use `subprocess` to execute an `aws s3 ls` command. We can then just check if "CL_000000.json" is in `captured_output_from_subprocess`. If yes, file exists, if no, file doesn't exist.

atarashansky: Though, it wouldn't hurt to use boto3 :)

backend/layers/thirdparty/openai_provider.py
joyceyan: I don't think it makes sense for the cellguide pipeline to have its own provider? It's true we're only using OpenAI for the cellguide pipeline, but I don't think it's correct to consider OpenAI to be a "cellguide pipeline provider". I can move it to `backend/common/providers` though, i didn't realize that `layers` wasn't a "common" module

backend/layers/thirdparty/s3_provider.py
joyceyan: Similar to the above comment I don't think it's really correct to have a separate S3 provider for each use case (ex: cellguide pipeline, data portal backend, etc). It would make more sense for this to live in the `common` module since the underlying logic (upload file, download file, check if file exists, etc) should be the same regardless of which client is using it and it doesn't really make sense to have a separate provider for each use case.

backend/layers/thirdparty/openai_provider.py
atarashansky: The primary concern is to not add code dependencies to the docker container. backend/common is already imported so that's better. But I still think it makes more sense to put the OpenAI provider in the cellguide backend (whether in cellguide/backend/pipeline or in cellguide/backend). The OpenAI API keys will be stored in the CellGuide AWS secret. If other portal pipelines decide they want to use OpenAI as well, then we can move it to a common location.

atarashansky: Could you elaborate on why you don't think it's correct to consider OpenAI to be a cellguide provider?

backend/cellguide/pipeline/gpt_descriptions/gpt_description_generator.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def upload_descriptions_to_s3(new_gpt_descriptions: dict[str, str]) -> None:
+    bucket_name = CellGuideConfig().bucket
+
+    for cell_type_id, description_str in new_gpt_descriptions.items():
+        object_key = f"gpt_descriptions/{cell_type_id}.json"
+
+        # Write to a local temp file
+        description_file = "/tmp/{cell_type_id}.json"
```
atarashansky, line 43: You can use the context manager instead of writing to `/tmp/`... Sometimes `/tmp/` might not be writeable (at least in my experience).

```
from tempfile import TemporaryDirectory
with TemporaryDirectory() as output_dir:
.....
```

backend/common/providers/s3_provider.py
```diff
@@ -0,0 +1,46 @@
...
+import boto3
+
+from backend.layers.thirdparty.s3_provider_interface import S3ProviderInterface
+
+AWS_S3_MAX_ITEMS_PER_BATCH = 1000
+
+logger = logging.getLogger(__name__)
+
+
+class S3Provider(S3ProviderInterface):
```
atarashansky, line 12: Is it necessary to use the S3Provider class instead of doing what the WMG and CellGuide pipelines do to upload their data (they use `subprocess` to invoke aws CLI directly)?

https://github.com/chanzuckerberg/single-cell-data-portal/blob/3ed9e0a7da5a4de26eda52a85cb8372fcb60ee32/backend/cellguide/pipeline/__init__.py#L68

Instead of uploading one file at a time, we could use `aws s3 sync` to upload the local directory with new GPT descriptions to the remote directory.

atarashansky: Granted, `S3Provider.does_object_exist` is nicer than parsing captured stdout from a `subprocess` run. Could you implement a sync function in this provider and use that instead of `upload_file`?

atarashansky (changes requested):

backend/common/providers/s3_provider.py
joyceyan: i think in general it's better to use a provider for a 3rd party service so that the calling service isn't familiar with the internal workings of the 3rd party service. as far as the cellguide pipeline is aware, it only needs to "upload files to storage service" or "check if object exists at this path in the storage service. a few benefits of this compartmentalization:

1. abstracting away the internal workings of a 3rd party provider makes it a bit cleaner if you ever want to switch cloud providers, say to google cloud or azure.
2. reduces code reuse. the same reason why we might want to package `output_json_per_key` into `utils.py` instead of copying and pasting the same code everywhere applies here.
3. reusing code reduces the odds of introducing bugs. in the case where you're executing the aws cli directly, you might accidentally miss a param in there or something.
4. code reviewers can have have confidence that `S3Provider.upload_file` works as expected without needing to review or understand any underlying logic backing that

backend/cellguide/pipeline/gpt_descriptions/gpt_description_generator.py
joyceyan: updated - thanks for the pointer, didn't know about that package!

backend/common/providers/s3_provider.py
atarashansky: Could you implement a sync function in the provider and use that for the GPT pipeline instead of uploading files one at a time?

joyceyan: updated!

backend/cellguide/pipeline/gpt_descriptions/gpt_description_generator.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # write to a local temp directory gpt_descriptions/
+    for cell_type_id, description_str in new_gpt_descriptions.items():
+        with TemporaryDirectory() as temp_dir:
+            local_file_path = os.path.join(temp_dir, f"{cell_type_id}.json")
+            with open(local_file_path, "w") as file:
+                file.write(description_str)
+
+    # upload to s3
+    S3Provider.sync_directory(src_dir=temp_dir, s3_uri=f"s3://{bucket_name}/gpt_descriptions/")
```
atarashansky, line 48: This needs to be in the `with TemporaryDirectory()` context because the tmp dir is deleted after leaving the context.

atarashansky (changes requested):

backend/cellguide/pipeline/gpt_descriptions/gpt_description_generator.py
atarashansky: I'd suggest moving TemporaryDirectory() outside the for loop, since you're making a new tmp dir every loop

joyceyan: good catch, updated

atarashansky (approved):

Issue #4311, opened 2023-03-07
Portal deployed_version endpoint not reporting version


atolopko-czi:
## Motivation

Cannot easily verify the deployed version due to endpoint not working:

https://api.cellxgene.cziscience.com/dp/v1/deployed_version responded:
```
{
"Data Portal": ""
}
```

## Definition of Done

Endpoint reports the deployed version

## Tasks

- [ ] Troubleshoot and fix

seve:
Wasn't able to get this done during my on-call. I added a test to check that the deployed_version !== "" and that was passing when running `make local-unit-test`. So this may be only occurring when the containers are built for deployment?

dsadgat:
@ebezzi was there progress made on this ticket?

ebezzi:
Yes, I did some more investigation, but wasn't able to understand the cause of the flakyness. I isolated the issue and the problem is at build time (if the endpoint returns empty, it means the Docker container was built with an empty variable). It is not clear to me yet what causes the Docker container to be built with the empty variable - likely a happy issue. I probably need to add some logs to the build pipeline and see if they add clarity. I am _happy_ to assign the ticket to me and keep working/monitoring while we do deployments.

metakuni:
This appears to be working now:

```
// 20230711143942
// https://api.cellxgene.cziscience.com/dp/v1/deployed_version

{
  "Data Portal": "21d07e9a9e78c881b42bdbe27cd5f7c90a04e07f"
}
```

but leaving this issue open until we figure out the root cause.

metakuni:
Case in point re: flakiness:

Staging is also working:
```
// 20230711144248
// https://api.cellxgene.staging.single-cell.czi.technology/dp/v1/deployed_version

{
  "Data Portal": "3646a4c30f74149da4e1d3bca8de35d20e60519f"
}
```

but not dev:
```
// 20230711144246
// https://api.cellxgene.dev.single-cell.czi.technology/dp/v1/deployed_version

{
  "Data Portal": ""
}
```

Preassigning to @ebezzi for further investigation on his next on-call rotation.

metakuni:
I think this can be closed in favor of #5479 .

cc: @atolopko-czi , @Bento007

metakuni:
Assigning to @Bento007 for consideration

atolopko-czi:
If we're giving up on trying to fix this, I would recommend this issue be changed to _remove_ the flakey endpoint.

[...]

PR #5617, opened 2023-08-25
fix: portal version endpoint

Bento007:
## Reason for Change

- #4311 

## Changes

- remove unused make recipe `local-backend`
- modify make recipe `.env.ecr` to add HAPPY variable for testing. 
- modify *docker_compose.yml* to explicitly assign HAPPY variables.

## Testing steps

- Update the function test for version to look for empty string rather than None.


danieljhegeman (commented):
why add the env var to _all_ Docker image configs when it's only being used by the backend? I'm wondering if this is potentially more confusing than useful, unless you anticipate us printing out one of those values in the processing pipeline, or in FE code...? I suppose?

edit: I guess it's only two and it's clear what they are. Doesn't hurt.

Does it work in rdev?

danieljhegeman (approved):

Bento007:
They docker recipes are already requesting the argument. The changes makes it more likely they will get the args requested.

PR #5638, opened 2023-08-29
chore(wmg): rebuild dataset dict if previous load_snapshot failed to do so

atarashansky:
## Reason for Change

- [Slack context](https://czi-sci.slack.com/archives/C023Q1APASK/p1693329070018309)

## Changes

- disentangle dataset_dict building from snapshot loading. Even if the snapshot exists, if dataset_dict does not exist as an attribute, it should be rebuilt.


prathapsridharan (approved):
lgtm

PR #5649, opened 2023-08-30
feat(cellguide): improve DAG filtering by tissue

atarashansky:
## Reason for Change

- Previously, when selecting a tissue, we were filtering the DAG in a CellGuide Card by taking the intersection of the expanded nodes in the Tissue tree and the expanded nodes in the Cell tree. This led to invalid tree states with the target cell type being collapsed for certain tissues.
- In retrospect, there was no justification for choosing to intersect the expanded nodes in the tree DAG and the cell type DAG.
- The correct approach is to use the cell tree state but inject tissue counts into the tree state. This activates logic used for pruning subtrees that do not belong in the tissue (which was previously only being used in the tissue card DAGs).


prathapsridharan (approved):
Very nice simplification!

PR #5655, opened 2023-08-30
chore: set fetch depth to 1 for all push test actions (2 for tests for codecov)

atarashansky:
## Reason for Change
Tests have been taking upwards of 20 minutes just at the checkout step due to our bloat in git commit history. Until such time that we prune our history and reduce size of the repo, we can decrease the time of our GHA jobs significantly by fetching only the tip of our commit history.

## Changes
 - set fetch depth to 1 for all actions except for the push tests (due to codecov), which i set to depth of 2


Bento007:
What are you making this change?

atarashansky:
> What are you making this change?

I'm making this change because the GHA steps have been taking forever (~20 minutes) at the checkout action because of the git bloat.

Bento007 (approved):

PR #5657, opened 2023-08-30
fix: fix staging e2e tests

atolopko-czi:
staging deploy failed due to a minor GHA bug; bitten again by the "stage" terminology of happy!


tihuan (approved):
Aw crap missed that! Thanks for the fix!

PR #5659, opened 2023-08-30
chore(submission): avoid using root logger, silence noisy boto libraries

Bento007:
## Reason for Change

- make logs easier to parse, by grouping them under a logger name. 
- silence noisy boto libraries, so debug setting can actually be used.

## Changes

- avoid using root logger
- name the processing logger
- Force log configuration for processing container, in case logs were configured somewhere else for example in a test fixture.
- Don't log debug message from botocore or s3transfer, because they are noisy and can leak session keys.

## Testing steps

- verified the log output is correct in the unit tests.


backend/layers/processing/logger.py
```diff
@@ -7,21 +7,25 @@
...
 
 def configure_logging(level=None):
     level = getattr(logging, os.environ.get("LOG_LEVEL", "INFO").upper())
     log_stdout_handler = logging.StreamHandler(stream=sys.stdout)
     formatter = jsonlogger.JsonFormatter(LOG_FORMAT, DATETIME_FORMAT)
     log_stdout_handler.setFormatter(formatter)
-    logging.basicConfig(handlers=[log_stdout_handler], level=level)
+    logging.basicConfig(handlers=[log_stdout_handler], level=level, force=True)
+    logging.getLogger("botocore").setLevel(max([logging.INFO, level]))  # don't set boto3 less than INFO
+    logging.getLogger("s3transfer").setLevel(max([logging.INFO, level]))
```
danieljhegeman, line 20: doesn't this require a `s3transfer` logger exist already? Where is it? I'm not seeing it in the repo

Bento007: It does **not** need to be instantiated by the `s3transfer` library before being configured.

danieljhegeman: Where is it being used? And how is it being accessed? I'm ignorant on this topic

danieljhegeman: Ah I see. The `s3transfer` python package will actually pick up this logger to use for its own logging, right?

danieljhegeman (approved):

backend/layers/processing/logger.py
Bento007: exactly, and it will be the same one we configure on this line

Issue #5437, opened 2023-08-10
Data mirroring for RDev stacks [CI/CD]


atolopko-czi:
For both ad hoc test and automated E2E and functional tests, RDev envs will need to be initialized with prod-representative data. We already data mirroring scripts to move data from prod to dev and staging, which can be adapted to also mirror to RDev. Both engineers and automated testing workflows should be able to invoke this script.

Implementation notes:
* For ad hoc testing needs, should support user-selectable data types to be mirrored: Portal DB (RDS), H5ADs (S3), CXGs (S3), WMG snapshot (S3) and possibly by collections/datasets (subsets of corpus).
* It may be sufficient for E2E tests to only test on subsets of corpus data, so the user-selectable data types feature may also be useful for automated tests.
* Any S3-related code that is not currently capable of finding its data using S3 paths that include the "stack prefix" will need to be updated (WMG only?)
* It is worth considering having this script run as a Batch job (or some other hosted compute), rather than in the localhost process, but that is a stretch goal.

Goal:
Allow engineers to initialize an RDev stack with selected production data, as needed for a particular dev/testing effort.

Provide a script that can mirror any combination of the following data types:
- [X] Portal database (RDS db) sync
- [X] S3 Portal datasets: .h5ad, .rds. Specify by list of datasets and or collection identifiers.
- [X] S3 Explorer datasets: CXGs.  Specify by list of datasets and or collection identifiers.
- [X] S3 WMG snapshot(s)
- [X] S3 CellGuide data (to s3://cellguide-data-public-dev/env-rdev-cellguide/{stack-name}/)

tihuan:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5437?planning-poker) with Zenhub @atolopko-czi @Bento007 @danieljhegeman @ebezzi

Bento007:
duplicate https://github.com/chanzuckerberg/single-cell-data-portal/issues/5274?

tihuan:
Hmm yeah feels like duplicate to me, maybe worth combining the descriptions since they're slightly different? Will defer to BE engs to decide what to do 😆

atolopko-czi:
We pointed the dup issue at 3 pts, so I pointed this issue the same.

danieljhegeman:
[...]

PR #5662, opened 2023-08-31
feat: enable mirroring of postgres data to arbitrary rdev stack

danieljhegeman:
- accept/handle 'rdev' as a valid DEST_ENV
- use STACK env var to specify rdev stack
- do not copy s3 assets from SRC_ENV
- asset uris will be incorrect; trivial for rdev

## Reason for Change

- #5437

## Changes

- Parallelize _all_ s3 sync and s3 cp operations, not just for rdev
- ***For rdev only***: Use sample dataset from `s3://env-rdev-dataportal/s3/` to populate all Dataset artifacts if COLLECTIONS argument is provided. Much faster. If _actual_ Dataset is required in explorer, developer can manually copy the asset file(s) to desired locale.

## Testing steps


backend/scripts/mirror_rds_data.sh
```diff
@@ -30,22 +30,34 @@ make db/dump OUTFILE=$DB_DUMP_FILE
...
 [[ $ANS == 'Y' ]] || exit 1
 
+function load_src_dump_to_dest_db() {
+  PGPASSWORD=${DB_PW} pg_restore --clean --if-exists --no-owner --no-privileges --no-comments --dbname=${DB_NAME} \
+  --host 0.0.0.0 --username ${DB_USER} --schema=persistence_schema ${DB_DUMP_FILE}
+}
+
 make db/tunnel/up
-PGPASSWORD=${DB_PW} pg_restore --clean --if-exists --no-owner --no-privileges --no-comments --dbname=${DB_NAME} --host 0.0.0.0 --username ${DB_USER} --schema=persistence_schema ${DB_DUMP_FILE}
+load_src_dump_to_dest_db || load_src_dump_to_dest_db  # Hack for rdev, where it fails on first and succeeds on retry
```
danieljhegeman, line 60: Is this a hack? Yes. Does it work flawlessly? Evidently, yes. The second attempt only runs if the first one fails; no effect on normal mirroring, and for rdev it satisfies the requirement that we run the script twice. User sees no difference. No error messages are surfaced.

atarashansky: Lol, I love the preemptive defensiveness 🔥

danieljhegeman: not my first rodeo when it comes to submitting suspicious code 😂

atolopko-czi: Curious, what error does it fail with on the first attempt?

backend/scripts/mirror_rds_data.sh
```diff
@@ -30,28 +30,48 @@ make db/dump OUTFILE=$DB_DUMP_FILE
...
 make db/tunnel/down
 
-DB_UPDATE_CMDS=$(cat <<EOF
--c "UPDATE persistence_schema.\"DatasetArtifact\" SET uri = regexp_replace(uri, '(s3:\\/\\/)([[:alpha:]]+-[[:alpha:]]+-)([[:alpha:]]+)(\\/.+)', '\\1\\2${DEPLOYMENT_STAGE}\\4') WHERE uri IS NOT NULL;"
+if [[ ! $DEST_ENV == 'rdev' ]]; then
```
atolopko-czi, line 63: nit: `!=`

danieljhegeman: ah thank you, my bash brain does weird things at times

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,7 +11,55 @@ set -e
...
+echo Copying WMG cube snapshot...
+
+latest_snapshot=$(aws s3 cp s3://cellxgene-wmg-${SRC_ENV}/latest_snapshot_run -)
+if [[ $DEST_ENV == 'rdev' ]]; then
+  s3_destination="env-rdev-wmg/${STACK}"
+else
+  s3_destination="cellxgene-wmg-${DEST_ENV}"
+fi
+echo -n $latest_snapshot | aws s3 cp - s3://${s3_destination}/latest_snapshot_run
+aws s3 cp --recursive s3://cellxgene-wmg-${SRC_ENV}/${latest_snapshot} s3://${s3_destination}/${latest_snapshot}
```
danieljhegeman, line 25: The above adds the mirroring of the latest snapshot for _all_ invocations, including the regular prod->(dev|staging) mirror.

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,80 @@ set -e
...
+  for uri in "${uris[@]}"; do
+    bucket=$(sed -E 's/s3:\/\/([^\/]+).*/\1/' <<< $uri)
+    if [[ ! -z `grep 'hosted-cellxgene' <<< $bucket` ]]; then
+      rdev_bucket_suffix="cellxgene"
+    elif [[ ! -z `grep 'corpora-data' <<< $bucket` ]]; then
+      rdev_bucket_suffix="artifacts"
+    fi
+    key=$(sed -E 's/s3:\/\/([^\/]+)\/(.*)/\2/' <<< $uri)
+    [[ "${key: -1}" == "/" ]] && s3_cmd=$S3_SYNC_CMD || s3_cmd=$S3_COPY_CMD  # use 'sync' if recursive, 'cp' otherwise
+    $s3_cmd $uri s3://env-rdev-${rdev_bucket_suffix}/${STACK}/${key} || echo "$uri does not exist -- skipping..."
```
atolopko-czi, line 51: github syntax highlighting stops here--is there a syntax error?

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,80 @@ set -e
...
+else
+  s3_destination="cellxgene-wmg-${DEST_ENV}"
+fi
+echo -n $latest_snapshot | aws s3 cp - s3://${s3_destination}/latest_snapshot_run
+$S3_SYNC_CMD s3://cellxgene-wmg-${SRC_ENV}/${latest_snapshot} s3://${s3_destination}/${latest_snapshot}
+
+
+echo Mirroring S3 Dataset data from $SRC_ENV to $DEST_ENV...
+
+if [[ $DEST_ENV == 'rdev' ]]; then
```
atolopko-czi, line 34: comment that rdev s3 copying support collection subsetting

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,80 @@ set -e
...
+      rdev_bucket_suffix="cellxgene"
+    elif [[ ! -z `grep 'corpora-data' <<< $bucket` ]]; then
+      rdev_bucket_suffix="artifacts"
+    fi
+    key=$(sed -E 's/s3:\/\/([^\/]+)\/(.*)/\2/' <<< $uri)
+    [[ "${key: -1}" == "/" ]] && s3_cmd=$S3_SYNC_CMD || s3_cmd=$S3_COPY_CMD  # use 'sync' if recursive, 'cp' otherwise
+    $s3_cmd $uri s3://env-rdev-${rdev_bucket_suffix}/${STACK}/${key} || echo "$uri does not exist -- skipping..."
+  done
+
+  # Copy public assets
```
atolopko-czi, line 54: a bigger issue than mirroring: since rdev doesn't support cloudfront proxying to public assets, once we copy them there, can they even be accessed publicly? What is actually needed for testing purposes? If public assets can't be used testing, maybe we don't even need to support mirroring them.

if it does make sense to mirror, should it be optional?

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,80 @@ set -e
...
+  DEPLOYMENT_STAGE=$DEPLOYMENT_STAGE make db/tunnel/down  # db access no longer needed
+  exts=("rds" "h5ad")
+  for dv_id in "${dataset_version_ids[@]}"; do
+    for ext in "${exts[@]}"; do
+      uri="s3://dataset-assets-public-${DEPLOYMENT_STAGE}/${dv_id}.${ext}"
+      $S3_COPY_CMD $uri s3://env-rdev-datasets/${STACK}/${dv_id}.${ext} || echo "$uri does not exist -- skipping..."
+    done
+  done
+else
+  # TODO: Add --delete once we confirm that is no data in the folders
```
atolopko-czi, line 66:
```suggestion
  # TODO: Add --delete once we confirm that there is no data in the folders
```
if I may correct my own, original error...

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,80 @@ set -e
...
+  ACCOUNT_ID=`aws sts get-caller-identity | jq -r '.Arn' | grep -o '[0-9]*'`  # First get AWS account id for dev account
+  SYNC_ROLE_CREDENTIALS=`aws sts assume-role --role-arn arn:aws:iam::${ACCOUNT_ID}:role/sync-datasets-dev \
+                                             --role-session-name sync-datasets \
+                         | jq -r '.Credentials'`
+  # Execute sync for datasets bucket with sync role credentials
+  AWS_SESSION_TOKEN=`echo $SYNC_ROLE_CREDENTIALS | jq -r '.SessionToken'` \
+  AWS_SECRET_ACCESS_KEY=`echo $SYNC_ROLE_CREDENTIALS | jq -r '.SecretAccessKey'` \
+  AWS_ACCESS_KEY_ID=`echo $SYNC_ROLE_CREDENTIALS | jq -r '.AccessKeyId'` \
+  $S3_SYNC_CMD s3://dataset-assets-public-${SRC_ENV}/ s3://dataset-assets-public-${DEST_ENV}/
+  set +x
```
atolopko-czi, line 89: assuming the else block is just a wholesale move, so not reviewed

backend/scripts/set_src_dest_envs.sh
```diff
@@ -3,20 +3,34 @@
...
     if [[ -z $NO_PROMPT ]]; then
         read -n 1 -p "ATTENTION: You are mirroring the dev env to itself. Useful for testing the mirroring operation, but otherwise useless. Proceed? (Y/n) " ANS
         [[ $ANS == 'Y' ]] || exit 1
     fi
+# allow dev->rdev for testing
+elif [[ "$SRC_ENV" == dev && "$DEST_ENV" == rdev ]]; then
+    if [[ -z $NO_PROMPT ]]; then
+        read -n 1 -p "ATTENTION: You are mirroring the dev env to rdev. Useful for testing fixes in rdev for data-related bugs found in dev, but otherwise useless. Proceed? (Y/n) " ANS
+        [[ $ANS == 'Y' ]] || exit 1
+    fi
```
atolopko-czi, line 25: same as earlier comment, mirroring _from_ dev not useful post-CI/CD

atolopko-czi (changes requested):
Overall, LGTM.  Added some questions and minor comments. 

Worth adding options (flags) to allow skip mirroring for WMG and/or Portal assets? Not sure how long it takes, and so whether it's worth doing.

backend/scripts/mirror_rds_data.sh
danieljhegeman:
```pg_restore: error: could not execute query: ERROR:  cannot drop constraint Dataset_pkey on table persistence_schema."Dataset" because other objects depend on it
DETAIL:  constraint DatasetVersion_id_fkey on table persistence_schema."DatasetVersion" depends on index persistence_schema."Dataset_pkey"
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
Command was: ALTER TABLE IF EXISTS ONLY persistence_schema."Dataset" DROP CONSTRAINT IF EXISTS "Dataset_pkey";```

Comment (atolopko-czi) [reply backend/scripts/mirror_rds_data.sh:60]: Thanks. The ordering of drop statements must be wrong on rdev, but no clue why that is. Or perhaps an earlier, required drop of the `Dataset_pkey` constraint is failing for some reason.

It could be worth adding the error as a comment for our future selves to resume debugging this.

File: backend/scripts/mirror_s3_data.sh (hunk 6 above)
Comment (danieljhegeman) [reply backend/scripts/mirror_s3_data.sh:69]: >since rdev doesn't support cloudfront proxying to public assets, once we copy them there, can they even be accessed publicly?

This is something @Bento007 is looking into in order to mimic prod behavior.

Comment (danieljhegeman) [reply backend/scripts/mirror_s3_data.sh:69]: >What is actually needed for testing purposes?

For Data Portal, the file must exist at that location; for rdevs with an Explorer component, then the file must be a legitimate h5ad. For these reasons, I've moved to simply using the (small) sample schema 3.0.0 Dataset for all cases. For every real asset in the prod env, an instance of the sample schema 3.0.0 Dataset will be copied to the appropriate s3 location in the rdev s3 path.

File: backend/scripts/mirror_s3_data.sh [hunk 10]
```diff
@@ -11,30 +11,80 @@ set -e
...
+  dataset_version_ids=($(PGOPTIONS='-csearch_path=persistence_schema' PGPASSWORD=${DB_PW} psql --dbname corpora_${DEPLOYMENT_STAGE} --username corpora_${DEPLOYMENT_STAGE} --host 0.0.0.0 --csv --tuples-only -c "$dv_id_query"))
+  DEPLOYMENT_STAGE=$DEPLOYMENT_STAGE make db/tunnel/down  # db access no longer needed
+  exts=("rds" "h5ad")
+  for dv_id in "${dataset_version_ids[@]}"; do
+    for ext in "${exts[@]}"; do
+      uri="s3://dataset-assets-public-${DEPLOYMENT_STAGE}/${dv_id}.${ext}"
+      $S3_COPY_CMD $uri s3://env-rdev-datasets/${STACK}/${dv_id}.${ext} || echo "$uri does not exist -- skipping..."
+    done
+  done
+else
```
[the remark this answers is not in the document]
danieljhegeman: added comment

backend/scripts/set_src_dest_envs.sh
danieljhegeman: Agreed -- shall we remove this only once we actually get rid of dev?

backend/scripts/mirror_s3_data.sh
danieljhegeman: It has some changes: `parallel`izing the sync cmds 🥲

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,109 @@ set -e
...
+  # jeopardy of losing anything permanently). This would make the
+  # operation more destructive, of course!
+
+  PARALLEL_CMD="parallel --ungroup --jobs 16"  # '--ungroup' permits stdout to stream continuously from all jobs
+
+  hex_chars="0 1 2 3 4 5 6 7 8 9 a b c d e f"  # Permits running 16 concurrent aws processes -- each one syncs only uuids starting with its hex char
+  set -x
+  echo -e "\n\n*** The following parallel aws s3 sync commands may appear to hang, but they are working. Patience is a virtue. ***\n\n"
+  $PARALLEL_CMD $S3_SYNC_CMD s3://corpora-data-${SRC_ENV}/ s3://corpora-data-${DEST_ENV}/ --exclude "'*'" --include "'{}*'" --exclude "'*loom'" ::: $hex_chars
+  $PARALLEL_CMD $S3_SYNC_CMD s3://hosted-cellxgene-${SRC_ENV}/ s3://hosted-cellxgene-${DEST_ENV}/ --exclude "'*'" --include "'{}*'" ::: $hex_chars
```
danieljhegeman, line 106: it's really great that github renders all of this ☝️  as though it is a comment block 😭

backend/scripts/mirror_s3_data.sh
```diff
@@ -11,30 +11,119 @@ set -e
...
+  for uri in "${uris[@]}"; do
+    bucket=$(sed -E 's/s3:\/\/([^\/]+).*/\1/' <<< $uri)
+    if [[ -n `grep 'hosted-cellxgene' <<< $bucket` ]]; then
+      rdev_bucket_suffix="cellxgene"
+    elif [[ -n `grep 'corpora-data' <<< $bucket` ]]; then
+      rdev_bucket_suffix="artifacts"
+    fi
+    key_dir=$(sed -E 's/s3:\/\/([^\/]+)\/(.*)/\2/' <<< $uri)
+    if [[ -n "$DATA" ]]; then
+      sem -j+0 $S3_SYNC_CMD s3://${bucket}/${key_dir} s3://env-rdev-${rdev_bucket_suffix}/${STACK}/${key_dir}
```
danieljhegeman, line 75: `sem` gives us concurrency, managing a queue of jobs for a fixed (default of 12 for all of us on standard macbooks) number of processes.

atolopko-czi: ah, `GNU sem is an alias for GNU parallel --semaphore.`

danieljhegeman: oh cool. I knew it use parallel under the hood but didn't realize you could invoke the paradigm from parallel itself

backend/scripts/set_src_dest_envs.sh
atolopko-czi: if you think we'll remember...

atolopko-czi:
One concern with the `s3://env-rdev-dataportal/s3/` test fixture data is that it's schema-dependent, and we'll need to remember to update these fixtures when schema's increment. Another concern is that it's always copying data for a specific schema version, and this means there will never be mixed schema data available to rdev stacks, even though the RDS data will have rows for mixed schema datasets. 

The user needs to be aware of these constraints at minimum, and ideally the user should be able to copy a subset of a real collection's assets if needed. Is that supported?

danieljhegeman:
> One concern with the `s3://env-rdev-dataportal/s3/` test fixture data is that it's schema-dependent, and we'll need to remember to update these fixtures when schema's increment. Another concern is that it's always copying data for a specific schema version, and this means there will never be mixed schema data available to rdev stacks, even though the RDS data will have rows for mixed schema datasets.
> 
> The user needs to be aware of these constraints at minimum, and ideally the user should be able to copy a subset of a real collection's assets if needed. Is that supported?

@atolopko-czi  All valid concerns; if the user passes any non-null value for `DATA` when invoking the top-level `make` command, then the _real_ assets are copied instead of the sample data.

Totally open to changing the default to be real, and allowing a `SAMPLE_DATA` flag var instead...?

atolopko-czi:
> @atolopko-czi All valid concerns; if the user passes any non-null value for `DATA` when invoking the top-level `make` command, then the _real_ assets are copied instead of the sample data.

Perfect, brilliant!
 
> Totally open to changing the default to be real, and allowing a `SAMPLE_DATA` flag var instead...?

I think we can leave the default and change it as engineers provide feedback.

atolopko-czi (approved):
Regretfully I haven't actually _tested_ the scripts myself, but from a code read and functionality standpoint, LGTM!

PR #5663, opened 2023-08-31
chore: skip failing functional test

atarashansky:
## Reason for Change

- This test is failing as deployed_version is not working. Skipping it to unblock deployments. [It should be re-added once the feature is fixed](https://github.com/chanzuckerberg/single-cell-data-portal/issues/5664).


atolopko-czi (approved):
Makes sense!

PR #5680, opened 2023-08-31
feat(cellguide): Use SDS components for organ and organism dropdown

prathapsridharan:
## Reason for Change

- #5670 

## Changes

- change the global organism and organ filter dropdowns to use SDS components

## Testing steps

- E2E tests pass
- manual testing on localhost

### DIFF
--- a/frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
+++ b/frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
@@ -34,18 +34,25 @@ import {
   CELL_GUIDE_CARD_HEADER_TAG,
   CELL_GUIDE_CARD_SYNONYMS,
 } from "src/views/CellGuide/components/CellGuideCard/constants";
-import DropdownSelect from "./components/common/DropdownSelect";
-import { SelectChangeEvent } from "@mui/material/Select";
 import { useOrganAndOrganismFilterListForCelltype } from "./components/MarkerGeneTables/hooks/common";
 import {
   ALL_TISSUES,
   NO_ORGAN_ID,
 } from "./components/MarkerGeneTables/constants";
+import {
+  DefaultDropdownMenuOption,
+  Dropdown,
+  InputDropdownProps,
+} from "@czi-sds/components";
 
 const RIGHT_SIDEBAR_WIDTH_PX = 400;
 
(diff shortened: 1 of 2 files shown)


atarashansky (approved):
LGTM!

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -101,20 +108,33 @@ export default function CellGuideCard({
   const { organismsList, organsMap } =
     useOrganAndOrganismFilterListForCelltype(cellTypeId);
 
-  const uniqueOrgans = Array.from(organsMap.keys());
+  const sdsOrganismsList = organismsList.map((organism) => ({
```
tihuan, line 111: let's memoize this, so we don't create a new array every time, which is a performance issue, since `[] === [] // false`

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -101,20 +108,33 @@ export default function CellGuideCard({
...
+    name: organism,
+  }));
 
-  const [selectedOrgan, setSelectedOrgan] = useState(ALL_TISSUES);
+  const sdsOrgansList = Array.from(organsMap.keys()).map((organ) => ({
```
tihuan, line 115: same here!

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -101,20 +108,33 @@ export default function CellGuideCard({
...
+    name: organ,
+  }));
+
+  const [selectedOrgan, setSelectedOrgan] = useState<DefaultDropdownMenuOption>(
+    {
```
tihuan, line 120: to ensure object equality and keeping the code DRY, instead of creating a new object for all tissue as default, let's just pick it from sdsOrgansList like so: `sdsOrgansList.find(organ => organ.name === ALL_TISSUES)`:

```ts
const [selectedOrgan, setSelectedOrgan] = useState<DefaultDropdownMenuOption>(
    sdsOrgansList.find(
      (organ) => organ.name === ALL_TISSUES
    ) as DefaultDropdownMenuOption
  );
```

tihuan (commented):
Looking great! Just some small perf optimization recommendations 😄 PTAL thank you!

PR #5699, opened 2023-09-05
feat: Make the top nav bar for cell card should be sticky

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5672

## Changes

makes the top nav bar sticky and also adds a gradient at the bottom

## Testing steps

https://github.com/chanzuckerberg/single-cell-data-portal/assets/5653616/8788da32-38a0-4825-96ce-de0824322c5d

## Notes for Reviewer
n/a


frontend/src/views/CellGuide/components/CellGuideCard/style.ts
```diff
@@ -28,16 +28,13 @@ export const CellGuideView = styled.div`
...
   ${(props: CellGuideViewProps) => {
     const { skinnyMode } = props;
     const spaces = getSpaces(props);
     const space = skinnyMode ? spaces?.xl : spaces?.xxl;
 
     return `
-    padding: ${TOP_PADDING_PX}px ${space}px 0px
-      ${space}px;
-
+    padding: 0px ${space}px 0px ${space}px;
```
joyceyan: i removed the top padding from `CellGuideView` and into `CellGuideCardHeader`, which will make sure that the padding sticks with the header as you scroll down but still shows up in the static view

frontend/src/views/CellGuide/components/CellGuideCard/style.ts
```diff
@@ -60,14 +61,15 @@ export const Wrapper = styled.div<CellGuideViewProps>`
   display: flex;
   flex-direction: column;
   align-self: stretch;
-  overflow-x: hidden;
```
joyceyan, line 63: i'm assuming that this was added for mobile view - so i'm changing the logic to just conditionally hide it depending on if we're in skinny mode or not. in order for the child `CellGuideCardHeader` to stay sticky within its parent container `Wrapper`, `Wrapper` needs to be a scrollable ancestor. removing the overflow in desktop view ensures that `Wrapper` stays scrollable

atarashansky (approved):

Issue #5439, opened 2023-08-10
Support high-resource testing on RDev [CI/CD]


atolopko-czi:
For RDev to more fully support all modes of testing (ad hoc, automated, QA testing), it must be provide compute resources that are capable of working with prod-equivalent data and with performance that is closer to our deployment environments. Currently, API calls and selected data can causes OOM errors or cause systems to operate too slowly, making RDev insufficient for some testing scenarios.

* ECS backend instances need to be up-scaled appropriately. 
* Batch instances need to be up-scaled appropriately

Implementation notes:
* Determine the appropriate levels of scaling needed
* Assess cost of upscaling to the desired levels. **Once we have estimates, let's get approval from Arathi before increasing AWS costs.**
* As needed, investigate the use of ECS soft and hard memory limits to prevent over-provisioning of RDev compute resources.
* As needed, work with the Happy Team to understand if RDev auto-scaling is possible.
* Document how RDev cost can be monitored so cost impact can be effectively managed by engineering team.

Tasks:
1. Once all the tests are in place, run the tests in current RDEV config to see if tests pass
2. If not, figure out how much to scale up given estimated parallel runs (includes ad-hocs and CI/CD provisioned stacks), and assess costs and get Arathi's approval budget

tihuan:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5439?planning-poker) with Zenhub @atolopko-czi @Bento007 @danieljhegeman @ebezzi

tihuan:
team agreed with 3

Bento007:
Here is the current configuration for rdev
https://github.com/chanzuckerberg/single-cell-infra/blob/d5783679da0611e020eef727e1fa7730a737c783/terraform/envs/dev/dp-rdev/main.tf#L71-L73

It uses a larger machine than dev 
https://github.com/chanzuckerberg/single-cell-infra/blob/d5783679da0611e020eef727e1fa7730a737c783/terraform/envs/dev/dp-dev/main.tf#L53-L55

Bento007:
Increasing the size of the ECS cluster will not significantly increase the cost of development. We could double it and it would only be an additional $200 a month. We are also configured to use 1-2 servers for happy and are only using 1. We have plenty of room to grow. There is no need to change the current configuration for running the data portal server.

[...]

PR #5702, opened 2023-09-06
fix(happy): prevent orphaned ECS tasks

Bento007:
## Reason for Change

When cleaning up an rdev environment sometime the ECS task would be left running. This will get expensive as we mmake more rdev environments.

This work is related to https://github.com/chanzuckerberg/single-cell-data-portal/issues/5439
## Changes

- add enable_dynamo_locking to prevent race conditions when cleaning up rdev environments

## Testing steps

- create an rdev and break it down.

### DIFF
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -80,5 +80,8 @@
   "tasks": {
     "migrate": ["migrate_db_task_definition_arn"],
     "delete": ["delete_db_task_definition_arn"]
+  },
+  "features": {
+    "enable_dynamo_locking": true
   }
 }


.happy/config.json
```diff
@@ -80,5 +80,8 @@
   "tasks": {
     "migrate": ["migrate_db_task_definition_arn"],
     "delete": ["delete_db_task_definition_arn"]
+  },
+  "features": {
+    "enable_dynamo_locking": true
```
danieljhegeman, line 85: The PR states that this is intended to solve the problem of not correctly cleaning up ECS tasks when tearing down an rdev instance. But what does this enabling this attribute actually do? What's the mechanism?

Bento007: This locks the db that stores information about each happy environment when someone is writing or deleting. This is to prevent a race condition which can causes the db to out of sync with reality.

danieljhegeman: Good to know

danieljhegeman (approved):

PR #5705, opened 2023-09-07
feat(dp): add primary cell count to dataset metadata

atarashansky:
## Reason for Change

 - [this is part of the work to automate the hero numbers on the landing page](https://docs.google.com/document/d/1UIw-nyNmRF6ob46WLOnxXhyT46J-jcEglzJ9bxGV2BU/edit).

## Changes

- add primary cell count field to curation and data portal apis
- update extract metadata with primary cell count

## Testing steps

- tests pass
- @brianraymor to check [rdev](https://primary-cell-count-backend.rdev.single-cell.czi.technology/)

### DIFF
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -222,7 +222,6 @@ def reshape_dataset_for_curation_api(
 
     ds["dataset_id"] = dataset_version.dataset_id.id
     ds["dataset_version_id"] = dataset_version.version_id.id
-    ds["primary_cell_count"] = dataset_version.metadata.primary_cell_count
     # Get none preview specific dataset fields
     if not preview:
         ds["assets"] = extract_dataset_assets(dataset_version)
@@ -305,6 +304,7 @@ class EntityColumns:
         "development_stage",
         "cell_type",
         "cell_count",
+        "primary_cell_count",
         "x_approximate_distribution",
         "batch_condition",
         "mean_genes_per_cell",
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -222,6 +222,7 @@ def reshape_dataset_for_curation_api(
 
     ds["dataset_id"] = dataset_version.dataset_id.id
     ds["dataset_version_id"] = dataset_version.version_id.id
+    ds["primary_cell_count"] = dataset_version.metadata.primary_cell_count
     # Get none preview specific dataset fields
     if not preview:
         ds["assets"] = extract_dataset_assets(dataset_version)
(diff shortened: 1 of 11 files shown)


backend/curation/api/v1/curation/collections/common.py
```diff
@@ -222,6 +222,7 @@ def reshape_dataset_for_curation_api(
 
     ds["dataset_id"] = dataset_version.dataset_id.id
     ds["dataset_version_id"] = dataset_version.version_id.id
+    ds["primary_cell_count"] = dataset_version.metadata.primary_cell_count
```
nayib-jose-gloria, line 225: remove, metadata fields are already included as part of this block: https://github.com/chanzuckerberg/single-cell-data-portal/pull/5705/files#diff-26c746d9f1e6e25a74bd8918b8773e6352c44219da15aeb38662790fbc721608R210-R221

nayib-jose-gloria: Add 'primary_cell_count' to https://github.com/chanzuckerberg/single-cell-data-portal/blob/631a0dfde8672d24272c9c09b169a04d11e9002c/backend/curation/api/v1/curation/collections/common.py#L299 instead

brianraymor:
For **any** changes to the public Discover API, I would like to review the swagger endpoint changes and naming prior to release.

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -304,6 +304,7 @@ class EntityColumns:
         "development_stage",
         "cell_type",
         "cell_count",
+        "primary_cell_count",
```
nayib-jose-gloria, line 307: I think we actually need to add a None default value for getattr [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5705/files#diff-26c746d9f1e6e25a74bd8918b8773e6352c44219da15aeb38662790fbc721608R216) to support the missing primary cell counts while we haven't backfilled

backend/layers/common/entities.py
```diff
@@ -172,6 +172,7 @@ class DatasetMetadata:
     development_stage: List[OntologyTermId]
     cell_type: List[OntologyTermId]
     cell_count: int
+    primary_cell_count: int
```
nayib-jose-gloria, line 175: Last thing to support the hybrid state where this isn't backfilled--I believe this type hint needs to be wrapped as Optional[int] and set to default to = None (i.e. `primary_cell_count: Optional[int] = None`)

nayib-jose-gloria (approved):
To adhere to our new CI/CD commitments, before merging, make sure this properly loads our seed data in an RDEV for both curation and portal API + properly processes and returns the expected primary cell count for a test dataset uploaded to that rdev.

PR #5707, opened 2023-09-07
feat: use the string "tissue agnostic" instead of "all tissues"

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5671

## Changes

updates only the view layer mapping `organsMap` to `sdsOrgansList`, which we use for display in the dropdown. also adds some logic to `handleChangeOrgan` to make sure we're mapping back from "tissue agnostic" to "all tissues" for setting the selected organ id + tracking the analytics event

## Testing steps

checked T cell which actually has canonical/computational markers now!

### DIFF
--- a/frontend/src/views/CellGuide/components/CellGuideCard/components/MarkerGeneTables/constants.ts
+++ b/frontend/src/views/CellGuide/components/CellGuideCard/components/MarkerGeneTables/constants.ts
@@ -2,6 +2,8 @@ export const HOMO_SAPIENS = "Homo sapiens";
 
 export const ALL_TISSUES = "All Tissues";
 
+export const TISSUE_AGNOSTIC = "Tissue Agnostic";
+
 export const NO_ORGAN_ID = "";
 
 export const CELL_GUIDE_CARD_CANONICAL_MARKER_GENES_TABLE =
--- a/frontend/src/views/CellGuide/components/CellGuideCard/components/MarkerGeneTables/hooks/canonical_markers.ts
+++ b/frontend/src/views/CellGuide/components/CellGuideCard/components/MarkerGeneTables/hooks/canonical_markers.ts
@@ -3,7 +3,12 @@ import {
   CanonicalMarkersQueryResponse,
   CanonicalMarkersQueryResponseEntry,
 } from "src/common/queries/cellGuide";
-import { ALL_TISSUES, HOMO_SAPIENS } from "../constants";
+import {
+  ALL_TISSUES,
+  HOMO_SAPIENS,
+  NO_ORGAN_ID,
+  TISSUE_AGNOSTIC,
+} from "../constants";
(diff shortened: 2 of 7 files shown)


frontend/src/views/CellGuide/components/CellGuideCard/components/MarkerGeneTables/constants.ts
```diff
@@ -2,6 +2,8 @@ export const HOMO_SAPIENS = "Homo sapiens";
 
 export const ALL_TISSUES = "All Tissues";
 
+export const TISSUE_AGNOSTIC = "Tissue Agnostic";
```
joyceyan, line 5: not sure if this is the best place to throw this constant - i can move it somewhere else

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -166,8 +167,11 @@ export default function CellGuideCard({
   const handleChangeOrgan = (option: DefaultDropdownMenuOption | null) => {
     if (!option) return;
     setSelectedOrgan(option);
-    setSelectedOrganId(organsMap.get(option.name) ?? "");
-    track(EVENTS.CG_SELECT_TISSUE, { tissue: option.name });
+    const optionName =
+      option.name === TISSUE_AGNOSTIC ? ALL_TISSUES : option.name;
+    setSelectedOrganId(organsMap.get(optionName) ?? "");
+    // Continue tracking the analytics event as All Tissues
+    track(EVENTS.CG_SELECT_TISSUE, { tissue: optionName });
```
joyceyan, line 174: i'm assuming it's better to continue tracking the analytics events as ALL_TISSUES

frontend/src/views/CellGuide/components/CellGuideCard/components/MarkerGeneTables/hooks/canonical_markers.ts
```diff
@@ -71,17 +76,15 @@ function _getReferenceData(
...
 }: {
   markerGene: CanonicalMarkersQueryResponseEntry;
   allTissuesLabelToIdMap: Map<string, string>;
-  selectedOrganLabel: string;
   selectedOrganId: string;
 }): boolean {
   // There are marker genes tissues labeled as "All Tissues"
   // so select them only when selectedOrganLabel is "All Tissues"
-  if (selectedOrganLabel === ALL_TISSUES) {
+  if (selectedOrganId === NO_ORGAN_ID) {
```
joyceyan, line 87: i'm using selected organ id instead of selected organ label because i think it's a bit better to use the id (which represents the actual entity) instead of a label (which is just for view purposes)

atarashansky (approved):
Nice - I like removing the dependency on the tissue label in the marker gene table hooks. GJ! LGTM!

Issue #5487, opened 2023-08-14
Create Github Release for each deployment to prod


Bento007:
After a successful deployment to **prod**, generate a new release in github with a change log (e.g. Github "Releases"). The change long will consist of the commits since last prod deployment. This should be an automated process.

Update deployment checklist in on-call log.

Bento007:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5487?planning-poker) with Zenhub @ebezzi @tihuan

tihuan:
P1 sounds good to me!

PR #5708, opened 2023-09-07
feat: automate github release upon prod deploy

nayib-jose-gloria:
## Reason for Change

- #5487 

## Changes

- add step in deploy-happy-stack to release a github version + autogenerate release notes compared to the previous release, only for prod deploys.

## Testing steps

- Testing with temp github workflow steps in this branch to trigger releases during push-tests (steps and test releases to be removed before merging)


.github/workflows/deploy-happy-stack.yml
```diff
@@ -95,6 +95,41 @@ jobs:
...
+                "method": "sort",
+                "filter": {
+                  "pattern": "sha-(.+)",
+                  "flags": "gu"
+                }
+              }
+            }
+      - name: Produce Github Release for Prod Deployments
+        uses: softprops/action-gh-release@v1
+        if: steps.tag_create.outputs.tag_exists == 'false' && github.event.deployment.environment == 'prod'
```
Bento007, line 128: Is it intended that we run ` github-tag-and-release` for all deployments? 

`if: github.event.deployment.environment == 'prod'` could be moved to the [job level](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#overview) so we only run this for the prod environment.

nayib-jose-gloria: good catch, this was a relic from initially putting this step within the deployment step. Will fix

Bento007 (approved):

Issue #5526, opened 2023-08-16
CxG Front Page - Newsletter Sticky


niknak33:
Currently, this is a bug - when a user closes the newsletter banner, the frontpage footer with the newsletter is removed and shows a large empty block. Please see image below:

Solution - is to have the newsletter banner be a sticky format so it's more visibly apparent for users and resolve the bug above. @hthomas-czi 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcUlOIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--c01ffa2293fbcc23d64d341604384ec814c0aa94/image.png)

Update - We will use the same sticky banner that is present on other pages i.e. CellGuide and GE, remove the survey links from the banner and roadmap links.

tihuan:
@hthomas-czi 
Hmm the visible gap seems to be part of the margin for the timeline section, so if we use the Gene Expression / Cell Guide banner, the gap will still be there. What is the ideal end state we want to have?  Is it just less margin bottom for the timeline?

20px margin bottom with banner:
<img width="1383" alt="Screenshot 2023-08-31 at 11 55 16 AM" src="https://github.com/chanzuckerberg/single-cell-data-portal/assets/6309723/3f552bb6-ea92-460f-8e7a-9fc60a9c6430">

20px margin bottom with banner:
<img width="865" alt="Screenshot 2023-08-31 at 11 57 22 AM" src="https://github.com/chanzuckerberg/single-cell-data-portal/assets/6309723/4a2a197c-9e3d-43e5-b1fc-95ec8a6b366f">

Current timeline section margin in orange:
<img width="1722" alt="Screenshot 2023-08-31 at 11 53 32 AM" src="https://github.com/chanzuckerberg/single-cell-data-portal/assets/6309723/9a0adef1-e8bf-49ec-84b2-d0db833cc352">

Current with banner:
<img width="1728" alt="Screenshot 2023-08-31 at 11 49 23 AM" src="https://github.com/chanzuckerberg/single-cell-data-portal/assets/6309723/58ef42c0-a86b-4f5d-bafe-6214d776182d">

Current without banner:
[...]

PR #5711, opened 2023-09-07
feat: 5526 Add newsletter banner in homepage

tihuan:
## Reason for Change

- #5526 

## Changes

1. This turned out to be a bigger change, because the homepage is the first time we need to support loading two BottomBanner on the same page
2. Add ` isHubSpotReady` and ` onHubSpotReady` on `BottomBanner` to optionally coordinate when HubSpot script is loaded, since `<Script onReady />` only notifies the first `<BottomBanner />` instance with the callback
3. Add `id` to distinguish instances on the same page
4. Use ` formInstanceId` on ` hbspt.forms.create` for HubSpot object to know which instance we're targeting
5. Add `formContainerQueryId` to target different banner instances
6. 

## Testing steps


tihuan:
actually found a bug - will fix tomorrow!

tihuan:
bug fixed!

The bug was that the observer wasn't connecting to the form, because I accidentally set isHubSpotReady back to `false` 🤦

frontend/src/components/BottomBanner/index.tsx
```diff
@@ -75,10 +111,13 @@ export default function BottomBanner({
...
 
-  // eslint-disable-next-line sonarjs/cognitive-complexity
+  /**
+   * Reads a query parameter from the URL to auto open the newsletter signup modal
+   * Allows sharing of a URL to lead directly to the newsletter signup, specifically for conferences
+   */
   useEffect(() => {
-    // Reads a query parameter from the URL to auto open the newsletter signup modal
-    // Allows sharing of a URL to lead directly to the newsletter signup, specifically for conferences
+    if (!isHubSpotReady) return;
```
tihuan, line 119: guard clause for readability - a long block nested in a condition caught me off guard

tihuan (commented):
Ready for review PTAL thank you!

seve (approved):

kaloster (approved):

Issue #5713, opened 2023-09-08
tech(dp): remove primary_cell_count from API and add a new endpoint serving the hero numbers


atarashansky:
[This PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5705/files) added primary_cell_count to the API, with the intent to have the FE consume the datasets/index endpoint and aggregate the cell counts to automate the hero numbers on the landing page.

A cleaner approach would be to set up a new endpoint that serves the hero numbers for the frontend to consume directly.

prathapsridharan:
@dsadgat @atarashansky - 

1. Is `primary_cell_count` the only thing that is to be removed from the `dataset` object? That is, the other hero numbers (number of distinct cell-types and number of datasets) will still be retrieved from `v1/datasets/index` and `v1/user-datasets/index`?
2. If so, what is the motivation to have a separate `GET` API call for `primary_cell_count`? Are there many other sites in the code where `primary_cell_count` is needed but we don't want to hit `v1/datasets/index` and `v1/user-datasets/index`? Logically it seems `primary_cell_count` would belong in a `DatasetMetadata` [object](https://github.com/chanzuckerberg/single-cell-data-portal/blob/2e4b3ad654e12473d3c237b3608f68c4327755a3/backend/layers/common/entities.py#L193). Look at, for example, how `cell_count` is [retrieved](https://github.com/chanzuckerberg/single-cell-data-portal/blob/2e4b3ad654e12473d3c237b3608f68c4327755a3/backend/portal/api/portal_api.py#L186) from a metadata object. This promotes tight cohesion but there might good reasons to break tight cohesion and I would be interested to know what they are.
3. By putting `primary_cell_count` in a separate API function, wouldn't we be duplicating code used to retrieve other metadata like `cell-count`? It seems `primary_cell_count` is [retrieved by reading the h5ad into memory and retrieving the count](https://github.com/chanzuckerberg/single-cell-data-portal/blob/2e4b3ad654e12473d3c237b3608f68c4327755a3/backend/layers/processing/process_validate.py#L194) and a separate API endpoint would need to first retrieve all datasets, then download and open each h5ad file to retrieve `primary_cell_count`? This almost certainly makes the "hero API" call less performant.

[...]

PR #5714, opened 2023-09-08
chore: remove primary cell count from API

atarashansky:
## Reason for Change

- #5713 
- Follow-up work will need to add an endpoint to compute the hero numbers.

## Changes

- add
- remove
- modify

## Testing steps

- Uploaded example dataset to rdev[LINK] and confirmed that `primary_cell_count` is not present any of the DP or Discover API responses.

### DIFF
--- a/backend/curation/api/curation-api.yml
+++ b/backend/curation/api/curation-api.yml
@@ -977,10 +977,6 @@ components:
           description: The number of cells in the Dataset.
           nullable: true
           type: integer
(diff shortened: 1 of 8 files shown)


nayib-jose-gloria (approved):

atarashansky:
Closing now in favor of speeding up delivery of the feature (see #5718 ). This can be re-opened once we address #5713

PR #5723, opened 2023-09-08
fix: WMG gene searchbar

tihuan:
## Reason for Change

1. We discovered today that the search bar needs a little more fix, since the marker gene right panel was blocked by the search bar

## Changes

1. Add `background-color: white` to RightSideBar component - this caused some overlap issue, since the background is transparent otherwise
2. Add z-index on the RightSideBar via `styled`, so we don't pollute the common RightSideBar component 

## Testing steps

BEFORE: Gene Search Bar is in front of the Marker Gene panel

Screenshot 2023-09-08 at 11 41 07 AM

AFTER: Gene Search Bar is behind the Marker Gene panel as expected

Screenshot 2023-09-08 at 1 06 01 PM

https://github.com/chanzuckerberg/single-cell-data-portal/assets/6309723/880ff12d-82c8-42b4-9113-1b0445adfc30

Test steps:

1. Go to https://localhost:3000/gene-expression
3. Expand tissue `Abdomen`
4. Click fist cell type's info icon to expand marker gene panel
5. Narrow viewport enough for the right side bar to overlap the gene search bar. The gene search bar should be behind the right side bar
6. Click "+ Add to Dot Plot" to add all marker genes to the heatmap
7. Scroll the heatmap horizontally. The gene search bar should stay fixed and not scrollable


frontend/src/components/common/SideBar/style.ts
```diff
@@ -36,6 +36,7 @@ export const SideBar = styled.div<Props>`
   width: ${(props) => `${props.sideBarWidth}px`};
   grid-area: ${(props) =>
     props.position === Position.LEFT ? "leftsidebar" : "rightsidebar"};
+  background-color: white;
```
tihuan, line 39: I think it's worth adding this in the common SideBar, so all call sites get it!

seve (approved):

kaloster (approved):

PR #5725, opened 2023-09-08
fix: secondary filters in gene expression

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5716

## Changes

adds a new field filteredCellTypeIds to the state, which we then pass in to WMG under `cell_type_ontology_ids`

## Testing steps

https://github.com/chanzuckerberg/single-cell-data-portal/assets/5653616/6327a7a0-ccf3-4c24-b470-546fef7b9053

### DIFF
--- a/frontend/src/common/queries/wheresMyGene.ts
+++ b/frontend/src/common/queries/wheresMyGene.ts
@@ -1115,6 +1115,7 @@ function useWMGFiltersQueryRequestBody(
     ethnicities,
     publications,
     sexes,
+    filteredCellTypeIds,
   ]);
 }
 
--- a/frontend/src/common/queries/wheresMyGene.ts
+++ b/frontend/src/common/queries/wheresMyGene.ts
@@ -1072,7 +1072,8 @@ function useWMGQueryRequestBody(version: 1 | 2) {
 function useWMGFiltersQueryRequestBody(
   version: 1 | 2 = 2
 ): FiltersQuery | null {
-  const { selectedOrganismId, selectedFilters } = useContext(StateContext);
+  const { selectedOrganismId, selectedFilters, filteredCellTypeIds } =
+    useContext(StateContext);
 
   const { data } = usePrimaryFilterDimensions(version);
 
@@ -1101,6 +1102,7 @@ function useWMGFiltersQueryRequestBody(
         sex_ontology_term_ids: sexes,
         tissue_ontology_term_ids: tissues,
         publication_citations: publications,
+        cell_type_ontology_term_ids: filteredCellTypeIds,
       },
     };
   }, [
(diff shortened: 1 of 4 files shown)


frontend/src/common/queries/wheresMyGene.ts
```diff
@@ -1101,6 +1102,7 @@ function useWMGFiltersQueryRequestBody(
         sex_ontology_term_ids: sexes,
         tissue_ontology_term_ids: tissues,
         publication_citations: publications,
+        cell_type_ontology_term_ids: filteredCellTypeIds,
```
joyceyan, line 1105: this doesn't seem to be working - not sure if there's something i'm missing but it seems like either it's not being sent to the WMG call

frontend/src/views/WheresMyGene/common/store/actions.ts
```diff
@@ -200,8 +200,13 @@ export function setXAxisHeight(
   };
 }
 
+interface SetFilteredCellTypesPayload {
```
joyceyan, line 203: not sure if this is the best place to declare this custom payload and importing it in the reducer

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -38,6 +39,7 @@ export interface State {
   compare?: CompareId;
   xAxisHeight: number;
   filteredCellTypes: string[];
+  filteredCellTypeIds: string[];
```
joyceyan, line 42: i considered updating `filteredCellTypes` from being a list of the cell type names to being a list of `CellType` entities, but i think it would have been annoying to update every existing use case (and future use cases) to filter it out into either a list of names or a list of ids

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -418,10 +421,11 @@ function setXAxisHeight(state: State, action: PayloadAction<number>): State {
 
 function setFilteredCellTypes(
   state: State,
-  action: PayloadAction<State["filteredCellTypes"]>
+  action: PayloadAction<SetFilteredCellTypesPayload>
```
joyceyan, line 424: wanted to use the same reducer/action so that we're updating the cell type names and ids at the same time

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -373,10 +382,20 @@ export default memo(function HeatMap({
 
   const handleCellTypeDelete = (cellTypeToDelete: string) => () => {
```
seve, line 383: nit:
s/cellTypeToDelete/cellTypeNameToDelete

seve (changes requested):

atarashansky (approved):
Fix lint, and force merge!

seve (approved):

Issue #5720, opened 2023-09-08
Group by functionality should only show options with data when cell type filter is being used


signechambers1:
Steps to reproduce:

1. Prod
2. Select tissue = kidney
3. Cell type = podocyte
4. Group by = publication
Rows show up empty with misaligned dots

Desired result: only publications with data are shown, dots are aligned. 

Please ensure this is fixed for ALL group by options.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa1dCIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--3fa6eaa26ba1870366050288b045b1cc07faff52/image%20%281%29.png)

atarashansky:
This is the same bug as #5719. Will only close if this test case passes.

Issue #5719, opened 2023-09-08
Marker genes cannot be added to the dot plot unless the marker gene panel is closed


signechambers1:
From Karen during usability study: 

He selected podocyte as a cell type and tried to add all the marker genes to the dot plot via marker gene panel and none of the dots would load on the dot plot.
The dots on the dot plot will load but only if you close the marker gene panel

[Slack thread](https://czi-sci.slack.com/archives/C0247APK621/p1694191617560819)

tihuan:
1. If no genes, bulk add marker genes will render heatmap as expected
2. If there are existing genes, bulk add marker genes won't trigger a render
3. Open/close marker gene panel will force re-render the heatmap

PR #5727, opened 2023-09-08
fix(WMG): correctly update heatmap data when filtering celltypes

seve:
## Reason for Change

- #5719 #5720

## Changes

- modify
  - Do not modify state in place, create new object and set state instead 

## Notes for Reviewer

@tihuan is looking into disallowing this behavior via eslint

### DIFF
--- a/frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
+++ b/frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
@@ -207,8 +207,10 @@ export default memo(function Chart({
         };
 
         setAllChartProps((allChartProps) => {
-          allChartProps[tissue] = result;
-          return { ...allChartProps };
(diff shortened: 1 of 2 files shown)


frontend/src/views/WheresMyGene/components/HeatMap/components/Chart/index.tsx
```diff
@@ -207,8 +207,10 @@ export default memo(function Chart({
         };
 
         setAllChartProps((allChartProps) => {
-          allChartProps[tissue] = result;
-          return allChartProps;
+          const newAllChartProps = { ...allChartProps };
```
tihuan, line 210: @seve I changed this to copy a new object before mutation, just so that we don't mutate the original object! 

I also tested locally to make sure it does work lol Thanks!

tihuan (approved):
LGTM!

tihuan:
Will force merge because of the known pipeline bug failing e2e test

PR #5729, opened 2023-09-08
fix(cellguide): fix mobile view tooltip and sticky header

atarashansky:
Some overflow scrolling was breaking the sticky (fixed) header in cellguide card pages.

Fixed by changing `width` to max-width and updating the left position of the sticky header in accordance.

### DIFF
--- a/frontend/src/views/CellGuide/components/CellGuideCard/style.ts
+++ b/frontend/src/views/CellGuide/components/CellGuideCard/style.ts
@@ -54,7 +54,6 @@ export const CellGuideView = styled.div<CellGuideViewProps>`
 
 export const CellGuideWrapper = styled.div<CellGuideViewProps>`
   margin: 0 auto 80px;
-  overflow-x: hidden;
   max-width: ${(props) =>
     props.skinnyMode
       ? `${DEFAULT_ONTOLOGY_WIDTH}px`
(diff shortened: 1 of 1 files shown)


frontend/src/views/CellGuide/components/CellGuideCard/style.ts
```diff
@@ -57,7 +54,7 @@ export const CellGuideView = styled.div<CellGuideViewProps>`
 
 export const CellGuideWrapper = styled.div<CellGuideViewProps>`
   margin: 0 auto 80px;
-  width: ${(props) =>
```
atarashansky, line 60: this was causing unintended x-scrolling when the viewport was smaller than the width

frontend/src/views/CellGuide/components/CellGuideCard/style.ts
```diff
@@ -115,10 +112,6 @@ export const CellGuideCardHeader = styled.div<CellGuideCardHeaderProps>`
   background: linear-gradient(180deg, #fff 89.82%, rgba(255, 255, 255, 0) 100%);
   z-index: 10;
   width: ${(props) => `${props.width + LEFT_RIGHT_PADDING_PX_XXL}px`};
-  left: ${(props) =>
```
atarashansky, line 118: after changing to use "max-width" this was no longer necessary

frontend/src/views/CellGuide/components/CellGuideCard/style.ts
```diff
@@ -211,6 +204,7 @@ export const MobileTooltipWrapper = styled.div`
   display: flex;
   flex-direction: column;
   height: calc(100vh - ${HEADER_HEIGHT_PX}px);
+  overscroll-behavior: contain;
```
atarashansky, line 207: this fixes a bug with the tooltip overlay where users could scroll past the bounds of the tooltip to scroll the container underneath

joyceyan (approved):

Issue #5574, opened 2023-08-21
RDEV testing stack provisioning strategy


tihuan:
Figure out a smart way to provision resources if needed:

Objectives:
1. @atolopko-czi whether we can create a new stack on demand (based on if an rdev has already been created based on rdev naming convention to find its corresponding stack) per PR and delete the rdev once PR/branch is deleted

metakuni:
Can this be considered a partial dup of #5439 ?

atolopko-czi:
> Can this be considered a partial dup of https://github.com/chanzuckerberg/single-cell-data-portal/issues/5439 ?

These are separate issues. #5439 is about creating/reusing a stack for running tests on rdev, while this issue #5574 is about scaling up the infra that runs _all_ of the stacks.

Bento007:
Over the past week I've been working on fixing issues in the rdev lifecycle that cause it to fail setup or teardown. 

What we want to do is spin up an rdev for e2e and functional tests then tear down afterwards. We will take the branch name and replace any characters that cannot be in a domain such at "./" and create an rdev with that name. Once the PR is merged, the environment will be torn down.

atolopko-czi:
@Bento007 ok to close this issue now?

Issue #5436, opened 2023-08-10
RDev for Auth0 authentication [CI/CD]


atolopko-czi:
For both ad hoc and automated testing, RDev backend APIs need to be made accessible when not being called via the FE. Using RDev first requires a Happy-level Auth0 authentication in order to access Happy-managed service endpoints. When accessing Happy via a browser, this automatically triggers an authentication flow. 

E2E tests will need to be able to navigate this authentication flow as well, using a "machine" user account, presumably.

Engineers and QA roles should also be able to _easily_ call RDev-hosted endpoints without hacks, such as "stealing" tokens from earlier RDev browser-based sessions.

Goal is to have reusable code that can be used to perform the Happy-level Auth0 authentication flow.

Implementation notes:
* Can we turn off CZI-level auth entirely? Will need to ask InfoSec/Central. Doing so makes this issue easy to resolve. Ask the these folks what other Happy-based teams are doing to test in RDev.
* If we can't just turn off auth, then investigate best ways to navigate the Auth0 auth flow programatically. While this may require a "headless browser interaction" strategy, there should be programmatic Auth0 APIs that can be used. Will need a machine token to make this work.
* If we can't do programmatic Auth0, then we may need to use a UI "interaction" framework (ala Puppeteer) or performing some token hackery.
* Work with Happy Team, as needed.

tihuan:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5436?planning-poker) with Zenhub @atolopko-czi @Bento007 @danieljhegeman @ebezzi

Bento007:
I started a Dialog with the happy team who made several suggested:

1. use Oauth proxy https://github.com/oauth2-proxy/oauth2-proxy/blob/738c09095bdd21d4816280e2480ec221a636d9b9/contrib/oauth2-proxy.cfg.example#L67 to disable auth for some endpoint
2. Using a cookie to bypass authentication
3. set https://github.com/chanzuckerberg/single-cell-data-portal/blob/a8981b7110ba5f4d23e86ce7395cf49d0ceecf08/.happy/terraform/envs/rdev/main.tf#L12 to False

I'm waiting to hear back from from the security team to see if 3 is ok. Since the application is also behind single-cell auth0, it might be enough.

tihuan:
[...]

PR #5748, opened 2023-09-12
feat(cicd): run functional on rdev with auth

Bento007:
## Reason for Change

- #5574  
- #5436
- Support running functional tests in an rdev environment.

## Changes
- Use Authentication header to get through oauth_proxy.
- Use STACK_NAME to identify the rdev environment in functional tests
- Support rdev environments for functional tests.
- use a tenacity timeout strategy for failing api key test.
- pass needed environment variables to docker image during test
- use a config dummy file for CorporaAuthConfig to avoid using AWS secret manager for the functional tests. Chamber already handles passing in all the needed secrets.

## Testing steps
- test against existing environment
   ```export STACK_NAME=auth DEPLOYMENT_STAGE=rdev AWS_PROFILE=single-cell-dev && make local-functional-test```
- [testing instructions](https://docs.google.com/document/d/1nynGcBS_TA55qlQo9WjINGkcMnE_xIBz7-inmop2bqo/edit#heading=h.t5qd7armv63z)


Makefile
```diff
@@ -211,13 +211,18 @@ local-unit-test-cxg-admin:
...
-local-functional-test: export AWS_PROFILE=$(TEST_AWS_PROFILE)
 local-functional-test: ## Run functional tests in the dev environment
 	if [ -n "$${BOTO_ENDPOINT_URL+set}" ]; then \
 		EXTRA_ARGS="-e BOTO_ENDPOINT_URL"; \
 	fi; \
 	chamber -b secretsmanager exec corpora/backend/$${DEPLOYMENT_STAGE}/auth0-secret -- \
-		docker-compose $(COMPOSE_OPTS) run --rm -T -e CLIENT_ID -e CLIENT_SECRET -e TEST_ACCOUNT_USERNAME -e TEST_ACCOUNT_PASSWORD -e DEPLOYMENT_STAGE -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN $${EXTRA_ARGS} \
+		docker-compose $(COMPOSE_OPTS) run --rm -T \
+		-e CLIENT_ID -e CLIENT_SECRET \
+		-e FUNCTEST_ACCOUNT_USERNAME -e FUNCTEST_ACCOUNT_PASSWORD \
```
danieljhegeman, line 221: this would be for actually logging the test user in (the normal cellxgene auth flow)...

Makefile
```diff
@@ -211,13 +211,18 @@ local-unit-test-cxg-admin:
...
+		-e TEST_AUTH0_USER_ACCOUNT_PASSWORD -e TEST_APP_ID -e TEST_APP_SECRET \
```
danieljhegeman, line 222: ...while this would be for getting past the outer auth0 layer / proxy, right?

danieljhegeman: My resulting question then would be: why wasn't the functest_account_username secret being used before now...

Makefile
Bento007: yes using cookies

Makefile
Bento007: > ...while this would be for getting past the outer auth0 layer / proxy, right?

yes for `-e TEST_APP_ID -e TEST_APP_SECRET`

Bento007: > y resulting question then would be: why wasn't the functest_account_username secret being used before now...

So they actually were being used, but they were getting introduced into the container through a different path. In the [GHA functional test](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/.github/workflows/deploy-happy-stack.yml.), the container was getting the secret directly from AWS Secrets manager using credentials supplied to the docker container. The AWS credentials were passed into the GHA container here https://github.com/chanzuckerberg/single-cell-data-portal/blob/3bca29bd6cef0b8d0eab6fdd49edf57cd8741364/Makefile#L220
but this didn't work when running the tests locally or when `DEPLOYMENT_STAGE=test` which is why this comment
https://github.com/chanzuckerberg/single-cell-data-portal/blob/3bca29bd6cef0b8d0eab6fdd49edf57cd8741364/Makefile#L206-L212

Now all of the secrets needed by the functional tests are introduced via the environment using the make recipe and chamber. No more sneaking secret getting introduced from multiple different sources. The functional tests no longer need to access AWS secrets to work.

Bento007: I'm going to remove the AWS credentials and the comment from the make recipe. They aren't needed.

danieljhegeman: excellent, thanks!

danieljhegeman (approved):
worked for me locally 👍

tests/functional/backend/common.py
```diff
@@ -15,22 +17,34 @@
...
         super().setUpClass()
         cls.deployment_stage = os.environ["DEPLOYMENT_STAGE"]
-        cls.config = CorporaAuthConfig()
+
+        # configure CorporaAuthConfig to use a temporary directory for the config file
+        cls.tempdir = tempfile.TemporaryDirectory()
+        dummy_config = f"{cls.tempdir.name}/dummy.json"
+        with open(dummy_config, "w") as fp:
+            json.dump({"api_base_url": os.getenv("API_BASE_URL")}, fp)
+        cls.config = CorporaAuthConfig(source=dummy_config)
```
seve, line 47: stinky

PR #5759, opened 2023-09-13
fix(cicd): build docker images using matrix strategy

Bento007:
## Reason for Change
When building the docker images in [build-images-and-create-deployment.yml](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5759/files#diff-e0f79301c397cd2b444a586cf77777bfe36bd14722a75aed390bed2d6a6d21bf), all of the images were being built in the jobs `build-frontend-backend-images` and `build-batch-job-images`. This was causing the GHA environment to crash by running out of disk space. This was causing a headache for deployment which would fail due to a lack do disk space. Now the images will be built in their own job, which does not share disk space. This avoid the disk space capacity error. The frontend and backend images are still built in the same job, but that is beyond the scope of this PR to address. 

As an added bonus, this reduces our image build time to under 3min in total.

## Changes

- modify build-image-and-create-deployment.yml to build the docker images in parallel using the GHA matrix strategy.
- Update happy version from 0.59.0->0.92.0 to give us access to slices. Now we are using slices to build the images in parallel jobs. Without the slices all of the images would be build for every job.
- Remove `add-image-tags` job. Now this step is included as part of the build-images job.
- Add frontend and backend docker images to the docker compose profile fullstack. This is required because the frontend depends on the backend according to our docker_compose.yml. Happy blocks us from building the frontend separately because of this. If we want to build the backend and front end in parallel jobs we will need to remove the [depends_on: backend](https://github.com/chanzuckerberg/single-cell-data-portal/blob/9162b0234fd1d1dabf624adad8bdfb88a983cc48/docker-compose.yml#L52-L53), and update our tests to spin up the backend when testing the frontend.

## Testing steps
Pointed the GHA at this branch and run. [Results](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6179082877)
Screenshot 2023-09-13 at 5 12 41 PM

### DIFF
--- a/.github/workflows/build-images-and-create-deployment.yml
+++ b/.github/workflows/build-images-and-create-deployment.yml
@@ -6,6 +6,7 @@ on:
       - main
       - staging
       - prod
+      - tsmith/image-matrix #TODO remove after PR approved
 
 env:
   CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/build-images-and-create-deployment.yml
+++ b/.github/workflows/build-images-and-create-deployment.yml
@@ -27,8 +27,7 @@ jobs:
     strategy:
       matrix:
         image:
-          - frontend
-          - backend
+          - frontend # pushed both the frontend and backend images
           - upload_failures
           - upload_success
           - dataset_submissions
@@ -58,7 +57,8 @@ jobs:
         run: |
           echo "HAPPY_COMMIT=$(git rev-parse --verify HEAD)" >> envfile
           echo "HAPPY_BRANCH=$(git branch --show-current)" >> envfile
-          happy push deploy --slice ${{ matrix.image }} --docker-compose-env-file envfile --aws-profile "" \
+          happy push devstack --env dev --slice ${{ matrix.image }} \
+          --docker-compose-env-file envfile --aws-profile "" \
           --tags sha-${GITHUB_SHA:0:8},branch-$(echo ${GITHUB_REF#refs/heads/} | sed 's/[\+\/]/-/g')
       - uses: 8398a7/action-slack@v3
         with:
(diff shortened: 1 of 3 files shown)


atolopko-czi (approved):
LGTM, and DRYer too!

.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -41,81 +52,14 @@ jobs:
...
-        uses: chanzuckerberg/github-actions/.github/actions/install-happy@install-happy-v1.4.2
-        with:
-          happy_version: "0.59.0"
-      - name: Docker re-tag
-        shell: bash
-        run: |
-          happy addtags --aws-profile "" --source-tag sha-${GITHUB_SHA:0:8} --dest-tag branch-$(echo ${GITHUB_REF#refs/heads/} | sed 's/[\+\/]/-/g')
+          happy push devstack --env dev --slice ${{ matrix.image }} \
+          --docker-compose-env-file envfile --aws-profile "" \
+          --tags sha-${GITHUB_SHA:0:8},branch-$(echo ${GITHUB_REF#refs/heads/} | sed 's/[\+\/]/-/g')
```
Bento007, line 62: Adding tags from the `add-image-tags` job

.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
       - staging
       - prod
+      - tsmith/image-matrix #TODO remove after PR approved
```
atarashansky, line 9: just a reminder

atarashansky (approved):

Bento007:
This is how you decipher what can be in the config [happy config](https://github.com/chanzuckerberg/happy/blob/d234338f25fea474294d31d6c7ff588b7b613f87/shared/config/happy_config.go#L67)

PR #5774, opened 2023-09-15
feat(cicd): manage rdev per pull request

Bento007:
## Reason for Change

- #5574
- Manages the creation, update, and destructions of Rdev environments base on the state of the pull request.

## Changes

- Add GHA to create or update an rdev environment when a PR is opened, reopened, or synchronized(commit added)
- Add GHA to delete an rdev environment when a PR is close or merged.
- Add a summary in the GHA that displays the frontend and backend link to the Rdev
- Remove unused env variables from build-images-and-create-deployment.yml

## Testing steps

- [x] rdev is create when PR is open
- [x] rdev is update when commit is added -> [workflow](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6203795852/job/16844949919)
- [x] rdev is delete when PR is closed -> [workflow](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6203717705)
- [x] rdev is created when PR is reopened -> [workflow](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6203795852?pr=5774)
- [x] no rdev is created when pull request base prod branches.
- [x] verify tags are set correctly in AWS ECR
- [x] The rdev is created under the authors name 
- [x] Summary is left as a comment to the RP  

## Notes for Reviewer
related infra work
- https://github.com/chanzuckerberg/single-cell-infra/pull/897 

- If a PR is closed and quickly opened, you need to cancel the delete rdev environment and the create rdev will fail


.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -8,8 +8,6 @@ on:
       - prod
 
 env:
-  CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
Bento007, line 11: These are use in this worflow.

.github/workflows/rdev-delete-for-pr.yml
```diff
@@ -0,0 +1,42 @@
+name: Delete rdev for PR
+
+on:
+  pull_request:
+    branches-ignore:
+      - staging
```
atolopko-czi, line 6: `main` instead of `staging` now? I don't think we'll need the `staging` branch for the new CI/CD flow, since main will be deployed directly to staging env.  same in next workflow, below

.github/workflows/rdev-update-for-pr.yml
```diff
@@ -0,0 +1,102 @@
...
+          happy_version: "0.92.0"
+      - name: Push images
+        run: |
+          echo "HAPPY_COMMIT=$(git rev-parse --verify HEAD)" >> envfile
+          echo "HAPPY_BRANCH=$(git branch --show-current)" >> envfile
+          happy push devstack --env dev --slice ${{ matrix.image }} \
+          --docker-compose-env-file envfile --aws-profile "" \
+          --tags sha-${GITHUB_SHA:0:8},pr-${{ github.event.number }},branch-$(echo ${GITHUB_REF_NAME} | sed 's/[\+\/]/-/g')
+
+  create-rdev:
```
atolopko-czi, line 64: since tests will be run on rdev by the `push-tests` workflow, will there be a race condition in setting up the rdev stack and running tests?

atolopko-czi (commented):
A couple initial questions.

.github/workflows/rdev-update-for-pr.yml
Bento007: The tests should run in the same workflow, this will avoid the race condition.

.github/workflows/rdev-delete-for-pr.yml
Bento007: As written this workflow will only run if the base branch is not staging or prod. I think you mean we should remove `-staging` because that branch wont be used anymore and main will deploy directly to the staging environment?

atolopko-czi: I guess it depends on what branch name is associated with the PR being closed. If it's the feature branch, then I think we want to ignore main. My main point is that we want to create and destroy rdev stacks only for feature branch PRs. So whatever that looks like here. :)

Bento007: So we only want rdevs for PRs when those PRs are targeting a feature branch(AKA not main), and not when the PR is targeting main or prod?

.github/workflows/rdev-update-for-pr.yml
```diff
@@ -0,0 +1,105 @@
+name: Create/Update rdev for PR
+
+on:
+  pull_request:
+    branches-ignore:
+      - prod
```
danieljhegeman, line 6: Instead of this, should we consider
```
branches:
  - main
 ```
 ?

Comment (danieljhegeman) [reply .github/workflows/rdev-update-for-pr.yml:6]: i.e., do we want to spin up an rdev for scenarios where someone creates a (minor) PR against someone else's feature branch?

**Review by danieljhegeman — APPROVED** [review]
LGTM 🚀

File: .github/workflows/rdev-update-for-pr.yml (hunk 4 above)
Comment (danieljhegeman) [reply .github/workflows/rdev-update-for-pr.yml:6]: (probably not; we probably want to create one every time... but just a thought)

Comment (Bento007) [reply .github/workflows/rdev-update-for-pr.yml:6]: most of our PRs are against main. If we ignore main then most PRs will not spin up an Rdev. The goal is to spin up and run tests on PRs before they get merge into main. I put prod in this list because we usually create PR to merge changes into prod from main. In this case we don't need to spin up an rdev because main is already associated with the dev environment. I had staging in this list be removed it, since that branch will be going away soon.

Comment (danieljhegeman) [reply .github/workflows/rdev-update-for-pr.yml:6]: No, my suggestion of
```
branches:
  - main
```
would spin up an rdev for all PRs against main, but not for any PRs against other branches such as prod or a feature branch

danieljhegeman: `branches` vs `branches-ignore`

.github/workflows/rdev-delete-for-pr.yml
danieljhegeman: No, he's saying that, in the case of a feature branch being merged into main, it depends on whether the branch name that gets fed into the GHA for such an action is `main` or `<feature_branch>`. If it is the former, then we want to ignore staging, because we don't want to try to find an rdev to delete when main is merging into staging; if it is the latter, then we want to ignore main, because main would be the branch name that would be passed to the GHA when main merges into staging.

.github/workflows/rdev-update-for-pr.yml
Bento007: If we did that, then if there was a feature branch it would lack rdevs and testing.

Issue #5438, opened 2023-08-10
Run E2E and functional tests on RDev stack [CI/CD]


atolopko-czi:
To ensure that `main` branch is kept "green" (passes all tests), we need to be able to run all E2E and functional tests on RDev. These tests are currently only run post-deploy on dev/stage/prod environments. Using RDev, we need to run as many of our tests as possible on feature branches that have PRs, before allowing them to be merged to main.

PR #5785, opened 2023-09-18
feat(cicd): run functional test on rdev after deploying

Bento007:
## Reason for Change

- #5438
- This allow functional and e2e test to be run on a a live environment during a pull request.

## Changes

- Add functional tests to run after updating the environment
- skip incompatible tests
- restore version test
- centralize test h5ad uri

## Testing steps

- see the functional tests running in after deploying the rdev in GHA.

## Notes for Reviewer
- When the PR is create, functional test might fail the first time while waiting for the rdev environment to be ready.

related infra work
- https://github.com/chanzuckerberg/single-cell-infra/pull/901


danieljhegeman (commented):
I may be ignorant - I am not seeing where the running of the `functional-test` job is gated such that it is only done _after_ deployment... this reads to me like it would be run on creation of any PR... which seems to me like it's what we want? By "deployment" do you mean "deployment of an rdev"? Slightly confusing

Edit: read description more closely and it sounds like that's what we want: run them every time for a PR. 👍

Bento007:
I mean deployment of an rdev

danieljhegeman (approved):
bombs away

PR #5790, opened 2023-09-20
test(wmg): filter tests

seve:
## Reason for Change

- #5447 

## Changes

- add
  - utils to add/remove cell type filters 
  - tests verifying cell type filter functionality


frontend/tests/features/wheresMyGene/wheresMyGene.test.ts
```diff
@@ -855,6 +857,18 @@ describe("Where's My Gene", () => {
       { page }
     );
   });
+  describe("Cell Type Filtering", () => {
+    test("Filter to single cell type and then clear", async ({ page }) => {
+      await goToWMG(page);
+      await searchAndAddFilterCellTypes(page, "B cell");
```
tihuan, line 863: nit: The function name implies plural cell types already, so maybe we can just supply an array of cell type names as the expected argument, so we don't need to worry about handling `string | string[]` in `searchAndAddFilterCellTypes()`?

e.g., `searchAndAddFilterCellTypes(page, ["B cell"])`

tihuan (approved):
LGTM! Just one non-blocking comment 👍 Thank you!

Issue #5758, opened 2023-09-13
Add asset endpoint to Portal API


MillenniumFalconMechanic:
#### Need
A dataset asset's permanent URL and file size is displayed on the download modal.

#### Approach
* Add an endpoint to the Portal API that calls `backend/curation/api/v1/curation/collections/common.py::extract_dataset_assets`.
* Note: [existing asset endpoint is to be deprecated](https://github.com/chanzuckerberg/single-cell-data-portal/issues/5697) and [Portal and Curation APIs may be consolidated in the future](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5712).

Issue #5697, opened 2023-09-05
Remove Data Portal pre-signed asset URL capability


MillenniumFalconMechanic:
See [Slack](https://clevercanary.slack.com/archives/C02CF7MQPGV/p1692165701393269).

PR #5794, opened 2023-09-21
feat: added get asset with permanent url.

MillenniumFalconMechanic:
## Reason for Change
- #5758

## Changes
- Added GET dataset asset endpoint that reuses existing Discover API functionality to calculate file size and permanent URL. (Supersedes POST dataset asset endpoint that generates pre-signed URL.)
- Modified Discover API's `extract_dataset_assets`: moved permanent URL builder code to business layer for reuse.

## Testing steps
- Added unit tests to `test_portal_api` and `test_business` test suites.
- Tested in rdev by hitting new GET endpoint and:
  - confirming file size matches existing POST file size for same asset, and,
  - confirming permanent URL yields download.

### DIFF
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -41,7 +41,7 @@ def extract_dataset_assets(dataset_version: DatasetVersion):
         filesize = get_business_logic().s3_provider.get_file_size(asset.uri)
         if filesize is None:
             filesize = -1
-        url = get_business_logic().generate_permanent_url(dataset_version.version_id, asset)
+        url = get_business_logic().generate_permanent_url(dataset_version.version_id, asset.type)
         result = {
             "filesize": filesize,
             "filetype": asset.type.upper(),
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -12,6 +12,7 @@ from backend.layers.common.entities import (
     CollectionVersionId,
     CollectionVersionWithDatasets,
     CollectionVersionWithPublishedDatasets,
+    DatasetArtifact,
     DatasetArtifactType,
     DatasetId,
     DatasetProcessingStatus,
@@ -33,7 +34,6 @@ def get_collections_base_url():
 
 
 def extract_dataset_assets(dataset_version: DatasetVersion):
-    base_url = CorporaConfig().dataset_assets_base_url
     asset_list = list()
     for asset in dataset_version.artifacts:
         if asset.type not in allowed_dataset_asset_types:
@@ -41,7 +41,7 @@ def extract_dataset_assets(dataset_version: DatasetVersion):
         filesize = get_business_logic().s3_provider.get_file_size(asset.uri)
         if filesize is None:
             filesize = -1
-        url = f"{base_url}/{dataset_version.version_id.id}.{asset.type}"
+        url = get_business_logic().generate_permanent_url(dataset_version.version_id, asset)
         result = {
             "filesize": filesize,
             "filetype": asset.type.upper(),
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -6,6 +6,7 @@ from uuid import UUID
 from backend.common.corpora_config import CorporaConfig
 from backend.common.utils.http_exceptions import ForbiddenHTTPException, GoneHTTPException, NotFoundHTTPException
 from backend.layers.auth.user_info import UserInfo
+from backend.layers.business.business import BusinessLogic
 from backend.layers.common.entities import (
     CollectionId,
     CollectionVersion,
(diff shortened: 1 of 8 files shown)


backend/layers/business/business.py
```diff
@@ -5,12 +5,14 @@
...
 from functools import reduce
 from typing import Dict, Iterable, List, Optional, Set, Tuple
+from backend.common.corpora_config import CorporaConfig
 
 from backend.layers.business.business_interface import BusinessLogicInterface
 from backend.layers.business.entities import (
     CollectionMetadataUpdate,
     CollectionQueryFilter,
     DatasetArtifactDownloadData,
+    DatasetArtifactDownloadDataDeprecated,
```
MillenniumFalconMechanic, line 15: Added `Deprecated` suffix to existing `DatasetArtifactDownloadData` specific to pre-signed URL functionality. The `XxDeprecated` entities and methods are to be removed in #5697.

backend/layers/business/business.py
```diff
@@ -5,12 +5,14 @@
 from datetime import datetime
 from functools import reduce
 from typing import Dict, Iterable, List, Optional, Set, Tuple
+from backend.common.corpora_config import CorporaConfig
```
ebezzi, line 8: This import should be moved to the method definition. Importing `CorporaConfig` has an implicit dependency on boto3 which we should avoid. I am not sure this is still enforced throughout the codebase but we should try to keep the business layer dependency free.

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -33,15 +33,14 @@ def get_collections_base_url():
...
-    base_url = CorporaConfig().dataset_assets_base_url
     asset_list = list()
     for asset in dataset_version.artifacts:
         if asset.type not in allowed_dataset_asset_types:
             continue
         filesize = get_business_logic().s3_provider.get_file_size(asset.uri)
         if filesize is None:
             filesize = -1
-        url = f"{base_url}/{dataset_version.version_id.id}.{asset.type}"
+        url = get_business_logic().generate_permanent_url(dataset_version.version_id, asset.type)
```
ebezzi, line 43: If `generate_permanent_url` is a static method, you may want to do `BusinessLogic.generate_permanent_url`.  Alternatively, it could be a method of `CollectionVersionBase` (more OOP compliant), or be in `helpers.py`.

backend/layers/business/business.py
```diff
@@ -101,6 +103,15 @@ def __init__(
         self.uri_provider = uri_provider
         super().__init__()
 
+    @staticmethod
+    def generate_permanent_url(dataset_version_id: DatasetVersionId, asset_type: DatasetArtifactType):
+        """
+        Return the permanent URL for the given asset.
+        """
+        base_url = CorporaConfig().dataset_assets_base_url
+        url = f"{base_url}/{dataset_version_id.id}.{asset_type}"
```
ebezzi, line 112: Return from here.

backend/layers/business/business.py
```diff
@@ -519,6 +528,24 @@ def get_dataset_artifacts(self, dataset_version_id: DatasetVersionId) -> Iterabl
...
+
+        if not artifact:
+            raise ArtifactNotFoundException(f"Artifact {artifact_id} not found in dataset {dataset_version_id}")
+
+        file_size = self.s3_provider.get_file_size(artifact.uri)
+        url = self.generate_permanent_url(dataset_version_id, artifact.type)
+
+        return DatasetArtifactDownloadData(file_size, url)
+
+    # Superseded by get_dataset_artifact_download_data. Remove with #5697.
```
ebezzi, line 545: Maybe add this as a `TODO: `

backend/layers/business/entities.py
```diff
@@ -14,6 +14,13 @@ class CollectionQueryFilter:
 
 @dataclass
 class DatasetArtifactDownloadData:
+    file_size: int
+    url: str
+
+
+# Superseded by DatasetArtifactDownloadData. Remove with #5697.
+@dataclass
+class DatasetArtifactDownloadDataDeprecated:
```
ebezzi, line 23: Mostly a nitpick, but I think `Deprecated` should be a prefix.

backend/portal/api/portal_api.py
```diff
@@ -647,6 +647,33 @@ def upload_relink(collection_id: str, body: dict, token_info: dict):
...
+    except ArtifactNotFoundException:
+        raise NotFoundHTTPException(detail=f"'dataset/{dataset_id}/asset/{asset_id}' not found.") from None
+
+    if download_data.file_size is None:
+        raise ServerErrorHTTPException() from None
+
+    if download_data.url is None:
+        raise ServerErrorHTTPException()
+
+    response = {"dataset_id": dataset_id, "file_size": download_data.file_size, "url": download_data.url}
```
ebezzi, line 672: Super nitpick, but for readability I would use one line per parameter.

backend/curation/api/v1/curation/collections/common.py
ebezzi: This is also fine btw.

ebezzi (approved):

PR #5801, opened 2023-09-21
fix(rdev): Parallel build for push-rdev.yml

Bento007:
## Reason for Change

- push-rdev.yml is running out of space during build. This is the same [issue](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5759) that was affecting the deployment of dev.

## Changes

- build the image using the matrix build strategy

## Testing steps

- look at the Push Remote Dev/ build-push-images GHA job

### DIFF
--- a/.github/workflows/push-rdev.yml
+++ b/.github/workflows/push-rdev.yml
@@ -24,6 +24,16 @@ permissions:
 
 jobs:
   build-push-images:
+    strategy:
+      matrix:
+        image:
+          - frontend # pushed both the frontend and backend images
+          - upload_failures
+          - upload_success
+          - dataset_submissions
+          - processing
+          - wmg_processing
+          - cellguide_pipeline
     runs-on: ubuntu-22.04
     steps:
       - name: Configure AWS Credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/push-rdev.yml
```diff
@@ -24,14 +24,24 @@ permissions:
 
 jobs:
   build-push-images:
+    strategy:
+      matrix:
+        image:
+          - frontend # pushed both the frontend and backend images
```
danieljhegeman, line 30: not doubting you but how is this also building the backend? Where's the dependency pick-up happening?

Bento007: by putting them in the same [profile](https://github.com/chanzuckerberg/single-cell-data-portal/blob/827abee7af280e8325322e2f11e0f62ef7d05a7e/.happy/config.json#L18-L22) for [frontend](https://github.com/chanzuckerberg/single-cell-data-portal/blob/827abee7af280e8325322e2f11e0f62ef7d05a7e/docker-compose.yml#L42) and [backend](https://github.com/chanzuckerberg/single-cell-data-portal/blob/827abee7af280e8325322e2f11e0f62ef7d05a7e/docker-compose.yml#L295C13-L295C13)

PR #5804, opened 2023-09-22
chore(wmg): added useEffect hook to expand filtered tissues (5739)

kaloster:
## Reason for Change

- (https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5739)

## Changes

- added useEffect to trigger auto expand of tissue nodes selected from the tissue filter

## Testing steps
- in Gene Expression - select tissue(s) from the tissue filter on left sidebar
- selected tissue nodes should auto-expand

## Notes for Reviewer
https://github.com/chanzuckerberg/single-cell-data-portal/assets/7976579/6cf537cb-4d22-47dc-a1ca-8f89caaa1d4a


frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -380,6 +380,28 @@ export default memo(function HeatMap({
...
   ]);
 
+  useEffect(() => {
+    if (filteredTissueIds.length === 0) return;
+    setExpandedTissues(new Set(filteredTissueIds));
+    Object.entries(sortedCellTypesByTissueName).forEach(
+      ([tissue, cellTypes]) => {
+        if (filteredTissueIds.includes(tissuesByName[tissue].id)) {
+          cellTypes.forEach((cellType) => {
+            setDisplayedCellTypes(
```
tihuan, line 390: maybe need @seve 's input here, but `handleExpandCollapse()` seems to handle adding newly added tissue's cellTypes via `setDisplayedCellTypes()` here: https://github.com/chanzuckerberg/single-cell-data-portal/blob/b7024e7af98bd110f4ba8b80f9e1b2ac03d76908/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx#L262

So perhaps potentially we could leverage `handleExpandCollapse()`?

Thank you!

seve: It could be possible, but we would have to modify that function to allow `tissueName` to be null. That function is also a bit more robust. Right now, this `useEffect` doesn't take filtered `cellTypes` into account.

seve (changes requested):

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
seve: To be clear, I requested changes on the stance that there's a bug, I think there's still discussion to be had on whether we should use that function.

PR #5805, opened 2023-09-22
feat(cicd): Add 30 minute timeout for all tests in GHA

Bento007:
## Reason for Change

- Some tests can run for several hours and always fail with error messages that aren't helpful. Rather than waste time waiting for the tests to fail, cancel the test once it exceeds the time out. This save time on compute and engineers wait or manually canceling the test.

## Changes

- Add 30 minute timeout for all tests in GHA


atolopko-czi:
Instead of a fixed 30m for all test types, what if it were 150% of the expected run time for each test type? I realize it's a moving target, but we could at least _start out_ with the correct values, and then adjust if we find it's being exceeded regularly (and we can't readily optimize the tests).

ebezzi (approved):

ebezzi:
> Instead of a fixed 30m for all test types, what if it were 150% of the expected run time for each test type? I realize it's a moving target, but we could at least _start out_ with the correct values, and then adjust if we find it's being exceeded regularly (and we can't readily optimize the tests).

I would keep it simple. 30m seems a good cutoff - if a series of tests take so long, they should probably be split and parallelized anyway.

PR #5816, opened 2023-09-22
feat(cicd): allow all conventional commit types

Bento007:
## Reason for Change

- The default has more options and reduce the configuration we need to manager.

## Changes
- allow all conventional commit types


tihuan (approved):
Thanks so much!

Issue #5809, opened 2023-09-22
feat(cellguide): add ontology tree state for cell types that are not children of animal cell


atarashansky:
The proposed solution is to set the root of the tree to be "cell" when building the ontology tree graph and associated initial tree state per cell type or tissue. The FE will be updated such that cell types that are descendants of animal cell will use animal cell as the root.

PR #5817, opened 2023-09-24
feat(cellguide): add ontology tree states for every cell type

atarashansky:
## Reason for Change

- #5809

## Changes

- updated the pipeline to build the ontology graph from "cell" as the root (`CL:0000000`).
- updated the frontend to set the root to be `CL:0000548__0` (animal cell) if the target cell type is a descendant of animal cell.

## Testing steps

- Added E2E test case for "abnormal cell" ontology dag view. If this test passes, it means that the tree view exists for abnormal cell (which is not a descendant of animal cell).

### DIFF
--- a/frontend/src/views/CellGuide/components/common/OntologyDagView/index.tsx
+++ b/frontend/src/views/CellGuide/components/common/OntologyDagView/index.tsx
@@ -51,6 +51,7 @@ import {
   CELL_GUIDE_CARD_ONTOLOGY_DAG_VIEW_TOOLTIP,
   CELL_GUIDE_CARD_ONTOLOGY_DAG_VIEW_DEACTIVATE_MARKER_GENE_MODE,
   MINIMUM_NUMBER_OF_HIDDEN_CHILDREN_FOR_DUMMY_NODE,
+  ANIMAL_CELL_ID,
 } from "src/views/CellGuide/components/common/OntologyDagView/constants";
 import {
   ALL_TISSUES,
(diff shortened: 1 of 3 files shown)


frontend/src/views/CellGuide/components/common/OntologyDagView/index.tsx
```diff
@@ -551,7 +557,7 @@ export default function OntologyDagView({
           </TableUnavailableHeader>
           <TableUnavailableDescription>
             The cell ontology visualization for this cell type is currently not
-            available as it is not a descendant of &quot;animal cell&quot;.
+            available.
```
seve, line 560: When are we going to run into this now?

frontend/src/views/CellGuide/components/common/OntologyDagView/index.tsx
```diff
@@ -728,3 +734,44 @@ function getShownData(
   }
   return notShownWhenExpandedNodes;
 }
+
+function getSubTree(
```
seve, line 738:
```suggestion
function getTargetSubTree(
```

seve (approved):
A light suggestion from me, got a bit confused on the functionality b/c I forgot about the target node aspect

joyceyan (approved):

frontend/src/views/CellGuide/components/common/OntologyDagView/index.tsx
joyceyan: if i'm following the ternary correctly, i believe when `data && initialTreeState` is false, so when we're unable to pull the data for whatever reason.

seve: yeah, i understand in theory, but should we add a test to make sure this message is never seen?

PR #5830, opened 2023-09-26
fix: fail happy deploy if service does not start successfully

danieljhegeman:
## Reason for Change

Data Portal complement of https://github.com/chanzuckerberg/single-cell-explorer/issues/662


danieljhegeman:
@prathapsridharan actually going to hold on merging this until I confirm that it behaves as expected with Explorer...

danieljhegeman:
This actually is not a solution to the problem that we encountered. Closing for now.

PR #5832, opened 2023-09-26
fix(cellguide): server side props for seo description

atarashansky:
## Reason for Change

- CellGuide pipeline was updated to generate SEO descriptions as strings, not dictionaries. Because SEO descriptions are only generated when new cell types are added, most of the SEO descriptions are still dictionaries. Newer cell types (like [this one](https://cellxgene.cziscience.com/cellguide/CL_4030026)) store the SEO descriptions as strings, leading to an error in parsing the server side props and undefined names.
- This PR updates the frontend server to get the cell type names from the `celltype_metadata.json` and updates the CellGuide query module to expect strings for the GPT SEO description.
- Once this is deployed, the old GPT SEO descriptions will be migrated to the correct format (strings).

### DIFF
--- a/frontend/src/common/queries/cellGuide.ts
+++ b/frontend/src/common/queries/cellGuide.ts
@@ -341,14 +341,9 @@ export const USE_GPT_SEO_DESCRIPTION_QUERY = {
   id: "cell-guide-gpt-seo-description-query",
 };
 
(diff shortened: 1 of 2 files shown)


prathapsridharan (approved):
lgtm

Issue #5739, opened 2023-09-11
Update tissue expansion functionality with tissue file


signechambers1:
We'd like to improve the functionality of the tissue expansion and tissue filter to make viewing gene expression easier for our users. The updates are:

1. If a user expands a tissue(s), then selects a tissue using the tissue filter, the tissue should remain expanded so that users can continue to see the underlying cell types.
2. Regardless of the state of the tissue before the tissue filter is applied, filtering to a tissue(s) should expand the selected tissue

PR #5834, opened 2023-09-26
feat: 5739 WMG auto expand selected tissue

tihuan:
## Reason for Change

- #5739

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️  Do NOT merge this PR until test cases are added! ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 

Test case ticket:

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5827

## Changes

1. Move `expandedTissueIds` from component state to reducer state
2. Add new reducer actions - `toggleExpandedTissueId` and `autoExpandTissues`
3. Use reducer to mutate `expandedTissueIds` when certain user events happen
4. Use new hook `useHandleExpandedTissueIds()` to mutate `expandedTissueIds` in cases where reducer cannot detect the cases
5. Update `expandedTissueIds` from Set to [], so it's easier to work with in reducer
6. Add `hooks.ts` and move `useTrackHeatMapLoaded()` to that file (@kaloster will lead refactor using this new file structure pattern!)
7. Change `displayedCellTypes` from state to `useMemo`, since its value is derived from other states it doesn't need to be `useState` itself. Also add `useMemo` `displayedTissues` and `displayedTissueIds` to cache calculations

## Testing steps

1. See https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5827 for test cases!

DEMO:

https://github.com/chanzuckerberg/single-cell-data-portal/assets/6309723/70522b1d-9865-451f-8603-3c720e73132f


frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -301,87 +295,58 @@ export default memo(function HeatMap({
...
-      setExpandedTissues(EMPTY_SET as Set<string>);
-    }
-  }, [
-    filteredCellTypes.length,
-    initialDisplayedCellTypeIds,
-    setExpandedTissues,
-  ]);
-
-  // Reset `displayedCellTypes` and `expandedTissues` when the user clears `filteredCellTypes`
-  useEffect(() => {
```
tihuan, line 334: `filteredCellTypes` reducer will update state `expandedTisssueIds` when `filteredCellTypes` changes, which will then trigger a recalculation of `displayedCellTypes` through `useMemo`. This way we have a single source of truth for `displayedCellTypes` calculation

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -301,87 +295,58 @@ export default memo(function HeatMap({
...
-            newDisplayedCellTypes.add(tissuesByName[tissue].id + cellType.name);
-            newExpandedTissues.add(tissuesByName[tissue].id);
-          }
-        });
-      }
-    );
-    const filteredCellTypeIds = filteredCellTypes.map(
-      (cellType) => cellTypesByName[cellType].id
-    );
-    track(EVENTS.WMG_SELECT_CELL_TYPE, {
```
tihuan, line 366: this tracking happens in reducer now

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -224,71 +218,71 @@ export default memo(function HeatMap({
...
+      })
+      .filter(({ name }) => {
+        if (!filteredCellTypes.length) return true;
 
-  useEffect(() => {
-    setDisplayedCellTypes(initialDisplayedCellTypeIds);
-  }, [initialDisplayedCellTypeIds]);
+        const tissueCellTypes = sortedCellTypesByTissueName[name];
 
-  const handleExpandCollapse = useCallback(
```
tihuan, line 249: this is now just dispatching an action!

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -404,11 +337,18 @@ export default memo(function HeatMap({
...
+    displayedTissueIds,
+    dispatch,
+  });
+
   /**
    * All tissue cell types to render in YAxisCharts
    */
   const allTissueCellTypes = useMemo(() => {
-    return Object.values(tissuesByName)
+    return displayedTissues
```
tihuan, line 351: use `displayedTissues` instead of `Object.values(tissuesByName)` for a shorter array to iterate, since `displayedTissues` is filtered and doesn't contain tissues with 0 cell types

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -476,26 +416,21 @@ export default memo(function HeatMap({
             {allTissueCellTypes.map(
               ({ tissueId, tissueName, tissueCellTypes }) => {
                 return (
-                  tissueCellTypes.length > 0 && (
```
tihuan, line 479: `tissueCellTypes.length > 0` remove this guard clause since `allTissueCellTypes` no longer contains tissues with 0 cell types

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -476,26 +416,21 @@ export default memo(function HeatMap({
...
+                    handleExpandCollapse={handleExpandCollapse}
+                  />
                 );
               }
             )}
           </YAxisWrapper>
           <ChartWrapper ref={chartWrapperRef} top={xAxisHeight}>
             {allTissueCellTypes.map(({ tissueName, tissueCellTypes }) => {
-              // Don't render anything if tissue has no cell types for some reason
-              if (!tissueCellTypes.length) return;
```
tihuan, line 497: no longer needed, since since `allTissueCellTypes` no longer contains tissues with 0 cell types

frontend/src/views/WheresMyGeneV2/components/Main/index.tsx
tihuan, line 1: use the new state `expandedTissueIds`, which is an array instead of a set

kaloster (approved):
LGTM! also branch tested locally, manually as well as e2e 👍🏽

frontend/src/views/WheresMyGene/common/store/reducer.ts
```diff
@@ -424,13 +477,87 @@ function setXAxisHeight(state: State, action: PayloadAction<number>): State {
...
+  });
+
+  return {
+    ...state,
+    filteredCellTypes,
+    filteredCellTypeIds,
+    expandedTissueIds: handleExpandedTissueIds(state, action),
+  };
+
+  function handleExpandedTissueIds(
```
seve, line 505: I know they're at different scopes, but having two functions in the same file with the same name threw me off

seve (approved):
I LOVE IT!!! Thanks @kaloster and @tihuan for all your hard work refactoring this! It really looks great now :))

frontend/src/views/WheresMyGene/common/store/reducer.ts
tihuan: Yeah I was debating whether to name them differently, but since they're scoped and considered "private", I thought that would be a good enough distinction. But it did throw me off as well, especially since they're both in the same reducer file

I think I'll just add a prefix to be super explicit, and when we migrate to RTK later and split reducers out to different files, I'll change the name back to `handleExpandedTissueIds()` since then we'll have file level namespace, which hopefully won't throw anyone off anymore!

tihuan: Updated!

PR #5851, opened 2023-09-28
test: Fix e2e tests

tihuan:
## Reason for Change

- #TICKET_NUMBER

## Changes

1. Update npm commands to add back e2e test commands to run against deployed environments from local machine
2. Update `Filter to multiple cell types and then clear` test to use `PP cell` instead of `NK cell`, since it seems like `NK cell` is not available in `dev`/`staging` data - CC: @seve 
3. Some tests were failing because we're grabbing tissue names in the heat map before the tissues are rendered in the heat map. So I updated the helper function to explicitly test for tissues to be present before the rest of the test steps can run

## Testing steps

1. All e2e tests should pass as long as dev WMG endpoint works


frontend/tests/utils/wmgUtils.ts
```diff
@@ -65,14 +67,17 @@ export async function goToWMG(page: Page, url?: string) {
   const targetUrl = url || `${TEST_URL}${ROUTES.WHERE_IS_MY_GENE}`;
   return await tryUntil(
     async () => {
-      await Promise.all([
-        page.waitForResponse(
-          (resp: { url: () => string | string[]; status: () => number }) =>
-            resp.url().includes("/wmg/v2/primary_filter_dimensions") &&
```
tihuan, line 71: waiting for this response to come back is actually not reliable, because in the app we fetch `primary_filter_dimensions` twice, where the second call keeps the heat map in loading state

The double fetching is caused by snapshotId changing from `null` to a number - but not a bug to fix today since it's been in the app since inception)

frontend/tests/utils/wmgUtils.ts
```diff
@@ -292,7 +297,7 @@ export async function searchAndAddGene(page: Page, geneName: string) {
 
 export async function searchAndAddFilterCellType(page: Page, cellType: string) {
   const beforeCellTypeNames = await getCellTypeNames(page);
-  await page.getByPlaceholder(CELL_TYPE_SEARCH_PLACEHOLDER_TEXT).type(cellType);
+  await page.getByPlaceholder(CELL_TYPE_SEARCH_PLACEHOLDER_TEXT).fill(cellType);
```
tihuan, line 300: `type()` API is deprecated

frontend/tests/features/wheresMyGene/wheresMyGene.test.ts
```diff
@@ -859,14 +860,18 @@ describe("Where's My Gene", () => {
   });
   describe("Cell Type Filtering", () => {
     test("Filter to multiple cell types and then clear", async ({ page }) => {
+      const CELL_TYPE_NAMES = ["B cell", "T cell", "PP cell"];
```
tihuan, line 863: keeping code DRY and use `PP cell` over `NK cell`, since `NK cell` is not available in dev/staging WMG data

frontend/tests/features/wheresMyGene/wheresMyGene.test.ts
```diff
@@ -159,11 +159,12 @@ describe("Where's My Gene", () => {
       page,
       locator: getDiseaseSelectorButton(),
     });
-    const datasetOptions = await page
+
+    const diseaseOption = await page
```
tihuan, line 163: typo fix

frontend/tests/features/wheresMyGene/wheresMyGene.test.ts
```diff
@@ -159,11 +159,12 @@ describe("Where's My Gene", () => {
...
       .getByRole("option")
-      .getByText("acute kidney failure")
-      .elementHandles();
```
tihuan, line 165: Playwright discourages usage of elementHandles

kaloster (approved):

Issue #5461, opened 2023-08-11
Auto-deploy main to both dev and staging envs


Bento007:
This involves having the main branch GHA deploy to staging (in addition to dev, which requires no change). Practically, this mean that upon merge to main, merge main to staging branch.

This applies to both 
- [ ] ~https://github.com/chanzuckerberg/single-cell-explorer~
- [ ] https://github.com/chanzuckerberg/single-cell-data-portal

Bento007:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5461?planning-poker) with Zenhub @tihuan

Bento007:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5461?planning-poker) with Zenhub @ebezzi

prathapsridharan:
@atolopko-czi @tihuan (cc: @Bento007 ) - In our meeting on 09/28/2023 we agreed that to enable a `staging` deploy when a commit is pushed to `main`, we can leverage our current workflow by merging the `main` branch into `staging` branch which would [create a deployment on a merge to staging branch](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d042bdadb17f2fa389f537ed26eb429698350596/.github/workflows/build-images-and-create-deployment.yml#L80) and then [kick off the creation of staging stack](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/.github/workflows/deploy-happy-stack.yml)

This is all well and good. But I have a couple of questions:

_What happens if there is a burst of commits to `main`?_ Per above, we would:

```
git checkout staging
git merge --no-ff main -m "Merge main into staging"
git push
```

Each of these merges to `staging` would trigger a creation of a staging deployment and creation of a stack. 

1. _Would these workloads just get queued up?_ This would mean stack creation could take a long time. I think this problem is no worse than it is currently for `dev` deployment when there is a burst of commits to `main` branch but just want to check with y'all.

2. _Is it possible for stacks to be created out of order due to a race condition?_ That is `commit A` _precedes_ `commit B` and _what you expect to see in staging is a build containing `commit B`_, but due to a race condition of some sort, `commit B` is processed first and then `commit A` is processed and what you have is `commit A` on staging.

[...]

Issue #5847, opened 2023-09-28
Turn off dev deployment from main [FINAL STEP]


atolopko-czi:
Turn off dev deployment from main

PR #5860, opened 2023-09-29
feat: Auto merge and deploy to staging branch

prathapsridharan:
## Reason for Change

- #5461

The idea is that we can leverage our existing workflow where a commit to `staging` branch creates a deployment to the `stage` environment. So, whenever new commits to `main` come in, we merge the commit to `staging` thus kicking off a deployment to `stage` environment.

Additionally, github action's [concurrency group feature is utilized](https://docs.github.com/en/actions/using-jobs/using-concurrency) to eliminate race conditions involving multiple builds and deployments run at the same time. Here, we cancel scheduled builds and deployments as well as currently running builds and deployments in favor of the newest build/deployment job.

## Changes

- add merging `main` branch to `staging` branch on new commits to `main`
- add GHA concurrency group feature

## Testing steps

- The following workflow was used to test a merge between 2 branches not named `[dev, staging, prod]`. Specifically the following github workflow merges commits from `psridharan/issue-5461` into `psridharan/test-auto-merge-2` triggered by a pushes to `psridharan/issue-5461`:

```
name: Trigger release candidate build and deploy

on:
  push:
    branches:
      - psridharan/issue-5461

jobs:
  merge:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout psridharan/issue-5461 into psridharan/test-auto-merge-2
        uses: actions/checkout@v2
        with:
          ref: psridharan/issue-5461
          fetch-depth: 0
      - name: Merge psridharan/issue-5461 into psridharan/test-auto-merge-2
        run: |
          git config user.name "GitHub Actions"
          git config user.email "actions@github.com"

          echo "BEFORE CHECKOUT: Listing current checked out branches:"
          echo "$(git branch --list)"

          echo "Checking out psridharan/test-auto-merge-2"
          git checkout psridharan/test-auto-merge-2
          echo "Confirming checked out branch receiving merge is: $(git branch --show-current)"

          echo "AFTER CHECKOUT: Listing current checked out branches:"
          echo "$(git branch --list)"

          echo "Current checked out receiving branch is: $(git branch --show-current)"
          echo "HEAD commit in current checked out receiving branch: $(git branch --show-current) is: $(git log -1 --format='%H')"

          echo "HEAD commit in incoming branch psridharan/issue-5461 is:"
          echo "$(git log -1 --format='%H' psridharan/issue-5461)"

          echo "About to merge psridharan/issue-5461 into psridharan/test-auto-merge-2"
          git merge --verbose --no-ff psridharan/issue-5461 -m "Merging psridharan/issue-5461 into psridharan/test-auto-merge-2"

          git push origin psridharan/test-auto-merge-2
```

- Here is the DEBUG log output that inspects the state before the `git merge` operation:

```
BEFORE CHECKOUT: Listing current checked out branches:
* psridharan/issue-5461
Checking out psridharan/test-auto-merge-2
Switched to a new branch 'psridharan/test-auto-merge-2'
branch 'psridharan/test-auto-merge-2' set up to track 'origin/psridharan/test-auto-merge-2'.
Confirming checked out branch receiving merge is: psridharan/test-auto-merge-2
AFTER CHECKOUT: Listing current checked out branches:
  psridharan/issue-5461
* psridharan/test-auto-merge-2
Current checked out receiving branch is: psridharan/test-auto-merge-2
HEAD commit in current checked out receiving branch: psridharan/test-auto-merge-2 is: baf69efb5f6f94c2d6645e3eedaf4d8703852ccc
HEAD commit in incoming branch psridharan/issue-5461 is:
4b12fc3e[27](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6386570986/job/17333448676?pr=5860#step:3:28)50f6af67bae5ca[28](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6386570986/job/17333448676?pr=5860#step:3:29)354502e9e[33](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6386570986/job/17333448676?pr=5860#step:3:34)bba
About to merge psridharan/issue-5461 into psridharan/test-auto-merge-2
Merge made by the 'ort' strategy.
 .../build-images-and-create-deployment.yml         |  4 +++
 .github/workflows/deploy-happy-stack.yml           |  5 +++
 .../trigger-release-candidate-build-and-deploy.yml | 41 ++++++++++++++++++++++
 3 files changed, 50 insertions(+)
 create mode 100644 .github/workflows/trigger-release-candidate-build-and-deploy.yml
To https://github.com/chanzuckerberg/single-cell-data-portal
   baf69efb5f6..64a[37](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/6386570986/job/17333448676?pr=5860#step:3:38)11d333  psridharan/test-auto-merge-2 -> psridharan/test-auto-merge-2
```

Attached is an image that shows the auto-merge

Screenshot 2023-10-02 at 4 31 47 PM

### DIFF
--- a/.github/workflows/build-images-and-create-deployment.yml
+++ b/.github/workflows/build-images-and-create-deployment.yml
@@ -10,7 +10,7 @@ on:
 # add a concurrency group to prevent multiple builds from running at the same time
 concurrency:
   group: $${{ github.ref }}-environment
-  cancel-in-progress: true
+  cancel-in-progress: false
 env:
   # Force using BuildKit instead of normal Docker, required so that metadata
   # is written/read to allow us to use layers of previous builds as cache.
--- a/.github/workflows/build-images-and-create-deployment.yml
+++ b/.github/workflows/build-images-and-create-deployment.yml
@@ -7,6 +7,10 @@ on:
       - staging
       - prod
 
+# add a concurrency group to prevent multiple builds from running at the same time
+concurrency:
+  group: $${{ github.ref }}-environment
+  cancel-in-progress: true
 env:
   # Force using BuildKit instead of normal Docker, required so that metadata
   # is written/read to allow us to use layers of previous builds as cache.
--- a/.github/workflows/merge-main-into-staging.yml
+++ b/.github/workflows/merge-main-into-staging.yml
@@ -1,0 +1,21 @@
+name: Merge main into staging
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  merge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Merge main into staging
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+          git checkout staging
+          git pull
+          git merge --no-ff main -m "Merge main into staging"
+          git push
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -9,20 +9,23 @@ jobs:
   merge:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
+      - name: Checkout code from psridharan/issue-5461
+        uses: actions/checkout@v2
+        with:
+          ref: psridharan/issue-5461
+      - name: Checkout code from psridharan/test-auto-merge-2
         uses: actions/checkout@v2
         with:
           ref: psridharan/test-auto-merge-2
-      - name: Fetch latest changes to psridharan/issue-5461
-        run: git fetch origin psridharan/issue-5461
       - name: Merge psridharan/issue-5461 into psridharan/test-auto-merge-2
         run: |
           git config user.name "GitHub Actions"
           git config user.email "actions@github.com"
           echo "Checkout psridharan/test-auto-merge-2"
           git checkout psridharan/test-auto-merge-2
-          echo "Current checkout branch is: $(git branch --show-current)"
-          echo "HEAD commit in current checkout branch is: $(git log -1 --format='%H')"
+          echo "Current checked out receiving branch is: $(git branch --show-current)"
+          echo "HEAD commit in current checked out receiving branch: $(git branch --show-current) is: $(git log -1 --format='%H')"
+          echo "HEAD commit in incoming branch psridharan/issue-5461 is: $(git log -1 --format='%H' psridharan/issue-5461)"
           echo "About to merge psridharan/issue-5461 into psridharan/test-auto-merge-2"
           git merge --verbose --no-ff psridharan/issue-5461 -m "Merging psridharan/issue-5461 into psridharan/test-auto-merge-2"
           git push origin psridharan/test-auto-merge-2
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -3,39 +3,37 @@ name: Trigger release candidate build and deploy
 on:
   push:
     branches:
-      - psridharan/issue-5461
+      - main
 
 jobs:
   merge:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout psridharan/issue-5461 into psridharan/test-auto-merge-2
+      - name: Checkout main branch and staging branch
         uses: actions/checkout@v2
         with:
-          ref: psridharan/issue-5461
-          fetch-depth: 0
-      - name: Merge psridharan/issue-5461 into psridharan/test-auto-merge-2
+          ref: main
+          fetch-depth: 0 # fetch all history to make merges work correctly
+      - name: Merge main branch into staging branch
         run: |
           git config user.name "GitHub Actions"
           git config user.email "actions@github.com"
 
-          echo "BEFORE CHECKOUT: Listing current checked out branches:"
+          echo "Listing locally available branches BEFORE 'staging' branch is checked out:"
           echo "$(git branch --list)"
 
-          echo "Checking out psridharan/test-auto-merge-2"
-          git checkout psridharan/test-auto-merge-2
+          echo "Checking out 'staging' branch"
+          git checkout staging
           echo "Confirming checked out branch receiving merge is: $(git branch --show-current)"
 
-          echo "AFTER CHECKOUT: Listing current checked out branches:"
+          echo "Listing locally available branches AFTER 'staging' branch is checked out:"
           echo "$(git branch --list)"
 
-          echo "Current checked out receiving branch is: $(git branch --show-current)"
           echo "HEAD commit in current checked out receiving branch: $(git branch --show-current) is: $(git log -1 --format='%H')"
 
-          echo "HEAD commit in incoming branch psridharan/issue-5461 is:"
-          echo "$(git log -1 --format='%H' psridharan/issue-5461)"
+          echo "HEAD commit in incoming branch: 'main' is: $(git log -1 --format='%H' main)"
 
-          echo "About to merge psridharan/issue-5461 into psridharan/test-auto-merge-2"
-          git merge --verbose --no-ff psridharan/issue-5461 -m "Merging psridharan/issue-5461 into psridharan/test-auto-merge-2"
+          echo "About to merge 'main' branch into 'staging' branch"
+          git merge --verbose --no-ff main -m "Merging main branch into staging branch"
 
-          git push origin psridharan/test-auto-merge-2
+          git push origin staging
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -17,8 +17,9 @@ jobs:
         run: |
           git config user.name "GitHub Actions"
           git config user.email "actions@github.com"
-          git branch --show-current
-          git log -1 --format='%H'
-          git rev-parse psridharan/issue-5461
+          echo "Current checkout branch is: $(git branch --show-current)"
+          echo "HEAD commit in current checkout branch is: $(git log -1 --format='%H')"
+          echo "HEAD commit in psridharan/issue-5461 branch is: "
+          echo "$(git rev-parse psridharan/issue-5461)"
           git merge --no-ff psridharan/issue-5461 -m "Merging psridharan/issue-5461 into psridharan/test-auto-merge-2"
           git push origin psridharan/test-auto-merge-2
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -19,7 +19,9 @@ jobs:
           git config user.email "actions@github.com"
           echo "Current checkout branch is: $(git branch --show-current)"
           echo "HEAD commit in current checkout branch is: $(git log -1 --format='%H')"
+          echo "Checking branches available in the repo:"
+          echo "$(git branch)"
           echo "HEAD commit in psridharan/issue-5461 branch is: "
-          echo "$(git rev-parse origin/psridharan/issue-5461)"
+          echo "$(git rev-parse psridharan/issue-5461)"
           git merge --no-ff psridharan/issue-5461 -m "Merging psridharan/issue-5461 into psridharan/test-auto-merge-2"
           git push origin psridharan/test-auto-merge-2
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -20,6 +20,6 @@ jobs:
           echo "Current checkout branch is: $(git branch --show-current)"
           echo "HEAD commit in current checkout branch is: $(git log -1 --format='%H')"
           echo "HEAD commit in psridharan/issue-5461 branch is: "
-          echo "$(git rev-parse psridharan/issue-5461)"
+          echo "$(git rev-parse origin/psridharan/issue-5461)"
           git merge --no-ff psridharan/issue-5461 -m "Merging psridharan/issue-5461 into psridharan/test-auto-merge-2"
           git push origin psridharan/test-auto-merge-2
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -9,8 +9,12 @@ jobs:
   merge:
     runs-on: ubuntu-latest
     steps:
-      - name: Merge psridharan/issue-5461 into psridharan/test-auto-merge-2
+      - name: Checkout psridharan/issue-5461 into psridharan/test-auto-merge-2
         uses: actions/checkout@v2
+        with:
+          ref: psridharan/issue-5461
+          fetch-depth: 0
+      - name: Merge psridharan/issue-5461 into psridharan/test-auto-merge-2
         run: |
           git config user.name "GitHub Actions"
           git config user.email "actions@github.com"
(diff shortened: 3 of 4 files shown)


.github/workflows/merge-main-into-staging.yml
```diff
@@ -0,0 +1,21 @@
...
+    branches:
+      - main
+
+jobs:
+  merge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Merge main into staging
```
atolopko-czi, line 14: Might rename this to indicate the underlying intention: "Trigger release candidate build and deploy". In fact maybe rename the entire workflow similarly?

.github/workflows/merge-main-into-staging.yml
```diff
@@ -0,0 +1,21 @@
...
+  push:
+    branches:
+      - main
+
+jobs:
+  merge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
atolopko-czi, line 13: consider using `ref: staging` instead of `git checkout staging` later.

We might also be able to get away with `fetch-depth: 1` (for checkout speed), if we later do a merge from main as `git pull origin main`

.github/workflows/merge-main-into-staging.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Merge main into staging
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+          git checkout staging
+          git pull
+          git merge --no-ff main -m "Merge main into staging"
```
atolopko-czi, line 20: Is the `--no-ff` strictly necessary? Just curious.

atolopko-czi (changes requested):
Exciting! A couple suggestions for your consideration.

.github/workflows/merge-main-into-staging.yml
prathapsridharan: _With the `--no-ff` option on it creates a merge commit in all cases_. This means every merge into `staging` will be merge commit even if the current HEAD of `staging` is an ancestor of the incoming commit from `main`. Without the `--no-ff` you won't see a merge commit on `staging` for the above case - instead the branch pointer would be moved up to point to the commit at the HEAD of `main`.

From the perspective of someone looking at the `staging` branch, seeing a merge commit like `Merge ...` more easily allows them to see that the GHA executed and what was added from `main` when this GHA executes.

I could remove it if you feel strongly that explicit merge commits are not helpful and are too noisy

.github/workflows/merge-main-into-staging.yml
prathapsridharan: will fix

.github/workflows/merge-main-into-staging.yml
prathapsridharan: will use `ref: staging`

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  merge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout main branch and staging branch
+        uses: actions/checkout@v2
+        with:
+          ref: main
+          fetch-depth: 0 # fetch all history to make merges work correctly
```
atarashansky, line 16: We set fetch-depth to 1 in order to prevent the checkout step from taking upwards of 10 minutes (our repo is super bloated and fetching all the commits takes forever). Is it possible to figure out how to fetch only the commits that are needed?

atarashansky (changes requested):

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: @atarashansky - I _think_ it is possible. Perhaps there is a way to use `git fetch --shallow-since`, but I am not sure yet.

But since this work is towards an MVP of continuous delivery to `staging`, I don't think this PR needs to be blocked on the fetch-depth issue. If it is taking too long, we can always turn off this workflow.

There are a couple of challenges with figuring out the issue of fetching only the commits from the incoming `main` branch that is:

1. Testing is pretty onerous for GHA stuff. As far as my knowledge is goes with GHA, I would have to keep pushing to this PR to demonstrate a test with test branches. Just look at the number of commits I had to make just to demonstrate that correctness of auto-merge

2. Concurrency poses problems. Say you have multiple commits into `main` in rapid succession that trigger this merge to `staging`. From my reading, there is no order guarantee of how these commits are processed. I don't think something like `fetch-depth=x` will work in this case. The fetching scheme under concurrency needs to be tested for correctness and the more complex the fetching scheme is, the harder it is to test, and the less sure we are about its correctness under concurrency. `fetch-depth=0` does not have this problem under concurrency.

My suggestion is to tackle the fetching scheme in a subsequent PR rather than hold this back for an MVP.

atarashansky: OK - sounds good!

atarashansky (approved):

.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -7,6 +7,10 @@ on:
       - staging
       - prod
 
+# add a concurrency group to prevent multiple builds from running at the same time
+concurrency:
```
Bento007, line 11: Is this a new feature? I love it! We should add something like this for rdev created for a PR.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Merge main branch into staging branch
```
Bento007, line 17: I thought the plan was deploy main to the staging environment and removing the staging branch?

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+
+          echo "Listing locally available branches BEFORE 'staging' branch is checked out:"
```
Bento007, line 22: Why do we need to do this list?

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -0,0 +1,39 @@
...
+          echo "$(git branch --list)"
+
+          echo "Checking out 'staging' branch"
+          git checkout staging
+          echo "Confirming checked out branch receiving merge is: $(git branch --show-current)"
+
+          echo "Listing locally available branches AFTER 'staging' branch is checked out:"
+          echo "$(git branch --list)"
+
+          echo "HEAD commit in current checked out receiving branch: $(git branch --show-current) is: $(git log -1 --format='%H')"
```
Bento007, line 32: > "HEAD commit in current checked out receiving branch:

I'm not sure what this means?

Bento007 (commented):
Some of the language you use in the echo statement is confusing to understand. Is there anyway to make it more clear.

prathapsridharan:
> Some of the language you use in the echo statement is confusing to understand. Is there anyway to make it more clear.

@Bento007 - I can clarify some of the echo messages. I initially had them for debugging to understand the state prior to merge but I will clean them up

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: will clean the echo messages up

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: @Bento007 (cc: @atolopko-czi ) -

Here we want to do the simplest thing to demonstrate the pipeline end-to-end. It is not the most efficient, but deploying `main` to `staging` and reusing the existing `deploy-happy-stack.yml` to deploy to `stage` env seems to be the simplest thing to do.

The most efficient route is to build only once and deploy the builds to different environments (`staging`, `prod`, etc) but that is a bit complex in terms of implementation and testing.

There are some additional benefits to this simple approach too:

1. Easier to test. GHA, as far as I understand, is pretty onerous to test. The number of commits in this PR to demonstrate auto-merge is an example of the pain point. The simplest path to the solution also happens to be the shortest path - just merge `main` to `staging` - which means easier to demonstrate a test of correctness.
2. Concurrency introduces complexity (ex: many commits to `main` coming in at once) and to the extent we can keep things straightforward, will allow us to understand the complexities of concurrency and then we can determine a more efficient solution
3. Easier to reason about in the beginning (Uniformity) - The merge-commit that goes from `main` to `staging` is replicates our documented manual steps for a `stage` deploy thus making continuous deployment to staging easier to reason about.
4. Easy to rollback. If continuous deployment to `stage` is running into problems, it is as simple as disabling the GHA workflow that merges `main` to `staging` and we are back to our starting point.

.github/workflows/build-images-and-create-deployment.yml
prathapsridharan: @Bento007 - I don't know if it is a new feature but this was suggested by @atolopko-czi when I brought up the issue of concurrency and order of operations:

https://github.com/chanzuckerberg/single-cell-data-portal/issues/5461#issuecomment-1740899948

Bento007 (approved):

.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -7,6 +7,10 @@ on:
...
+  group: $${{ github.ref }}-environment
+  cancel-in-progress: false
```
atolopko-czi, line 13: curious, what is the issue with canceling "in progress" workflows?

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: agreed, reasonble to take on later. can you add an issue to the [post-mvp](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/5457) epic?

atolopko-czi: some possible solutions usings --merge-base and --deepen options, referenced in these [search results](https://www.google.com/search?q=git+clone+sufficient+history+to+merge+two+branches), but no clear immediate win.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: I agree w/most of this and the overall decision.

For 2, haven't we addressed the concurrency issue with `concurrency` config? If not, isn't the concurrency concern for `staging` branch merges going to be the same as it would be for `main`-based staging deploys?

For 1, most of the GHA-dev/debug cycle was for the main->staging auto-merge right? So possibly might have been shorted dev time if deployed to stagin from `main`? (I haven't studies all the commits!)

.github/workflows/build-images-and-create-deployment.yml
atolopko-czi, line 1: Shouldn't we also remove `main` from `on: push: branches:` now?

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: Another way entirely around merging would be to destroy and recreate the staging branch every time.

.github/workflows/build-images-and-create-deployment.yml
Bento007: I don't know the exact reason but there is a lot of state changes that happens as a result of the deployment. At the very least there are test fixtures created and happy deployment configuration files that are modified. Interrupting a deployment could leave it in an unknown state. If we do cancel in progress workflows we should be mindful of when and where we allow it. I think determining that would be outside the scope of this PR.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: @atolopko-czi 

For concurrency (2), we have a provisional solution with `concurrency` feature but we haven't tested it. My point is that it is hard to see right now what kinds of concurrency gotchas we might run into.

For the testing point made in (1), if I understand you correctly, you are proposing committing the change to `main` and then see if it actually auto-merges to `staging`? In that case it would take the same number of attempts (see the number of commits to test auto-merge between test branches) and moreover, pollute `main` with a bunch of these "attempting-fix" commits.

In the absence of unit tests, I think it is preferable to demonstrate tests in testing sandbox - in this case, using a couple of test branches

.github/workflows/build-images-and-create-deployment.yml
prathapsridharan: @atolopko-czi @Bento007 - `cancel-in-progress: true` seems like a blunt force instrument. It might open the [possibility of corrupting state](https://github.com/orgs/community/discussions/26311). I think it `cancel-in-progress: true` could work out for us, but I feel starting conservatively here is the best way to go.

Note that not setting `cancel-in-progress` (`false` is the default), won't prevent the `concurrency` feature from [canceling pending tasks](https://docs.github.com/en/actions/using-jobs/using-concurrency)

So we still save on time here by using the default `concurrency` feature

.github/workflows/build-images-and-create-deployment.yml
prathapsridharan: We still want a `dev` deployment right as per the ticket: https://github.com/chanzuckerberg/single-cell-data-portal/issues/5461

Removing `dev` deployment is in [another ticket](https://github.com/chanzuckerberg/single-cell-data-portal/issues/5847) and that is when we remove `main` from `on: push: branches:` in this file I think

atolopko-czi: 💯 Forgot about the issue split!

.github/workflows/build-images-and-create-deployment.yml
atolopko-czi: Thanks, both. Good justifications. Maybe just add a TODO and/or issue to revisit this later?

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: > For the testing point made in (1), if I understand you correctly, you are proposing committing the change to main and then see if it actually auto-merges to staging?  

Not actually proposing anything :) We can take this offline, since I was just thinking aloud.

.github/workflows/build-images-and-create-deployment.yml
prathapsridharan: @atolopko-czi - Added a ticket here: https://github.com/chanzuckerberg/single-cell-data-portal/issues/5895

atolopko-czi (approved):
LGTM, with 2 requests to add follow-up issues for future improvements:
- remove need for fetch-depth=0 (full clone)
- revisit canceling in-progress builds

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: @atolopko-czi - I have created a ticket here to capture the fact that we might want to improve the speed of a `stage` deploy: https://github.com/chanzuckerberg/single-cell-data-portal/issues/5896

prathapsridharan:
> LGTM, with 2 requests to add follow-up issues for future improvements:
> 
> * remove need for fetch-depth=0 (full clone)
> * revisit canceling in-progress builds

@atolopko-czi  - Added issues #5895 and #5896 to address these

.github/workflows/merge-main-into-staging.yml
danieljhegeman: i might argue against it since we will be deploying ~continuously~ 😄

atolopko-czi: Agreed--since _every commit_ to main is going to be merged to staging, it's probably not adding much to see the merge commit.

prathapsridharan: will remove

PR #5886, opened 2023-10-03
fix: Add concurrency to rdev managing GHAs

Bento007:
## Reason for Change

-  There is a chance the rdev environment can get into an unknown state if multiple happy updates are run on the same environment concurrently. This PR adds some locking to prevent happy from running for the same rdev in different GHA at the same time. 

## Changes

- add concurrency locking of the happy update/delete jobs to prevent concurrent happy modification.

## Testing steps

- Created an rdev using a PR. Adding additional commits and check that the GHA triggered by the latest commit waited for the previous GHA to finish.

### DIFF
--- a/.github/workflows/rdev-delete-for-pr.yml
+++ b/.github/workflows/rdev-delete-for-pr.yml
@@ -7,11 +7,6 @@ on:
     types:
       - closed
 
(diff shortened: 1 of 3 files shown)


prathapsridharan (approved):
lgtm

PR #5891, opened 2023-10-04
feat: persist additional dataset metadata in DB and return in Discover API

nayib-jose-gloria:
## Reason for Change

- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/420

## Changes

- added DatasetMetadata entity optional fields
    - feature count (var length)
    - feature_biotype (set of values present)
    - feature_reference (set of values present)
    - default_embedding 
    - embeddings present (obsm keys)
    - raw_data_location (name of matrix in H5AD that contains raw data)
- extract fields above during dataset processing to persist in DB dataset_metdata JSON blob 
- added fields to API response + OpenAPI spec for Discover API GET /v1/collections/{collection_id} and GET /v1/collections/{collection_id}/datasets/{dataset_id} and corresponding dataset_version endpoints (not index endpoint)

## Testing steps
- unit tests + rdev upload successful (API returns expected responses)

## Notes for Reviewer
- waiting on response from product for whether these fields should appear in any other API endpoint response (e.g. dataset index)


backend/layers/common/entities.py
```diff
@@ -178,6 +178,11 @@ class DatasetMetadata:
     donor_id: List[str]
     is_primary_data: str
     x_approximate_distribution: Optional[str]
+    default_embedding: Optional[str] = None
+    embeddings: Optional[List[str]] = None
+    feature_biotype: Optional[List[str]] = None
+    feature_count: Optional[int] = None
+    feature_reference: Optional[List[str]] = None
```
nayib-jose-gloria, line 185: note: I'm realizing now that, without backfilling the previous DatasetVersions, these will need to stay as Optionals with None as the default arg otherwise we'll get errors fetching prior versions. Not sure if its worth considering a backfill plan as a follow-up (could we get away with just appending these attributes with a null field for all previous versions, so we don't need to reprocess old datasets?)

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -202,13 +202,18 @@ def reshape_dataset_for_curation_api(
     dataset_version: DatasetVersion,
     use_canonical_url: bool,
     preview=False,
+    index=False,
```
nayib-jose-gloria, line 205: going to tackle a refactor in a separate to see if we can avoid adding this new flag, otherwise this seems like the path of least resistance for the functional change

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -202,13 +202,18 @@ def reshape_dataset_for_curation_api(
...
 ) -> dict:
     ds = dict()
 
     # Determine what columns to include from the dataset
-    columns = EntityColumns.dataset_metadata_preview_cols if preview else EntityColumns.dataset_metadata_cols
+    columns = EntityColumns.dataset_metadata_cols
+    if preview:
+        columns = EntityColumns.dataset_metadata_preview_cols
+    elif index:
+        columns = EntityColumns.dataset_metadata_index_cols
```
Bento007, line 216:
```suggestion
        columns = EntityColumns.dataset_metadata_index_cols
    else:
        columns = EntityColumns.dataset_metadata_cols
    
```

backend/layers/processing/process_download_validate.py
```diff
@@ -154,11 +154,16 @@ def _get_batch_condition() -> Optional[str]:
             mean_genes_per_cell=numerator / denominator,
             is_primary_data=_get_is_primary_data(),
             cell_type=_get_term_pairs("cell_type"),
-            x_approximate_distribution=_get_x_approximate_distribution(),  # TODO: pay attention
```
Bento007, line 157: 🙅 what a helpful TODO message?

backend/layers/common/entities.py
Bento007: We can skip back filling as long as they are optional. These will get added when we migrate anyways right? We can default to None during the migration if it doesn't exists.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -151,6 +151,11 @@ def mock_config_fn(name):
             donor_id=["test_donor_1"],
             is_primary_data="BOTH",
             x_approximate_distribution="normal",
+            default_embedding="X_embedding_1",
```
Bento007, line 154: If we were using pytest fixture this would be easy to factor out and avoid repeating.

Bento007 (commented):
Still reviewing. Here are my initial thoughts

backend/layers/processing/process_download_validate.py
nayib-jose-gloria: I know LOL

I have no idea what this is referring to, so I guess I was not paying sufficient attention

backend/curation/api/curation-api.yml
```diff
@@ -1313,15 +1341,47 @@ components:
...
+      nullable: true
     explorer_url:
       description: |
         The URL at which a given Dataset may be explored using CELLxGENE Explorer.
       nullable: true
       type: string
-    donor_id:
-      description: Free-text that identifies a unique individual that data were derived from.
+    feature_biotype:
+      description: List of feature biotypes in dataset. Must be
```
danieljhegeman, line 1362: typo?

backend/curation/api/v1/curation/collections/common.py
danieljhegeman: Agreed

danieljhegeman (commented):
Approve if @Bento007 also approves; see minor suggestion

backend/curation/api/curation-api.yml
nayib-jose-gloria: The diff is showing up weirdly, but the donor_id isn't being deleted. I just moved it above 'explorer_url' to maintain alphabetical ordering

Bento007: > List of feature biotypes in dataset. Must be

Why does this line just end?

Bento007 (approved):

PR #5894, opened 2023-10-04
feat: implement expression summary arrays builder

atarashansky:
## Reason for Change

- #5866 

## Notes for reviewer
- This is being merged into a feature branch and not intended to be complete.
- Please focus the review on the builder logic and architecture.


backend/wmg/data/rankit.py
```diff
@@ -15,7 +15,7 @@ def quantiles(max_rank: int, ranks: np.ndarray) -> np.ndarray:
     return np.array([np.round((i - 0.5) / max_rank, 5) for i in ranks])
 
 
-def rankit(Xraw: sc.sparse.spmatrix, offset: float = 3.0) -> sc.sparse.csr_matrix:
+def rankit(X: sparse.csr_matrix, offset: float = 3.0):
```
atarashansky, line 18: rankit was modified to alter the input in-place

backend/wmg/data/schemas/cube_schema.py
```diff
@@ -22,8 +22,6 @@
     "cell_type_ontology_term_id",
     "tissue_original_ontology_term_id",
     "dataset_id",
-    "assay_ontology_term_id",
```
atarashansky, line 25: removing these to shrink the cube by 66% as they are not needed by the application

backend/wmg/data/schemas/cube_schema.py
```diff
@@ -33,7 +31,7 @@
 # The full set of logical cube dimensions by which the cube can be queried.
 expression_summary_logical_dims = expression_summary_indexed_dims + expression_summary_non_indexed_dims
 
-filters = [tiledb.ZstdFilter(level=+22)]
+filters = [tiledb.DictionaryFilter(), tiledb.ZstdFilter(level=+19)]
```
atarashansky, line 34: applying the dictionary filter to dictionary-encode strings. this reduces cube size from 8gb to 2.7gb.

backend/wmg/data/schemas/cube_schema.py
```diff
@@ -43,9 +41,11 @@
 )
 
 # The cube attributes that comprise the core data stored within the cube.
+filters_numeric = [tiledb.ByteShuffleFilter(), tiledb.ZstdFilter(level=+5)]
 expression_summary_logical_attrs = [
-    tiledb.Attr(name="nnz", dtype=np.uint64, filters=filters),  # TODO: Why uint64?
-    tiledb.Attr(name="sum", dtype=np.float32, filters=filters),
+    tiledb.Attr(name="nnz", dtype=np.uint64, filters=filters_numeric),  # TODO: Why uint64?
+    tiledb.Attr(name="sum", dtype=np.float32, filters=filters_numeric),
+    tiledb.Attr(name="sqsum", dtype=np.float32, filters=filters_numeric),
```
atarashansky, line 48: adding sqsum to expression_summary cube - this is required for generating the default and fmg cubes (the latter two cubes are now being generated directly from the expression_summary cube).

adding new attributes only negligibly affects query times. considering that the expression summary cube is much smaller now after removing assay and development logical dimensions, query times will still be significantly faster compared to what they are currently.

backend/wmg/data/snapshot.py
```diff
@@ -11,10 +11,6 @@
 
 from backend.common.utils.s3_buckets import buckets
 from backend.wmg.config import WmgConfig
-from backend.wmg.data.schemas.corpus_schema import (
```
atarashansky, line 14: this module no longer exists

backend/wmg/data/schemas/cube_schema.py
prathapsridharan: 1. why did you choose to greater compression? Typically, greater compression comes at the cost of query speed

2. What made you bump the compression level of Zstandard from 22 to 19?

backend/wmg/data/schemas/cube_schema.py
```diff
@@ -43,9 +41,11 @@
 )
 
 # The cube attributes that comprise the core data stored within the cube.
+filters_numeric = [tiledb.ByteShuffleFilter(), tiledb.ZstdFilter(level=+5)]
```
prathapsridharan, line 44: Related to the compression question earlier - what made you choose this combination of byte shuffling and the applying zstandard at level 5?

backend/wmg/pipeline/summary_cubes/builder.py
```diff
@@ -0,0 +1,435 @@
+import logging
```
prathapsridharan, line 1: where is the `builder.py` module invoked in this PR?

prathapsridharan (changes requested):
If I am not mistaken your general strategy is to first remove the old way of building cubes and secondly add the new way of building cubes iteratively.

The reason why I think that is because you have commented out `pipelines/summary_cubes/__init__.py` and added `builder.py` - which builds the expression summary cubes

I am not able to see how this will be executed yet (perhaps, there are previous PRs you have merged that has the background context?). But let me ask a couple of high level question here without strongly advocating for any particular solution:

1. Delivering this feature iteratively is good, but what happens when there is an unforeseen injection in production while this transition is happening that requires the old WMG cube to be fixed and regenerated (ex: data corruption or OOM error due to some new data characteristics)? 
2. What happens when you run into a snag in the next PR (say, `cell_counts_cube` builder). This increases the timeline to deliver the transition to census-based cubes and if we have removed code to generate the old-style WMG cube and the problem I have stated in point (1) arises, then we might be in for a scramble
3. The old pipeline code has tests, but the new one does not. Would it be safer to keep the old code to rollback to as a safety net?

The general strategy to address this scenario, to add new code alongside the old code and switch portions of the old code to the new code as and when it becomes ready using config flags in the code.

One justification to "burn the bridge behind you" (delete old code) is that the new code and accompanying tests and demonstration of correctness is very close

backend/wmg/pipeline/summary_cubes/builder.py
atarashansky: It's not invoked in this PR.

backend/wmg/data/schemas/cube_schema.py
atarashansky: Bruce gave me performance optimization suggestions - these are the filters used in census for numeric attributes.

backend/wmg/data/schemas/cube_schema.py
atarashansky: 1. In this case, DictionaryFilter will speed up queries because it dictionary encodes bulky variable-length strings. Queries on integer arrays are much faster than queries on arrays with variable-length types.
2. A small performance optimization suggested by Bruce. Typically, the disk space savings you get with increasing levels of compression follow diminishing returns. I chose 19, but it could very well be 5. I haven't empirically searched across compression levels yet. In any case, 22 is overkill.

backend/wmg/pipeline/summary_cubes/builder.py
```diff
@@ -0,0 +1,435 @@
...
+            corpus_path (str): The path to the corpus.
+            organismInfo (dict): Information about the organism.
+                ex:
+                organismInfo = {
+                    "id": "UBERON:0000955",
+                    "label": "brain",
+                }
+        """
+        organism = organismInfo["label"]
+        organismId = organismInfo["id"]
```
prathapsridharan, line 97: do you mean `organ` instead of `organism` here? The example of `brain` suggests to organ to me. Or perhaps the example is wrong it should be `homo sapiens` or something?

backend/wmg/pipeline/summary_cubes/builder.py
```diff
@@ -0,0 +1,435 @@
...
+        self,
+        *,
+        schema: tiledb.ArraySchema,
+        cube_index: pd.DataFrame,
+        other_cube_attrs: list,
+        cube_sum: np.ndarray,
+        cube_nnz: np.ndarray,
+        cube_sqsum: np.ndarray,
+    ):
+        """
```
prathapsridharan, line 374: nit: missing return type annotation which is a tuple

prathapsridharan (approved):
I just have some minor comments but LGTM

backend/wmg/pipeline/summary_cubes/builder.py
atarashansky: Oh shoot, lol, nice catch! That's what I get for having copilot autocomplete my docstring.

Issue #5940, opened 2023-10-09
Prevent DDoS from FE


tihuan:
Context
1. Recently we discovered that there's an infinite loop in FE that would send infinite API calls to DDoS BE

Solution:
1. FE now adds a network guard to detect excessive API calls within a 5 second sliding window and shut down API calls
2. Add a global assertion to catch DDoS in every e2e test

dsadgat:
@tihuan can you move this to the appropriate swim lane pls? TY

tihuan:
Done! Sorry was hoping this would get merged on Monday. Added to the current  sprint!

I'll ask Ronen to help get this reviewed and merged!

PR #5941, opened 2023-10-09
feat: Add network guard

tihuan:
## Reason for Change

- #5940

## Changes

1. Add a network guard in `window.fetch()` to catch excessive API calls
2. Add network guard in every e2e test to detect excessive API calls, so we're aware of it in test logs

## Testing steps

1. Temporarily change `handleChange()` in frontend/src/views/WheresMyGene/components/Filters/components/Compare/connect.ts to depend on `optionLabel`, which will generate a new `handleChange` function every time `optionLabel` changes. Which is the crux of the infinite loop
2. Navigate to https://localhost:3000/gene-expression?compare=disease&sexes=PATO%3A0000383&tissues=blood%2Clung&genes=DPM1%2CTNMD%2CTSPAN6&ver=2
3. Observe console that it will now throw many "Too many API calls" errors, but the `network` tab will stop making requests

### DIFF
--- a/frontend/src/common/networkGuard.ts
+++ b/frontend/src/common/networkGuard.ts
@@ -1,0 +1,67 @@
+import { isSSR } from "src/common/utils/isSSR";
+
+export const TOO_MANY_REQUESTS_ERROR_MESSAGE_PREFIX =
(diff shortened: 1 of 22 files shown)


kaloster (approved):
LGTM!

Issue #571, opened 2020-08-31
Consumers want a matrix download to have a meaningful filename


brianraymor:
For [Scientists can download annotated expression matrices from the portal dataset-view](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/16), there was agreement to use the **filename** for the download name. For this increment, the "-29-remixed" (-dsuid-remixed) must be eliminated from **filename**.

```
      "dataset_assets": [
        {
          "created_at": 1598479174.003882, 
          "dataset_id": "8e3bf114-5491-4855-99a1-83770a3f75fa", 
          "filename": "Single_cell_longitudinal_analysis_of_SARS_CoV_2_infection_in_human_bronchial_epithelial_cells-29-remixed.h5ad", 
          "filetype": "H5AD", 
```

---

Other options can be explored for customizing filenames based on consumer-defined formats and dataset metadata - [simple example](https://wiki.hydrogenaud.io/index.php?title=XLD_Configuration#.22File_Naming.22_tab).

tihuan:
@mckinsel can you help check what needs to be done to meet the requirement, so the ticket is ready for estimate 😄? Thank you!

mweiden:
@ambrosejcarr suggested we close this during the 2020-09-10. @mckinsel choose your own adventure!

PR #5947, opened 2023-10-10
feat(4.0): Mass revision publish script

Bento007:
## Reason for Change

- #571

## Changes

- add script to publish all current revisions. This is for helping curator publish collections after a major schema migration.
- factor out shared logic between `PublishRevisions` and `SchemaMigrate`
- move function that use cellxgene_schema_cli into `schema_validation_provider`
- modify

## Testing steps

- add unit tests for new `PublishRevisions` script
- ran the new batch job and verified the output

## Notes for Reviewer
- Do we have a new schema version we can test migration on? This will let me test this on an actual data in rdev.


nayib-jose-gloria:
@Bento007 re: testing a migration, all of our seeded rdev data is still 3.0.0 last I checked. So we can test using 3.1.0 migration

backend/layers/processing/publish_revisions.py
```diff
@@ -0,0 +1,128 @@
...
+                    )
+                else:
+                    self.logger.info(
+                        "Publishing collection version.",
+                        extra={"collection_version_id": collection_version.version_id.id},
+                    )
+                    self.business_logic.publish_collection_version(collection_version.version_id)
+        self.cleanup()
+
+    def cleanup(self) -> None:
```
nayib-jose-gloria, line 111: The auto_publish = False step in the main schema migration batch job only skips publish, it still runs the cleanup. So this (and the other code tracking which objects to delete) I don't think are necessary

nayib-jose-gloria (changes requested):
One note about some logic I think we can remove. otherwise looks good!

nayib-jose-gloria (approved):
ready to merge after a successful rdev migration followed by bulk publish run

PR #5951, opened 2023-10-10
feat: persist tissue_type in APIs and DB + add schema 4 feature flag

nayib-jose-gloria:
## Reason for Change

- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/578

## Changes

- extract tissue_type from h5ads during dataset processing, and group with corresponding tissue_ontology_term_id and tissue label fields.
- refactor 'tissue' field in DatasetMetadata entity to be a 'TissueOntologyTermId' dataclass--which extends the OntologyTermId dataclass and adds tissue-specific field, 'tissue_type'
- Support schema 4 feature flag, accessed via CorporaConfig. Extracts tissue_type from h5ad when flag is TRUE, else uses old mechanism to support CI/CD while working on other migration issues. Set to False in all envs except rdev right now. 
- Update OpenAPI spec to support tissue_type being returned as part of existing tissue field in dataset response objects

## Testing steps

- unit test + rdev testing of uploads + GET endpoints (with schema 4 feature flag on and off)


backend/portal/api/portal-api.yml
```diff
@@ -1175,6 +1175,16 @@ components:
       type: array
       items:
         $ref: "#/components/schemas/ontology_element"
+    tissue_element_array:
+      type: array
+      items:
+        allOf:
+          - $ref: "#/components/schemas/ontology_element"
+          - type: object
```
Bento007, line 1183: once schema 4.0 is in will this need to be combined? We should add some TODO comments to remind us to cleanup after after we finalize 4.0

nayib-jose-gloria: no, this is the intended final form. Only tissue ontology elements have tissue_type, but are otherwise the same as other ontology elements. This is how 'inheritance' is formatted in openAPI spec

backend/portal/api/portal_api.py
```diff
@@ -134,23 +135,13 @@ def _dataset_asset_to_response(dataset_artifact: DatasetArtifact, dataset_id: st
...
-
-
 def _ontology_term_ids_to_response(ontology_term_ids: List[OntologyTermId]):
     """
     Converts a list of OntologyTermId objects to an object compliant to the API specifications.
     This is useful because dataset metadata contain all the possible values for that ontology
     that exist in the dataset proper.
     """
-    return [_ontology_term_id_to_response(otid) for otid in ontology_term_ids]
+    return [dataclasses.asdict(otid) for otid in ontology_term_ids]
```
Bento007, line 144: 👍

tests/unit/processing/test_extract_metadata.py
```diff
@@ -16,6 +16,13 @@ def setUp(self):
             self.business_logic, self.uri_provider, self.s3_provider, self.downloader, self.schema_validator
         )
 
+        def mock_config_fn(name):
+            if name == "schema_4_feature_flag":
+                return "True"
+
+        mock_config = patch("backend.common.corpora_config.CorporaConfig.__getattr__", side_effect=mock_config_fn)
+        mock_config.start()
```
Bento007, line 24: add a tear down for this mock

Bento007 (changes requested):
looks good, a few minor things

tests/unit/processing/test_extract_metadata.py
nayib-jose-gloria: done! went through and added it to some other test classes missing them as well

Bento007 (approved):

PR #5955, opened 2023-10-11
fix(cellguide): determine tissue label-to-id lookup table based on computational marker gene table

atarashansky:
## Reason for Change

- The computational marker gene table is used to generate the list of valid filter options, but the tissue-to-id  lookup table was being calculated from the source data. The two aren't necessarily overlapping, so there were marker genes being filtered out even though they shouldn't have been (e.g. respiratory system marker genes were being filtered out because source data doesn't contain a tissue label with respiratory system).

## Changes

- add
- remove
- modify

## Testing steps


prathapsridharan (approved):
lgtm

PR #5964, opened 2023-10-12
fix: properly fetch previous prod release tag to produce changelog diff for prod release notes

nayib-jose-gloria:
## Reason for Change

- #5487
- Changelog diff builder GHA wasn't properly retrieving the most recent prod release tag when comparing to the new tag; thus, changelog diffs were not being generated correctly.

## Changes

- workflow fetches the most recent prod tag release (those matching the pattern `sha-*`) and passes it directly to the changelog builder github action, rather than relying on its built-in tag resolver (which can only fetch tags in lexicographic or semver order, rather than by date)
- changelog builder github action will check the diff between the new tag and the previous tag we fetch/pass in and generate release notes accordingly

### DIFF
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -109,6 +109,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{ github.event.deployment.sha }}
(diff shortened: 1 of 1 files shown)


atolopko-czi (approved):
Thanks for the fix!

Issue #5484, opened 2023-08-14
Fix format of publication in publication filter and group by


signechambers1:
The existing format of publications in the publication filter and group by functionality does not match the format used on the Collections page. 

Existing format:
Vazquez-Garcia et al. Nature 2022

Desired format:
Vazquez-Garcia et al. (2022) Nature

Please update both the publication filter and group by functionality.

PR #5969, opened 2023-10-12
chore(wmg): fix publication format 

kaloster:
## Reason for Change

- #5484


tihuan (approved):
LGTM! Thanks for updating the test too 👍 

Is there an opportunity to share the citation util function with Collections page, so we have single source of truth?

Thank you!

PR #5995, opened 2023-10-15
implement marker genes cube builder

atarashansky:
The computational marker gene pipeline from CellGuide is invoked and the results are formatted into a pandas dataframe and written to the marker genes tiledb array.


backend/wmg/pipeline/marker_genes.py
```diff
@@ -0,0 +1,89 @@
...
+    create_empty_cube_if_needed,
+    load_pipeline_state,
+    write_pipeline_state,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@log_func_runtime
+def create_marker_genes_cube(*, corpus_path: str):
```
joyceyan, line 36: i don't think we need the `*` decorator here if there's only one param

joyceyan (approved):

backend/wmg/pipeline/marker_genes.py
atarashansky: I will open a separate PR to remove that from all the different "create" functions :D

PR #5997, opened 2023-10-15
implement expression summary default cube builder

atarashansky:
The expression summary default cube is constructed by reading the entire expression summary cube into a dataframe, groupby-summing across non-default attributes, and then writing to the tiledb array.


backend/wmg/pipeline/expression_summary_default.py
```diff
@@ -0,0 +1,66 @@
...
+    ctx = create_ctx()
+    with tiledb.scope_ctx(ctx):
+        dfs = []
+        with tiledb.open(expression_summary_uri, "r") as cube:
+            for row in cube.query(return_incomplete=True).df[:]:
+                dfs.append(row)
+        expression_summary_df = pd.concat(dfs, axis=0)
+
+        expression_summary_df_default = (
+            expression_summary_df.groupby(expression_summary_indexed_dims + expression_summary_non_indexed_dims)
```
atolopko-czi, line 56: PyArrow tables also can perform `groupby`. I'm curious if it would be faster to stay fully in the pyarrow "space". Just a thought, and hardly important if this working efficiently enough.

backend/wmg/data/schemas/cube_schema_default.py
```diff
@@ -27,27 +29,25 @@
...
         for cube_indexed_dim in expression_summary_indexed_dims
     ]
 )
 
 # The cube attributes that comprise the core data stored within the cube.
 expression_summary_logical_attrs = [
-    tiledb.Attr(name="nnz", dtype=np.uint64, filters=filters),  # TODO: Why uint64?
-    tiledb.Attr(name="sum", dtype=np.float32, filters=filters),
-    tiledb.Attr(name="sqsum", dtype=np.float32, filters=filters),
+    tiledb.Attr(name="nnz", dtype=np.uint64, filters=filters_numeric),  # TODO: Why uint64?
```
atolopko-czi, line 41: `Why uint64?` is a super-old comment. Would uint32 suffice? Since this is a cube, the attr size is probably irrelevant. If irrelevant, it's a good time to remove the comment.

atolopko-czi (commented):
Two peripheral comments, but otherwise LGTM.

backend/wmg/pipeline/expression_summary_default.py
atarashansky: Left a TODO! I think the runtime is fast enough, would be interesting to see if Pyarrow groupby incurs less memory overhead, though.

PR #5999, opened 2023-10-16
feat: wmg snapshot v3

atarashansky:
## Performance improvements
There are two performance improvements included in this PR.
 1. Update `build_ordered_cell_types_by_tissue` to iterative over numpy arrays instead of pandas DataFrame rows.
   - This reduces runtime from 8s to 50ms for large queries.
 1. Re-write the gene expression rollup function. Instead of rolling up a tidy pandas dataframe, the dataframe is reshaped into a 2D numpy array per numeric column, where the rows correspond to groups excluding gene and the columns correspond to genes. Rollup is performed on this 2D array and then reshaped into a pandas dataframe.
   - This also includes a performance optimization to `find_descendants_per_cell_type` where a hash table is used to store descendants for cell types so that the operation does not need to be needlessly repeated.
   - This reduces runtime for queries, especially large ones with compare dimension selected (e.g. 11s before to 3s after). it also dramatically reduces memory used (e.g. 5GB before to 500MB after).

## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps


prathapsridharan (approved):
lgtm

PR #6009, opened 2023-10-16
feat: Trigger staging deployment on auto-merge to staging branch

prathapsridharan:
## Reason for Change

- #5461 
- Use `repository_dispatch` to trigger a `stage` environment deploy after `main` is auto-merged to `staging` branch

## Changes

- add `repository_dispatch` calls to workflow

## Testing steps

- Manually done on this PR branch or `main`

### DIFF
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -1,6 +1,9 @@
 name: Deploy Happy
 
-on: deployment
+on:
+  deployment:
+  repository_dispatch:
+    types: [deploy-images]
 
 # add a concurrency group to prevent multiple builds from running at the same time
 concurrency:
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -1,9 +1,6 @@
 name: Deploy Happy
 
-on:
-  deployment:
-  repository_dispatch:
-    types: [deploy-images]
+on: deployment
 
 # add a concurrency group to prevent multiple builds from running at the same time
 concurrency:
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -7,6 +7,14 @@ on:
   push:
     branches:
       - "main"
+      - staging
+  repository_dispatch:
+    types: [run-unit-tests]
+
+# add a concurrency group to prevent multiple builds from running at the same time
+concurrency:
+  group: $${{ github.ref }}-environment
+  cancel-in-progress: false
 
 env:
   CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -7,9 +7,6 @@ on:
   push:
     branches:
       - "main"
-      - staging
-  repository_dispatch:
-    types: [run-unit-tests]
 
 # add a concurrency group to prevent multiple builds from running at the same time
 concurrency:
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -34,6 +34,8 @@ jobs:
 
           git push origin staging
 
+      # TODO(prathap): remove this dummy comment
+      #
       # We need to use repository_dispatch because by default github actions
       # prevents actions triggered by 'GITHUB_TOKEN' from triggering other
       # workflows. Details are here:
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -34,8 +34,6 @@ jobs:
 
           git push origin staging
 
-      # TODO(prathap): remove this dummy comment
-      #
       # We need to use repository_dispatch because by default github actions
       # prevents actions triggered by 'GITHUB_TOKEN' from triggering other
       # workflows. Details are here:
--- a/.github/workflows/trigger-release-candidate-build-and-deploy.yml
+++ b/.github/workflows/trigger-release-candidate-build-and-deploy.yml
@@ -6,10 +6,10 @@ on:
       - main
 
 jobs:
-  merge:
+  continuous_deploy_to_staging:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout main branch and staging branch
+      - name: Checkout main branch
         uses: actions/checkout@v2
         with:
           ref: main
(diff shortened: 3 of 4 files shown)


.github/workflows/push-tests.yml
```diff
@@ -7,6 +7,14 @@ on:
   push:
     branches:
       - "main"
+      - staging
+  repository_dispatch:
+    types: [run-unit-tests]
```
atolopko-czi, line 12: I'm not sure we _need_ to run unit tests on staging, since they'll have been run on main.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -28,9 +28,30 @@ jobs:
...
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          event-type: build-images
+      #- name: Trigger repository_dispatch to deploy images
+      #  if: success()
+      #  uses: peter-evans/repository-dispatch@v2
+      #  with:
+      #    token: ${{ secrets.GITHUB_TOKEN }}
+      #    event-type: deploy-images
```
nayib-jose-gloria, line 51: I believe a successful build image workflow will trigger the deployment on its own without a repository dispatch; it already gets around the GHA protection by using a third party GHA for triggering deployment workflows (avakar/create-deployment@v1)

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -28,9 +28,30 @@ jobs:
...
 
-          echo "About to merge 'main' branch into 'staging' branch"
-          git merge --verbose main -m "Merging main branch into staging branch"
+          # echo "About to merge 'main' branch into 'staging' branch"
+          echo "About to merge 'psridharan/trigger-stage-deploy' branch into 'staging' branch"
+          # git merge --verbose main -m "Merging main branch into staging branch"
+          git merge --verbose psridharan/trigger-stage-deploy -m "Merging psridharan/trigger-stage-deploy branch into staging branch"
 
           git push origin staging
+      - name: Trigger repository_dispatch to build images
```
atolopko-czi, line 40: Should add comment here explaining why repository_dispatch is necessary (https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow)

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -28,9 +28,30 @@ jobs:
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          event-type: build-images
+      #- name: Trigger repository_dispatch to deploy images
+      #  if: success()
+      #  uses: peter-evans/repository-dispatch@v2
+      #  with:
+      #    token: ${{ secrets.GITHUB_TOKEN }}
+      #    event-type: deploy-images
+      #- name: Trigger repository_dispatch to run unit tests
```
atolopko-czi, line 52: Although I don't think we need to run unit tests on staging (per earlier comment), if we _do_ run them, we would want to do that before building and deploying images, right?

.github/workflows/deploy-happy-stack.yml
```diff
@@ -1,6 +1,9 @@
 name: Deploy Happy
 
-on: deployment
+on:
+  deployment:
+  repository_dispatch:
+    types: [deploy-images]
```
atolopko-czi, line 6: maybe name this "deploy-happy-stack" for consistency, even if we don't like the name :)

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi, line 51: I'm curious whether the deployment might actually work when `build-images` is run, since it is using the special "deployment" event. Maybe it's subject to the same github workflow triggering restrictions...but maybe not?

atolopko-czi (changes requested):
Looks workable, but added a couple suggestions. 

Also, I noticed [this](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch):
```
Note: This event will only trigger a workflow run if the workflow file is on the default branch.
```
So do we need to merge this to main before any testing can be done? No concerns about that, just pointing it out in case we're expecting it to be testable pre-merge.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: I see @nayib-jose-gloria had the [same thought](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6009#discussion_r1362126686).

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: Maybe also link out to https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch for quick reference.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -3,16 +3,16 @@ name: Trigger release candidate build and deploy
 on:
   push:
     branches:
-      - main
+      - psridharan/trigger-stage-deploy # main
 
 jobs:
-  merge:
+  continuous_deploy_to_staging:
```
atolopko-czi, line 9: If decide to _not_ run unit tests on staging branch (redundantly with main), then we will need to check that unit tests passed on main before proceeding.

prathapsridharan:
> this

Nice catch @atolopko-czi ! Will get this ready for merge to `main` to test it.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: @atolopko-czi - will do!

.github/workflows/push-tests.yml
prathapsridharan: @atolopko-czi - I can remove it. And perhaps it is better to remove it to reduce the complexity.

My thinking, however, is that `staging` branch is the branch from which we will deploy to `prod` and it would be useful to see that we have all our tests passing there. We can certainly revisit this later.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: @atolopko-czi - Will remove it entirely and we can revisit later whether we want to run tests or not based on what we observe

.github/workflows/deploy-happy-stack.yml
prathapsridharan: will do!

.github/workflows/trigger-release-candidate-build-and-deploy.yml
nayib-jose-gloria: yup! it won't trigger push tests tho, so that would need a separate dispatch (though it looks like you and Andrew are discussing whether its necessary to do so)

nayib-jose-gloria (approved):

atolopko-czi (approved):
:shipit:

PR #6011, opened 2023-10-17
fix(wmg): newsletter test revert to empty string

kaloster:
## Changes

revert temp fix for newsletter validate email from 
`emailInput.fill(" ")`  to `emailInput.fill("")`


tihuan (approved):
Oh got it perfect! Thanks so much for the revert 🎉 !

PR #6015, opened 2023-10-17
fix: Add script to cancel all active schema migration batch jobs.

Bento007:
## Reason for Change

- This is a convenience script to simplify schema-migration development

## Changes

- Add script to cancel all active schema migration batch jobs in the rdev schema migration batch job queue.

## Testing steps

- Ran on rdev during migration and verified the batch job queue was emptied after running


nayib-jose-gloria (approved):
one nit--mind putting 'rdev' upfront in the script name? just for extra clarity about use

PR #6026, opened 2023-10-17
feat: Initiate stage env deploy only after tests pass on main

prathapsridharan:
## Reason for Change

- #5461 

Initiates `stage` deploy only if `push-tests.yml` pass on `main` branch

## Testing steps

Manual testing on `main` branch


tihuan (approved):
LGTM thanks so much!

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -46,3 +55,25 @@ jobs:
...
           event-type: build-images
           client-payload: '{"ref": "refs/heads/staging"}'
+
+      - name: Send slack notification if main not merged into staging
+        if: failure()
+        uses: 8398a7/action-slack@v3
+        with:
+          status: ${{ job.status }}
+          fields: repo,commit,author,eventName,workflow,job,mention
+          mention: "here"
```
atolopko-czi, line 65: just curious, how does the notification (textually) indicate that the merge-to-staging failed?

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -46,3 +55,25 @@ jobs:
...
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
+  notify_on_test_failure_on_main:
```
atolopko-czi, line 68: Will this be redundant with notifications already sent by push-tests.yml?

atolopko-czi (changes requested):
Looks good, with one minor concern re:failure notification.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
```diff
@@ -4,10 +4,19 @@ on:
...
     branches:
       - main
-
+  workflow_run:
+    workflows:
+      - "Push Tests"
+    types:
+      - completed
+    branches:
+      - main
```
atolopko-czi, line 13: Do we know that the `on:push:branches:main` _will not_ trigger this workflow _before_ the the "push tests" completes? I'm just not sure how the precedence of these two triggers works out in practice. In other words, are these two conditions evaluated as AND or OR?

prathapsridharan: That's a good point. I can start by removing `on:push:branches:main` and see if it triggers after `Push Tests` runs on `main`. If so, I think that is sufficient.

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: Good catch! Will fix!

.github/workflows/trigger-release-candidate-build-and-deploy.yml
prathapsridharan: These are the fields and their meaning: https://action-slack.netlify.app/usage/fields/

and this is what it looks like: https://action-slack.netlify.app/

Screenshot 2023-10-18 at 9 54 39 AM

.github/workflows/trigger-release-candidate-build-and-deploy.yml
atolopko-czi: I interpret that a little differently. Namely that whatever version of the workflow file that is on main will be the version that is run, and it won't run an updated version of the workflow file that happens to be on the development branch.

.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -16,11 +16,6 @@ env:
   DOCKER_REPO: ${{ secrets.ECR_REPO }}/
   GITHUB_BRANCH: ${{ github.event.client_payload.ref || github.ref }}
 
-# add a concurrency group to prevent multiple builds from running at the same time
```
prathapsridharan, line 19: @atolopko-czi @tihuan - I am removing `concurrency` so as to complicate things. I was confused by dependency between _"staging  happy stack deploy" and "push tests" on main_ which was caused due to the concurrency groups (see image below)
Screenshot 2023-10-18 at 11 25 19 AM



I will address concurrency in the next PR.

tihuan (approved):

PR #6027, opened 2023-10-18
chore(wmg): refactor CellInfoSideBar

kaloster:
## Reason for Change

- #6012

## Changes

- refactored component as per epic description
- moved images to common/images folder
- moved component to WMG V2 folder


tihuan (approved):
Wow so much cleaner! Thanks so much for taking the time to replace <div> with <> too to cut down the depth of our DOM tree!! 🌴 🔪

Issue #5416, opened 2023-08-07
Gene Expression consistency quick fixes


niknak33:
- [ ] Tissues shouldn't be in bold - This is when you come to the GE landing page, you notice all the tissues are listed in bold. Only tissues that have their accordions in an open state should be bold. (Please consider picking up the padding fix between tissues if easy to do in #5837)
- [ ] For consistency, Cell Type tags from the Autocomplete should match the style of Filter Tags.

tihuan:
@niknak33 please help adding a priority tag thank you!

signechambers1:
@niknak33 added P1 since this is gene expression related. let me know if you disagree!

niknak33:
Totally agree - P1. Thanks Signe.

signechambers1:
@dsadgat @niknak33 updated this ticket so removing nik as the assignee

PR #6028, opened 2023-10-18
fix: 5416 GE quick fixes

tihuan:
## Reason for Change

- #5416 

## Changes

1. Use `TagFilter` for cell type filter tag
2. Update tissue label font weight based on `expanded`

## Testing steps

1. Go to rdev
2. Select some cell types and see the new styled tags
3. Collapse a tissue, the tissue name should no longer be bold

Screenshot 2023-10-18 at 9 24 31 AM


kaloster (approved):
LGTM!

Issue #5917, opened 2023-10-05
Analytics (update): <Cell Ontology Tree Misfiring Ho> 


niknak33:
### Overview

**Tool**: `<CellGuide>`
Feature: `<Ontology Tree>`

**Description**: `<For any events related to hovering, we usually require the user to hover over for at least a few seconds rather than their cursor accidentally passed over one of the nodes but they never even saw the info box pop up. The latter should not make the event fire. >`

### Updates needed

`<Event name>`

- Current: `<Description of how the event works currently>`
- Requested change: `<Description of the update needed>`

Checklist

- [ ] Once event updates are deployed to Staging ping @ainfeld to check analytics
- [ ] Verify events are showing up in the Staging space in Plausible correctly
- [ ] Once event updates are deployed to Prod ping @ainfeld
- [ ] Confirm Plausible events are still showing up in the Prod space in Plausible correctly

#### People to cc

**PM**: @niknak33 
**DS**: @ainfeld

Privacy policy: https://cellxgene.cziscience.com/privacy

PR #6031, opened 2023-10-18
fix(cellguide): fix tooltip analytic firing timing

atarashansky:
## Reason for Change

- #5917

## Testing steps

- Tested locally


frontend/src/views/CellGuide/components/common/OntologyDagView/components/AnimatedNodes/index.tsx
```diff
@@ -82,12 +82,14 @@ export default function AnimatedNodes({
     event: React.MouseEvent<HTMLDivElement>,
     datum: TreeNodeWithState
   ) => {
-    const id = setTimeout(() => {
-      track(EVENTS.CG_TREE_NODE_HOVER, {
-        cell_type: datum.name,
-      });
-    }, 2 * 1000);
-    setTimerId(id);
+    if (!timerId) {
```
tihuan, line 85: do we need to setTimerId(null) after timeout callback function is called?

e.g.,

```ts
setTimeout(() => {
        track(EVENTS.CG_TREE_NODE_HOVER, {
          cell_type: datum.name,
        });
        setTimerId(null); // <--- this?
      }, 2 * 1000);
```

tihuan: Also can we write a test for this please 🙏 ?

atarashansky: No we don't, because the mouseout function handles that.

atarashansky: Can you point me to an example where we have an e2e test for analytics? That would be helpful for me.

tihuan: I don't think we have an example yet, so you'll be the first one!

My initial thought is that I think we can detect console.log, since we do log that when `track()` is called

tihuan: > No we don't, because the mouseout function handles that.

Ah I see!  I was thinking about colocating the create and destroy in the same place, so we don't run into this kind of confusion. What do you think?

atarashansky: We still need to destroy in handleMouseOut anyway to handle the case where users exit out of the node before the setTimeout triggers. Which means that adding setTimerId(null) in the setTimeout is redundant.

tihuan: Ah good point, because we want to cancel it within 2 seconds. Got it thank you!

atarashansky:
E2E tests for analytics will be added in a separate story.

Issue #5916, opened 2023-10-05
Fix missing get_tombstoned parameter in persistence_mock.get_collection_version_with_datasets 


MillenniumFalconMechanic:
#### Need
The `get_tombstoned` parameter in `persistence_mock.get_collection_version_with_datasets` is not passed through to `self._update_version_with_canonical`. See [code](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/layers/persistence/persistence_mock.py#L270) and [suggested fix](https://github.com/chanzuckerberg/single-cell-data-portal/pull/5824/files#diff-766d914dd42c86f642d416c3e803a78eab7c8c8a14ec64f2b5f31d1358f4be70R270).

CC @danieljhegeman.

danieljhegeman:
This is a bug that should be fixed before any further development is done on DP service code! Thanks Mim!

danieljhegeman:
@MillenniumFalconMechanic any chance this can be tackled this week?

PR #6032, opened 2023-10-18
fix: added pass-thru of get_tombstoned.

MillenniumFalconMechanic:
## Reason for Change

- #5916


danieljhegeman (approved):
Thanks Mim! 🚀

Issue #4821, opened 2023-05-09
Add zebra striping to minimal tables


karenliang0:
The minimal tables that we have within Gene Expression FMG panel should utilize zebra striping instead of dividers. This design change will provide more differentiation between rows for ease of reading. 

In the near future, zebra striping will also be utilized for the minimal tables within CellCards.

[Figma Design](https://www.figma.com/file/nUbwlxcSPAGoRGWyYaIkyv/Q2-2023-Gene-Expression-Improvements?node-id=114%3A10261&t=gGX5DMyKe1XurP9T-1)

PR #6034, opened 2023-10-18
chore(wmg): add zebra striping to table

kaloster:
## Reason for Change

- #4821

## Changes

https://github.com/chanzuckerberg/single-cell-data-portal/assets/7976579/4901bff7-718c-4097-abf7-bb515ab5758e

- replaced table markup with div table
- added striping

### DIFF
--- a/frontend/src/views/WheresMyGeneV2/components/CellInfoSideBar/style.ts
+++ b/frontend/src/views/WheresMyGeneV2/components/CellInfoSideBar/style.ts
@@ -1,17 +1,29 @@
-import { HTMLTable } from "@blueprintjs/core";
 import styled from "@emotion/styled";
 import { Button, fontBodyS } from "@czi-sds/components";
 import Image from "next/image";
 import { fontBodyXxs, getColors } from "@czi-sds/components";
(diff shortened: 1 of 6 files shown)


tihuan:
Woohoo thanks for the PR, Ronen!

To help with the reviewers, for UI changes like this can we please add screenshot and/or video please 🙏 

Thanks so much!

frontend/src/views/WheresMyGeneV2/components/CellInfoSideBar/style.ts
```diff
@@ -129,3 +130,47 @@ export const NoMarkerGenesDescription = styled("span")`
     `;
   }}
 `;
+
+export const DivTable = styled.div`
```
tihuan, line 134: Love the div table too 🥇 🙏

tihuan (approved):
LGTM! Thanks so much, @kaloster 🙏 🚀 

Also confirmed that we have some tests (`describe("Find Marker Genes", () => {`) for the marker gene table, so I don't think we need more tests here. But @kaloster please double check in case I'm wrong. Thank you!

PR #6052, opened 2023-10-20
test: Update collection revision e2e test timeout

tihuan:
## Reason for Change

1. e2e collection revision test was timing out at 30 mins, because there's a bug in the waitUntil block logic that triggers a 5 min wait every time it can't find a suitable collection

### DIFF
--- a/frontend/tests/features/collection/revision.test.ts
+++ b/frontend/tests/features/collection/revision.test.ts
@@ -158,6 +158,11 @@ describe("Collection Revision @loggedIn", () => {
   });
 });
 
+/**
+ * (thuang): Wait for 1 min instead of the default 3 minutes, so we fail faster
+ */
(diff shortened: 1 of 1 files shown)


frontend/tests/features/collection/revision.test.ts
```diff
@@ -183,16 +188,23 @@ async function startRevision(page: Page): Promise<string> {
           { page }
         );
       } catch {
-        await page
-          .getByTestId(COLLECTION_ROW_WRITE_REVISION_ID)
-          .first()
-          .getByTestId(COLLECTION_VIEW_REVISION)
-          .click();
```
tihuan, line 190: this unconditionally wait for a suitable collection to appear til the max test timeout of 5 mins, which is why the test takes 30 mins to timeout in GHA

seve (approved):

tihuan:
Thanks so much, Seve!

PR #6057, opened 2023-10-21
feat(cellguide): add marker gene mode dropdown

atarashansky:
## Reason for Change

- Discoverability of the marker gene mode was low due to the icon only appearing on hover in the computational marker gene row.
- This change makes the feature more convenient and discoverable for users.
## Changes

- added a dropdown for activating marker gene mode to allow users to search for their marker gene of interest

## Screenshots
image
image

## Testing steps

- Added e2e tests for dropdown


hthomas-czi:
Looks great, @atarashansky! I have a couple more thoughts:

1. I find the auto-scroll behavior when selecting a gene from the dropdown a bit jarring. I understand having it when selecting a gene from the table, but it seem unnecessary for the dropdown.
2. Should we automatically navigate to the page with the selected gene when users color by gene from the dropdown?

Minor note:
1. The gene info button and ontology button in the Marker Gene table should have the same hover behavior: 
`gray[500]` → `common.black`

hthomas-czi (approved):
Looks great, @atarashansky! I have a couple more thoughts:

1. I find the auto-scroll behavior when selecting a gene from the dropdown a bit jarring. I understand having it when selecting a gene from the table, but it seem unnecessary for the dropdown.
2. Should we automatically navigate to the page with the selected gene when users color by gene from the dropdown?

Minor note:
1. The gene info button and ontology button in the Marker Gene table should have the same hover behavior:
`gray[500]` → `common.black`

atarashansky:
> 2. Should we automatically navigate to the page with the selected gene when users color by gene from the dropdown?
i'm not sure what you mean by this - which page?

frontend/src/common/queries/cellGuide.ts
```diff
@@ -487,15 +487,19 @@ export function useAllOrgansLookupTables(): Map<string, string> {
...
+} {
   const { data: tissueData } = useTissueMetadata();
   const { data: computationalMarkers } = useComputationalMarkers(cellTypeId);
 
   return useMemo(() => {
     if (!tissueData || !computationalMarkers) {
-      return new Map<string, string>();
+      return {
+        allTissuesLabelToIdLookup: new Map<string, string>(),
+        computationalMarkers: [],
```
tihuan, line 501: nit: prefer `EMPTY_ARRAY` instead of a new `[]`

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -216,6 +221,49 @@ export default function CellGuideCard({
...
+      organId: selectedOrganId,
+    });
+
+  const sdsGenesList = useMemo(() => {
+    const genes = computationalMarkerGeneTableData.map((gene) => ({
+      name: gene.symbol,
+    }));
+
+    // Insert an element at the beginning
+    genes.unshift({ name: "No Gene" });
```
tihuan, line 237: maybe extract "No Gene" as a constant, so we have single source of truth for reference?

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -216,6 +221,49 @@ export default function CellGuideCard({
...
+
+    return genes;
```
tihuan, line 239: let's do the following for clarity 🙏 

Reading the code, it's kinda jarring to see an unshift to insert an element to the front unnecessarily

```ts
const NO_GENE_OPTIONS = { name: "No Gene" };

return [NO_GENE_OPTIONS, ... genes];
```

frontend/src/views/CellGuide/components/CellGuideCard/index.tsx
```diff
@@ -216,6 +221,49 @@ export default function CellGuideCard({
...
+  }, [computationalMarkerGeneTableData]);
+
+  const handleChangeGene = (option: DefaultDropdownMenuOption | null) => {
+    if (option?.name === "No Gene") {
```
tihuan, line 243: we can use the new constant `if (option?.name === NO_GENE_OPTIONS.name) { ... }`

frontend/tests/features/cellGuide/cellGuide.test.ts
```diff
@@ -759,11 +759,59 @@ describe("Cell Guide", () => {
...
+          .getByTestId(CELL_GUIDE_ONTOLOGY_VIEW_LEGEND_TEST_ID)
+          .textContent();
+        expect(legendText).toContain("Marker Score");
+        expect(legendText).toContain("Expressed in Cells(%)");
+
+        // deactivate marker gene mode and check that the legend and tooltips reverted
+        await page
+          .getByTestId(CELL_GUIDE_CARD_GLOBAL_MARKER_GENE_DROPDOWN)
+          .click();
+        await page.getByRole("option", { name: "No Gene" }).click();
```
tihuan, line 786: maybe `NO_GENE_OPTIONS` should be in a `constants.ts` file per @kaloster 's new pattern, so we can import that here instead of hard coding it here

tihuan (commented):
Looks really great! Thanks for updating the e2e tests too, @atarashansky 😄 🙏 Just some small nits PTAL thank you!

ainfeld:
@niknak33 I think we need to instrument this new drop down since it will contribute to visitors engaging with cell guide. I expect it will be used more since users no longer have to scroll down to the marker gene section and click into the computational tab to activate the new feature. 

If you agree, please add an instrumentation ticket for this and assign Alec? Is that right @atarashansky ?

hthomas-czi:
> Reference

@atarashansky Sorry. I'm referring to the Marker Gene table page vis-a-vis pagination

Issue #6036, opened 2023-10-18
tech: improve performance of build_ordered_cell_types_by_tissue by removing pandas df iteration with iloc


atarashansky:
https://github.com/chanzuckerberg/single-cell-data-portal/blob/3fd65811b39a2c8f39e790ece16dc246d0d114f6/backend/wmg/api/v2.py#L392

The above line is iterating through the pandas DF row by row, which is slow for large DFs.

A performance optimization was previously applied for `build_expression_summary`:
 - https://github.com/chanzuckerberg/single-cell-data-portal/blob/3fd65811b39a2c8f39e790ece16dc246d0d114f6/backend/wmg/api/v2.py#L319

This optimization iterates over numpy arrays (columns) instead of the pandas dataframe. The same technique can be applied to the cell types by tissue function.

atarashansky:
After implementing the above fix ([rdev here](https://new-wmg-frontend.rdev.single-cell.czi.technology/gene-expression)) the runtime of the `build_ordered_cell_types_by_tissue` function improved by 160X (8s to 50ms)

PR #6059, opened 2023-10-23
feat(wmg): performance improvements

atarashansky:
## Performance improvements
There are two performance improvements included in this PR.

 - Update build_ordered_cell_types_by_tissue to iterative over numpy arrays instead of pandas DataFrame rows.
   - This reduces runtime from 8s to 50ms for large queries.
 - Re-write the gene expression rollup function. Instead of rolling up a tidy pandas dataframe, the dataframe is reshaped into a 2D numpy array per numeric column, where the rows correspond to groups excluding gene and the columns correspond to genes. Rollup is performed on this 2D array and then reshaped into a pandas dataframe.
   - This also includes a performance optimization to find_descendants_per_cell_type where a hash table is used to store descendants for cell types so that the operation does not need to be needlessly repeated.
   - This reduces runtime for queries, especially large ones with compare dimension selected (e.g. 11s before to 3s after). it also dramatically reduces memory used (e.g. 5GB before to 500MB after).

## Reason for Change

- #6036

## Changes

- add
- remove
- modify

## Testing steps

### DIFF
--- a/backend/common/utils/rollup.py
+++ b/backend/common/utils/rollup.py
@@ -46,15 +46,21 @@ def find_descendants_per_cell_type(cell_types):
     descendants_per_cell_type : list
         List of lists of descendants for each cell type in the input list.
     """
-
+    lookup_table = {}
+    cell_types_set = set(cell_types)
     relatives_per_cell_type = []
     for cell_type in cell_types:
-        relatives = descendants(cell_type)
-        relatives_per_cell_type.append(relatives)
+        if ";;" in cell_type:
+            prefix, suffix = cell_type.split(";;")
+            suffix = ";;" + suffix
+        else:
+            prefix = cell_type
+            suffix = ""
+        if cell_type not in lookup_table:
+            relatives = [f"{i}{suffix}" for i in descendants(prefix)]
+            lookup_table[cell_type] = list(cell_types_set.intersection(relatives))
 
-    for i, children in enumerate(relatives_per_cell_type):
-        # remove descendent cell types that are not cell types in the WMG data
-        relatives_per_cell_type[i] = list(set(children).intersection(cell_types))
+        relatives_per_cell_type.append(lookup_table[cell_type])
 
     return relatives_per_cell_type
 
--- a/backend/common/utils/rollup.py
+++ b/backend/common/utils/rollup.py
@@ -34,7 +34,16 @@ def ancestors(cell_type):
 
 def find_descendants_per_cell_type(cell_types):
     """
-    Find the ancestors or descendants for each cell type in the input list.
+    Find the descendants for each cell type in the input list.
+    Terms are only considered descendants if they are present in the input list.
+
+    Terms can be suffixed with ";;{A_0}--{B_0}--..." to indicate that only descendants within the same
+    group specified by the suffix should be considered. For example, if the term is
+    "CL:0000540;;{tissue_0}--{disease_0}" then only descendants with the same tissue_0 and disease_0
+    will be considered.
+
+    This is useful for rolling up cell types within groups (e.g. tissues) but not across groups.
+
 
     Parameters
     ----------
@@ -46,10 +55,15 @@ def find_descendants_per_cell_type(cell_types):
     descendants_per_cell_type : list
         List of lists of descendants for each cell type in the input list.
     """
+    # a lookup table to avoid redundant ontology lookups
     lookup_table = {}
+
+    # a set of all cell types in the input list
+    # used to compute intersections with descendants
     cell_types_set = set(cell_types)
     relatives_per_cell_type = []
     for cell_type in cell_types:
+        # if the input cell types are suffixed, only consider descendants within the same group
         if ";;" in cell_type:
             prefix, suffix = cell_type.split(";;")
             suffix = ";;" + suffix
@@ -57,7 +71,9 @@ def find_descendants_per_cell_type(cell_types):
             prefix = cell_type
             suffix = ""
         if cell_type not in lookup_table:
+            # find descendants of cell type and re-suffix them
             relatives = [f"{i}{suffix}" for i in descendants(prefix)]
+            # only consider descendants (+suffixes) that are in the input list
             lookup_table[cell_type] = list(cell_types_set.intersection(relatives))
 
         relatives_per_cell_type.append(lookup_table[cell_type])
--- a/backend/wmg/api/common/rollup.py
+++ b/backend/wmg/api/common/rollup.py
@@ -83,23 +83,33 @@ def _rollup_gene_expression(gene_expression_df, universal_set_cell_counts_df) ->
         and likely greater size than the input gene expression dataframe that includes combinations
         for which numeric values should be aggregated during the rollup operation.
     """
+    # if the gene expression dataframe is empty, then there is nothing to roll up
     if gene_expression_df.shape[0] == 0:
         return gene_expression_df
 
+    # get the numeric columns in the gene expression dataframe
     numeric_columns = list(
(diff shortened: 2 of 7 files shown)


tests/unit/backend/wmg/api/test_v2.py
```diff
@@ -66,12 +67,9 @@ def generate_expected_term_id_labels_dictionary(
...
     result = {}
     result["cell_types"] = {}
-    # assume tissues are sorted, and cell types are sorted within each tissue
-    # assume the length of cell types is the dimensionality of each column in the
-    # test snapshot (i.e. each tissue contains all fake cell types).
-    # this line determines the starting order for each cell type in each tissue.
-    orders = [int(tissue.split("_")[-1]) * len(cell_types) for tissue in tissues]
-    for tissue, order in zip(tissues, orders):
+
+    order = -1
```
atarashansky, line 71: Temporary - working on this test

backend/common/utils/rollup.py
```diff
@@ -46,15 +46,21 @@ def find_descendants_per_cell_type(cell_types):
...
-        relatives_per_cell_type[i] = list(set(children).intersection(cell_types))
+        if ";;" in cell_type:
+            prefix, suffix = cell_type.split(";;")
+            suffix = ";;" + suffix
+        else:
+            prefix = cell_type
+            suffix = ""
+        if cell_type not in lookup_table:
+            relatives = [f"{i}{suffix}" for i in descendants(prefix)]
+            lookup_table[cell_type] = list(cell_types_set.intersection(relatives))
```
atarashansky, line 61: Using a hash table to avoid needing to repeat this operation

backend/common/utils/rollup.py
```diff
@@ -46,15 +46,21 @@ def find_descendants_per_cell_type(cell_types):
...
+    cell_types_set = set(cell_types)
     relatives_per_cell_type = []
     for cell_type in cell_types:
-        relatives = descendants(cell_type)
-        relatives_per_cell_type.append(relatives)
-
-    for i, children in enumerate(relatives_per_cell_type):
-        # remove descendent cell types that are not cell types in the WMG data
-        relatives_per_cell_type[i] = list(set(children).intersection(cell_types))
+        if ";;" in cell_type:
```
atarashansky, line 53: Cell types can be suffixed with ";;" if descendants need to be calculated within groups only.

For example, `"CL:0000540;;UBERON:0002048--PATO:0000461"` means that CL:0000540 is in tissue UBERON:0002048 and disease PATO:0000461. Descendants of CL:0000540 will be restricted to those that are in the above group.

backend/wmg/api/common/rollup.py
```diff
@@ -85,53 +83,58 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
         and likely greater size than the input gene expression dataframe that includes combinations
         for which numeric values should be aggregated during the rollup operation.
     """
-    # extract group-by terms and queried genes from the input dataframes
-    # if a queried gene is not present in the input dot plot dataframe, we can safely
-    # ignore it as it need not be rolled up anyway.
-    group_by_terms = list(universal_set_cell_counts_df.index.names)
-    genes = list(set(gene_expression_df["gene_ontology_term_id"]))
+    if gene_expression_df.shape[0] == 0:
+        return gene_expression_df
```
atarashansky, line 87: do nothing if df is empty.

backend/wmg/api/common/rollup.py
```diff
@@ -85,53 +83,58 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
...
+    group_by_terms = list(universal_set_cell_counts_df.index.names)
+    available_combinations = set(universal_set_cell_counts_df.index.values)
 
-    # get the total number of cells per tissue to populate the n_cells_tissue in the added entries
-    n_cells_tissue_dict = gene_expression_df.groupby("tissue_ontology_term_id").first()["n_cells_tissue"].to_dict()
+    numeric_columns.remove("n_cells_tissue")
 
-    # get the set of available combinations of group-by terms from the aggregated cell counts
-    available_combinations = set(universal_set_cell_counts_df.index.values)
+    pivoted = gene_expression_df.pivot_table(
```
atarashansky, line 98: pivot the tidy dataframe into a 2D array with indices corresponding to `group_by_terms` and columns corresponding to genes.

backend/wmg/api/common/rollup.py
```diff
@@ -85,53 +83,58 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
...
+        index=group_by_terms, columns=["gene_ontology_term_id"], fill_value=0, values=numeric_columns, aggfunc="sum"
+    )
+    genes = np.array(pivoted["sum"].columns.tolist())
+    missing_combinations = list(available_combinations.difference(pivoted.index.values))
+    dense_tables = np.stack(
```
atarashansky, line 103: Stack the 2D tables for sum, nnz, etc (N x M x P) where P is the number of numeric metadata

backend/wmg/api/common/rollup.py
```diff
@@ -85,53 +83,58 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
...
+    multi_index = pd.MultiIndex.from_tuples(pivoted.index.tolist() + missing_combinations, names=group_by_terms)
+    group_by_terms.remove("cell_type_ontology_term_id")
+    group_by_terms = ["cell_type_ontology_term_id"] + group_by_terms
 
-    # for each gene, get the set of available combinations of group-by terms from the input expression dataframe
-    entries_to_add = []
-    for gene in genes:
-        gene_expression_df_per_gene = gene_expression_df[gene_expression_df["gene_ontology_term_id"] == gene]
-        available_combinations_per_gene = set(zip(*gene_expression_df_per_gene[group_by_terms].values.T))
+    multi_index = multi_index.reorder_levels(group_by_terms)
```
atarashansky, line 115: Make cell type first in every multi-index tuple

backend/wmg/api/common/rollup.py
```diff
@@ -85,53 +83,58 @@ def _add_missing_combinations_to_gene_expression_df_for_rollup(
...
+        if col > 1:
+            cell_types_for_rollup = np.char.add(cell_types_for_rollup, "--")
+        cell_types_for_rollup = np.char.add(cell_types_for_rollup, multi_index_array[:, col])
 
-    # add the missing entries to the input expression dataframe
-    gene_expression_with_missing_combos_df = pd.concat((gene_expression_df, pd.DataFrame(entries_to_add)), axis=0)
+    dense_tables = rollup_across_cell_type_descendants_array(dense_tables, cell_types_for_rollup)
 
-    return gene_expression_with_missing_combos_df
+    row, col = dense_tables[:, :, 0].nonzero()
```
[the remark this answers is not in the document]
prathapsridharan: By the `0` in the last index, you are inferring that if one of the numeric columns is non-zero then all numeric columns are non-zero? If so, perhaps a comment might help?

backend/wmg/api/common/rollup.py
prathapsridharan: A comment stating that `N is the size of the multi-dim index (tissue, cell-type, compare-dim)`, `M is the number of genes for each [tissue, cell-type, compare-dim]`, and `P is the number of non-numeric columns for [tissue, cell-type, compare-dim, gene]`.

Doesn't have to be exactly as I state but something gives an idea for the input structure and the output would be helpful

prathapsridharan (approved):
Great work here!

Just some comments about adding some explanatory comments. If you could expand upon your inline PR comments and add it to the code, it would help understanding quite a bit.

These ndarray transformations are a bit tough to wrap one's head around. I imagine it will be even harder if some time passes. It would be pretty helpful to have just enough explanatory comments in the code to help the reader

PR #6060, opened 2023-10-23
feat(wmg): wmg v3 api

atarashansky:
## Reason for Change

- #6065
- #6071  

## Changes

- added dynamic scaling to legend and colorscale
- updated backend to read v3 snapshot
- Re-activated skipped unit tests now that v3 has been activated

## Dynamic scaling demo 
https://czi-sci.slack.com/archives/C0247APK621/p1698163069890259
## Testing steps

- All tests pass, including previously skipped tests
- E2E tests pass

### DIFF
--- a/backend/wmg/data/snapshot.py
+++ b/backend/wmg/data/snapshot.py
@@ -75,9 +75,6 @@ class WmgSnapshot:
 
     dataset_metadata: Optional[Dict] = field(default=None)
 
-    def __hash__(self):
-        return hash(None)  # hash is not used for WmgSnapshot
-
 
 # Cached data
 cached_snapshot: Optional[WmgSnapshot] = None
--- a/backend/wmg/data/snapshot.py
+++ b/backend/wmg/data/snapshot.py
@@ -62,10 +62,8 @@ class WmgSnapshot:
     # backend/wmg/data/schemas/cube_schema.py.
     cell_counts_cube: Optional[Array] = field(default=None)
 
-    # Pandas DataFrame containing per-tissue ordering of cell types.
-    # Columns are "tissue_ontology_term_id", "cell_type_ontology_term_id", "order"
-
-    cell_type_orderings: Optional[DataFrame] = field(default=None)
+    # Dictionary of (cell type, tissue) tuples as keys and order as values.
+    cell_type_orderings: Optional[dict] = field(default=None)
 
     # precomputed list of ids for all gene and tissue ontology term ids per organism
     primary_filter_dimensions: Optional[Dict] = field(default=None)
--- a/backend/wmg/data/snapshot.py
+++ b/backend/wmg/data/snapshot.py
@@ -20,7 +20,6 @@ EXPRESSION_SUMMARY_CUBE_NAME = "expression_summary"
 EXPRESSION_SUMMARY_DEFAULT_CUBE_NAME = "expression_summary_default"
 CELL_COUNTS_CUBE_NAME = "cell_counts"
 MARKER_GENES_CUBE_NAME = "marker_genes"
-DATASET_TO_GENE_IDS_FILENAME = "dataset_to_gene_ids.json"
 FILTER_RELATIONSHIPS_FILENAME = "filter_relationships.json"
 DATASET_METADATA_FILENAME = "dataset_metadata.json"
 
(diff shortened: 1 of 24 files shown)


frontend/src/views/WheresMyGeneV2/components/Main/index.tsx
```diff
@@ -158,7 +167,42 @@ export default function WheresMyGene(): JSX.Element {
           const { cellTypeGeneExpressionSummaries } = geneExpressionSummary;
 
           for (const cellTypeGeneExpressionSummary of cellTypeGeneExpressionSummaries) {
+            // get term before $
+            const cellTypeGeneExpressionSummaryId =
+              cellTypeGeneExpressionSummary.viewId.split("$")[0];
```
tihuan, line 172: you can use `getOntologyTermIdFromCellTypeViewId(viewId)`!

frontend/src/views/WheresMyGeneV2/components/Main/index.tsx
```diff
@@ -158,7 +167,42 @@ export default function WheresMyGene(): JSX.Element {
...
+
+            /*
+            This is to dynamically set the min and max based on the data that is visible to users.
+            Skip conditions:
+            - If the cell type is not part of an expanded tissue
+            - If the tissue does not contain any of the filtered cell types and at least one cell type is filtered
+            - If the cell type is not included in the filtered cell types and at least one cell type is filtered
+            - If the tissue id is not included in the filtered tissue ids and at least one tissue is filtered
+            - If the cell type gene expression summary view id is not included in the available view ids
+            The above conditions capture all possible scenarios where the data is not visible to users.
```
tihuan, line 184: awesome documentation 🙏 Thanks so much!

frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
```diff
@@ -255,6 +255,7 @@ export default memo(function HeatMap({
                 echartsRendererMode={echartsRendererMode}
                 setAllChartProps={setAllChartProps}
                 chartProps={allChartProps[tissue]}
+                maxExpression={6.0}
```
tihuan, line 258: can we add some comment in code why we use 6.0 here even though we're not using RankIt anymore? Thanks!

tihuan (commented):
Looks really great! Thanks for fixing the scaling, @atarashansky 🙏 

Can you add the videos you have in Slack in this PR for paper trail please? Since Slack messages don't stay around for long

Also any e2e tests you can think of we should add?

Thank you!

frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
atarashansky: This component is no longer used, so I didn't think any comments were necessary - when are we deleting it?

atarashansky:
> Also any e2e tests you can think of we should add?

This seems like it would be better covered by FE unit tests. Do we have any test framework set up to run those?

backend/wmg/data/snapshot.py
```diff
@@ -73,8 +72,12 @@ class WmgSnapshot:
     filter_relationships: Optional[Dict] = field(default=None)
 
     # dataset metadata dictionary
+
     dataset_metadata: Optional[Dict] = field(default=None)
 
+    def __hash__(self):
```
prathapsridharan, line 78: if the object doesn't have to hashable then why is this there? Could it be removed?

kaloster (approved):

prathapsridharan (approved):
Looks good. Just a small comment about removing `__hash__` on `WMGSnapshot` if it is not needed

frontend/src/views/WheresMyGene/components/HeatMap/index.tsx
tihuan: ah I see, so 6.0 is just for passing TS. @kaloster is working on removing WMGv1!

tihuan:
> > Also any e2e tests you can think of we should add?
> 
> This seems like it would be better covered by FE unit tests. Do we have any test framework set up to run those? In any case, we can write an E2E test that just makes sure that the unscaled max value is changing with the filters.

CC has unit test examples here 👍 

https://github.com/chanzuckerberg/single-cell-data-portal/blob/cb44c8c395ff73df23060ec43f5a6b488e6c82d5/frontend/tests/features/collection/collection.test.ts#L35

Issue #5660, opened 2023-08-31
RDS and CXG conversions should be independent operations


brianraymor:
# Context

See [**single-cell-data-wrangling**](https://czi-sci.slack.com/archives/C024HCSH9PT/p1693261728537429).

Recently, RDS conversions have been especially _brittle_. Currently, a RDS conversion failure will result in the cancellation of a **pending** CXG conversion although there should be no correlation between potential RDS and CXG failures. As a result, curators and submitters are unable to view the Explorer visualization until the RDS conversion is addressed. This is currently blocking a submitter review.

I do not remember if there was a rationale for this design such as _fast failing_. 

CC: @Bento007 since this issue is based on a personal slack conversation. @atolopko-czi

atolopko-czi:
> I do not remember if there was a rationale for this design such as fast failing.

I'm not aware of any technical reason why fast failing behavior is required here. But we may have to consider the processing status logic to ensure that _partial_ failures (e.g. just RDS) can be retried appropriately.

We actually want to eventually have the [CXG generation be moved into a separate pipeline](https://docs.google.com/document/d/1oeNXVrNn8PyBjkY_v4kiNobmE97bEMx-nGBq-Vm-X7s/edit#bookmark=id.oo23v4xea6z5), so it can be owned by the Viz team.  It would only be loosely coupled to the dataset submission pipeline via an event notification when a "labeled" h5ad is ready, and not an RDS. This might be worth considering when we take on this issue.

brianraymor:
@nayib-jose-gloria - added to schema 4 epic for visibility. Please consider.

Bento007:
https://dev.to/aws-builders/parallel-task-error-handling-in-step-functions-4f1c

danieljhegeman:
>As a result, curators and submitters are unable to view the Explorer visualization until the RDS conversion is addressed.

Noting that, currently, it also blocks _publishing_.

@brianraymor I am unclear on whether you would want to start allowing publishing in cases where seurat conversion failed.

metakuni:
On a side note, I think I recall from a long time ago that despite being defined as parallel steps in the step function that the Seurat and CXG conversions actually run sequentially.

@ebezzi , is this still the case?

danieljhegeman:
If the reported `TaskStarted` times are to be believed in the step function logs, then the cxg and seurat steps/tasks are indeed being launched in parallel rather than in sequence.

ebezzi:
[...]

PR #6068, opened 2023-10-23
feat: do not fail fast if Seurat conversion fails

danieljhegeman:
## Reason for Change

- #5660 

## Changes

- add
- remove
- modify

## Testing steps


backend/layers/processing/upload_success/app.py
```diff
@@ -3,23 +3,48 @@
...
+    cxg_job, seurat_job = events["cxg_job"], events["seurat_job"]
+    cxg_job["execution_id"], seurat_job["execution_id"] = events["execution_id"], events["execution_id"]
 
-    business_logic.update_dataset_version_status(
-        DatasetVersionId(dataset_id), DatasetStatusKey.PROCESSING, DatasetProcessingStatus.SUCCESS
-    )
+    if cxg_job.get("error"):
+        handle_failure(cxg_job, context)
+    elif seurat_job.get("error"):
+        # Same as handle_failure except do not delete artifacts (because cxg succeeded)
```
nayib-jose-gloria, line 33: what do you think of passing a flag to handle_failure to avoid deleting artifacts instead of moving this code here?

nayib-jose-gloria (commented):
1 thought; otherwise looks good. Curious to get @Bento007 's opinion around the terraform change

backend/layers/processing/upload_success/app.py
danieljhegeman: ✅

Bento007 (approved):
With this change we can remove the error handler all together and pipe all error message through the success handler. If it's not too much work and fresh in our mind, could you add that as a followup PR.

nayib-jose-gloria (approved):

PR #6081, opened 2023-10-24
feat(4.0): Dynamic machine size for ingest

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/543

The new step function
Screenshot 2023-11-01 at 9 53 06 AM

## Changes

- add a batch job that is created by the ingest SFN
- Add function to estimate the memory to needed by batch job

## Testing steps

- added unit tests.

## Notes for Reviewer
- there is no limit according to AWS docs for batch job definitions


.happy/terraform/modules/sfn/main.tf
```diff
@@ -66,7 +68,111 @@ resource "aws_sfn_state_machine" "state_machine" {
               },
               {
                 "Name": "STEP_NAME",
-                "Value": "download-validate"
+                "Value": "download"
```
danieljhegeman, line 71: I'm a little hazy on the details here, but can we confirm that we want `"Value"` here and that the above values are correctly either `"Value"` or `"Value.$"`?

danieljhegeman: `"Value"` includes it as an env variable but doesn't add it to the Context object... is that right? Whereas `"Value.$"` _will_ add it to the Context object?

Bento007: "Value.$" to my understand will mean we are getting the value from a json object. The target would have to be `$.foo`

backend/layers/processing/process_download.py
```diff
@@ -75,6 +81,41 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+        MEMORY_MODIFIER = 1.1  # add 10% overhead
+        MEMORY_PER_VCPU = 4000
+        MIN_MEMORY_MB = 4000
+        MIN_VCPU = 1
+        # The largest machine we are allocating is r5ad.2xlarge. This machine has 64GB of memory and 16 vCPUs.
+        MAX_MEMORY_MB = 64000
+        MAX_VCPU = 16
+        SWAP_MEMORY_MB = 300000
+
+        adata = scanpy.read_h5ad(local_filename)
```
ebezzi, line 94: Shouldn't we use `backed='r'` here to avoid loading X in memory?

backend/layers/processing/process_download.py
```diff
@@ -75,6 +81,41 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+
+        # Note: this is a rough estimate of the uncompressed size of the dataset. This method avoid loading the entire
+        # dataset into memory.
+        uncompressed_size_MB = adata.n_obs * adata.n_vars // MB
```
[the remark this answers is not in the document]
Bento007: Why a factor of 4?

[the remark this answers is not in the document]
ebezzi: Each element is an int64 (or a float32, for the normalized layer), so 4 bytes.

.happy/terraform/modules/sfn/variables.tf
```diff
@@ -39,3 +39,45 @@ variable max_attempts {
...
+  type        = string
+  description = "Datasets public-access bucket name"
+}
+
+variable frontend_url {
+  type        = string
+  description = "url for the frontend app"
+}
+
+variable "batch_job_log_group" {
```
ebezzi, line 80:
```suggestion
variable batch_job_log_group {
```

backend/layers/processing/process_download.py
```diff
@@ -16,12 +22,22 @@
 from backend.layers.processing.logger import logit
 from backend.layers.processing.process_logic import ProcessingLogic
 from backend.layers.thirdparty.s3_provider_interface import S3ProviderInterface
+from backend.layers.thirdparty.step_function_provider import StepFunctionProvider
 from backend.layers.thirdparty.uri_provider import UriProviderInterface
 
+MEMORY_MODIFIER = 1.1  # add 10% overhead
+MEMORY_PER_VCPU = 4000
+MIN_MEMORY_MB = 4000
+MIN_VCPU = 1
```
ebezzi, line 31: I think it's good to start with 1, but there is a chance that only using a single CPU, for a compute intensive process, may cause issues (e.g. send the machine into a resource starved state). Recommend revisiting this only if we see that processing of small datasets takes too long.

backend/layers/processing/process_download.py
```diff
@@ -75,37 +91,106 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+    def estimate_resource_requirements(
+        adata: scanpy.AnnData,
+        memory_modifier: float = MEMORY_MODIFIER,
+        min_memory_MB: int = MIN_MEMORY_MB,
+        min_vcpu: int = MIN_VCPU,
+        max_memory_MB: int = MAX_MEMORY_MB,
+        max_vcpu: int = MAX_VCPU,
+        swap_memory_MB: int = SWAP_MEMORY_MB,
+        memory_per_vcpu: int = MEMORY_PER_VCPU,
+    ) -> Dict[str, Any]:
```
ebezzi, line 114: Looks like the values are all integers here.

Bento007:
```suggestion
    ) -> Dict[str, int]:
```

.happy/terraform/modules/sfn/main.tf
```diff
@@ -81,14 +86,69 @@ resource "aws_sfn_state_machine" "state_machine" {
             "Next": "HandleErrors",
             "ResultPath": "$.error"
           }
-        ]
+        ],
+        "ResultPath": "$.batch"
```
danieljhegeman, line 90: I'm a little confused as to where `.batch` is coming from. What is producing/populating this output?

danieljhegeman: I understand the nature of `ResultPath` well enough I think. I am just confused as to when/where `.batch` is being assigned on the path.

danieljhegeman: Oh, this is actually _assigning_ the output as being on `.batch`. Nvm.

Bento007: If the original input was
```
{
 "foo": "bar",
}
```

The result would go
```
{
 "foo": "bar",
 "batch": "result
}
```

Bento007: the response that populates it is from here https://github.com/chanzuckerberg/single-cell-data-portal/blob/1ea8e43f4e0c8721e545dcf8532b969e7bd9211c/backend/layers/processing/process_download.py#L193

danieljhegeman: Yep

backend/layers/processing/process_download.py
```diff
@@ -75,37 +91,106 @@ def remove_prefix(string: str, prefix: str) -> str:
...
 
-    def process(self, dataset_id: DatasetVersionId, dataset_uri: str, artifact_bucket: str, sfn_task_token: str):
+    @staticmethod
+    def get_job_definion_name(dataset_vid: str) -> str:
+        if os.getenv("REMOTE_DEV_PREFIX"):
+            stack_name = os.environ["REMOTE_DEV_PREFIX"].replace("/", "")
+            prefix = f"{os.environ['DEPLOYMENT_STAGE']}-{stack_name}"
+        else:
+            prefix = f"{os.environ['DEPLOYMENT_STAGE']}"
+        job_definition_name = f"dp-{prefix}-ingest-process-{dataset_vid}"
```
danieljhegeman, line 101: I'm thinking we can improve on this name. Perhaps replace `ingest-process` with `process-validate` instead?

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -268,6 +268,16 @@ module upload_sfn {
...
+
+  # use for defining the batch job definition
+  image             = "${local.upload_image_repo}:${local.image_tag}"
+  batch_role_arn    = local.batch_role_arn
+  remote_dev_prefix = local.remote_dev_prefix
+  artifact_bucket   = local.artifact_bucket
+  cellxgene_bucket  = local.cellxgene_bucket
+  datasets_bucket   = local.datasets_bucket
+  frontend_url      = local.frontend_url
+  batch_job_log_group = module.upload_batch.batch_job_log_group
```
danieljhegeman, line 280: I feel like this must have been fed in already somewhere. Where was that? Something's not adding up in my head.

danieljhegeman, line 280: I feel like this must have been fed in already somewhere. Where was that? Something's not adding up in my head.

backend/layers/processing/process_download.py
```diff
@@ -75,37 +91,106 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+        estimated_memory_MB = max([uncompressed_size_MB * memory_modifier, min_memory_MB])
+        if estimated_memory_MB > max_memory_MB:
+            estimated_memory_MB = max_memory_MB
+            estimated_vcpus = max_vcpu
+            max_swap = swap_memory_MB
+        else:
+            estimated_vcpus = max([estimated_memory_MB / memory_per_vcpu, min_vcpu])
+            max_swap = 0
+
+        return {"Vcpus": int(ceil(estimated_vcpus)), "Memory": int(ceil(estimated_memory_MB)), "MaxSwap": max_swap}
```
danieljhegeman, line 140: move the `int(ceil())` logic up to ln 137, where you are actually calculating the `estimated_vcpus`. That var should always hold a whole number.

tests/unit/processing/test_process_download.py
```diff
@@ -1,55 +1,81 @@
...
-        self.assertEqual(status.processing_status, DatasetProcessingStatus.INITIALIZED)
-        self.assertEqual(status.upload_status, DatasetUploadStatus.WAITING)
+        # Mock SFN client
+        mock_sfn = Mock()
+        mock_sfn_provider.return_value = mock_sfn
 
+        # This is where we're at when we start the SFN
         pdv = ProcessDownload(self.business_logic, self.uri_provider, self.s3_provider)
-        pdv.process(dataset_version_id, dropbox_uri, "fake_bucket_name", "fake_sfn_task_token")
+        pdv.process(dataset_vid, dropbox_uri, bucket_name, "fake_sfn_task_token")
```
danieljhegeman, line 44: I'm not opposed to changing the way we customarily refer to various entities in terms of var naming, but given the minute nature of this change, and the fact that it isn't being applied repo-wide, I would prefer we not introduce yet another way to var-name a `dataset_version_id`. We can do a find/replace at some point later if you really want to.

danieljhegeman (changes requested):
Some minor change requests

.happy/terraform/modules/ecs-stack/main.tf
Bento007: this is created in `module.upload_batch` so it doesn't exist anywhere else.

backend/layers/processing/process_download.py
Bento007: I was trying to avoid overloading those lines with nested function calls.

backend/layers/processing/process_download.py
Bento007: it is used in more that validate. It's sued for validte, cxg, and seurat

backend/layers/processing/process_download.py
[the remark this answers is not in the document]
ebezzi: Also, since this is likely to be a sparse matrix, this is probably an over-estimate anyway. If I recall correctly, the reason for loading the matrix into memory is to determine the sparsity - but iirc, curators aren't really updating dense matrices anymore. Therefore, we might have an opportunity to optimize even more. Maybe worth a discussion with them at some point.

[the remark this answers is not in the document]
nayib-jose-gloria: schema migration will automatically update everything in the corpus, so we do need to account for any dense matrices that are already in the corpus

backend/layers/processing/process_download.py
```diff
@@ -75,37 +91,108 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+                "MaxSwap": batch_resources["MaxSwap"],
+            },
+        }
+
+    def process(
+        self, dataset_version_id: DatasetVersionId, dataset_uri: str, artifact_bucket: str, sfn_task_token: str
+    ):
         """
         1. Download the original dataset
         2. Upload the labeled dataset to the artifact bucket
```
nayib-jose-gloria, line 162: remove step 2 from the docstring, this is now done in process_validate. move to process_validate.process if not already there

nayib-jose-gloria (approved):
the exact resource allocations might need to be tweaked but I think this looks good to start testing things out

.happy/terraform/modules/ecs-stack/main.tf
danieljhegeman: I mean we aren't adding new batch steps. These parameters, like `artifact_bucket`, would have already been needed in the previous configuration of the SFN... and I'm wondering where they came in previously. Won't worry about it

tests/unit/processing/test_process_download.py
```diff
@@ -75,7 +105,87 @@ def test_download_unknown_uri(self):
...
+
+
+@pytest.mark.parametrize(
+    "environ,expected",
+    [
+        ({"REMOTE_DEV_PREFIX": "/stack/", "DEPLOYMENT_STAGE": "test"}, "dp-test-stack-ingest-process-fake_dataset_id"),
+        ({"REMOTE_DEV_PREFIX": "stack", "DEPLOYMENT_STAGE": "test"}, "dp-test-stack-ingest-process-fake_dataset_id"),
+        ({"DEPLOYMENT_STAGE": "test"}, "dp-test-ingest-process-fake_dataset_id"),
+    ],
+)
```
danieljhegeman, line 154: So this parametrize is used to pass the first element from a given tuple in as the first parameter... which is defined as the first element in a comma-delineated str... so `{"REMOTE_DEV_PREFIX": "/stack/", "DEPLOYMENT_STAGE": "test"}` is assigned to `environ`... and `"dp-test-stack-ingest-process-fake_dataset_id"` is `expected`. 

I honestly don't really see what this gets us, other than complexity in the form of adding more function decorators. If anything this is more hacky than inlining these parameters in the function itself.

tests/unit/processing/test_process_download.py
```diff
@@ -75,7 +105,87 @@ def test_download_unknown_uri(self):
...
+    mock_ProcessDownload.get_job_definion_name.assert_called_once_with("fake_dataset_id")
+    assert resp == {
+        "JobDefinitionName": "fake_job_definition_name",
+        "Vcpus": 1,
+        "Memory": 4000,
+        "LinuxParameters": {
+            "Swappiness": 60,
+            "MaxSwap": 0,
+        },
+    }
```
danieljhegeman, line 191: I'm not sure this is terribly useful... sort of just testing function implementation. I would just remove it

tests/unit/processing/test_process_download.py
Bento007: It is funky for sure, and I don’t agree with all of pytest syntax. Sometimes there is too much magic, but if you trust the magic it makes things easier. Also using pytest syntax integrates better with the plugins we use for pytest.

The fact that the tests are short and to the point makes it easier to reason about what specifically is being tested.

Also pytest doesn't play will with python unittest subtests so using parameterize is what we have.

danieljhegeman (approved):
Approving but a couple minor comments, also I would encourage you to remove the last unit test that I commented on before merging. I do not think it is testing anything real.

tests/unit/processing/test_process_download.py
Bento007: It might seem excessive but it establish a contract with the implementation that this is the behavior we expect. If the implementation changes so should the test.

Bento007:
In the future we could put the resource allocation parameters in a config file. That way it can be configured without redeploying a container.

tests/unit/processing/test_process_download.py
prathapsridharan: `parameterize` is a general implement that is useful to name test cases. So I think you could include a name for the test case (the input and expected output) and you when you run your tests on the command line, you will see something like `test-case-1` (if that is the name of the test case) failed.

This is more interpretable than including the all the test cases in a data structure like a `list` and looping through it and executing the test. A real run here:

```
tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py::test_sanitize_api_query_dict[input_query0-expected_sanitized_query0] PASSED                                                                                                    [ 81%]
tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py::test_sanitize_api_query_dict[input_query1-expected_sanitized_query1] PASSED                                                                                                    [ 83%]
tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py::test_sanitize_api_query_dict[input_query2-expected_sanitized_query2] PASSED                                                                                                    [ 85%]
tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py::test_sanitize_api_query_dict[input_query3-expected_sanitized_query3] PASSED                                                                                                    [ 87%]
tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py::test_sanitize_api_query_dict[input_query4-expected_sanitized_query4] PASSED
```

danieljhegeman: @prathapsridharan given that, how would you rearrange this particular test?

backend/layers/processing/process_download.py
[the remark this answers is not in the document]
danieljhegeman: @Bento007 did you try rerunning your migration and allocating 4 bytes per value?

Separately, iff that works, then it might be worth, as a followup, taking into account the sparsity rather than assuming dense.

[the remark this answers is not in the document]
Bento007: I did not but I did find that using getsizeof to be a slightly better approach to estimating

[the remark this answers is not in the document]
danieljhegeman: getsizeof applied to what exactly?

PR #6095, opened 2023-10-25
chore(wmg): remove wmgv1 

kaloster:
## Reason for Change

- #6092
- removed unused/dead code, merged mixed code into wmgv2, removed wmgv1 tree


seve (commented):
I never considered how much git history we'd be eradicating by developing this way, is there any way to prevent that?

kaloster:
@seve the majority of changes are just moving files around - so that will keep the existing git history in tact and contributions will stay attributed to their respective contributors. the code that was removed was unused code - if you look at the total lines added/removed here it's not that much - the lines that were added were mostly due to mixed resources (i.e styles being pulled from older files that need to be removed) - I actually tried to avoid copying as much as possible and move and rename instead to keep the history intact. 

if there's a cleaner way of doing it I'm happy to do it.

tihuan:
> @seve the majority of changes are just moving files around - so that will keep the existing git history in tact and contributions will stay attributed to their respective contributors. the code that was removed was unused code - if you look at the total lines added/removed here it's not that much - the lines that were added were mostly due to mixed resources (i.e styles being pulled from older files that need to be removed) - I actually tried to avoid copying as much as possible and move and rename instead to keep the history intact.
> 
> if there's a cleaner way of doing it I'm happy to do it.

Ooh yeah really good point!

It seems like we can preserve git history by using `git mv` command!

------ CGPT says -------

When you move a file in a Git repository, you can preserve the file's history by using the `git mv` command. Git is designed to handle file renames and moves efficiently while maintaining the file's history.

Here's how you can move a file while preserving its history:

1. Use the `git mv` command to rename or move the file. For example:

   ```bash
   git mv oldfile.txt newfile.txt
   ```

   If you're moving the file to a different directory, you can specify the new directory in the command:

   ```bash
   git mv oldfile.txt path/to/newfile.txt
   ```

2. Commit the changes:

   ```bash
   git commit -m "Moved oldfile to newfile"
   ```

This will ensure that Git tracks the move as a rename/move operation, and the history of the file will be preserved.

If you've already moved the file without using `git mv`, Git will still try to detect renames and moves based on the content similarity between the files. However, it might not always correctly detect the move, especially if the changes between the files are substantial. In such cases, you can use the `--find-renames` option with `git log` to see the history of the moved file:

```bash
git log --follow path/to/newfile.txt
```

The `--follow` option makes Git track the history of the file even if it was moved or renamed.

In summary, using `git mv` is the best practice for preserving file history when moving or renaming files in Git. If you've already moved the file, Git's similarity-based rename detection can help, but using `git mv` is more reliable.

kaloster:
Thanks for this @tihuan! I find git is pretty good at tracking moves and changes and will use `git mv` where possible 🙏🏽

tihuan (approved):
LGTM! Thanks so much, @kaloster for cleaning this up!

@seve - @kaloster and I have discussed about keeping the v1 history vs. keeping v2 history and decided that keeping v2 history is more valuable. Do you agree?

Thanks!

seve (approved):

PR #6102, opened 2023-10-25
feat: Add support for schema-4 format for "self_reported_ethnicity_ontology_term_id"

prathapsridharan:
## Reason for Change

- The specific ticket being addressed is here: https://github.com/chanzuckerberg/single-cell/issues/528
- For the full story see the containing epic here: https://github.com/chanzuckerberg/single-cell/issues/460

## Changes

- add support for comma separated `self_reported_ethnicity_ontology_term_id`

## Testing steps

- Unit tests

### DIFF
--- a/backend/wmg/data/ontology_labels.py
+++ b/backend/wmg/data/ontology_labels.py
@@ -22,10 +22,10 @@ SUFFIXES_TO_STRIP = ["organoid", "cell culture"]
 
 def ontology_term_label(ontology_term_id: str) -> Optional[str]:
     """
-    Returns the label for an ontology term, given its id. This excludes gene ontology term, which are handled
-    separately by gene_gene_term_label(). Return None if ontology term id is invalid.
+    Returns the label for an ontology term, given its id. This excludes gene ontology term
+    and self reported ethnicity term, which are handled separately by gene_gene_term_label()
+    and ethnicity_term_label() respectively. Return None if ontology term id is invalid.
     """
-    global ontology_term_id_labels
 
     # this catches the organoid tissue edge case (e.g. UBERON:0000995 (organoid)) or the cell culture edge case
     # (e.g. CL:0000082 (cell culture))
@@ -44,18 +44,48 @@ def ontology_term_label(ontology_term_id: str) -> Optional[str]:
     return ontology_term_id_label
 
 
+def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> str:
+    """
+    Return the label for self reported ethnicity ontology term id.
+
+    This function is compatible with schema-4 and schema-3 format of
+    `self_reported_ethnicity_ontology_term_id`.
+
+    NOTE: It is assumed that self_reported_ethnicity_ontology_term_id has been validated against
+    a schema (ex: schema-3 or schema-4) before this function is called. Therefore,
+    no error error checking on the format of the input is done.
+
+    Parameters
+    ----------
+    self_reported_ethnicity_ontology_term_id: An ontology_term_id string that can
+                                              encode multiple ethnicities with comma delimiter
+                                              in Schema-4.
+                                              NOTE: Schema-3 does not support comma delimited
+                                              ontology term IDs to encode multiple ethnicities
+                                              in a single string.
+
+    Returns
+    -------
+    ethnicity_term_label: A comma delimited ethnicity label corresponding to the input ethnicity_term_id
+    """
+    # In schema-4, self_reported_ethnicity_ontology_term_id can be a comma
+    # separated string to denote multiple ethnicities.
+    individual_term_ids = self_reported_ethnicity_ontology_term_id.split(",")
+    term_labels = [ontology_term_id_labels.get(term_id) for term_id in individual_term_ids]
+    ethnicity_term_label = ",".join(term_labels)
+    return ethnicity_term_label
+
+
 def gene_term_label(gene_ontology_term_id: str) -> Optional[str]:
     """
     Returns the label for a gene ontology term, given its id. Return None if ontology term id is invalid.
     """
-    global gene_term_id_labels
     return gene_term_id_labels.get(gene_ontology_term_id)
 
 
 def __load_ontologies() -> None:
     global ontology_term_id_labels
 
-    ontology_term_id_labels = {}
     all_ontologies = json.loads(__open_ontology_resource(all_ontologies_json_file).read().decode("utf-8"))
     for _, ontology_dict in all_ontologies.items():
         for term_id, term in ontology_dict.items():
@@ -65,7 +95,6 @@ def __load_ontologies() -> None:
 def __load_genes() -> None:
     global gene_term_id_labels
 
-    gene_term_id_labels = {}
     for genes_file in genes_files:
         for row in csv.DictReader(
             __open_ontology_resource(genes_file).read().decode("utf-8").split("\n"),
@@ -80,6 +109,9 @@ def __open_ontology_resource(file) -> IO:
     file_path = root_path.joinpath("common", "ontology_files", file)
     return gzip.open(file_path)
 
+
+ontology_term_id_labels = {}
+gene_term_id_labels = {}
 
 __load_ontologies()
 __load_genes()
--- a/backend/wmg/data/ontology_labels.py
+++ b/backend/wmg/data/ontology_labels.py
@@ -47,7 +47,7 @@ def ontology_term_label(ontology_term_id: str) -> Optional[str]:
     return ontology_term_id_label
 
 
-def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> str:
+def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> Optional[str]:
     """
     Return the label for self reported ethnicity ontology term id.
 
@@ -70,6 +70,8 @@ def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> str:
     Returns
     -------
     ethnicity_term_label: A comma delimited ethnicity label corresponding to the input ethnicity_term_id
+                          or None if the ethnicity term id "unknown", "na", "multiethnic" or some
+                          value that is not recognized.
     """
     # In schema-4, self_reported_ethnicity_ontology_term_id can be a comma
     # separated string to denote multiple ethnicities.
@@ -80,7 +82,7 @@ def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> str:
     logger.info(f"PRATHAP! ethnicity_term_label() func - individual_term_ids: {individual_term_ids}")
     term_labels = [ontology_term_id_labels.get(term_id) for term_id in individual_term_ids]
     logger.info(f"PRATHAP! ethnicity_term_label() func - term_labels: {term_labels}")
-    ethnicity_term_label = ",".join(term_labels)
+    ethnicity_term_label = ",".join(term_labels) if all([isinstance(t, str) for t in term_labels]) else None
     return ethnicity_term_label
 
 
--- a/backend/wmg/data/ontology_labels.py
+++ b/backend/wmg/data/ontology_labels.py
@@ -20,7 +20,7 @@ genes_files = [
 
 SUFFIXES_TO_STRIP = ["organoid", "cell culture"]
 
-logger = logging.getLogger("wmg")
+logger = logging.getLogger("wmg-ontology-labels")
 
 
 def ontology_term_label(ontology_term_id: str) -> Optional[str]:
@@ -74,10 +74,12 @@ def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> str:
     # In schema-4, self_reported_ethnicity_ontology_term_id can be a comma
     # separated string to denote multiple ethnicities.
     individual_term_ids = self_reported_ethnicity_ontology_term_id.split(",")
-    logger.info(f"PRATHAP! self_reported_ethnicity_ontology_term_id: {self_reported_ethnicity_ontology_term_id}")
-    logger.info(f"PRATHAP! individual_term_ids: {individual_term_ids}")
+    logger.info(
+        f"PRATHAP! ethnicity_term_label() func - self_reported_ethnicity_ontology_term_id: {self_reported_ethnicity_ontology_term_id}"
+    )
+    logger.info(f"PRATHAP! ethnicity_term_label() func - individual_term_ids: {individual_term_ids}")
     term_labels = [ontology_term_id_labels.get(term_id) for term_id in individual_term_ids]
-    logger.info(f"PRATHAP! term_labels: {term_labels}")
+    logger.info(f"PRATHAP! ethnicity_term_label() func - term_labels: {term_labels}")
     ethnicity_term_label = ",".join(term_labels)
     return ethnicity_term_label
 
--- a/backend/wmg/data/ontology_labels.py
+++ b/backend/wmg/data/ontology_labels.py
@@ -1,7 +1,6 @@
 import csv
 import gzip
 import json
-import logging
 import pathlib
 from typing import IO, Optional
 
@@ -19,8 +18,6 @@ genes_files = [
 ]
 
 SUFFIXES_TO_STRIP = ["organoid", "cell culture"]
-
-logger = logging.getLogger("wmg-ontology-labels")
 
 
 def ontology_term_label(ontology_term_id: str) -> Optional[str]:
@@ -76,12 +73,7 @@ def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> Optio
     # In schema-4, self_reported_ethnicity_ontology_term_id can be a comma
     # separated string to denote multiple ethnicities.
     individual_term_ids = self_reported_ethnicity_ontology_term_id.split(",")
-    logger.info(
-        f"PRATHAP! ethnicity_term_label() func - self_reported_ethnicity_ontology_term_id: {self_reported_ethnicity_ontology_term_id}"
-    )
-    logger.info(f"PRATHAP! ethnicity_term_label() func - individual_term_ids: {individual_term_ids}")
     term_labels = [ontology_term_id_labels.get(term_id) for term_id in individual_term_ids]
-    logger.info(f"PRATHAP! ethnicity_term_label() func - term_labels: {term_labels}")
     ethnicity_term_label = ",".join(term_labels) if all([isinstance(t, str) for t in term_labels]) else None
     return ethnicity_term_label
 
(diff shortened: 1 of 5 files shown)


tests/unit/backend/wmg/api/test_v2.py
```diff
@@ -707,17 +714,19 @@ def test__filter_request_with_query_key_not_specified_in_api_spec_returns_400(se
 
     @patch("backend.wmg.api.v2.fetch_datasets_metadata")
     @patch("backend.wmg.api.v2.gene_term_label")
+    @patch("backend.wmg.api.v2.ethnicity_term_label")
     @patch("backend.wmg.api.v2.ontology_term_label")
     @patch("backend.wmg.api.v2.load_snapshot")
     def test__filter_request_with_filter_dims__returns_valid_filter_dims__base_case(
-        self, load_snapshot, ontology_term_label, gene_term_label, fetch_datasets_metadata
+        self, load_snapshot, ontology_term_label, ethnicity_term_label, gene_term_label, fetch_datasets_metadata
```
prathapsridharan, line 721: @atarashansky - Should there be a test case for a request to `/filters` for comma separated `self_reported_ethnicity_ontology_term_id`?

```
filter: {
self_reported_ethnicity_ontology_term_ids: ["self_reported_ethnicity_ontology_term_id_1,self_reported_ethnicity_ontology_term_id_2"]
```

It is not clear what the `"self_reported_ethnicity_terms"` key in the response would contain because that involves a call to a rather dense function `find_dim_option_values()`: https://github.com/chanzuckerberg/single-cell-data-portal/blob/1a8f0992d80e97933072e2262e02712500cea382/backend/wmg/api/v2.py#L218C29-L218C29

tests/unit/backend/wmg/api/test_v2.py
```diff
@@ -756,14 +765,16 @@ def test__filter_request_with_filter_dims__returns_valid_filter_dims__base_case(
...
             self.assertEqual(json.loads(response.data)["filter_dims"], expected_filters)
 
     @patch("backend.wmg.api.v2.fetch_datasets_metadata")
     @patch("backend.wmg.api.v2.gene_term_label")
+    @patch("backend.wmg.api.v2.ethnicity_term_label")
     @patch("backend.wmg.api.v2.ontology_term_label")
     @patch("backend.wmg.api.v2.load_snapshot")
     def test__filter_request_with_filter_dims__returns_valid_filter_dims(
-        self, load_snapshot, ontology_term_label, gene_term_label, fetch_datasets_metadata
+        self, load_snapshot, ontology_term_label, ethnicity_term_label, gene_term_label, fetch_datasets_metadata
```
prathapsridharan, line 777: @atarashansky - Should a similar test case including comma separated `self_reported_ethnicity_ontology_term_id` be added here too?

backend/wmg/data/ontology_labels.py
```diff
@@ -44,18 +44,48 @@ def ontology_term_label(ontology_term_id: str) -> Optional[str]:
...
+def ethnicity_term_label(self_reported_ethnicity_ontology_term_id: str) -> str:
+    """
+    Return the label for self reported ethnicity ontology term id.
+
+    This function is compatible with schema-4 and schema-3 format of
+    `self_reported_ethnicity_ontology_term_id`.
+
+    NOTE: It is assumed that self_reported_ethnicity_ontology_term_id has been validated against
+    a schema (ex: schema-3 or schema-4) before this function is called. Therefore,
+    no error error checking on the format of the input is done.
```
joyceyan, line 56: nit: error error :)

backend/wmg/data/ontology_labels.py
```diff
@@ -22,10 +22,10 @@
 
 def ontology_term_label(ontology_term_id: str) -> Optional[str]:
     """
-    Returns the label for an ontology term, given its id. This excludes gene ontology term, which are handled
-    separately by gene_gene_term_label(). Return None if ontology term id is invalid.
+    Returns the label for an ontology term, given its id. This excludes gene ontology term
+    and self reported ethnicity term, which are handled separately by gene_gene_term_label()
```
joyceyan, line 26: when is gene_term_label used?

joyceyan:
Code seems good but would be nice to have this tested against an rdev with schema 4 multi-ethnic values. I think @nayib-jose-gloria said there are schema 4 datasets for each test scenario, so I think we should be able to find a dataset that only has comma-delimited self_reported_ethnicity string.

backend/wmg/data/ontology_labels.py
prathapsridharan: Right here: https://github.com/chanzuckerberg/single-cell-data-portal/blob/1a8f0992d80e97933072e2262e02712500cea382/backend/wmg/api/v2.py#L339

joyceyan: can you also rename this comment to gene_term_label lol

prathapsridharan:
> Code seems good but would be nice to have this tested against an rdev with schema 4 multi-ethnic values. I think @nayib-jose-gloria said there are schema 4 datasets for each test scenario, so I think we should be able to find a dataset that only has comma-delimited self_reported_ethnicity string.

@nayib-jose-gloria - Following up on the above

1. Do you have an entire schema-4 data layer (that is, all our relational DBs and tile-db cubes) containing comma-separated string values for `self_reported_ethnicity_ontology_term_id` that this change can be tested against? For example, given a schema-4 database (relational DBs and cubes), I would like to see what `/wmg/v2/filters` or `/wmg/v2/query` returns
2. The core of the implementation is just that `self_reported_ethnicity_ontology_term_id` is just transformed into english labels by first splitting on `,` (ex: `self_reported_ethnicity_ontology_term_id.split(',')`). Because of how `split()` method on string works, this is backward compatible with schema-3 where there are no comma delimited values for `self_reported_ethnicity_ontology_term_id`. I think we can skip gating this modification on the feature flag. What do you think?

nayib-jose-gloria:
> > Code seems good but would be nice to have this tested against an rdev with schema 4 multi-ethnic values. I think @nayib-jose-gloria said there are schema 4 datasets for each test scenario, so I think we should be able to find a dataset that only has comma-delimited self_reported_ethnicity string.
> 
> @nayib-jose-gloria - Following up on the above
> 
> 1. Do you have an entire schema-4 data layer (that is, all our relational DBs and tile-db cubes) containing comma-separated string values for `self_reported_ethnicity_ontology_term_id` that this change can be tested against? For example, given a schema-4 database (relational DBs and cubes), I would like to see what `/wmg/v2/filters` or `/wmg/v2/query` returns
> 2. The core of the implementation is just that `self_reported_ethnicity_ontology_term_id` is just transformed into english labels by first splitting on `,` (ex: `self_reported_ethnicity_ontology_term_id.split(',')`). Because of how `split()` method on string works, this is backward compatible with schema-3 where there are no comma delimited values for `self_reported_ethnicity_ontology_term_id`. I think we can skip gating this modification on the feature flag. What do you think?

The [rdev](https://schema-4-backend.rdev.single-cell.czi.technology/curation/v1/collections/9035c5da-d5fa-4445-8908-0e3a972221f4) I provided a bit ago (schema-4) will validate schema 4 compliant uploads. It includes 1 dataset with a comma-delimited self reported ethnicity value. It should have generated a valid cxg as well. I'm not sure how WMG generation / testing in an rdev works, but you're free to leverage that dataset and/or upload other schema 4 compliant datasets to that rdev stack that you'd like to test with. The relevant cxg will be uploaded to s3://env-rdev-cellxgene/schema-4/{dataset_version_id}.cxg 

I believe Alec/Emanuele have alluded to creating rdevs where the input to the WMG cube builders are JSONs with mock Discover API values that they'd like to test with. That also seems like a good route if its not feasible to generate an rdev WMG cube using the rdev linked above

nayib-jose-gloria:
> Code seems good but would be nice to have this tested against an rdev with schema 4 multi-ethnic values. I think @nayib-jose-gloria said there are schema 4 datasets for each test scenario, so I think we should be able to find a dataset that only has comma-delimited self_reported_ethnicity string.
> 
> @nayib-jose-gloria - Following up on the above
> 
> 2. The core of the implementation is just that `self_reported_ethnicity_ontology_term_id` is just transformed into english labels by first splitting on `,` (ex: `self_reported_ethnicity_ontology_term_id.split(',')`). Because of how `split()` method on string works, this is backward compatible with schema-3 where there are no comma delimited values for `self_reported_ethnicity_ontology_term_id`. I think we can skip gating this modification on the feature flag. What do you think?

That seems reasonable to me

prathapsridharan:
> The [rdev](https://schema-4-backend.rdev.single-cell.czi.technology/curation/v1/collections/9035c5da-d5fa-4445-8908-0e3a972221f4) I provided a bit ago (schema-4) will validate schema 4 compliant uploads. It includes 1 dataset with a comma-delimited self reported ethnicity value. It should have generated a valid cxg as well. I'm not sure how WMG generation / testing in an rdev works, but you're free to leverage that dataset and/or upload other schema 4 compliant datasets to that rdev stack that you'd like to test with. The relevant cxg will be uploaded to s3://env-rdev-cellxgene/schema-4/{dataset_version_id}.cxg
> 
> I believe Alec/Emanuele have alluded to creating rdevs where the input to the WMG cube builders are JSONs with mock Discover API values that they'd like to test with. That also seems like a good route if its not feasible to generate an rdev WMG cube using the rdev linked above

@atarashansky - Do you have any ideas about testing here?

1.[ I mentioned unit testing for the current backend change](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6102#discussion_r1372403002). We could modify the [create_temp_wmg_snapshot()](https://github.com/chanzuckerberg/single-cell-data-portal/blob/08e5da82e05ff2facce3be6d17949d7be80060cb/tests/unit/backend/wmg/fixtures/test_snapshot.py#L255) fixture generating function to be able to test comma-delimited values [in unit tests like these](https://github.com/chanzuckerberg/single-cell-data-portal/blob/08e5da82e05ff2facce3be6d17949d7be80060cb/tests/unit/backend/wmg/api/test_v2.py#L754). Is that sufficient as a test of backend functionality rather than generating a WMG cube for testing (at least for this comma-delimited value support)? Is that easier to do than generating WMG cube for testing?
2. The WMG cubes generation on an rdev approach seems a bit unclear to me given that we need some new way to generate the cube to ingest the single schema-4 dataset that @nayib-jose-gloria [has provided in his rdev](https://schema-4-backend.rdev.single-cell.czi.technology/curation/v1/collections/9035c5da-d5fa-4445-8908-0e3a972221f4) AND it is unclear how you would we would isolate this single dataset to test its implication on WMG frontend

I am in favor (1) more than (2) but wanted your thoughts

joyceyan:
(1) sounds reasonable to me if (2) is prohibitively difficult

prathapsridharan:
@atarashansky @joyceyan - After thinking through this some more, the unit test I have provided in this PR are sufficient. Add unit tests for higher level functions like `/filters` are not useful because their logic (at least for the case `self_reported_ethnicity_ontology_term_id`) are agnostic to the value of the field.

I believe this PR is ready in terms of functionality and test coverage

joyceyan (changes requested):
> Add unit tests for higher level functions like /filters are not useful because their logic (at least for the case self_reported_ethnicity_ontology_term_id) are agnostic to the value of the field.

can you explain why testing the endpoint wouldn't be useful? there are plenty of cases i've seen where you might write tests for the lower level function, and you'll encounter an issue because the full end to end flow isn't tested. if the concern is that writing the unit tests is prohibitively difficult and the cost/benefit analysis isn't worth it, then that's i think a separate issue. but i don't think it's correct to say that unit tests on the endpoint level are not useful.

atarashansky (changes requested):

backend/wmg/data/ontology_labels.py
```diff
@@ -22,10 +22,10 @@
 
 def ontology_term_label(ontology_term_id: str) -> Optional[str]:
     """
-    Returns the label for an ontology term, given its id. This excludes gene ontology term, which are handled
-    separately by gene_gene_term_label(). Return None if ontology term id is invalid.
+    Returns the label for an ontology term, given its id. This excludes gene ontology term
```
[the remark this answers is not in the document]
prathapsridharan: @atarashansky - Per the documentation for this function, this function should not be called for `gene_ontology_term_id` and `self_reported_ethnicity_ontology_term_id`.

If you want to translate `self_reported_ethnicity_ontology_term_id` to a label, you should call `ethnicity_term_lablel()` below.

The caller that determines which of these functions to call to translate a `*_ontology_term_id` to term label is [build_ontology_term_id_label_mapping()](https://github.com/chanzuckerberg/single-cell-data-portal/blob/32b8b0edbdb75fb065d9b72ed8b7c4ceadd7037d/backend/wmg/api/v2.py#L357C6-L357C6) for which there also [accompanying unit tests](https://github.com/chanzuckerberg/single-cell-data-portal/blob/32b8b0edbdb75fb065d9b72ed8b7c4ceadd7037d/tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py)

backend/wmg/data/ontology_labels.py
```diff
@@ -22,10 +22,10 @@
...
+    and self reported ethnicity term, which are handled separately by gene_term_label()
+    and ethnicity_term_label() respectively. Return None if ontology term id is invalid.
     """
-    global ontology_term_id_labels
```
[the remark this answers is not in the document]
prathapsridharan: @atarashansky - Why do you say that? Python operates by _lexical scoping_ which means that it will look in its parent frame for `ontology_term_id_labels` which is the `ontology_lables.py` file where, upon import, declares `ontology_term_id_labels = {}` and `gene_term_id_labels = {}` in [lines 115-116](https://github.com/chanzuckerberg/single-cell-data-portal/blob/32b8b0edbdb75fb065d9b72ed8b7c4ceadd7037d/backend/wmg/data/ontology_labels.py#L115)

I think `global` is only required when you want to mutate a variable that is in an ancestor frame as in [__load_ontologies() function](https://github.com/chanzuckerberg/single-cell-data-portal/blob/32b8b0edbdb75fb065d9b72ed8b7c4ceadd7037d/backend/wmg/data/ontology_labels.py#L89) and [__load_genes() function](https://github.com/chanzuckerberg/single-cell-data-portal/blob/32b8b0edbdb75fb065d9b72ed8b7c4ceadd7037d/backend/wmg/data/ontology_labels.py#L98C8-L98C8)

In the case of mutation in functions like `__load_ontologies()`, if you didn't declare `global ontologies_term_id_labels` you would get an `UnboundLocalError`.

More info here: https://realpython.com/python-use-global-variable-in-function/#the-global-keyword

Moreover, passing tests and a functioning rdev indicate that this refactor is working

backend/wmg/data/ontology_labels.py
```diff
@@ -44,18 +44,50 @@ def ontology_term_label(ontology_term_id: str) -> Optional[str]:
...
     Returns the label for a gene ontology term, given its id. Return None if ontology term id is invalid.
     """
-    global gene_term_id_labels
     return gene_term_id_labels.get(gene_ontology_term_id)
 
 
 def __load_ontologies() -> None:
     global ontology_term_id_labels
 
-    ontology_term_id_labels = {}
```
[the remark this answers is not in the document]
prathapsridharan: My reasoning here: https://github.com/chanzuckerberg/single-cell-data-portal/pull/6102/files#r1376607274

backend/wmg/data/ontology_labels.py
```diff
@@ -65,7 +97,6 @@ def __load_ontologies() -> None:
 def __load_genes() -> None:
     global gene_term_id_labels
 
-    gene_term_id_labels = {}
```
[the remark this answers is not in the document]
prathapsridharan: My reasoning here: https://github.com/chanzuckerberg/single-cell-data-portal/pull/6102/files#r1376607274

prathapsridharan:
> > > Add unit tests for higher level functions like /filters are not useful because their logic (at least for the case self_reported_ethnicity_ontology_term_id) are agnostic to the value of the field.
> > 
> > 
> > can you explain why testing the endpoint wouldn't be useful? there are plenty of cases i've seen where you might write tests for the lower level function, and you'll encounter an issue because the full end to end flow isn't tested. if the concern is that writing the unit tests is prohibitively difficult and the cost/benefit analysis isn't worth it, then that's i think a separate issue. but i don't think it's correct to say that unit tests on the endpoint level are not useful.
> 
> If the tests are dependent on realistic test snapshot, then updating them to include the comma-separated ethnicities is not feasible until the test fixtures are updated to contain those entries. Currently, the test fixtures are generated as subsets of the WMG snapshot, so we would need to wait for schema 4 wmg snapshot.
> 
> If the tests are using `create_tmp_wmg_snapshot` then we could create an ethnicity metadata generator that creates multi-value terms. Though, I personally don't see the API-level tests as being that important in this instance. I think as long as we test the function that the API calls to perform the ontology label mapping and include a multi-ethnic test case, then we should be fine.

@joyceyan @atarashansky - Which is what these tests do (test the ontology id to label mapping):

1. https://github.com/chanzuckerberg/single-cell-data-portal/pull/6102#issuecomment-1785646502
2. https://github.com/chanzuckerberg/single-cell-data-portal/blob/32b8b0edbdb75fb065d9b72ed8b7c4ceadd7037d/tests/unit/backend/wmg/api/common/test_wmg_api_helpers.py

backend/wmg/data/ontology_labels.py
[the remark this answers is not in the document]
atarashansky: Oh, lol, I didn't see lines 115-116 (which is funny considering that I was the one who added it there). I think the "global" keywords are vestigial, then...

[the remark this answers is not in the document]
atarashansky: Thanks for the clarification!

tests/unit/backend/wmg/api/test_v2.py
prathapsridharan: @atarashansky - Yea that is the conclusion I came to and [added this comment](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6102#issuecomment-1783539492) stating that unit tests that test the conversion from `self_reported_ethnicity_ontology_term_id` to label _for multi-valued ethnicity ontology IDs_ are sufficient

prathapsridharan:
Closing this PR because in a code review with @joyceyan @atarashansky on 10/30/2023, it was decided that we are _filtering out comma-delimited ethnicity term IDs_ from grouping and filtering if they are done by `self_reported_ethnicity_ontology_term_id`. 

This filtering work is the subject of this ticket: https://github.com/chanzuckerberg/single-cell/issues/596

Thus no translation of term ID to term label is needed for comma-delimited term IDs for `self_reported_ethnicity_ontology_term_id`

PR #6116, opened 2023-10-30
feat(wmg): remove redundant cell type nodes in wmg

atarashansky:
## Reason for Change

- [Context](https://czi-sci.slack.com/archives/C0247APK621/p1698344084419589)
- This PR will improve performance of GE due to dataframe sizes being much smaller

## Changes

- added a function to check if nodes are redundant by seeing if they have any descendants that contain the same number of cells. these entries are filtered out from the cell counts dataframe.

## Testing steps

- Added a unit test and updated existing tests


backend/common/utils/rollup.py
```diff
@@ -32,6 +32,56 @@ def ancestors(cell_type):
...
+        prefix = cell_type
+        suffix = ""
+
+    # find descendants of cell type and re-suffix them
+    relatives = [f"{i}{suffix}" for i in descendants(prefix)]
+    valid_relatives = list(valid_cell_types.intersection(relatives))
+    if cell_counts is not None:
+        relative_cell_counts = [
+            cell_counts.get(relative)
+            for relative in valid_relatives
```
sidneymbell, line 76: The indentation looks broken here (and it's missing a `:`)

atarashansky: Haha, this is a multi-line list comprehension.

backend/common/utils/rollup.py
```diff
@@ -32,6 +32,56 @@ def ancestors(cell_type):
     return ancestors
 
 
+def get_valid_descendants(
+    cell_type: str, valid_cell_types: frozenset[str], cell_counts: Optional[dict[str, int]] = None
+):
+    """
+    Get valid descendants for a cell type.
```
prathapsridharan, line 39: Adding a bit more description here would be really useful later. This function is used in 2 ways:

1. To get a simple list of descendants for a cell-type _pre-rollup_ with no pruning
2. To identify cell-types to exclude _post-rollup_ per the heuristic by return `[]` for its descendants, which the caller (`are_cell_types_not_redundant_nodes()`) will interpret as a cell-type to exclude post rollup.

I think it would also be useful to state in the docstring to look at `are_cell_types_not_redundant_nodes()` for a description of the heuristic

backend/common/utils/rollup.py
```diff
@@ -55,30 +105,36 @@ def find_descendants_per_cell_type(cell_types):
...
+        valid_descendants.append(lookup_table[cell_type])
+    return valid_descendants
 
-        relatives_per_cell_type.append(lookup_table[cell_type])
 
-    return relatives_per_cell_type
+def are_cell_types_not_redundant_nodes(cell_types, cell_counts):
+    """
+    Determines whether each cell type in a list of cell types is not a redundant node.
+
```
prathapsridharan, line 121: You should say a bit about the heuristic because it is not obvious at all. Something like:

If a cell-type `A` has the same cell-count as its descendant cell-type `B`, then we conclude that `A` and `B` must be connected by a single relationship chain where `A` and the cell-types leading `A` to `B` in the chain have zero cell-count and therefore the entire chain `[A,B)` (that is excluding `B`) will be excluded.

And then refer the reader to `get_valid_descendants()` to understand how the exclusion is done

backend/common/utils/rollup.py
```diff
@@ -32,6 +32,56 @@ def ancestors(cell_type):
...
+    # find descendants of cell type and re-suffix them
+    relatives = [f"{i}{suffix}" for i in descendants(prefix)]
+    valid_relatives = list(valid_cell_types.intersection(relatives))
+    if cell_counts is not None:
+        relative_cell_counts = [
+            cell_counts.get(relative)
+            for relative in valid_relatives
+            if relative != cell_type and cell_counts.get(relative) is not None
+        ]
+        if cell_counts.get(cell_type) in relative_cell_counts:
```
prathapsridharan, line 79: Should we check that `len(relative_cell_counts) == 1` and raise an exception otherwise? Because are now looking for a linear chain here right?

Also a comment here to better explain why we are doing this will be really useful for someon reading this later

prathapsridharan (changes requested):
The code looks good. My comments are around adding a basic test and about writing some explanatory comments. The heuristic is not obvious and would benefit from some comment writing over code reading because the code doesn't reveal the heuristic (ie a linear chain of cell-types where nodes in the chain that have zero cell-count pre-rollup are pruned)

tests/unit/backend/wmg/data/test_rollup.py
```diff
@@ -55,3 +59,42 @@ def test__cell_types_in_same_lineage_are_colinear(self):
...
             a, b = cell_types
             assert are_cell_types_colinear(a, b) == expected
+
+    def test__cell_types_are_not_redundant(self):
+        cell_counts = {
+            "CL:0000127;;UBERON:0000955--HANCESTRO:0005": 10,
+            "CL:0000127;;UBERON:0000955--HANCESTRO:0006": 10,
+            "CL:0000127;;UBERON:0000955--HANCESTRO:0008": 50,
+            "CL:0000127;;UBERON:0000955--multiethnic": 70,
+            "CL:0000127;;UBERON:0000955--unknown": 410,
```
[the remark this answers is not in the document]
atarashansky: What if there were 250 cells labeled as `CL:0000127;;UBERON:0000955--unknown`? ;) It's not invalid.

tests/unit/backend/wmg/data/test_rollup.py
```diff
@@ -55,3 +59,42 @@ def test__cell_types_in_same_lineage_are_colinear(self):
...
+                "CL:0002605;;UBERON:0000955--multiethnic",
+                "CL:0002627;;UBERON:0000955--HANCESTRO:0006",
+                "CL:0002627;;UBERON:0000955--HANCESTRO:0008",
+                "CL:0002627;;UBERON:0000955--multiethnic",
+                "CL:0002627;;UBERON:0000955--unknown",
+            ],
+            dtype="<U46",
+        )
+
+        result = are_cell_types_not_redundant_nodes(cell_type_groups, cell_counts)
```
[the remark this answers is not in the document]
atarashansky: I understand your point, but I don't want to have to invert the output unecessarily when I can just return exactly what is needed for the downstream step.

backend/common/utils/rollup.py
atarashansky: Why does the length of `relative_cell_counts` have to be 1?

atarashansky: I can't think of any trivial assertions we can make that would ensure the sums are calculated properly. That's why we have separate unit tests for the rollup operation.

tests/unit/backend/wmg/data/test_rollup.py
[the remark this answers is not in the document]
prathapsridharan: Yea its why my comment was about readability. I would have to recall that several descendants are not shown here. Moreover in such a scenario, those 250 cells would be available in the input to this function. So when thinking about the input to represent the whole set, you expect to see numbers that add up.

When reading this I had to ask myself "why is this number 410?" and then reason about it.

Anyways, it is not a show stopper. I leave it to your discretion. You might reason about things differently.

tests/unit/backend/wmg/data/test_rollup.py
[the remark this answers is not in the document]
prathapsridharan: fair enough

backend/common/utils/rollup.py
prathapsridharan: Yea sorry I was mistaken about the assertion that `len(relative_cell_counts) == 1`. 

You could have the following graph (expressed as an adjacency list here):

```
A: B
B: C
C: D
D: [E, F]
```

Here `[A, B, C, D]` form a linear chain. 

Assume _pre-rollup_ numbers of:

```
A: 0
B: 0
C: 0
D: 1
E: 10
F: 2
```

_Unpruned post-rollup_ numbers:

```
A: 13
B: 13
C: 13
D: 13
E: 10
F: 2
```

The descendants of `A` would be `[B, C, D, E, F]`. Then _post-rollup_ the descendant cell-counts list for `A` (ie `relative_cell_counts`) would have entries `[B: 13, C: 13, D: 13, E: 10, F: 2]`, which has length greater than 1. In this case, you would exclude `A, B, C` when processing each in turn

prathapsridharan (approved):
LGTM

PR #6121, opened 2023-10-31
feat: Add lightweight FeatureFlagService

joyceyan:
## Reason for Change

The goal of this is to provide a lightweight layer for us to access feature flags set up in `CorporaConfig`. This will allow us to, if we ever desire to, swap out the underlying implementation of feature flags. I provided some basic documentation within `feature_flag.py` for how to use this.

Feature flags use a design pattern called string literals. String literals use the `Literal` provided in Python typing to ensure type strictness similar to enums. Since `FeatureFlagService.is_enabled` requires passing in a `FeatureFlag` type, we can ensure that only strings that exist within this `Literal` are allowed. Since they're simply strings, they can be easily passed into functions like `getattr(CorporaConfig(), feature_flag)` without needing to convert between strings and enums.

## Changes

- Adds the FeatureFlagService, and declares `schema_4_feature_flag` as the only valid feature flag
- Updates all instances of `CorporaConfig().schema_4_feature_flag.lower() == "true"` to use `FeatureFlagService.is_enabled(FeatureFlagValues.SCHEMA_4`

## Testing steps

- Adds a test, `TestFeatureFlag` that mocks `CorporaConfig.__getattr__` to verify that the service is working as expected

## Notes for Reviewer
n/a

### DIFF
--- a/backend/curation/api/v1/curation/collections/common.py
+++ b/backend/curation/api/v1/curation/collections/common.py
@@ -4,6 +4,7 @@ from urllib.parse import urlparse
 from uuid import UUID
 
 from backend.common.corpora_config import CorporaConfig
+from backend.common.feature_flag import FeatureFlagService, FeatureFlagValues
 from backend.common.utils.http_exceptions import ForbiddenHTTPException, GoneHTTPException, NotFoundHTTPException
 from backend.layers.auth.user_info import UserInfo
 from backend.layers.business.business import BusinessLogic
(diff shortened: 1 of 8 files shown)


nayib-jose-gloria (approved):
thank you!

backend/curation/api/v1/curation/collections/common.py
```diff
@@ -4,6 +4,7 @@
 from uuid import UUID
 
 from backend.common.corpora_config import CorporaConfig
+from backend.common.feature_flag import FeatureFlagService, FeatureFlagValues
```
Bento007, line 7: Is there ever a situation where `FeatureFlagService` and not `FeatureFlagValues` would be imported? `FeatureFlagValues` could be a property  of `FeatureFlagService` to simplify or maybe something similar to.

joyceyan: I don't think so, but there are cases where you'd want to import `FeatureFlagValues` and not `FeatureFlagService`.

PR #6122, opened 2023-10-31
feat(4.0): Split download-validate into download and validate

Bento007:
## Reason for Change

Splitting download-validate into download and validate step function steps. The download step will be used prepare the future steps for a dynamically size batch job.

- https://github.com/chanzuckerberg/single-cell/issues/543

## Changes

- - Remove unused code from downloader.py
- remove
- modify

## Testing steps

## Checklist 🛎️


.happy/terraform/modules/sfn/main.tf
```diff
@@ -22,20 +22,74 @@ resource "aws_sfn_state_machine" "state_machine" {
...
+                "Name": "DATASET_ID",
+                "Value.$": "$.dataset_id"
+              },
+              {
+                "Name": "COLLECTION_ID",
+                "Value.$": "$.collection_id"
+              },
+              {
+                "Name": "STEP_NAME",
+                "Value": "download"
```
danieljhegeman, line 69: Asked this in the other PR, but just checking to make sure this is correct. `"Value"` vs `"Value.$"`.

backend/layers/processing/downloader.py
danieljhegeman, line 1: Since this will comprise the single `download` function, which is used in only one place, let's simply consolidate everything in `process_download.py`.

backend/layers/processing/process.py
```diff
@@ -82,6 +81,7 @@ def log_batch_environment(self):
             "MAX_ATTEMPTS",
             "MIGRATE",
             "REMOTE_DEV_PREFIX",
+            "TASK_TOKEN",
```
danieljhegeman, line 84: This is not included in the sfn definition yet

backend/layers/processing/process_download.py
```diff
@@ -0,0 +1,108 @@
...
+            raise ValueError(f"Download for URI scheme '{file_url.scheme}' not implemented")
+        return local_path
+
+    # TODO: after upgrading to Python 3.9, replace this with removeprefix()
+    @staticmethod
+    def remove_prefix(string: str, prefix: str) -> str:
+        if string.startswith(prefix):
+            return string[len(prefix) :]
+        else:
+            return string[:]
```
danieljhegeman, line 76:
```suggestion
            return string
```

.happy/terraform/modules/sfn/main.tf
Bento007: download is not a json object so `$`'s are not needed

backend/layers/processing/process.py
Bento007: Good catch.

danieljhegeman (changes requested):
Need to consolidate the download/process_download files into one. A few minor comments.

backend/layers/processing/process_download.py
```diff
@@ -0,0 +1,137 @@
...
+from backend.common.utils.corpora_constants import CorporaConstants
+from backend.layers.business.business_interface import BusinessLogicInterface
+from backend.layers.common.entities import (
+    DatasetArtifactType,
+    DatasetProcessingStatus,
+    DatasetStatusKey,
+    DatasetUploadStatus,
+    DatasetVersionId,
+)
+from backend.layers.processing.downloader import logger
```
danieljhegeman, line 15: this file no longer exists

.happy/terraform/modules/sfn/main.tf
danieljhegeman: LOL great point

danieljhegeman (changes requested):
import errors after removing `downloader.py`

tests/unit/processing/test_process_download.py
danieljhegeman, line 1: I would add one simple error test like the last one, but for `http` scheme.

danieljhegeman (changes requested):
one last simple test request

tests/unit/processing/test_process_download.py
```diff
@@ -57,6 +57,18 @@ def test_download_from_s3_uri(self):
...
+        Call process download using a dropbox uri
+        """
+
+        dropbox_uri = "https://www.dropbox.com/s/fake_location/test.h5ad?dl=1"
+        pdv = ProcessDownload(self.business_logic, self.uri_provider, self.s3_provider)
+        pdv.download = Mock()
+
+        assert pdv.download_from_source_uri(dropbox_uri, "fake_local_path") == "fake_local_path"
+        pdv.download.assert_called_once_with(dropbox_uri, "fake_local_path")
+
```
danieljhegeman, line 71: Perhaps you added this test intentionally, separate from my comment on this file, but what I had meant to request was a test that checks for [the error raised on ln 70 of process_download.py](https://github.com/chanzuckerberg/single-cell-data-portal/blob/bbebe52ef2a369662f54eab917ecb2b8a602c7bb/backend/layers/processing/process_download.py#L70), which is why I suggested `http` since it is not an accepted URI scheme for our ingestion pipeline

danieljhegeman (changes requested):
see comment(s) on test_process_download.py

tests/unit/processing/test_process_download.py
Bento007: That line isn't actually reachable. Everything that fails will fail at line 49. I plan to remove most of this logic in a refactor anyways.

Bento007: the refactor https://github.com/chanzuckerberg/single-cell-data-portal/pull/6129

backend/layers/processing/process_download.py
```diff
@@ -0,0 +1,134 @@
...
+from backend.layers.processing.exceptions import UploadFailed
+from backend.layers.processing.logger import logit
+from backend.layers.processing.process_logic import ProcessingLogic
+from backend.layers.thirdparty.s3_provider_interface import S3ProviderInterface
+from backend.layers.thirdparty.uri_provider import UriProviderInterface
+
+
+class ProcessDownload(ProcessingLogic):
+    """
+    Base class for handling the `Download and Validate` step of the step function.
```
nayib-jose-gloria, line 24: nit: just 'Download' step now

tests/unit/processing/test_processing.py
nayib-jose-gloria, line 1: rename to test_process_validate for clarity when adding future tests?

danieljhegeman (approved):

PR #6125, opened 2023-10-31
feat(schema-migration): in report step, return list of changes logged by migrate command in cellxgene-schema

nayib-jose-gloria:
## Reason for Change

- in addition to errors, curators would like the migration report to also list any tracked changes to datasets due to the migration script. The migrate command in cellxgene-schema (single-cell-curation) will be updated to return a list of strings with changes to report, if any, for each given dataset. For schema 4, as an example, the migrate script will report any` *_colors` keys in the uns that must be deleted to be validated against new `*_colors` validation rules. 

## Changes

- If any reported_changes are returned by the migrate command when running in dataset_migrate, store them under a`report/migrate_changes` subdirectory in the artifact bucket directory for this migration execution 
- In turn, store errors in a `report/errors` subdirectory rather than in `report` to distinguish the two portions of the migration report 
- small refactor for DRYness when retrieving these two sets of files when building the migration report in the report step. 
- renamed `_store_sfn_response` and `_retrieve_sfn_response` parameter from `step_name` to `directory` in response to using subdirectories instead of just step_name when storing step function outputs
- added an unrelated fix to `gather_collections` step of migrations; when `auto_publish` is set to `false` we were accidentally skipping code block that filtered out published collections with active revisions from the `gather_collections` output. Refactored that block to always go through that logic, and then check `auto_publish`.
- fix pytest fixture that resulted in unit tests hiding the bug above

## Testing steps

- Unit Tests
- rdev test


backend/layers/processing/schema_migration.py
```diff
@@ -277,32 +284,37 @@ def wrapper(*args, **kwargs):
...
                 )
-            )
-            self.logger.info("Error files found", extra={"error_files": len(error_s3_keys)})
-            for s3_key in error_s3_keys:
-                local_file = os.path.join(self.local_path, "data.json")
-                self.s3_provider.download_file(self.artifact_bucket, s3_key, local_file)
-                with open(local_file, "r") as f:
-                    jn = json.load(f)
-                report["errors"].append(jn)
+                self.logger.info(f"{directory} files found", extra={"files": len(s3_keys)})
```
Bento007, line 304: nit: This seems like a log message that should go in s3_provider

backend/layers/processing/schema_migration.py
```diff
@@ -277,32 +285,37 @@ def wrapper(*args, **kwargs):
...
                 )
-            )
-            self.logger.info("Error files found", extra={"error_files": len(error_s3_keys)})
-            for s3_key in error_s3_keys:
-                local_file = os.path.join(self.local_path, "data.json")
-                self.s3_provider.download_file(self.artifact_bucket, s3_key, local_file)
-                with open(local_file, "r") as f:
-                    jn = json.load(f)
-                report["errors"].append(jn)
+                self.logger.info(f"{len(s3_keys)} {directory} files found")
```
Bento007, line 305: I think the variable name `directory` is confusing since it's not a full directory and only the suffix.  Maybe `message_type` or something similar.

nit: 
```suggestion
                self.logger.info("Files Found", extra={"directory": directory, "count": len(s3_keys)}")
```

Bento007 (approved):

Issue #6128, opened 2023-10-31
feat(nav-bar): update navbar to new layout that accounts for census projects page


seve:
https://www.figma.com/file/DMoBg0zsvktKrOZ3akAnFc/Census?type=design&node-id=774-21604&mode=dev

![image](https://github.com/chanzuckerberg/single-cell-data-portal/assets/8716829/1e058c60-0236-4a53-9db3-3b68cea1f5b2)

PR #6137, opened 2023-11-01
feat: update navbar to match new design

seve:
#6128 
image

### DIFF
--- a/frontend/src/components/Header/components/Nav/index.tsx
+++ b/frontend/src/components/Header/components/Nav/index.tsx
@@ -6,7 +6,13 @@ import { track } from "src/common/analytics";
 import { EVENTS } from "src/common/analytics/events";
 import NavDivider from "src/components/Header/components/Nav/components/NavDivider";
 import { isRouteActive } from "src/components/Header";
-import { LinkWrapper, Nav as NavWrapper } from "./style";
+import {
+  LinkWrapper,
+  NavItemContainer,
+  NavSection,
+  NavSectionTitle,
+  Nav as NavWrapper,
+} from "./style";
 import { BetaChip } from "../../style";
 
 export const CENSUS_LINK = "https://chanzuckerberg.github.io/cellxgene-census/";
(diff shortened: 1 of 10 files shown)


hthomas-czi (commented):
Looks good! I have a couple requests:

- Increase the `font-weight` for the tabs from `medium` to `semibold` (`600`) to better differentiate between the new section labels and the tabs.
- The overall height of the nav bar should increase from `48px` to `56px` to maintain some vertical padding while accounting for the new "Applications" and "Census" labels.
- Increase the logo size to `28px` to maintain the same proportionality with the increased nav bar height.

hthomas-czi:
@seve A couple other general comments while I'm waiting for rdev to refresh:

- The main content div should have `box-sizing: content-box` and `padding: 0px 40px` in addition to the `margin: 80px auto`, so the content doesn't hit the edge when the viewport is smaller
- The `font-weight` for the Embedding and Mode buttons should be `medium` (`500`) not `semibold`

frontend/src/components/Header/components/Nav/index.tsx
```diff
@@ -6,7 +6,13 @@ import { track } from "src/common/analytics";
...
+import {
+  LinkWrapper,
+  NavItemContainer,
+  NavSection,
+  NavSectionTitle,
+  Nav as NavWrapper,
+} from "./style";
 import { BetaChip } from "../../style";
 
 export const CENSUS_LINK = "https://chanzuckerberg.github.io/cellxgene-census/";
```
kaloster, line 18: can we please move this into a constants file?

frontend/src/components/Header/components/Nav/index.tsx
```diff
@@ -17,7 +23,8 @@ interface Props {
 }
```
kaloster, line 24: and this to a types file? thank you!

kaloster (commented):
Solid work! Just a couple of comments. Thanks!

kaloster (approved):
LGTM! 🚀

Issue #5881, opened 2023-10-03
Implement Metadata-Update Batch Job


nayib-jose-gloria:
Full [context](https://docs.google.com/document/d/1wXP4BVD6Js-Lj9T8nwzkBA931lutUbTLu56pNpSatwc/edit) proposal

add as branching step in process file in single-cell-data-portal, same as the other steps for single dataset processing (i.e. downloadvalidate)

Inputs: dataset_version_id, metadata_key, new_value

1. Fetch current DatasetVersion from DB using dataset_version_id
2. Fetch h5ad + rds artifact s3 URIs from DB for current DatasetVersion
3. Create new DatasetVersion for canonical Dataset that is a duplicate of the current DatasetVersion
4. Set DatasetVersion processing status to PENDING and upload status to UPLOADING
5. For this CollectionVersion, replace its mapping to the current DatasetVersion with this new DatasetVersion.
6. Update metadata_key to new_value in 'uns' metadata dict for h5ads and rds files (load in 'backed' mode if possible for minimal memory usage)
           - TODO: Look into Python libraries to easily handle calls to R to handle rds update (from Emanuele: look into rpy2)
7. Store updated artifacts in S3 under s3://corpora-data-<env>/<new_dataset_version_id>/ (same pattern for regular dataset artifact processing)
8. Create new entries in DatasetArtifact DB table for updated artifacts
9. In the new DatasetVersion DB object, update:
         - dataset_metadata JSON field to also map metadata_key to new_value
         - artifacts field to list of UUIDs mapping to the new entries in DatasetArtifact DB
10. status back to processing status SUCCESS and upload status to UPLOADED

nayib-jose-gloria:
Note: Reach out to R expert on TileDB team (Dirk)

Bento007:
For ease of review I would break it into 4 PRs:
- h5ad updates.
- cxg updates.
- seurat updates.
- batch job.

PR #6144, opened 2023-11-02
feat: code for 'light reprocessing' batch job to only update metadata of artifacts without full processing/conversion

nayib-jose-gloria:
## Reason for Change

- #5881

## Changes

- create light reprocessing batch job class, inheriting from ProcessDownload, and updating just the metadata dict of our artifacts while creating a new DatasetVersion (and updating dataset status in the DB as normal)
- go through steps to reprocess h5ad and DatasetMetadata while only doing backed anndata reads and writes to the uns
- use rpy2 to load existing rds, update only the metadata dict (called Misc) in SeuratObject, and save back to rds as an artifact of the new Dataset Version
- Copy previous cxg dir to new s3 location using new DatasetVersionID, then update relevant metadata keys and DB
- accepts JSON string, can update multiple metadata keys in one job
- new batch job definition in terraform
- defined DatasetArtifactMetadataUpdate entity that defines artifact metadata we will allow to be updated with this batch job (included 3 use-cases we are either actively planning on implementing or have issues written for implementing in the future)

## Testing steps

- Unit Tests + rdev runs

## Notes for Reviewer


backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,149 @@
...
+        # TODO: factor out the logic we want from ingest dataset from the URI check, so we don't need to pass a uri
+        new_dataset_version_id, dataset_id = self.business_logic.ingest_dataset(
+            collection_version_id, raw_h5ad_s3_uri, None, dataset_version_id, start_step_function=False
+        )
+        self.process(new_dataset_version_id, raw_h5ad_s3_uri, self.artifact_bucket)
+
+        key_prefix = self.get_key_prefix(new_dataset_version_id)
+        # update anndata
+        # TODO: factor out any shared logic
+        if h5ad_s3_uri:
```
Bento007, line 70: I would turn each of these ifs into a new function to make it easier to test. There will be less things to mock.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,190 @@
...
+
+        if DatasetArtifactType.RDS in artifact_uris:
+            self.update_rds(
+                artifact_uris[DatasetArtifactType.RDS], key_prefix, new_dataset_version_id, metadata_update_dict
+            )
+
+        if DatasetArtifactType.CXG in artifact_uris:
+            self.update_cxg(artifact_uris[DatasetArtifactType.CXG], key_prefix, new_dataset_version_id)
+
+        self.update_processing_status(new_dataset_version_id, DatasetStatusKey.H5AD, DatasetConversionStatus.CONVERTED)
```
Bento007, line 166: why isn't this in update_h5ad

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,190 @@
...
+        new_dataset_version_id: DatasetVersionId,
+        metadata_update_dict: Dict[str, str],
+    ):
+        seurat_filename = self.download_from_source_uri(
+            source_uri=rds_s3_uri,
+            local_path=CorporaConstants.LABELED_H5AD_ARTIFACT_FILENAME,
+        )
+        self.update_processing_status(new_dataset_version_id, DatasetStatusKey.RDS, DatasetConversionStatus.CONVERTING)
+
+        base = importr("base")
```
Bento007, line 90: this looks promising

backend/layers/processing/dataset_metadata_update.py
nayib-jose-gloria: I had interpreted this as 'we've now converted H5AD to the other two formats' but I code-searched and you're right that we use this to mean we've finished labeling the H5AD. will fix

danieljhegeman (commented):
@nayib-jose-gloria need to also copy the h5ad and rds (if exists) to the public artifacts bucket!

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,220 @@
...
+    def update_cxg(
+        self,
+        cxg_s3_uri: str,
+        key_prefix: str,
+        dataset_version_id: DatasetVersionId,
+        metadata_update_dict: Dict[str, str],
+    ):
+        new_cxg_dir = f"s3://{self.cellxgene_bucket}/{key_prefix}.cxg/"
+        self.s3_provider.upload_directory(cxg_s3_uri, new_cxg_dir)
+        ctx = tiledb.Ctx(
```
nayib-jose-gloria, line 126: I think I'll factor this out somewhere or at least add a comment. I just copied this from the cxg conversion code

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,220 @@
...
+        key_prefix: str,
+        new_dataset_version_id: DatasetVersionId,
+        metadata_update_dict: Dict[str, str],
+    ):
+        h5ad_filename = self.download_from_source_uri(
+            source_uri=h5ad_s3_uri,
+            local_path=CorporaConstants.LABELED_H5AD_ARTIFACT_FILENAME,
+        )
+
+        adata = scanpy.read_h5ad(h5ad_filename)
```
nayib-jose-gloria, line 58: Using backed="r+" to read the anndata should work in theory to be able to modify metadata without reading into memory. Sadly, there is a known [anndata bug](https://github.com/scverse/anndata/issues/446) that makes this not possible, so we will have to read into memory after alla. Knowing this, @Bento007, what changes should be made to the infra to support all datasets?

Bento007: The resource allocated should be enough already.

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -1765,7 +1765,7 @@ def test__get_all_datasets_for_user_index(self):
                 # append the dataset_id to the list of dataset ids
                 dataset_ids.append(dataset["dataset_id"])
 
-        self.assertEqual(len(dataset_ids), 6)
+        self.assertEqual(len(dataset_ids), 8)
```
danieljhegeman, line 1768: what's going on here? 🔍 🤔

tests/unit/backend/layers/common/base_test.py
```diff
@@ -283,6 +283,7 @@ def generate_dataset(
             )
         if artifacts is None:
             artifacts = [
+                DatasetArtifactUpdate(DatasetArtifactType.RAW_H5AD, f"s3://fake.bucket/{dataset_version_id}/raw.h5ad"),
```
danieljhegeman, line 286: Evidently we weren't testing for the actual artifacts present in the fake bucket anywhere before, were we... 😏

tests/unit/processing/test_dataset_metadata_update.py
```diff
@@ -0,0 +1,356 @@
...
+    @patch("scanpy.read_h5ad")
+    def test_update_h5ad(self, mock_read_h5ad, *args):
+        collection_version = self.generate_unpublished_collection(add_datasets=1)
+        original_dataset_version = collection_version.datasets[0]
+        new_dataset_version_id, _ = self.business_logic.ingest_dataset(
+            collection_version_id=collection_version.version_id,
+            url=None,
+            file_size=0,
+            existing_dataset_version_id=original_dataset_version.version_id,
+            start_step_function=False,
```
danieljhegeman, line 56: add this parameter to corresponding stub function in BusinessLogicInterface

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,233 @@
...
+            source_uri=h5ad_uri,
+            local_path=CorporaConstants.LABELED_H5AD_ARTIFACT_FILENAME,
+        )
+
+        adata = scanpy.read_h5ad(h5ad_filename)
+        metadata = original_dataset_version.metadata
+        for key, val in metadata_update_dict.items():
+            adata.uns[key] = val
+            if hasattr(metadata, key):
+                setattr(metadata, key, val)
```
danieljhegeman, line 66: I haven't looked closely before: we're sure we have exactly matching str names for all keys for metadata in 1) the anndata uns dict, and 2) the metadata json in the corresponding db row? I would assume so but just want to make sure.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,233 @@
...
+        )
+        self.update_processing_status(
+            new_dataset_version_id, DatasetStatusKey.VALIDATION, DatasetValidationStatus.VALID
+        )
+        self.update_processing_status(new_dataset_version_id, DatasetStatusKey.H5AD, DatasetConversionStatus.CONVERTED)
+
+    def update_rds(
+        self,
+        rds_uri: str,
+        key_prefix: str,
```
danieljhegeman, line 87: I think the `new` vs `original` approach you've taken for some of the var names in this PR is a good one. I'd say apply it broadly: I'd rename this to `new_key_prefix`.

Also consider `old` instead of `original`. Though tbh I'm not sure I care too much between those two.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,233 @@
...
+
+        if DatasetArtifactType.RDS in artifact_uris:
+            self.update_rds(
+                artifact_uris[DatasetArtifactType.RDS], key_prefix, new_dataset_version_id, metadata_update_dict
+            )
+        elif original_dataset_version.status.rds_status == DatasetConversionStatus.SKIPPED:
+            self.update_processing_status(new_dataset_version_id, DatasetStatusKey.RDS, DatasetConversionStatus.SKIPPED)
+        else:
+            self.logger.error(
+                f"Cannot find RDS artifact uri for {dataset_version_id}, " f"and Conversion Status is not SKIPPED."
```
danieljhegeman, line 195:
```suggestion
                f"Cannot find RDS artifact uri for {dataset_version_id}, and Conversion Status is not SKIPPED."
```

danieljhegeman: That's a new one for me 😂

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,233 @@
...
+            {
+                "sm.consolidation.buffer_size": consolidation_buffer_size(0.1),
+                "py.deduplicate": True,  # May reduce memory requirements at cost of performance
+            }
+        )
+        array_name = f"{new_cxg_dir}/cxg_group_metadata"
+        with tiledb.open(array_name, mode="r", ctx=ctx) as metadata_array:
+            cxg_metadata_dict = json.loads(metadata_array.meta["corpora"])
+            for key, value in metadata_update_dict.items():
+                cxg_metadata_dict[key] = value
```
danieljhegeman, line 136:
```suggestion
            cxg_metadata_dict.update(metadata_update_dict)
```

danieljhegeman (commented):
Haven't covered tests yet but generally code looks good. Some minor nits. Also consider [this suggestion PR for consolidating shared tiledb ctx config in one place](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6233/files) rather than copying. We don't want edits only being applied in one spot.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,233 @@
...
+        else:
+            self.logger.error(f"Cannot find raw H5AD artifact uri for {dataset_version_id}.")
+            raise ValueError
+
+        new_dataset_version_id, dataset_id = self.business_logic.ingest_dataset(
+            collection_version_id=collection_version_id,
+            url=raw_h5ad_uri,
+            file_size=0,
+            existing_dataset_version_id=dataset_version_id,
+            start_step_function=False,
```
danieljhegeman, line 168: add this parameter to corresponding stub function in BusinessLogicInterface

danieljhegeman (commented):
@nayib-jose-gloria what is the higher-level plan for actually applying this update sfn to a Dataset? Is a revision going to be instantiated for each Collection first?

backend/layers/processing/dataset_metadata_update.py
Bento007: Now that we are no longer using swap, the dowload step should run before this to spit out a memory estimation for this step.

backend/layers/processing/dataset_metadata_update.py
Bento007: If we want to be extra certain we can make a new dataclass for the metadata.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def update_h5ad(
+        self,
+        h5ad_uri: str,
+        original_dataset_version: DatasetVersion,
+        key_prefix: str,
+        new_dataset_version_id: DatasetVersionId,
+        metadata_update_dict: Dict[str, str],
+    ):
+        h5ad_filename = self.download_from_source_uri(
```
Bento007, line 56: Downloading and uploading the updated dataset uses very little CPU and memory. We should run these steps in a separate thread to make the most out of the machines.

backend/layers/processing/dataset_metadata_update.py
Bento007: old/new, previous/current

Bento007 (commented):
Since we need to read the h5ad into memory, compress, and save it's going to be slow and memory hungry. As for seurat and cxg, are those resource hungry? I wondering if we could run these all in parallel.

nayib-jose-gloria:
> @nayib-jose-gloria what is the higher-level plan for actually applying this update sfn to a Dataset? Is a revision going to be instantiated for each Collection first?

This batch job assumes the collection is in a non-published, editable state (private or revision). I can add an explicit check for that if one of the nested calls doesn't already (I think ingest_dataset might?)

For the specific use-case of auto-updating citations, a batch job will be triggered per dataset in response to a DOI update in the collection metadata. In that case, the user must've already had a revision open in order to write a DOI update.

tests/unit/backend/layers/api/test_portal_api.py
nayib-jose-gloria: This test counts how many dataset assets appear when fetching dataset metadata. Previously, our mock did not include RAW_H5AD. Now it does, which reflects the state of real datasets

backend/layers/processing/dataset_metadata_update.py
nayib-jose-gloria: Great point. There is one diff--'title' in the schema is 'name' in the DB metadata. 

We aren't implementing that use-case yet, but I will make a change to account for these potential discrepancies. Probably by making a new dataclass as Trent suggested

danieljhegeman (changes requested):
need to use `multiprocessing` rather than `threading`

backend/layers/processing/dataset_metadata_update.py
Bento007: I'd also recommend using a Processing pool to avoid running over resources.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,284 @@
...
+            raise ValueError
+
+        self.process(new_dataset_version_id, raw_h5ad_uri, self.artifact_bucket)
+
+        new_artifact_key_prefix = self.get_key_prefix(new_dataset_version_id.id)
+
+        with multiprocessing.Pool() as pool:
+            if DatasetArtifactType.H5AD in artifact_uris:
+                self.logger.info("Main: Starting thread for h5ad update")
+                pool.starmap_async(
```
Bento007, line 172: I think `apply_async` over `starmap_async` would be better suited , since there is not iterating over multiple jobs and doesn't block

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,284 @@
...
+        for key, val in metadata_update.as_dict_without_none_values().items():
+            adata.uns[key] = val
+
+            db_field = ARTIFACT_TO_DB_FIELD.get(key) if key in ARTIFACT_TO_DB_FIELD else key
+            setattr(metadata, db_field, val)
+
+        adata.write(h5ad_filename, compression="gzip")
+        self.business_logic.set_dataset_metadata(new_dataset_version_id, metadata)
+
+        self.create_artifact(
```
Bento007, line 76: After we upload the artifacts we should delete the files locally downloaded and generated. This is to avoid running out of disk space on the EC2.

backend/layers/processing/dataset_metadata_update.py
danieljhegeman: This is true but also the results are not being awaited anywhere; statuses will be marked as success/failure before the subprocesses complete. [My fix for it](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6274).

danieljhegeman (changes requested):
[fix for handling async objects for artifact subprocesses](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6274)

tests/unit/processing/test_dataset_metadata_update.py
```diff
@@ -0,0 +1,526 @@
...
+def mock_apply_async(func, args=()):
+    return Mock(return_value=func(*args))
+
+
+mock_enter = Mock(apply_async=Mock(side_effect=mock_apply_async))
+mock_pool = Mock(__enter__=Mock(return_value=mock_enter), __exit__=Mock())
+mock_pool_constructor = Mock(return_value=mock_pool)
+
+
+@patch("backend.layers.processing.dataset_metadata_update.multiprocessing.Pool", mock_pool_constructor)
```
Bento007, line 45: If only some of the tests needs this mock, pull those tests out into their own test class. I don't think update_h5ad, update_cxg, or update_seurat need it.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -82,6 +82,7 @@ def update_h5ad(
             DatasetStatusKey.H5AD,
             datasets_bucket=self.datasets_bucket,
         )
+        os.remove(h5ad_filename)
```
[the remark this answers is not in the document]
Bento007: I see, you're keeping the file around so you verify the change.

Bento007 (changes requested):
Please apply the test change suggestions

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+            self.logger.error(f"Cannot find raw H5AD artifact uri for {current_dataset_version_id}.")
+            raise ValueError
+
+        self.process(new_dataset_version_id, raw_h5ad_uri, self.artifact_bucket)
+
+        new_artifact_key_prefix = self.get_key_prefix(new_dataset_version_id.id)
+
+        artifact_jobs = []
+
+        with multiprocessing.Pool() as pool:
```
Bento007, line 182: I would process each file in it's own directory. I'm concern there is a change we try uploading the h5ad or seurat file to the CXG directory when we run upload_directory.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+        self.create_artifact(
+            seurat_filename,
+            DatasetArtifactType.RDS,
+            new_key_prefix,
+            new_dataset_version_id,
+            self.artifact_bucket,
+            DatasetStatusKey.RDS,
+            datasets_bucket=self.datasets_bucket,
+        )
+        os.remove(seurat_filename)
```
danieljhegeman, line 123: I'm gathering that this helps in scenarios where multiple datasets are being processed at the same time on the same machine...? 🤔

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+            )
+            return
+
+        if DatasetArtifactType.RAW_H5AD in artifact_uris:
+            raw_h5ad_uri = artifact_uris[DatasetArtifactType.RAW_H5AD]
+        else:
+            self.logger.error(f"Cannot find raw H5AD artifact uri for {current_dataset_version_id}.")
+            raise ValueError
+
+        self.process(new_dataset_version_id, raw_h5ad_uri, self.artifact_bucket)
```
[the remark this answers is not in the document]
danieljhegeman: Generally speaking I think we should consider renaming some of our various `process`-named functions =)

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+                artifact_jobs.append(
+                    pool.apply_async(
+                        self.update_h5ad,
+                        (
+                            artifact_uris[DatasetArtifactType.H5AD],
+                            current_dataset_version,
+                            new_artifact_key_prefix,
+                            new_dataset_version_id,
+                            metadata_update,
+                        ),
```
danieljhegeman, line 194:
```suggestion
                        )
```

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+                self.logger.info("Main: Starting thread for rds update")
+                artifact_jobs.append(
+                    pool.apply_async(
+                        self.update_rds,
+                        (
+                            artifact_uris[DatasetArtifactType.RDS],
+                            new_artifact_key_prefix,
+                            new_dataset_version_id,
+                            metadata_update,
+                        ),
```
danieljhegeman, line 213:
```suggestion
                        )
```

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+                self.logger.info("Main: Starting thread for cxg update")
+                artifact_jobs.append(
+                    pool.apply_async(
+                        self.update_cxg,
+                        (
+                            artifact_uris[DatasetArtifactType.CXG],
+                            f"s3://{self.cellxgene_bucket}/{new_artifact_key_prefix}.cxg",
+                            new_dataset_version_id,
+                            metadata_update,
+                        ),
```
danieljhegeman, line 238:
```suggestion
                        )
```

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,288 @@
...
+                    )
+                )
+            else:
+                self.logger.error(f"Cannot find cxg artifact uri for {current_dataset_version_id}.")
+                self.update_processing_status(
+                    new_dataset_version_id, DatasetStatusKey.CXG, DatasetConversionStatus.FAILED
+                )
+
+        # blocking call on async jobs before checking for valid artifact statuses
+        [j.get() for j in artifact_jobs]
```
danieljhegeman, line 248: I think you should include a comment about why we want to `.get()` and not `.wait()` here. Subtle distinction that is important which could get errantly changed in code gets shuffled around in the future.

backend/layers/processing/dataset_metadata_update.py
danieljhegeman: Understand the concern but this isn't going to be an issue, I don't think. To my eye, the cxg components are all copied s3-to-s3, and then the metadata update is actually applied via tiledb directly (perhaps with local in-mem intermediary under the hood to some degree but that is irrelevant) to the appropriate file in s3. @nayib-jose-gloria can confirm

nayib-jose-gloria: +1 to Dan's comment

tests/unit/processing/test_dataset_metadata_update.py
nayib-jose-gloria: I am tempted to keep it given that a majority of the tests need it, and it doesn't negatively impact those that don't. But I can be swayed if you think there is a downside

nayib-jose-gloria: Oh nvm I misread your comment--yeah I can put those in a separate test class

backend/layers/processing/dataset_metadata_update.py
nayib-jose-gloria: It was an OOM safeguard suggestion from @Bento007

danieljhegeman: OOM? It's removing a file from _disk_...

danieljhegeman: I mean I don't think it's going to do anything fundamentally bad but it seems unnecessary and may just cause the entire operation to take a few seconds longer than otherwise required 🤷‍♂️ 

Unless we have reason to believe we'll have disk space issues. I assume no. Anyways nbd either way

nayib-jose-gloria: Sorry, OOM on the brain. It was a suggestion to save on disk space. I don't know how likely it is that we would have a disk space problem, though I agreed it wouldn't hurt

backend/layers/processing/process_download.py
```diff
@@ -187,5 +187,6 @@ def process(
         )
         self.update_processing_status(dataset_version_id, DatasetStatusKey.UPLOAD, DatasetUploadStatus.UPLOADED)
 
-        sfn_client = StepFunctionProvider().client
-        sfn_client.send_task_success(taskToken=sfn_task_token, output=json.dumps(response))
+        if sfn_task_token:
```
danieljhegeman, line 190: for unit test? The only thing is this could cause a silent failure in deployed scenario. Patch instead?

nayib-jose-gloria: this job isn't part of a step function, so there's nothing to send task success to nor a need to init a StepFunctionClient. But I wanted to re-use the rest of the `process_download.process` logic

I can extract everything above this line into a separate function and call that instead, though

danieljhegeman: Ah, right. I would just add a comment explaining that for lesser minds like mine 😅

backend/layers/processing/dataset_metadata_update.py
Bento007: We are using swap again. Still no need to change anything

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,305 @@
...
+
+        new_artifact_key_prefix = self.get_key_prefix(new_dataset_version_id.id)
+
+        artifact_jobs = []
+
+        with multiprocessing.Pool(processes=3) as pool:
+            if DatasetArtifactType.H5AD in artifact_uris:
+                self.logger.info("Main: Starting thread for h5ad update")
+                artifact_jobs.append(
+                    pool.apply_async(
```
danieljhegeman, line 202: Get rid of the pool and just use `Process` and `.start()` it.

Replace `.get()` below with `.join()`.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -173,79 +232,76 @@ def update_metadata(
...
+            artifact_jobs.append(h5ad_job)
+            h5ad_job.start()
+        else:
+            self.logger.error(f"Cannot find labeled H5AD artifact uri for {current_dataset_version_id}.")
+            self.update_processing_status(new_dataset_version_id, DatasetStatusKey.H5AD, DatasetConversionStatus.FAILED)
+
+        if DatasetArtifactType.RDS in artifact_uris:
+            self.logger.info("Main: Starting thread for rds update")
+            rds_job = Process(
+                target=DatasetMetadataUpdater.update_rds,
```
Bento007, line 264: clever how you got this to work. It's confusing, but it works.

tests/unit/processing/test_dataset_metadata_update.py
```diff
@@ -55,212 +54,14 @@ def mock_download(source_uri, local_path):
...
-        assert self.updater.has_valid_artifact_statuses(dataset_version_id) is False
-
-    @patch("backend.common.utils.dl_sources.uri.downloader")
-    @patch("scanpy.read_h5ad")
-    @patch("backend.layers.processing.dataset_metadata_update.DatasetMetadataUpdater.update_h5ad")
-    @patch("backend.layers.processing.dataset_metadata_update.DatasetMetadataUpdater.update_rds")
-    @patch("backend.layers.processing.dataset_metadata_update.DatasetMetadataUpdater.update_cxg")
+    @patch("backend.layers.processing.dataset_metadata_update.S3Provider", Mock(side_effect=MockS3Provider))
+    @patch("backend.layers.processing.dataset_metadata_update.DatabaseProvider", Mock(side_effect=DatabaseProviderMock))
+    @patch("backend.layers.processing.dataset_metadata_update.DatasetMetadataUpdaterWorker.update_h5ad")
```
Bento007, line 62: Why not mock the class `DatasetMetadataUpdaterWorker` and check if the functions are called? Then you don't you don't need to mock each one individually

Bento007 (commented):
you can cutdown on the amount of patching if you mock the class rather than class members

danieljhegeman (changes requested):
Good that the multiprocessing is working but we need to check exit codes for subtasks.

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -0,0 +1,344 @@
...
+                ),
+            )
+            artifact_jobs.append(cxg_job)
+            cxg_job.start()
+        else:
+            self.logger.error(f"Cannot find cxg artifact uri for {current_dataset_version_id}.")
+            self.update_processing_status(new_dataset_version_id, DatasetStatusKey.CXG, DatasetConversionStatus.FAILED)
+
+        # blocking call on async functions before checking for valid artifact statuses
+        [j.join() for j in artifact_jobs]
```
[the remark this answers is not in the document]
danieljhegeman: Actually, I'm wrong. After reading the logic more closely, it is evident that the ConversionStatus is being set at the end of every subprocess worker job, and this is then interpreted by the main thread which will appropriately update the processing status as a failure. We do not need additional checks.

But, since this is a step function and we'll want to be able to quickly find the failed runs in the console, we should probably throw an error [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/88a8eec8926ea32af5d333a8ab0b41165bec6a88/backend/layers/processing/dataset_metadata_update.py#L311-L313).

Bento007 (approved):

danieljhegeman (approved):
gtg 👍

PR #6150, opened 2023-11-03
feat(projects): dynamically pull project data from static file and mock API

seve:
- #6149
- #6151 

## Changes

- add
   - static files (`census-projects.json`, `census-notebook-links.json`)
   - query to backend
   - pull out project rendering code into modular function
   - query to crossref API


frontend/src/common/queries/censusDirectory.ts
```diff
@@ -0,0 +1,117 @@
...
+  // const response = await fetch(url, {
+  //   ...DEFAULT_FETCH_OPTIONS,
+  //   ...JSON_BODY_FETCH_OPTIONS,
+  // });
+
+  // const result = await response.json();
+  // if (!response.ok) throw result;
+
+  // return result.data;
+  const data: ProjectResponse = {
```
seve, line 42: mocked data for now, will be removed during integration

frontend/src/common/queries/censusDirectory.ts
```diff
@@ -0,0 +1,117 @@
...
+  Object.entries(data).forEach(
+    async ([id, project]: [
+      keyof ProjectResponse,
+      ProjectResponse[keyof ProjectResponse],
+    ]) => {
+      if (!project.DOI) return;
+
+      // include a mailto: query param to insure reliable service
+      const url = apiTemplateToUrl(
+        "https://api.crossref.org/works/{DOI}?mailto=cellxgene@cziscience.com",
```
seve, line 86: crossref request will be cached alongside our other queries

frontend/src/common/API.ts
```diff
@@ -20,4 +20,5 @@ export enum API {
   WMG_FILTERS_QUERY = "/wmg/v1/filters",
   WMG_MARKER_GENES = "/wmg/v2/markers",
   WMG_GENE_INFO = "/gene_info/v1/gene_info",
+  CENSUS_DIRECTORY = "/census/v1/directory", // TODO: replace with actual endpoint
```
seve, line 23: Final endpoint won't live in this file, similar to cell guide API,

frontend/src/common/queries/censusDirectory.ts
bkmartinjr: just wanted to confirm you saw the "polite tips" on crossref, which supposedly will help avoid blocking in some cases.

https://www.crossref.org/documentation/retrieve-metadata/rest-api/tips-for-using-the-crossref-rest-api/

brianraymor: I was under the impression that we have a crossref API key that was inherited from Meta. CC: @metakuni 

https://github.com/chanzuckerberg/single-cell-data-portal/blob/bc584a92c7585b511e70dc8822a7836b9a54a0b0/backend/common/providers/crossref_provider.py#L34

metakuni: Correct, we have access to [Metadata Plus](https://www.crossref.org/services/metadata-retrieval/metadata-plus/). The credentials can be found in the team vault in 1Password.

seve: Hmm we might want to change some code over in cell guide too @atarashansky

seve: Since these requests are from the browser, I don't believe we should be using the API key, correct?

seve: Since these requests are from the browser, I don't believe we should be using the API key, correct?

brianraymor: I don't know what you're trying to do. Is there a spec that explains the use of crossref in this context?

seve: No spec.

DOI's are being provided to the client via manifest, we need to convert these DOI's into the format shown[ in the design:](https://www.figma.com/file/DMoBg0zsvktKrOZ3akAnFc/Census?type=design&node-id=774-21604&mode=dev)

image

frontend/src/common/queries/censusDirectory.ts
```diff
@@ -0,0 +1,116 @@
...
+        {
+          DOI: encodeURIComponent(project.DOI),
+        }
+      );
+      const response = await fetch(url);
```
kaloster, line 90: I wonder if we can clean this up a bit by wrapping the request in a try-catch? Something like this will have a more robust error handling:

```
  try {
    const response = await fetch(url);
    const { message: result } = await response.json();
    return await responseHandler(result); // extract the handler to a helper function 
  }
  catch (e) {
    console.log((e as Error).message);
    // Handle the error.
  }
```

frontend/src/views/CensusDirectory/index.tsx
```diff
@@ -21,7 +27,10 @@ import {
   textSecondary,
 } from "src/common/theme";
 
-function DetailItem(props: { label: string; children: string }) {
+import staticProjects, { type StaticProject } from "census-projects.json";
+import notebookLinks from "census-notebook-links.json";
+
+function DetailItem(props: { label: string; children: string; link?: string }) {
```
kaloster, line 33: Possible to extract styles to a separate style file?

kaloster (commented):
Looks good! Just a couple of comments. Thanks!

frontend/src/views/CensusDirectory/index.tsx
seve: this is done in a following PR

kaloster (approved):
Thanks for the quick iteration @seve ! 

LGTM! 🚀

Issue #559, opened 2020-08-27
Remove corpora.cziscience.com domain


Bento007:
# Need
As part of the merging into a single cellxgene team we need to remove the old frontend website and api website.
The sites https://corpora.cziscience.com and https://api.corpora.cziscience.com must both be removed. This will require making changes to single-cell-infra as well as finding a new domain for https://api.corpora.cziscience.com.

# Definition of Done
- https://corpora.cziscience.com is removed
- https://api.corpora.cziscience.com is removed
- https://cellxgene.cziscience.com continues to display the datasets using the data-portal api.

brianraymor:
Per September 24 Sprint Planning, @maniarathi and @mckinsel agreed that we should merge the three remaining issues related to corpora deprecation into one issue and assign an estimate of 1 (rather than three tickets each estimated at 1). Closing.

PR #6164, opened 2023-11-06
fix: use CustomJSONEncoder to ensure JSON serializability of objects

danieljhegeman:
## Reason for Change

- [single-cell #559](https://github.com/chanzuckerberg/single-cell/issues/559)

## Changes

- modify `json.dump` and `json.dumps` calls to use `CustomJSONEncoder` class

## Testing steps

- Unit tests updated to make a side-effect call to test JSON serialization
- Confirmed test fails when the CustomJSONEncoder is not used (i.e., JSON serialization _is_ being tested appropriately)


Bento007 (approved):
Great changes! Thank you for fixing this.

PR #6190, opened 2023-11-08
feat(4.0): adjust memory needs for schema migration and ingestion.

Bento007:
## Reason for Change

- Adjust the memory allocated for schema migration and ingest to run a full migration

## Changes

- add swap_modifier parameter to allow the amount of swap allocated to be modified per batch job
- decouple tests for memory settings from the actual settings used. This allow the tests to be less brittle.
- increase the amount of memory that can be allocated for an ingest job.
- Adjust swap_modifier to 0 to not use swap
- Increase resource for migrating a dataset and stop using swap.
- remove min_vcpu. VCPU is now always base on a calculated value.
- Round the memory to the next highest multiple of `MEMORY_PER_CPU`

## Testing steps

- running schema migration to test.


backend/layers/processing/process_download.py
```diff
@@ -23,14 +23,14 @@
...
 
-MEMORY_MODIFIER = 1.1  # add 10% overhead
-MEMORY_PER_VCPU = 4000
-MIN_MEMORY_MB = 4000
-MIN_VCPU = 1
-# The largest machine we are allocating is r5ad.2xlarge. This machine has 64GB of memory and 16 vCPUs.
-MAX_MEMORY_MB = 64000
-MAX_VCPU = 16
-SWAP_MEMORY_MB = 300000
+MEMORY_MODIFIER = 2  # add % overhead
```
nayib-jose-gloria, line 26: nit: what % overhead? 100?

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -44,15 +44,15 @@ resource aws_batch_job_definition schema_migrations_swap {
     resourceRequirements = [
       {
         type= "VCPU",
-        Value="6"
+        Value="38"
```
nayib-jose-gloria, line 47: don't need to leave a comment, but curious for myself why these numbers? educated estimates based on testing? some calculation?

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -44,15 +44,15 @@ resource aws_batch_job_definition schema_migrations_swap {
...
       },
       {
         Type="MEMORY",
-        Value = "47500"
+        Value = "304000"
```
nayib-jose-gloria, line 51: don't need to leave a comment, but curious for myself why these numbers? educated estimates based on testing? some calculation?

backend/layers/processing/process_download.py
Bento007: in this case it would be 200%. Would "add x% additional memory"

nayib-jose-gloria: got it. Maybe good to just be explicit with 'adds (x*100)% memory overhead'

.happy/terraform/modules/schema_migration/main.tf
Bento007: I chose this number because it's the amount of swap that was previously allocated + `MEMORY` that we previously had. I rounded up to 300000 to add more buffer, but migration worked before with it being less, so i'll change it to 250000. It'll save us some compute.

Bento007, line 51: 256000
```suggestion
        Value = "256000"
```

.happy/terraform/modules/schema_migration/main.tf
Bento007: the 38 is MEMORY/8GB. All of AWS machines in `r5a` family follow this pattern. I don't have a strong reason for following this, but if are going to use a machine to the point where nothing else can run than we mine as well use the whole thing.

```suggestion
        Value="32"
```

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -44,15 +44,15 @@ resource aws_batch_job_definition schema_migrations_swap {
...
       }
     ]
     linuxParameters= {
-     maxSwap= 200000,
+     maxSwap= 0,
      swappiness= 60
```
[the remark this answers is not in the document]
Bento007: No it is not needed if maxSwap is 0, but i left it there in case we decided we need swap again. It'll be easy to figure out what parameter to adjust.

backend/layers/processing/process_download.py
Bento007:
```suggestion
MEMORY_MODIFIER = 2  # adds (x*100)% memory overhead
```

nayib-jose-gloria (approved):

PR #6213, opened 2023-11-13
fix: slack dataset failure notification reports collection version ID

nayib-jose-gloria:
## Reason for Change

- We recently changed the SFN to use collection version ID instead of collection ID. We missed swapping the env var name in the slack failure notification module, therefore failures were being reported with collection version ID as None.


Bento007 (approved):
Weird i thought i fixed that one.

PR #6218, opened 2023-11-13
feat: quickly modify ingest memory configuration without redeploying,

Bento007:
## Reason for Change

- The memory settings for the ingest process are dynamic and can sometimes be off. This results in fail upload. To make it easier to fine tune and adjust the memory setting for ingest, the values have been move into the CorporaConfig. This allow us to quickly change the memory setting for ingest using the corpora config secret. This saves us the need to redeploy prod.

## Changes

- move ingest memory configuration secret to corpora config
- set default values for the ingest memory configuration
- make all memory calculation based on the number of VCPUs allocated.

## Testing steps

- updated tests

### DIFF
--- a/backend/layers/processing/process_download.py
+++ b/backend/layers/processing/process_download.py
@@ -97,7 +97,7 @@ class ProcessDownload(ProcessingLogic):
         :param memory_modifier: A multiplier to increase/decrease the memory requirements by
         :param min_vcpu: The minimum number of vCPUs to allocate.
         :param max_vcpu: The maximum number of vCPUs to allocate.
-        :param memory_per_vcpu: The amount of memory to allocate per vCPU.
+        :param memory_per_vcpu: The amount of memory to allocate per vCPU. 8000 MB is what AWS uses as the ratio
         :param swap_modifier: The multiplier to increase/decrease the swap memory requirements by
         :param max_swap_memory_MB: The maximum amount of swap memory to allocate.
         :return: A dictionary containing the resource requirements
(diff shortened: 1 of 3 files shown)


backend/layers/processing/process_download.py
```diff
@@ -89,21 +83,19 @@ def get_job_definion_name(dataset_version_id: str) -> str:
...
-        max_vcpu: int = MAX_VCPU,
-        max_swap_memory_MB: int = MAX_SWAP_MEMORY_MB,
-        swap_modifier: int = SWAP_MODIFIER,
-        memory_per_vcpu: int = MEMORY_PER_VCPU,
+        memory_modifier: float = config.ingest_memory_modifier,
+        min_vcpu: int = config.ingest_min_vcpu,
+        max_vcpu: int = config.ingest_max_vcpu,
+        max_swap_memory_MB: int = config.ingest_max_swap_memory_mb,
+        swap_modifier: int = config.ingest_swap_modifier,
+        memory_per_vcpu: int = 8000,
```
nayib-jose-gloria, line 91: any reason this one is still set here rather than the config?

Bento007: because this is based on what AWS does and not based on our config.  AWS allocates 8GB per VCPU so we are following that pattern in our estimation.

nayib-jose-gloria: gotcha, thanks

nayib-jose-gloria (approved):

PR #6228, opened 2023-11-14
fix: Use built in set and reset in SecretConfig for testing

Bento007:
## Reason for Change

- Fixes tests that are failing because of how CorporaConfig is mocked
- This simplifies how configuration setting are mocked.

## Changes

- modify how SecretConfig classes are mocked


nayib-jose-gloria (approved):
thank you!

Issue #5880, opened 2023-10-03
Trigger metadata-update processing job in collection metadata update endpoints


nayib-jose-gloria:
In PUT dp/v1/collections/{collection_id] and PATCH  curation/v1/collections/{collection_id], IF Publication DOI is added or updated and new value is valid:
- Fetch list of dataset_version_id in CollectionVersion from DB, then kickoff metadata-update SFN with dataset_version_id, metadata_key_to_upload, new_value arg.

PR #6230, opened 2023-11-15
feat: trigger batch job to update dataset artifact citations upon DOI update

nayib-jose-gloria:
## Reason for Change

- #5880 

## Changes

- Added BatchJobProvider and made it parameter for BusinessLogic
- Added method to BatchJobProvider to kickoff dataset artifact metadata update batch job for a given dataset + map of metadata update changes to persist
- Added static method to BusinessLogic for generating dataset citations (+ refactored for DRYness)
- Trigger dataset artifact metadata update batch job to update the 'citation' for each dataset in a collection upon a detected DOI update in the business logic method that handles collection metadata updates
- Renamed instances of existing_dataset_version --> current_dataset_version in business logic for consistency

## Testing steps

- unit tests + rdev


backend/layers/processing/dataset_metadata_update.py
```diff
@@ -137,68 +136,65 @@ def update_cxg(
 
     def update_metadata(
         self,
-        collection_version_id: CollectionVersionId,
-        dataset_version_id: DatasetVersionId,
+        old_dataset_version_id: DatasetVersionId,
```
Bento007, line 139: old, current, existing, we now have 3 ways of identifying nearly the same thing. Please take the time to consolidate the names.

backend/layers/business/business.py
```diff
@@ -374,6 +458,25 @@ def create_empty_dataset(self, collection_version_id: CollectionVersionId) -> Tu
 
         return (new_dataset_version.version_id, new_dataset_version.dataset_id)
 
+    def create_empty_version_for_existing_dataset(
+        self, collection_version_id: CollectionVersionId, current_dataset_version_id: DatasetVersionId
```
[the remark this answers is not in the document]
Bento007: I'm starting to lean towards old, current, new. In general I think only current and new will be used. Current would be the dataset that is already part of the collection.  New would be the new dataset that will be added to the collection. Old would be the version before the current version, which is never used.

backend/layers/processing/dataset_metadata_update.py
Bento007:
```suggestion
        current_dataset_version_id: DatasetVersionId,
```

Bento007 (changes requested):
Let's converge on naming

backend/layers/business/business.py
```diff
@@ -297,22 +350,38 @@ def update_collection_version(
...
+        if current_collection_version.published_at is not None:
             raise CollectionUpdateException(["Cannot update a published collection"])
 
         # Determine if publisher metadata should be unset, ignored or set at the end of the method.
         # Note: the update needs to be done at the end to ensure atomicity
         unset_publisher_metadata = False
         publisher_metadata_to_set = None
 
+        new_doi = None
+        old_doi = None
```
Bento007, line 363:
```suggestion
        current_doi = None
```

danieljhegeman:
@nayib-jose-gloria you're going to need to generate a new business logic class in each subtask, and pass into the `update_h5ad` call. Right now that function is referencing a (shared) business logic as an instance attribute.

backend/layers/business/business.py
```diff
@@ -409,7 +411,11 @@ def update_collection_version(
             self.database_provider.save_collection_publisher_metadata(version_id, publisher_metadata_to_set)
         self.database_provider.save_collection_metadata(version_id, new_metadata)
 
-        if not ignore_doi_update and new_doi != old_doi and FeatureFlagService.is_enabled(FeatureFlagValues.SCHEMA_4):
+        if (
```
Bento007, line 414:
```suggestion
                if all(
            not ignore_doi_update,
            new_doi != current_doi,
            FeatureFlagService.is_enabled(FeatureFlagValues.SCHEMA_4)
        ):
```

tests/unit/backend/layers/business/test_business.py
```diff
@@ -710,6 +763,29 @@ def test_add_empty_dataset_ok(self):
         self.assertEqual(new_dataset_version.status.upload_status, DatasetUploadStatus.NA)
         self.assertEqual(new_dataset_version.status.processing_status, DatasetProcessingStatus.INITIALIZED)
 
+    def test_create_empty_dataset_version_for_current_dataset(self):
+        collection, revision = self.initialize_collection_with_an_unpublished_revision(num_datasets=1)
+        with self.subTest("Success Case"):
```
Bento007, line 768: Avoid using subtest in new tests.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -2201,5 +2277,72 @@ def add_artifact():
...
+class TestDatasetArtifactMetadataUpdates(BaseBusinessLogicTestCase):
+    """
+    Test methods related to updating metadata fields on dataset artifacts
+    """
+
+    def test_generate_dataset_citation(self):
+        collection = self.initialize_unpublished_collection(num_datasets=1)
+        dataset_version_id = collection.datasets[0].version_id
+
+        with self.subTest("Without DOI"):
```
Bento007, line 2289: Avoid using subtest in new tests.

Bento007 (changes requested):
avoid subtests please

backend/layers/business/business.py
```diff
@@ -297,34 +350,54 @@ def update_collection_version(
...
             raise CollectionUpdateException(["Cannot update a published collection"])
 
         # Determine if publisher metadata should be unset, ignored or set at the end of the method.
         # Note: the update needs to be done at the end to ensure atomicity
         unset_publisher_metadata = False
         publisher_metadata_to_set = None
 
+        new_doi = None
+        current_doi = None
         if not ignore_doi_update:
```
danieljhegeman, line 364: I realize you didn't commit this, but can you change this to a positive-usage variable `update_doi`. It makes more sense given the logic of the usage. Trying to avoid `if not`s. @Bento007 will agree

backend/layers/business/business.py
```diff
@@ -340,6 +413,27 @@ def update_collection_version(
...
 
+        if all(
+            [not ignore_doi_update, new_doi != current_doi, FeatureFlagService.is_enabled(FeatureFlagValues.SCHEMA_4)]
+        ):
+            for dataset in current_collection_version.datasets:
+                if dataset.status.processing_status != DatasetProcessingStatus.SUCCESS:
+                    self.logger.info(
+                        f"Dataset {dataset.version_id.id} is not successfully processed. Skipping metadata update."
+                    )
+                    continue
```
danieljhegeman, line 424: @nayib-jose-gloria this silent failure to update the doi here makes me antsy. I realize the updated Collection metadata is returned to the user, so they technically can verify that the metadata look as they wish, but the doi will be hidden amongst other features (and the others will have been updated!). Especially when using the Discover API...

nayib-jose-gloria: I should update this to read if processing_status == FAILURE

if a dataset makes it to this loop, it has either status SUCCESS or FAILURE, and if it's FAILURE, then its current metadata state is irrelevant since it requires a reupload to fix anyway. Unless there's an edge case I'm not considering?

backend/layers/business/business.py
nayib-jose-gloria: I will handle this in a separate PR to avoid increasing this PR's scope

backend/layers/business/business.py
```diff
@@ -340,6 +413,27 @@ def update_collection_version(
             self.database_provider.save_collection_publisher_metadata(version_id, publisher_metadata_to_set)
         self.database_provider.save_collection_metadata(version_id, new_metadata)
 
+        if all(
+            [not ignore_doi_update, new_doi != current_doi, FeatureFlagService.is_enabled(FeatureFlagValues.SCHEMA_4)]
```
Bento007, line 417: `not ignore_doi_updated` would be easier to understand if positive logic is used. ex: `updated_doi`

backend/layers/business/business.py
```diff
@@ -297,34 +350,54 @@ def update_collection_version(
...
+                    if dataset.status.processing_status not in [
+                        DatasetProcessingStatus.SUCCESS,
+                        DatasetProcessingStatus.FAILURE,
+                    ]:
+                        errors.append(
+                            {
+                                "link_type": CollectionLinkType.DOI,
+                                "reason": "Cannot update DOI while a dataset is processing or awaiting upload.",
+                            }
+                        )
```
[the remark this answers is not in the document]
danieljhegeman: Or, better yet: introduce a single error for each such dataset, but _include a reference to the Dataset id_ in the error. That way they'll at least see which Datasets' processings are holding up the change.

danieljhegeman (changes requested):
see error comment

backend/layers/business/business.py
```diff
@@ -374,13 +468,32 @@ def create_empty_dataset(self, collection_version_id: CollectionVersionId) -> Tu
 
         return (new_dataset_version.version_id, new_dataset_version.dataset_id)
 
+    def create_empty_dataset_version_for_current_dataset(
+        self, collection_version_id: CollectionVersionId, current_dataset_version_id: DatasetVersionId
+    ) -> Tuple[DatasetVersionId, DatasetId]:
```
Bento007, line 473: I don't see any functions using DatasetId. Do we need to return it? We could just return the a `DatasetVersion` and parse `DatasetId`. If needed.

Bento007 (approved):

danieljhegeman (approved):
🚀

danieljhegeman (approved):
gtg 👍 

Whoops. Meant to approve the other PR. Darn tabs. Already approved this one.

Issue #5905, opened 2023-10-04
GE Make load state more obvious


signechambers1:
In our most recent user research sessions for Gene Expression we found that the loading indicator was easily missed, people were not sure if the application was responding to their feedback or just broken.

This work ticket captures the work to improve the loading state to make it more obvious when results are loading.

cc @hthomas-czi for design, @ainfeld for analytics, @signechambers1 for product questions

**UX Design**
[Figma](https://www.figma.com/file/Mfo9nQ1m3jQzdPbwoEZvOP/Gene-Expression?type=design&node-id=7019-413187&mode=dev)
[Skeleton Loader Animation Example (MUI)](https://mui.com/material-ui/react-skeleton/)

dsadgat:
Figure out what is feasible technical wise, and scope prior to taking on ticket! Thanks!

PR #6239, opened 2023-11-16
feat: add loading state to heatmap

kaloster:
## Reason for Change

- #5905

## Changes
- added memoized value to count Total rows
- added Skeleton component and count rows and columns to create loader effect
- utilized `isLoading` and `isLoadingAPI` to hide/show heatmap/loader
- added prop to heat map to hide it using css so `isAnyTissueLoading` can still be utilized

** there are 2 types of animation here - around the 27 second mark this changes into 'wave' animation

https://github.com/chanzuckerberg/single-cell-data-portal/assets/7976579/b7fb292a-d4c2-4d7b-b128-83a0e3d13d5a

### DIFF
--- a/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
+++ b/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
@@ -139,7 +139,6 @@ export default memo(function HeatMap(props: Props): JSX.Element {
                         variant="circular"
                         width={19}
                         height={19}
-                        animation="wave"
                         key={i}
                       />
                     ))}
--- a/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
+++ b/frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
@@ -101,7 +101,7 @@ export default memo(function HeatMap(props: Props): JSX.Element {
           <CellCountLabel>Cell Count</CellCountLabel>
         </TopLeftCornerMask>
         <Container {...{ className }} id={HEATMAP_CONTAINER_ID}>
-          {isLoadingAPI || isAnyTissueLoading(isLoading) ? <Loader /> : null}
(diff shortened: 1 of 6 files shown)


frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -122,7 +126,32 @@ export default memo(function HeatMap(props: Props): JSX.Element {
...
             )}
           </YAxisWrapper>
-          <ChartWrapper ref={chartWrapperRef} top={xAxisHeight}>
+          {isLoadingAPI || isAnyTissueLoading(isLoading) ? (
+            <ChartWrapper
+              top={xAxisHeight}
+              hidden={!isAnyTissueLoading(isLoading)}
+            >
+              <SkeletonContainer>
+                {[...Array(totalElementsCount)].map((i) => (
```
tihuan, line 135:
```suggestion
                {[...Array(totalElementsCount)].map((_, index) => (
```
if the intention is to use the array `index`! Otherwise I think currently `i` is `undefined`?

frontend/src/views/WheresMyGeneV2/components/HeatMap/index.tsx
```diff
@@ -122,7 +126,32 @@ export default memo(function HeatMap(props: Props): JSX.Element {
               }
             )}
           </YAxisWrapper>
-          <ChartWrapper ref={chartWrapperRef} top={xAxisHeight}>
+          {isLoadingAPI || isAnyTissueLoading(isLoading) ? (
```
tihuan, line 129: instead of ternary with `null` maybe just `&&` ?

tihuan (commented):
Wow looks really polished with the loading state 🤯 Amazing work, Ronen and Harley 🔥

tihuan (approved):
LGTM! Thanks for adding the new linter too ✨ 🙏 🚀 !!

PR #6242, opened 2023-11-16
fix: Track mismatches in compare_dict

Bento007:
## Reason for Change

- Some times test fails on this check and there is no information to help debug. This track the mismatches between two dicts and prints them out if the any are found. This also use `unittest` syntax to keep error reporting consistent.
## Changes

- make compare_dicts errors visible to unittest 
- rename compare_dicts to assert_dicts_equal

### DIFF
--- a/tests/test_utils/__init__.py
+++ b/tests/test_utils/__init__.py
@@ -26,9 +26,9 @@ class CompareDictsAddin:
         @contextmanager
         def path(key):
             try:
-                yield path.append(key)
+                yield _path.append(key)
             finally:
-                path.pop()
+                _path.pop()
 
         def _compare_dict(_dict1, _dict2):
             if len(_dict1) != len(_dict2):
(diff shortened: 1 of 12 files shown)


tests/test_utils/__init__.py
```diff
@@ -1,36 +1,82 @@
...
+                with path(key):
+                    if isinstance(value1, dict) and isinstance(value2, dict):
+                        _compare_dict(value1, value2)
+                    elif isinstance(value1, list) and isinstance(value2, list):
+                        if len(value1) != len(value2):
+                            update_mismatch(
+                                "{}: Lengths differ: {} vs {}".format(".".join(path), len(value1), len(value2))
+                            )
+                        # check if the lists contain dictionaries as elements
+                        elif len(value1) > 0 and isinstance(value1[0], dict) and isinstance(value2[0], dict):
```
joyceyan, line 53: i think it's technically possible for a list to contain a dictionary and other things right? so you could have
```
list_of_things = [{"key": "value"}, 3, "string"]
```
so i think you'd have to iterate through all values in the list, and run `_compare_dict` for each value if they're different :|

joyceyan (approved):

Bento007:
Closing because of issues getting unit tests to pass with this change.

PR #6252, opened 2023-11-20
fix: The name of the sfn created uses the new dataset version id

Bento007:
## Reason for Change

- This will make it easier to find datasets that failed during migration.

## Changes

- modify the sfn name to use the new dataset_version_id

## Testing steps

- updated tests


nayib-jose-gloria (approved):

nayib-jose-gloria (commented):
looks like there's a few unit tests to fix, but they look straightforward. feel free to merge once those are fixed

danieljhegeman (approved):

Issue #6234, opened 2023-11-15
feat(SFN Errors): Mark ingest sfn with errors as failed


Bento007:
## Need

When an ingest step function runs, if an error occurs and everything is handled using error handler, the step functions is marked as successful in the AWS Console. This can make it difficult to go to find a dataset that failed during a seurat, cxg, or validation step.

 <img width="518" alt="Screenshot 2023-11-15 at 3 45 54 PM" src="https://github.com/chanzuckerberg/single-cell-data-portal/assets/1429913/5f10277d-0426-47cc-820c-de5391bdc011">

If an error is detected that results in a CXG, seurat, or validation failure, the SFN should handle the error and mark it's self as failed.
<img width="1011" alt="Screenshot 2023-11-15 at 3 49 38 PM" src="https://github.com/chanzuckerberg/single-cell-data-portal/assets/1429913/2749879d-d3b3-49ad-8dcd-8b07ba5f4336">

This will mark the SFN as failed and will be clear when looking at the AWS SFN Console

## Approach

- Add a step to check if there are errors in the output and set SFN to failed if any were found. This is require modifying the outputs of some steps to pass along the error message.

## Definition of Done
- An ingest SFN that fails any of the previous step will be marked as failed.

brianraymor:
Says "flowchart!" in Minion's voice.

PR #6255, opened 2023-11-21
feat: rds error should fail loudly in sfn

danieljhegeman:
## Reason for Change

- #6234 

## Changes

- exit(1) when cxg failure or seurat failure is present. Logging is already occurring locally as well as notification to slack.

## Testing steps

- Tests will come in follow-up but this should be incorporated asap for schema migration testing purposes.

## Checklist 🛎️

### DIFF
--- a/.happy/terraform/modules/sfn/main.tf
+++ b/.happy/terraform/modules/sfn/main.tf
@@ -304,14 +304,14 @@ resource "aws_sfn_state_machine" "state_machine" {
       },
       "DeregisterJobDefinition": {
         "Type": "Task",
-        "Next": "CheckForCxgOrSeuratErrors",
+        "Next": "CheckForErrors",
         "Parameters": {
           "JobDefinition.$": "$[0].batch.JobDefinitionName"
         },
         "Resource": "arn:aws:states:::aws-sdk:batch:deregisterJobDefinition",
         "ResultPath": null
       },
-      "CheckForCxgOrSeuratErrors": {
+      "CheckForErrors": {
         "Type": "Choice",
         "Choices": [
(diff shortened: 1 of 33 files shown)


nayib-jose-gloria (commented):
This change works for cxg + seurat, but the original ticket mentions marking the sfn as failed upon errors in the validate step also. I'm not 100% sure if we should, since some validate failures will be intended failures (i.e. the curator submits an invalid dataset), but it would be helpful for debugging cases where the failure is unexpected. I'm not sure how simple it'd be to determine whether a validation error is expected or unexpected and handle accordingly. @Bento007 your thoughts?

We can decouple that change to unblock this, if we make a separate ticket for it

nayib-jose-gloria (commented):
@Bento007, I believe, mentioned that it might be problematic to fail before we deregister the Job Definition. Am I recalling correctly? This step runs before then, so that'd be a problem if so

danieljhegeman:
> @Bento007, I believe, mentioned that it might be problematic to fail before we deregister the Job Definition. Am I recalling correctly? This step runs before then, so that'd be a problem if so

@nayib-jose-gloria that's right, I'll fix that.

Bento007:
When I was imaging the solution for this, i was assuming we could do it within in the SFN definition. When a step fails in the SFN it puts the error in the result passed to the next step. All we'd need to do is check for the existence of this error field in the resulting JSON and set the state to Failed. This check could then be done right after we deregister.

Bento007 (approved):
Will you also be catching error from download and validate in future work?

Bento007 (approved):
🔥

Issue #5352, opened 2023-07-28
"No Publication" updates


atarashansky:
Users were confused in our usability study by "No Publication." We'd like to make a couple updates to reduce confusion:

- [ ] Pin "Publication Pending" as the last option in the publication secondary filter
- [ ] Pin "Publication Pending" as the last option when Grouping By publication

Additional context here: https://czi-sci.slack.com/archives/C0247APK621/p1690517857542799

PR #6256, opened 2023-11-21
chore: pin no publication to the bottom of publication filter

kaloster:
## Reason for Change

- #5352

## Changes

- this code is creating a sorted version of publication_citations where any object with a name of "No Publication" is moved to the end, and the rest are sorted alphabetically by name. The sorted array is memoized, so it will only be recomputed when publication_citations changes.
Screenshot 2023-11-21 at 9 38 36 AM


tihuan (approved):
Perfect! Love the test :1st_place_medal: Thanks so much for doing that, @kaloster !! LGTM!

PR #6262, opened 2023-11-21
fix: use getsizeof to estimate memory requirements of h5ad.

Bento007:
## Reason for Change
- The current method for estimating `adata.n_var *adata.n_obs` is not accurate. Using `sys.getsizeof` uses the h5ad implementation of estimates the memory size of an adata which should be more accurate across more datasets.

## Changes

- increase memory requirements to 256GB for the downloaded step during ingest
- use `getsizeof` during download step to estimate memory requirement 
- adjust single threaded application to a single VCPU.

## Testing steps

- update unit tests

## Notes for Reviewer

- This change to memory estimation will make the download step slow because it will load the dataset into memory.


.happy/terraform/modules/batch/variables.tf
```diff
@@ -53,5 +53,5 @@ variable frontend_url {
 variable batch_container_memory_limit {
   type        = number
   description = "Memory hard limit for the batch container"
-  default     = 63500
+  default     = 256000
```
Bento007, line 56: This the the memory used for the download job

.happy/terraform/modules/batch/main.tf
```diff
@@ -42,9 +42,9 @@ resource aws_batch_job_definition batch_job_def {
       "value": "${var.frontend_url}"
     }
   ],
-  "vcpus": 8,
+  "vcpus": 4,
```
Bento007, line 45: This is the VCPU use for the download job. It's single threaded and doesn't make use of multiple CPUs. Adjusting it down to save resources.

nayib-jose-gloria (approved):
This looks good, but lets verify with the dev migration before pushing thru to prod

Bento007:
This is not much better than the original estimation.

PR #6267, opened 2023-11-22
fix: mark step function as failed when failure occurs in download / validate

danieljhegeman:
## Reason for Change

- #6234
- Followup to handle case where error occurs in Download or Validate steps (ahead of actual conversion)

## Changes

- add separate path for Download and Validate errors
- add separate Deregister step to account for different json shape of input

## Testing steps

- tested in rdev ✅ 

## Checklist 🛎️


nayib-jose-gloria (approved):
reminder to coordinate prod deploy such that no uploads are going on while this change is deploying! and give a brief explanation that these changes will make reporting errors + debugging easier for engineers

danieljhegeman:
Confirmed no runs in dev or staging; merged.

.happy/terraform/modules/sfn/main.tf
```diff
@@ -299,7 +299,16 @@ resource "aws_sfn_state_machine" "state_machine" {
             "MaxAttempts": 3,
             "BackoffRate": 2.0
         } ],
-        "Next": "DeregisterJobDefinition",
+        "Next": "DeregisterJobDefinitionAfterHandleErrors",
+        "ResultPath": null
+      },
+      "DeregisterJobDefinitionAfterHandleErrors": {
+        "Type": "Task",
+        "Next": "CheckForErrors",
```
Bento007, line 307: what if the downloader fails and there is not batch job registered?

Bento007 (commented):
Can you addd a picture of the machine to make it easier to understand the change

.happy/terraform/modules/sfn/main.tf
danieljhegeman: If the downloader fails and there is no task to deregister, then the deregistration will fail, which is fine; when an engineer comes to look at the step function, they'll see the Download failure.

danieljhegeman:
> Can you addd a picture of the machine to make it easier to understand the change

Screenshot 2023-11-28 at 3 04 14 PM
Screenshot 2023-11-28 at 3 03 55 PM

.happy/terraform/modules/sfn/main.tf
danieljhegeman: But I'm going to run a test scenario where Download fails...

danieljhegeman: I _could_ add an empty default value in the input for $.batch.JobDefinitionName so that we send an empty value to `DeregisterJobDefinitionAfterHandleErrors`... but that would probably just cause the Deregister step to fail, which would mean it would still report as a failure albeit without going all the way to the end and getting to `DownloadValidateError`.

I don't think it's worth worrying about given the intention of these changes, which is to simply cause the overall sfn to fail, which it's doing. The cause is clear enough once you go to investigate, which you have to anyways since we're grouping Download errors with Validate errors.

Thoughts?

Bento007: As long as the step function is marked as failed, i don't mind.

PR #575, opened 2020-09-02
Make the width of the cookie button flexible by removing restriction on width.

maniarathi:

PR #6311, opened 2023-12-04
feat: fetch published doi if preprint

danieljhegeman:
- default to preprint if published doi cannot be fetched
- remove unused backend.common.providers.crossref_provider
- change crossref unit tests to use correct crossref provider

## Reason for Change

- [#575](https://github.com/chanzuckerberg/single-cell/issues/575)

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️


backend/layers/thirdparty/crossref_provider.py
```diff
@@ -143,19 +146,11 @@ def fetch_metadata(self, doi: str) -> dict:
...
-            except Exception:
-                pass
+    def fetch_published_metadata(self, doi_response_message: dict) -> dict:
+        try:
+            published_doi = doi_response_message["relation"]["is-preprint-of"]
+            # the new DOI to query for ...
+            if published_doi[0]["id-type"] == "doi":
+                return self.fetch_metadata(published_doi[0]["id"])
+        except Exception:  # if fetch of published doi errors out, just use preprint doi
+            pass
```
nayib-jose-gloria, line 156: nit: since we are expecting a return value and assigning the output elsewhere, I think its clearer to explicitly return None here instead. But that's a personal preference

backend/layers/thirdparty/crossref_provider.py
```diff
@@ -143,19 +146,11 @@ def fetch_metadata(self, doi: str) -> dict:
...
-                published_doi = message["relation"]["is-preprint-of"]
-                if published_doi[0]["id-type"] == "doi":
-                    return published_doi[0]["id"]
-            except Exception:
-                pass
+    def fetch_published_metadata(self, doi_response_message: dict) -> dict:
+        try:
+            published_doi = doi_response_message["relation"]["is-preprint-of"]
+            # the new DOI to query for ...
+            if published_doi[0]["id-type"] == "doi":
```
nayib-jose-gloria, line 153: I'm not sure if they ever actually put anything other than a DOI, but since the API returns a list here, we should iterate and look for an entry with id-type 'doi' rather than just checking the first entry

nayib-jose-gloria, line 153: I'm not sure if they ever actually put anything other than a DOI, but since the API returns a list here, we should iterate and look for an entry with id-type 'doi' rather than just checking the first entry

backend/layers/thirdparty/crossref_provider.py
```diff
@@ -143,19 +146,11 @@ def fetch_metadata(self, doi: str) -> dict:
...
+                return self.fetch_metadata(published_doi[0]["id"])
```
nayib-jose-gloria, line 154: I noticed that the CrossRef API returns DOIs as strings with the /'s escaped (i.e. "10.3389\/fcvm.2020.00052" from https://api.crossref.org/works/10.1101/2019.12.31.892166). I think we'd need to strip these esc characters if they exist (unless urlparse and/or requests.get already implicitly account for that)

danieljhegeman: Not needed; the single `\` in the context of `\/` gets removed once you load from json into dict.

danieljhegeman: `curl https://api.crossref.org/works/10.1101/2019.12.31.892166 | jq '.message.DOI'`

nayib-jose-gloria (approved):
approving pending an rdev test confirming functionality

PR #6319, opened 2023-12-05
fix(wmg): Remove links to outdated data downloads

sidneymbell:
## Reason for Change

We got an email from a user flagging a data integrity issue. I've filed a separate bug to update these and populate the correct links asap. For now, this PR just removes the download links to the incorrect data.


atarashansky (approved):
🩹

PR #6350, opened 2023-12-14
chore: adhoc metadata reprocess -- DO NOT MERGE

danieljhegeman:
## Reason for Change

- Temporary deployment to allow adhoc metadata reprocessing
- Goal: create updated artifacts with `_updated.(h5ad|rds|cxg)` suffix.

## Changes

    - do not create new db entities -- ONLY create new artifacts
    - start process by PATCH /curation/v1/collections/{collection_id}
    - works for published and private alike
    - no interaction with db rows -- no status updates
    - write files to prefix ending with "_updated"
    - no writes to public datasets bucket -- these will fail due to write constraints for key name -- can copy from artifacts bucket to public datasets bucket later.

## Testing steps

## Checklist 🛎️

### DIFF
--- a/backend/curation/api/v1/curation/collections/collection_id/actions.py
+++ b/backend/curation/api/v1/curation/collections/collection_id/actions.py
@@ -2,7 +2,11 @@ from dataclasses import asdict
 
 from flask import Response, jsonify, make_response
 
-from backend.common.utils.http_exceptions import InvalidParametersHTTPException, MethodNotAllowedException
+from backend.common.utils.http_exceptions import (
+    ForbiddenHTTPException,
+    InvalidParametersHTTPException,
+    MethodNotAllowedException,
+)
 from backend.curation.api.v1.curation.collections.common import (
     extract_doi_from_links,
     get_inferred_collection_version,
@@ -39,7 +43,10 @@ def get(collection_id: str, token_info: dict) -> Response:
 
 
 def patch(collection_id: str, body: dict, token_info: dict) -> Response:
+
     user_info = UserInfo(token_info)
+    if not user_info.is_cxg_admin():
+        raise ForbiddenHTTPException("Unauthorized") from None
 
     if "links" in body and not body["links"]:
         raise InvalidParametersHTTPException(detail="If provided, the 'links' array may not be empty")
--- a/backend/layers/business/business.py
+++ b/backend/layers/business/business.py
@@ -1,4 +1,3 @@
-import copy
 import logging
 import os
 from collections import defaultdict
@@ -7,7 +6,6 @@ from functools import reduce
 from typing import Dict, Iterable, List, Optional, Set, Tuple
 
 from backend.common.corpora_config import CorporaConfig
-from backend.common.feature_flag import FeatureFlagService, FeatureFlagValues
 from backend.layers.business.business_interface import BusinessLogicInterface
 from backend.layers.business.entities import (
     CollectionMetadataUpdate,
@@ -61,7 +59,6 @@ from backend.layers.common.entities import (
     DatasetValidationStatus,
     DatasetVersion,
     DatasetVersionId,
-    Link,
     PublishedDatasetVersion,
 )
 from backend.layers.common.helpers import (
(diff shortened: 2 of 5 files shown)


backend/curation/api/v1/curation/collections/collection_id/actions.py
```diff
@@ -39,7 +43,10 @@ def get(collection_id: str, token_info: dict) -> Response:
 
 
 def patch(collection_id: str, body: dict, token_info: dict) -> Response:
+
     user_info = UserInfo(token_info)
+    if not user_info.is_cxg_admin():
+        raise ForbiddenHTTPException("Unauthorized") from None
```
danieljhegeman, line 49: Kick off reprocessing for a whole collection with a `PATCH` to `/curation/v1/collections/{collection_id}`. Works for private and public Collections.

backend/layers/business/business.py
```diff
@@ -155,14 +151,8 @@ def trigger_dataset_artifact_update(
...
             )
-        if new_dataset_version_id is None:
-            new_dataset_version_id = self.create_empty_dataset_version_for_current_dataset(
-                collection_version.version_id, current_dataset_version_id
-            ).version_id
 
-        self.batch_job_provider.start_metadata_update_batch_job(
-            current_dataset_version_id, new_dataset_version_id, metadata_update
-        )
+        self.batch_job_provider.start_metadata_update_batch_job(current_dataset_version_id, metadata_update)
```
danieljhegeman, line 155: want the same dataset version id

backend/layers/business/business.py
```diff
@@ -340,98 +330,23 @@ def update_collection_version(
...
-                        break
-
-            if current_doi and new_doi is None:
-                # If the DOI was deleted, remove the publisher_metadata field
-                unset_publisher_metadata = True
-            elif (new_doi is not None) and new_doi != current_doi:
-                # If the DOI has changed, fetch and update the metadata
-                publisher_metadata_to_set = self._get_publisher_metadata(new_doi, errors)
+        # Determine if the DOI has changed
+        current_doi = next((link.uri for link in current_collection_version.metadata.links if link.type == "DOI"), None)
```
danieljhegeman, line 336: use the existing DOI

backend/layers/business/business.py
```diff
@@ -340,98 +330,23 @@ def update_collection_version(
...
+                self.logger.info(
+                    f"Dataset {dataset.version_id.id} is not successfully processed. Skipping metadata update."
                 )
+                continue
+
+            citation = self.generate_dataset_citation(
+                current_collection_version.collection_id, dataset.version_id, current_doi
+            )
+            metadata_update = DatasetArtifactMetadataUpdate(citation=citation)
+            self.trigger_dataset_artifact_update(current_collection_version, metadata_update, dataset.version_id)
```
danieljhegeman, line 349: skip straight to updating _all_ datasets

backend/layers/business/business.py
```diff
@@ -340,98 +330,23 @@ def update_collection_version(
...
+        for dataset in current_collection_version.datasets:
+            if dataset.status.processing_status != DatasetProcessingStatus.SUCCESS:
+                self.logger.info(
+                    f"Dataset {dataset.version_id.id} is not successfully processed. Skipping metadata update."
                 )
+                continue
+
+            citation = self.generate_dataset_citation(
+                current_collection_version.collection_id, dataset.version_id, current_doi
+            )
```
danieljhegeman, line 347: generate citation with 
- original dataset version id
- current doi

nayib-jose-gloria (commented):
overall thought: I think we should adapt this approach into its own batch job that we manually kickoff rather than kicking it off via PATCH and using the existing dataset_metadata_update job. Its not very much work to create a new batch definition, and we can avoid commenting out / changing existing business logic. Thus we can avoid gating new submissions or worrying about the pitfalls of kicking off a batch job per dataset (i.e. the DockerTimeouts we saw in testing dataset metadata reprocessing in large collections)

I can work on a branch

nayib-jose-gloria:
> overall thought: I think we should adapt this approach into its own batch job that we manually kickoff rather than kicking it off via PATCH and using the existing dataset_metadata_update job. Its not very much work to create a new batch definition, and we can avoid commenting out / changing existing business logic. Thus we can avoid gating new submissions or worrying about the pitfalls of kicking off a batch job per dataset (i.e. the DockerTimeouts we saw in testing dataset metadata reprocessing in large collections)
> 
> I can work on a branch

working on modified solution here, will do some testing to determine feasibility. Though I am also going to work on a rollback plan as an alternative, should this approach not give us confidence when testing. https://github.com/chanzuckerberg/single-cell-data-portal/pull/6351

Issue #6345, opened 2023-12-13
Bug: Dataset Citations Published with wrong collections URL


nayib-jose-gloria:
Tracking issue for Data Portal corpus being migrated to schema 4.0.0 with a typo in dataset citations generated for each Dataset.

Example Published Citation:
'Publication: https://doi.org/10.1016/j.cell.2022.11.005 Dataset Version: https://datasets.cellxgene.cziscience.com/3effe418-5174-4365-ae79-0b68bfa127f1.h5ad curated and distributed by CZ CELLxGENE Discover in Collection: https://cellxgene.cziscience.com/2d2e2acd-dade-489f-a2da-6c11aa654028'

The Collection link should be: https://cellxgene.cziscience.com/collections/2d2e2acd-dade-489f-a2da-6c11aa654028

Make fix in citation generator for new uploads, AND backfill update in metadata AND artifacts for datasets.

nayib-jose-gloria:
Fixed

PR #6351, opened 2023-12-14
feat: reprocess dataset citations in-place

nayib-jose-gloria:
## Reason for Change

- #6345
- Proposed solution, built off of proposed approach in https://github.com/chanzuckerberg/single-cell-data-portal/pull/6350/files#diff-df36ae23a2a7f265486f5569606db86dbd40811ca8b99fda3aafb293d1b1d379

## Changes
- new batch job for reprocessing dataset citations modeled after dataset_metadata_update (i.e. 'light' reprocessing by doing point updates of our artifacts' metadata where necessary, rather than re-validating, re-converting our datasets)
- 1 batch job per dataset
- old files are not overwritten as part of job, new artifacts will be created in a new file in the same directory as existing datasets in corpora-data-prod. Then validated and then copied to replace old artifacts with a bulk cp job [This approach is To Be Finalized]
- Does not create new dataset versions, to retain existing dataset version IDs.
- Updates dataset metadata in DB for current dataset version IDs to include new citation

## Testing steps

- rdev
- Test [Run](https://us-west-2.console.aws.amazon.com/batch/home?region=us-west-2#jobs/ec2/detail/0a7bebb7-3286-4d50-80ca-f60ab67ce60b) + S3 [Output](https://s3.console.aws.amazon.com/s3/buckets/env-rdev-artifacts?region=us-west-2&prefix=pr-6351/45fd8020-45e5-40a0-8562-e4fc405f0a9a_updated/&showversions=false)
- requires refinement of vCPU + swap allotted for batch job, need to test with collection with multiple large datasets

### DIFF
--- a/backend/layers/processing/reprocess_dataset_metadata.py
+++ b/backend/layers/processing/reprocess_dataset_metadata.py
@@ -208,7 +208,7 @@ class DatasetMetadataReprocess(ProcessDownload):
 
         artifact_uris = {artifact.type: artifact.uri for artifact in current_dataset_version.artifacts}
 
-        new_artifact_key_prefix = self.get_key_prefix(current_dataset_version_id) + "_updated"
+        new_artifact_key_prefix = self.get_key_prefix(current_dataset_version_id.id) + "_updated"
 
         artifact_jobs = []
 
--- a/backend/layers/processing/reprocess_dataset_metadata.py
+++ b/backend/layers/processing/reprocess_dataset_metadata.py
@@ -1,0 +1,308 @@
+"""
+Updates citations in-place across dataset artifacts for a Collection
+"""
+import json
+import logging
+import os
+from multiprocessing import Process
+
+import scanpy
+import tiledb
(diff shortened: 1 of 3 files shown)


backend/layers/processing/reprocess_dataset_metadata.py
```diff
@@ -0,0 +1,308 @@
...
+        # blocking call on async functions before checking for valid artifact statuses
+        [j.join() for j in artifact_jobs]
+
+    def update_dataset_citations_in_collection(
+        self,
+        collection_version_id: CollectionVersionId,
+    ):
+        collection_version = self.business_logic.get_collection_version(collection_version_id)
+        doi = next((link.uri for link in collection_version.metadata.links if link.type == "DOI"), None)
+        for dataset in collection_version.datasets:
```
nayib-jose-gloria, line 284: we'll likely need some blocking / batching / threadpooling for this for larger collections, looking into it

danieljhegeman (approved):
🚀

PR #6356, opened 2023-12-14
chore: remove parameterize from requirements-processing.txt

Bento007:
## Reason for Change

- remove parameterize a superfluous requirement from requirements-processing.txt

## Changes

- remove parameterize a superfluous requirement from requirements-processing.txt


Bento007:
Closing, because `pytest.mark.parameterize` does not work with `unittest.TestCase` classes

PR #6358, opened 2023-12-15
fix: add schema_reference + citation to RDS artifacts

nayib-jose-gloria:
## Reason for Change

- #6345
- Fix logic for assigning new keys to RDS artifacts

## Changes

- support adding new keys, not just re-assigning values of existing ones 

## Testing steps

- rdev

### DIFF
--- a/backend/layers/processing/reprocess_dataset_metadata.py
+++ b/backend/layers/processing/reprocess_dataset_metadata.py
@@ -8,7 +8,7 @@ from multiprocessing import Process
 
 import scanpy
 import tiledb
-from rpy2.robjects import StrVector
+from rpy2.robjects import ListVector, StrVector, r
 from rpy2.robjects.packages import importr
 
 from backend.common.utils.corpora_constants import CorporaConstants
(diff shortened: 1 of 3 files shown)


backend/layers/processing/reprocess_dataset_metadata.py
```diff
@@ -114,11 +114,22 @@ def update_rds(
...
+            else:
+                new_keys.append((key, val))
+
+        if new_keys:
+            new_key_vector = ListVector({k: v for k, v in new_keys})
+            seurat_metadata += new_key_vector
+            r.assign("rds_object", rds_object)
+            r.assign("seurat_metadata", seurat_metadata)
+            r("rds_object@misc <- rds_misc")
+            rds_object = r["rds_object"]
```
danieljhegeman, line 132: Not totally clear to me what's going on here

danieljhegeman (commented):
I'll approve if you can just address my [question](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6358/files#r1427467291)

backend/layers/processing/reprocess_dataset_metadata.py
nayib-jose-gloria: rpy2 is an interface for running R in a Python process. It provides a number of functions / objects that allow you to perform R operations pythonically, but it doesn't integrate well with all R libraries. So, it also provides an interface that lets you write R directly. 

With Seurat, I found that the rpy2 python representation of the 'Misc' dict (this is the Seurat 'uns' equivalent) doesn't support adding new fields (if you tried to saveRDS after the transformation above, it does not persist the additions). Though it does persist editing values of existing keys (existing code). 

So, this block adds the new metadata to the Misc object, assigns the python objects to R objects, and runs the R code to update the Misc dict for the rds_object. Then passes it back into a python object to be saved below.

danieljhegeman: Fair enough!

danieljhegeman (approved):

PR #6384, opened 2023-12-20
fix: add relevant containers to make recipe so they start.

Bento007:
## Reason for Change

- this make it easier to setup the local environment agian


Bento007:
Is this helpful to anyone or just me?

seve:
@Bento007 you tagged the wrong user, the one previously tagged for review sends an email to cellxgene@

Makefile
```diff
@@ -120,11 +120,11 @@ local-sync: local-rebuild local-init  ## Re-sync the local-environment state aft
 
 .PHONY: local-start
 local-start: .env.ecr ## Start a local dev environment that's been stopped.
-	docker-compose $(COMPOSE_OPTS) up -d
+	docker-compose $(COMPOSE_OPTS) up -d backend frontend database oidc localstack
 
 .PHONY: local-stop
 local-stop: ## Stop the local dev environment.
-	docker-compose stop
+	docker-compose stop frontend backend frontend database oidc localstack
```
atolopko-czi, line 127: one frontend too many?

atolopko-czi (commented):
I've done this manually in this past, so this seems like a useful change. Should the target name be explicit about it's altered function? e.g. `{start,stop}-backend`. And maybe a comment, "excludes batch processing pipelines".

Makefile
Bento007:
```suggestion
	docker-compose stop frontend backend database oidc localstack
```

Bento007:
> I've done this manually in this past, so this seems like a useful change. Should the target name be explicit about it's altered function? e.g. `{start,stop}-backend`. And maybe a comment, "excludes batch processing pipelines".

@atolopko-czi can we assume if they are running make recipes, that they will look at the definition if needed. I think the recipe is simple and self explanatory. I don't see a lot of value in create a new recipe to be more explicit. If we did then we should make one for all of the containers which i think is overkill.

atolopko-czi:
> > I've done this manually in this past, so this seems like a useful change. Should the target name be explicit about it's altered function? e.g. `{start,stop}-backend`. And maybe a comment, "excludes batch processing pipelines".
> 
> @atolopko-czi can we assume if they are running make recipes, that they will look at the definition if needed. I think the recipe is simple and self explanatory. I don't see a lot of value in create a new recipe to be more explicit. If we did then we should make one for all of the containers which i think is overkill.

Wasn't suggesting a new recipe, just a rename. But it's a very minor suggestion, so feel free to ignore. :)

PR #6410, opened 2024-01-03
test: Fix CellGuide tree view e2e test

tihuan:
## Reason for Change

The test is failing right now because the node is half hidden and not hoverable

Screenshot 2024-01-03 at 9 25 40 AM

## Changes

1. Update test to zoom out of the tree view, so more nodes are in the view

## Testing steps

All tests should pass now!

## Checklist 🛎️


atarashansky (approved):
LGTM! Elegant fix 👍

PR #6413, opened 2024-01-03
feat: notify @sc-pm and @sc-eng with each deploy

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/6100

## Changes

three things this changes:
1. notifies @sc-eng if there are issues with any happy deploy
2. notifies @sc-eng and @sc-pm for all prod deploys, regardless of status
3. notifies @sc-eng for all other deploys that fail

the [documentation](https://action-slack.netlify.app/usage/with/) isn't 100% clear, but i believe we need to specify `if_mention` if we want any of the things specified in `mention` to work. i'm pretty sure this is why our existing slack alerts are not notifying @here. regardless, i think @sc-eng is better than @here 

note that (2) notifies #cell-science-ops, which is where we currently do the manual notification, so that seems right to me.

## Testing steps

not sure how to test this.. is it possible? it may make sense to split out the 3 changes into separate PR's so we can look at them individually

### DIFF
--- a/.github/workflows/deploy-happy-stack.yml
+++ b/.github/workflows/deploy-happy-stack.yml
@@ -90,7 +90,8 @@ jobs:
         with:
           status: ${{ job.status }}
           fields: repo,commit,author,eventName,workflow,job,mention
-          mention: "here"
+          mention: "subteam^S02KZL4SQM6"
+          if_mention: "always"
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK }}
         if: failure() && github.ref == 'refs/heads/main'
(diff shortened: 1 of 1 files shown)


nayib-jose-gloria (commented):
re: testing, we can either:

1) merge-and-fix as we go, since every merge will trigger this GHA
2) add a (temp, branch-only) trigger to this GHA on push to this branch, and replace all instances of 'github.event.deployment.<attribute>' with hard-coded values (i.e. 'dev' for environment, HEAD of dev tag for tag, 'devstack' for stackname, etc.). This should allow us to run a deployment to dev on any push to this branch, to verify the alerts are working as expected, which could be a less messy/shorter feedback loop than option 1 if we require a lot of small fixes. Rollback temp changes once an approach is verified.

nayib-jose-gloria (approved):

PR #6423, opened 2024-01-05
chore: use poetry for dependency management of docker containers.

Bento007:
## Reason for Change
- Libraries we depend on sometime have conflicting dependency requirements. [Poetry](https://python-poetry.org/) helps to alleviate part of the problem by locking dependencies as well as providing tools to keep the dependences up to date with versions that work with all other dependencies. 
- Locking the dependences helps eliminate variability in our build for reproducible results.
- Our dependencies for developing and testing are mixed with run time code. This adds more variability to our builds as well as bloat. Using `pyproject.toml` makes it easier to keep them separated and install as needed.

## Changes

- update dockerfile* build recipes to install dependencies from requirements files generated by poetry
- update dockerfile build recipes to install depedncies 
- consolidate requirements for upload_failure and upload_success into upload_handler
- remove old *requirements.txt* files
- update *.dockerignore* to allow files in *./python_dependencies* to be copies and remove *requirements.txt* references.
- update instruction for install WMG test environment.
- add PIP flags to make docker builds more efficient
- separate out developer requirements from run time requirements
- add make recipes to lock and update requirements.

## Testing steps
- testing using rdev

## Checklist 🛎️

## Notes for Reviewer
- *poetry.lock* files are generated and do not need review. 
- the previous requirements files were moved into *pyproject.toml* files


prathapsridharan:
@Bento007 - Thanks for this PR! A few questions:

1. Maybe I missed the context and reasoning but for my own edification, what is the dependency management problem we are trying to solve here?
2. Should reviewer's in this group and perhaps other engineers explicitly talk about what the problem is, whether it is annoying enough to solve, and how poetry solves this?

My suggestion is that if the engineering team isn't clear about the particular dependency management problem and/or not familiar with `poetry` perhaps it is worth talking about this first before adding a dependency like `poetry`. What do you think?

Bento007:
> @Bento007 - Thanks for this PR! A few questions:
> 
> 1. Maybe I missed the context and reasoning but for my own edification, what is the dependency management problem we are trying to solve here?
> 2. Should reviewer's in this group and perhaps other engineers explicitly talk about what the problem is, whether it is annoying enough to solve, and how poetry solves this?
> 
> My suggestion is that if the engineering team isn't clear about the particular dependency management problem and/or not familiar with `poetry` perhaps it is worth talking about this first before adding a dependency like `poetry`. What do you think?

[point 4](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6424#issuecomment-1879362219). is a good example of what this helps to mitigate. If we start using dependabot to upgrade our dependencies more frequently we will want a stronger guarantee of exactly what packages are install for reproducibility.

Bento007:
There is no need to change how we install dependencies at this time

PR #6424, opened 2024-01-05
feat: enable dependabot for GHA and frontend packages

Bento007:
## Reason for Change

- dependabot will automatically check for package updates and open pull request to update them.
- it will also make recommendation when a security vulnerability is discovered in a package we depend on.

## Changes

- add dependabot.yml to check for updated to GHA and frontend packages.


prathapsridharan:
I have a few questions:

1. What happens if there are many such pull requests for updated packages? 
2. Will engineers actually go through each PR and consider whether to update or not?
3. What if engineers ignore them and stale package update PRs build up?
4. Is it possible that an old update for package A was missed (due to an ignored PR) but then a PR for update package B which depends on an updated version of package A is opened and the reviewer accepts the package B update? Would that cause a failure in the build?

Bento007:
@prathapsridharan 
1. we can configure how many dependabot will have open at a time. The default is 5 per requirements file. It will give us a change to gradually burn down the backlog of requirements without being swamped.
2. If there are no errors from our automated tests, I don't see a strong reason to not accept the PR. If errors occur we can decide how to proceed.
3. We have a reaper that closes stale PRs after two weeks if they are not resolved.
4. In that case we should pin package A to a version so B does not try to update. We should also see our tests fails as a result. This is something the package manager poetry is helpful at preventing in python.

prathapsridharan:
> @prathapsridharan
> 
> 1. we can configure how many dependabot will have open at a time. The default is 5 per requirements file. It will give us a change to gradually burn down the backlog of requirements without being swamped.
> 2. If there are no errors from our automated tests, I don't see a strong reason to not accept the PR. If errors occur we can decide how to proceed.
> 3. We have a reaper that closes stale PRs after two weeks if they are not resolved.
> 4. In that case we should pin package A to a version so B does not try to update. We should also see our tests fails as a result. This is something the package manager poetry is helpful at preventing in python.

For (4), in the scenario I have in mind, the dependabot PR to _update package A did not happen due to an engineer forgetting about the PR and the reaper closing that PR_. So for example, package A was at version `1.0` and did not get updated to `1.1` by _involuntary omission_. Now, package B update PR comes but it depends on package A being at `1.1` which didn't happen. But an engineer accepted this PR to update package B which might then cause a failed build right?

I don't know how likely this scenario is and whether it is worth agonizing over. Just want to point it out because this does seem to rely on some level of developer diligence but it might be manageable if there aren't too many PRs coming in.

Bento007:
> > @prathapsridharan
> > 
> > 1. we can configure how many dependabot will have open at a time. The default is 5 per requirements file. It will give us a change to gradually burn down the backlog of requirements without being swamped.
> > 2. If there are no errors from our automated tests, I don't see a strong reason to not accept the PR. If errors occur we can decide how to proceed.
> > 3. We have a reaper that closes stale PRs after two weeks if they are not resolved.
> > 4. In that case we should pin package A to a version so B does not try to update. We should also see our tests fails as a result. This is something the package manager poetry is helpful at preventing in python.
> 
> For (4), in the scenario I have in mind, the dependabot PR to _update package A did not happen due to an engineer forgetting about the PR and the reaper closing that PR_. So for example, package A was at version `1.0` and did not get updated to `1.1` by _involuntary omission_. Now, package B update PR comes but it depends on package A being at `1.1` which didn't happen. But an engineer accepted this PR to update package B which might then cause a failed build right?
> 
> I don't know how likely this scenario is and whether it is worth agonizing over. Just want to point it out because this does seem to rely on some level of developer diligence but it might be manageable if there aren't too many PRs coming in.

I agree it will require developer diligence. This exact scenario is why i'm in favor of switching how we manage our python dependencies to something that locks the current requirements such a poetry.

prathapsridharan:
@Bento007 - Sounds good. Is this PR necessary since you have another PR that seems to already alert on `npm` packages as well as `python` packages?

https://github.com/chanzuckerberg/single-cell-data-portal/pull/6425

Bento007:
@prathapsridharan this one should be merged before #6425

prathapsridharan (approved):
LGTM

Issue #6247, opened 2023-11-20
Cell Types Descriptions - Batch 2 for ready for release 


niknak33:
Second batch of descriptions are ready for upload to CellGuide

Please see the following Google Sheet - https://docs.google.com/spreadsheets/d/1HOeyfPbv_ncJrgNP5kkYdJZYZuMUkDoqpoqPVKgQcgc/edit#gid=1294147951

A separate sheet called `BATCH 2 for release: GPT4 cl_descriptions` contains the next 100 descriptions.

kaloster:
More form Alec:

Nik should give you a spreadsheet with the new descriptions that are ready for release. If you have 100 entries in that spreadsheet, format them into 100 jsons with the necessary attributes described in your endpoint, write a script that reads the authentication credentials from an environment variable to include in the request headers, and iterate through the different jsons you created, POSTing them one at a time

This script should belong to the scripts/ directory

It's up to you to decide what input the script receives - it could be a folder containing already-processed jsons, or the table itself

PR #6426, opened 2024-01-08
feat(cellguide): upload json script

kaloster:
## Reason for Change

- #6247

This is a script to upload cellguide descriptions using the `/cellguide/v1/upload` endpoint.

It requires a csv file to be present in the `scripts` directory with the following columns:
`Cell Ontology ID, Final version (QC'd), Supporting reference 1, Supporting reference 2, Supporting reference 3, 
Supporting reference 4, Supporting reference 5`

The script will convert the csv to json and post each row to the endpoint.

## Testing steps

Step 1 - Login: `python upload_cellguide_descriptions.py --deployment rdev login --api_key <api_key>`
Step 2 - Upload: `python upload_cellguide_descriptions.py --deployment rdev upload --path cl_descriptions.csv`

## Notes for Reviewer
- rdev option is there for testing purposes - once reviewed and approved will be removed before merging.


scripts/upload_cellguide_descriptions.py
```diff
@@ -0,0 +1,103 @@
...
+@click.group()
+@click.option("--deployment", default="test", show_default=True, help="The name of the deployment to target")
+@click.pass_context
+def cli(ctx, deployment):
+    """
+    Set deployment stage and api base url
+    """
+    os.environ["DEPLOYMENT_STAGE"] = deployment
+    api_base = {
+        "rdev": "https://pr-6426-backend.rdev.single-cell.czi.technology",
```
kaloster, line 30: rdev option is here for testing purposes - once reviewed and approved this line will be removed before merging.

scripts/upload_cellguide_descriptions.py
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return json.dumps(json_row)
+
+
+def format_references(csv_row):
+    """
+    Format references list
+    """
+    references = []
+    for i in range(1, 6):
```
atarashansky, line 93: My memory is murky so I might not be remembering the spec properly - are there known to be only a maximum of 5 supporting references?

atarashansky (approved):
Left one small question, but otherwise this looks great! I'm trusting that you've tested this in rdev already..

scripts/upload_cellguide_descriptions.py
kaloster: yes, it's up to not including the upper limit when range is defined

Issue #6405, opened 2024-01-02
[BUG] GeneExpression SVG download misalignment


tihuan:
Context:

SVG download is currently misaligned - dots for cell types and genes are not aligned properly. See video and image below

Screenshot:
<img width="757" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBenFUQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1d0a55c5e835f4e5752ac3d4eb52142ea2e83508/Screenshot%202024-01-02%20at%2012.14.40%E2%80%AFPM.png" alt="Screenshot 2024-01-02 at 12.14.40 PM.png" />

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBemFUQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d18d4c62440a1528c1d9b11760bf1902a7ce5d59/Screen%20Recording%202024-01-02%20at%2011.04.15%E2%80%AFAM.mov)

tihuan:
submitting a PR today

Issue #6381, opened 2023-12-20
GeneExpression Image Diffing E2E tests


tihuan:
Context:
@niknak33 pointed out that GeneExpression image download is a very important feature that we should ensure its functionality and reliability as much as possible

We have a few image download bugs caused by not having image diffing tests in place, since these things are much harder for humans to manually check frequently

Action Items:
1. Investigate different ways to do image diffing across MacOS and Linux, so we can hook this up with existing Github Action e2e tests
2. Options:
    1. Chromatic e2e visual test - this is Timmy's most ideal solution, but it's currently in preview and CZI doesn't have access to it yet. (I do have contacts at Chromatic that I can reach out if we don't get access in the new year)
    2. Set up Docker to run image diffing test locally, so we can generate images needed for GHA/Linux to use
    3. Datadog to see if they can download image files and use md5 hash to check consistency

tihuan:
@seve @dsadgat @niknak33 for awareness: I just created this ticket to track the work we need to do image diffing for GE download image feature. Not sure which sprint we'd want to pull this in in the new year

Thank you!

niknak33:
thanks @tihuan I would like to see if we can fit it into the first few sprints of the new year, if possible, thanks for creating a ticket.

tihuan:
Awesome! Look forward to implementing this 🚀 Thanks, Nik!

tihuan:
Great news! My contact at Chromatic gave us early access to the feature already 😁 Will look into setting this up now.

A few things I need to check are:

1. How to add Chromatic to Data Portal repo and set that up
2. How to set up Chromatic e2e test per their docs here: https://www.chromatic.com/docs/e2e-visual-tests/
2. Chromatic budget. Currently I think CZI has one Chromatic billing and the quota is used by all Edu and Science projects, so I'm not sure how CZI teams share this budget cost internally (i.e. does Edu eat the whole Chromatic cost, does Science already share 50% cost because SDS uses it, etc.). Maybe just something we need to keep in mind for the future! (CC: @dsadgat 🙏)

tihuan:
@dsadgat I'm thinking about adding this to the coming sprint. What do you think? Thank you!

tihuan:
@dsadgat I removed this ticket from the sprint, because instead of adding image diffing in GE, it's more beneficial to spend the time doing so in Explorer for spatial work. Thank you!

PR #6428, opened 2024-01-08
fix: 6405 GE SVG download

tihuan:
## Reason for Change

- #6405

Followup - to prevent image download regression, we have #6381 in the pipeline to run visual tests

## Changes

1. To fix the SVG height not taking filtered cell types into account, resulting in the SVG having huge blank spaces `<SaveExport />` SVG download function is now using `allChartProps` like CSV download, in order to get the single source of truth on the tissues, genes, and expressions are rendered
2. Update SVG download to use the new constant for heat map cell height `HEAT_MAP_TOTAL_CELL_HEIGHT_PX` that was introduced 2 months ago
3. Update a fixture test URL to remove `compare` URL param, since it's not supposed to have it

SVG download now looks like this:

Screenshot 2024-01-08 at 8 05 32 AM

## Testing steps

1. Go to https://pr-6428-frontend.rdev.single-cell.czi.technology/gene-expression
2. Add some filters and genes
3. Download SVG
4. The downloaded file should be aligned now with no buggy huge white spaces like described in the ticket above

## Checklist 🛎️

- [x] Add product, design, and eng as reviewers for rdev review
- [x] For UI changes, add screenshots/videos, so the reviewers know what you expect them to see


joyceyan (approved):

tihuan:
Thanks so much for the review, @joyceyan ! Will automerge 🚀

Issue #5795, opened 2023-09-21
Investigate and upgrade Node from 16.14.2 to latest [SPIKE]


tihuan:
We're currently locking Node version to 16.14.2, IIRC due to errors we encountered running `npm i` in GHA. Maybe it's time to revisit to see if the issues can be resolved with some affected package upgrades

tihuan:
We're pretty behind node versions (latest LTS is v20, and we're still on v16) 

Newer versions of dependencies we use now require at least Node v18 (such as Next.js), so we should see if we can upgrade to the latest version now

https://nextjs.org/docs/getting-started/installation

tihuan:
Actually unsure how much effort this will be yet. Might need a tech spike first to investigate how gnarly it is to upgrade to the latest node version.

tihuan:
Timeboxing to 2 days

PR #6431, opened 2024-01-08
chore: 5795 Upgrade node LTM version

tihuan:
## Reason for Change

- #5795

⚠️⚠️⚠️ ACTION REQUESTED ⚠️⚠️⚠️

1. To run `npm run dev` properly, you may need to run `npm rebuild` after this PR is merged into `main`

## Changes

1. Update Node version from 16.14.2 to 20.10.0 (the current LTM)
2. Update its corresponding type file and fix some type issues caused by TypeScript defaulting to NodeJS' API instead of DOM's equivalents, which is what we actually want

## Testing steps

1. The whole app should work as intended. Since this is a node related upgrade, it shouldn't affect client side runtime in the browser but mostly server side runtime and build time
2. I've checked server side rendering, app obviously builds for both dev and prod modes, and `npm i` works for me

## Checklist 🛎️


tihuan:
WMG rdev BE is currently down, thus e2e tests are failing

tihuan:
All tests should be passing now 💯

kaloster (approved):

tihuan:
Thanks so much for rerunning the GHA for me, @kaloster 🙏 !!

Issue #5482, opened 2023-08-14
On-demand trigger of prod deployment via cmd that invokes via GHA workflow dispatch


Bento007:
Use same branch merge strategy, but without using a PR, possibly with a script now that can be invoked via `make` command.

The prod deploy script should verify that staging HEAD commit has passed all tests before merging and pushing the prod branch.

Update deployment checklist in on-call log.

Bento007:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/5482?planning-poker) with Zenhub @ebezzi @tihuan

Bento007:
I assume this would be a new recipe in our make file for data portal, as well as one for explorer.

prathapsridharan:
@atolopko-czi @Bento007 @tihuan @ebezzi - I want to talk about the user flow for initiating a `prod` deploy. 

First, I want to make very obvious, the general approach so that we are all on the same page:

_Because we want to have this deploy triggered by a human, I think creating a GH workflow using `workflow_dispatch` that can be triggered manually from the command line or from the "Actions" tab on the web UI seems reasonable. This manual workflow would be triggered from a source branch and presently, that source branch is `staging`. We will leverage what we have already by overwriting the `prod` branch with some commit, say X, that is selected for deployment from the `staging`. Creating `prod` branch off commit X would then leverage our existing GHA automation and trigger a `prod` deploy._

_As far as determining what should be deployed to `prod`_, I can think of 3 broad approaches:

_A) Take the latest commit on `staging` branch and mark that as the candidate commit to deploy. If all GHA checks on this latest commit are not passing, then the `prod` branch will not be created and the `prod` deploy will not be initiated._

_B) Find the latest commit on `staging` with all GHA checks passing and create the `prod` branch off that commit_

_C) Have the manually triggered workflow, accept a commit hash from the user off which to create the `prod` branch_

My view is that we should go with **(A)**. We are using `staging` as our continuous deployment target as practice (like minor league) to eventually graduate to real continuous deployment to `prod` (major league). We should then take failing GHA on `staging` seriously and fix them by making commits to `main`. Not being able to deploy to `prod` is a forcing function to fix `staging` if it is broken. This approach also has the benefit of simplicity in implementation and will work nicely in the MVP and beyond.

[...]

PR #6434, opened 2024-01-08
feat(CICD): create GHA that promotes HEAD of staging to prod + triggers deploy

nayib-jose-gloria:
## Reason for Change

- #5482

## Changes

- add GHA that 1) checks if most recent stage deploy is green via Github API, 2) if so, merges HEAD of staging into prod and triggers our regular deployment process, obviating the need for prod deployments via PR (though, it is still available as an option)
- leverage reusable workflows for image build + deployment trigger, and refactors main -> staging merge GHA to use reusable workflow pattern rather than repository dispatch pattern as well
- GHA can be manually dispatched via github action UI ("workflow_dispatch"), or via make command in command line

## Testing steps

- merge and test

## Notes for Reviewer
- Implementation assumes we will maintain the current state where the Deployment GHAs are only run by czibuiltbot + Deployments are the only type of GHA czibuiltbot runs in our repo, as we call the Github API and query by 'actor' czibuiltbot to parse for the status of the most recent staging Deployment action. If not success, we fail to promote staging to prod.
- Note: Achieving this ticket requires that we loosen the prod branch protection to require a PR to merge; we can keep that protection for non-authorized users, but users in the czi-single-cell-eng group must be allowed to merge to prod without a PR.


joyceyan:
I looked at the ticket and it seems like @Bento007 and @prathapsridharan have more context so I'll leave the PR review to them, but I'm happy to take a look if they're busy or something!

Makefile
```diff
@@ -229,3 +229,7 @@ coverage/report-xml: coverage/combine
 .PHONY: coverage/report
 coverage/report-html: coverage/combine
 	docker-compose $(COMPOSE_OPTS) run --rm -T backend bash -c "cd /single-cell-data-portal && coverage html --data-file=$(COVERAGE_DATA_FILE) -i --skip-empty"
+
+.PHONY: promote-staging-to-prod # set GITHUB_PAT env var OR call as "make GITHUB_PAT=<github_pat> promote-staging-to-prod"
```
danieljhegeman, line 233: can we work this into the recipe so the user doesn't have to think about it?

Makefile
```diff
@@ -229,3 +229,7 @@ coverage/report-xml: coverage/combine
...
+promote-staging-to-prod:
+	curl -X POST https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/dispatches -H 'Accept: application/vnd.github.everest-preview+json' --data '{"event_type": "deployment"}' --header 'authorization: Bearer $(GITHUB_PAT)'
```
danieljhegeman, line 235: So using the api is a bit annoying because it requires a token, whereas simply using a `git` command, a la promoting a branch and then pushing, does not (uses user's local ssh creds). I would tend to favor the git cli approach instead for this reason.

danieljhegeman (commented):
A bit skeptical of introducing the need for a personal access token unless it is truly necessary; see suggestion

.github/workflows/promote-staging-to-prod.yml
```diff
@@ -0,0 +1,77 @@
...
+
+          echo "Latest commit on 'prod' branch is: $(git log -1 --format='%H' prod)"
+          echo "Latest commit on 'staging' branch is: $(git log -1 --format='%H' staging)"
+
+          echo "About to merge 'staging' branch into 'prod' branch"
+          git merge --verbose staging -m "Merging staging branch into prod branch"
+
+          git push origin prod
+
+      # We need to use repository_dispatch because by default github actions
```
Bento007, line 55: Did you consider [reusing](https://docs.github.com/en/actions/using-workflows/reusing-workflows) the build-image-for-prod workingflow here? It would keep everything connect within the same summary, and reduce our dependency on third party libraries.

Makefile
Bento007: If it's simple, go for it, otherwise save it for a future improvement.

.github/workflows/promote-staging-to-prod.yml
```diff
@@ -0,0 +1,77 @@
...
+          event-type: build-images-for-prod
+          client-payload: '{"ref": "refs/heads/prod"}'
+
+      - name: Send slack notification if staging not merged into prod
+        if: failure()
+        uses: 8398a7/action-slack@v3
+        with:
+          status: ${{ job.status }}
+          fields: repo,commit,author,eventName,workflow,job,mention
+          mention: "subteam^S02KZL4SQM6"
```
Bento007, line 74: who does this notify and where do you find this magic?

.github/workflows/promote-staging-to-prod.yml
```diff
@@ -0,0 +1,77 @@
...
+
+          echo "Checking out 'prod' branch"
+          git checkout prod
+          echo "Confirming checked out branch receiving merge is: $(git branch --show-current)"
+
+          echo "Latest commit on 'prod' branch is: $(git log -1 --format='%H' prod)"
+          echo "Latest commit on 'staging' branch is: $(git log -1 --format='%H' staging)"
+
+          echo "About to merge 'staging' branch into 'prod' branch"
+          git merge --verbose staging -m "Merging staging branch into prod branch"
```
danieljhegeman, line 51: this operation is what should be done in the Makefile rather than bothering to use a personal access token to use the github API to call this GHA

danieljhegeman, line 51: this operation is what should be done in the Makefile rather than bothering to use a personal access token to use the github API to call this GHA

danieljhegeman (changes requested):

.github/workflows/promote-staging-to-prod.yml
nayib-jose-gloria: it notifies the sc-oncall-eng slack group, and I got it from Joyce's work in [this](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6413) PR updating the slack notifications

.github/workflows/promote-staging-to-prod.yml
nayib-jose-gloria: I like having the GHA set-up because I like having the UI button push option, but I'll factor this out into a script that both command line + UI can use

danieljhegeman: 🙏 yeah I think having the GHA is good because the UI button feature is nice, but I would replace the `make` recipe with a regular `git`-based operation so that we don't introduce the personal access token dependency

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,31 @@
...
+  echo "All checks passed in most recent staging deployment!"
+else
+  echo "Some checks failed in most recent staging deployment. Please check GHA runs."
+  exit 1
+fi
+
+echo "fetch origin branch history"
+git fetch
+echo "Stash any local changes"
+git stash
```
danieljhegeman, line 14: I don't totally like having this script mutate the local working tree. I think, in order of preference:
1) fail with message if the working tree is not clean
or
2) pop the stash at the end after stashing

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Checking out 'staging' branch and pull latest"
+git checkout staging
+git pull origin staging
+echo "Checking out 'prod' branch and pull latest"
+git checkout prod
+git pull origin prod
```
danieljhegeman, line 20: We often have hiccups on staging and prod branches due to hot fixes and (accidental) merge commits -- a `pull` might fail for either of these branches. I think it might be better to delete the staging/prod branch and create anew from origin after a `fetch`.

.github/workflows/promote-staging-to-prod.yml
danieljhegeman: Resolved with added script 👍

danieljhegeman (changes requested):

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+checks=$(curl -s  "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage") | .conclusion' | head -n 1)
```
Bento007, line 3: This is almost identical to `Check if Latest Staging Post-Deploy Tests are Passing`. You could consolidate by moving it into it's own bash script that takes the -H parameter.

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,34 @@
...
+  echo "Some checks failed in most recent staging deployment. Please check GHA runs."
+  exit 1
+fi
+
+if [[ $(git diff HEAD) ]]; then
+  echo "Local has uncommitted changes, please commit or stash and try again."
+  exit 1
+fi
+
+echo "fetch origin branch history"
```
Bento007, line 16: L16-L34 could also be used in the GHA. I'd like to avoid having these processes evolve separately.

Bento007 (changes requested):
There is some room for some easy code consolidation.

Bento007 (approved):

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+if [ ! -z $1 ]; then
+  checks=$(curl -s -H "Authorization: Bearer $1" "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage") | .conclusion' | head -n 1)
```
danieljhegeman, line 4: what is `czibuildbot` exactly, and why are we including this as a query param? Does it narrow the workflows returned to only happy deploys (i.e. is this the only action run by that actor?)?

danieljhegeman (changes requested):
@nayib-jose-gloria a couple items:
1. The local script does not confirm `head_sha` as being the correct hash before proceeding with the merge locally (and pushing). I think we should add that to prevent race conditions.
2. *scope creep idea*: Should we consider deploying _up to_ the latest _passing_ commit? Edit: this is a terrible idea given that we're still iterating on things. Ignore this.

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,48 @@
...
+else
+  checks=$(curl -s  "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage") | .conclusion' | head -n 1)
+fi
+
+if [ "$checks" == "success" ]; then
+  echo "All checks passed in most recent staging deployment!"
+else
+  echo "Some checks failed in most recent staging deployment. Please check GHA runs."
```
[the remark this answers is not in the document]
nayib-jose-gloria: I think we should still fail + encourage the on-call to check out the runs themselves in that case. So I will keep the behavior but update the message to be more clear about the possible scenarios.

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+if [ ! -z $1 ]; then
+  resp=$(curl -s -H "Authorization: Bearer $1" "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage")')
+else
+  resp=$(curl -s  "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage")')
+fi
+
+checks=$(echo "$resp" | jq -r '.conclusion' | head -n 1)
```
danieljhegeman, line 9: move the `head -n 1` operation back up to the end of the `curl` cmd, where it was before. We just want the `.conclusion` attribute grab on this line.

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,61 @@
...
+gha_head_sha=$(echo "$resp" | jq -r '.head_sha' | head -n 1)
```
danieljhegeman, line 10: same as above

danieljhegeman (changes requested):
almost...

scripts/promote_staging_to_prod.sh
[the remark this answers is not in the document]
danieljhegeman: Yep, perfect

scripts/promote_staging_to_prod.sh
danieljhegeman: I'm a little wary of using that as the selector. Can we make use of [this](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#:~:text=List%20all%20workflow%20runs%20for%20a%20workflow.%20You%20can%20replace%20workflow_id%20with%20the%20workflow%20file%20name.%20For%20example%2C%20you%20could%20use%20main.yaml.%20You%20can%20use%20parameters%20to%20narrow%20the%20list%20of%20results.%20For%20more%20information%20about%20using%20parameters%2C%20see%20Parameters.) instead? Use the name of the happy deploy workflow file.

scripts/promote_staging_to_prod.sh
nayib-jose-gloria: this should work as-written, the `.conclusion` will grab the conclusion attribute on everything in the response list, then `head -n 1` will grab the first item in the list (most recent)

danieljhegeman: Ah, sorry, my mistake. My suggestion was wrong. I agree it works as written but it is a bit indirect. My intended suggestion:

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+if [ ! -z $1 ]; then
+  resp=$(curl -s -H "Authorization: Bearer $1" "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage")')
```
danieljhegeman, line 4:
```suggestion
  resp=$(curl -s -H "Authorization: Bearer $1" "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '[.workflow_runs[] | select(.display_title == "stage")][0]')
```

scripts/promote_staging_to_prod.sh
```diff
@@ -0,0 +1,61 @@
...
+else
+  resp=$(curl -s  "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '.workflow_runs[] | select(.display_title == "stage")')
```
danieljhegeman, line 6:
```suggestion
  resp=$(curl -s  "https://api.github.com/repos/chanzuckerberg/single-cell-data-portal/actions/runs?actor=czibuildbot" | jq -r '[.workflow_runs[] | select(.display_title == "stage")][0]')
```

danieljhegeman: and then remove `head -n 1` from the code below

danieljhegeman (changes requested):
suggested a more logical structure for the parsing of the curl resp. Also consider renaming `resp` var to `most_recent_deploy_run` or something.

scripts/promote_staging_to_prod.sh
nayib-jose-gloria: totally missed that you could use the workflow file name instead of the ID, good find

danieljhegeman (approved):
Lgtm. Thanks @nayib-jose-gloria

Issue #6403, opened 2024-01-02
Implement local disk caching of WMG snapshot


atarashansky:
We want to implement a scheme to cache the tiledb data files closer to the application containers.

- [Experiments](https://github.com/chanzuckerberg/single-cell-data-portal/issues/6146#issuecomment-1834402991) suggest that caching the tiledb data files on host (or close to the host) on which the web app container is running is much faster than querying S3 for the tiledb datafiles
- [More experiments](https://github.com/chanzuckerberg/single-cell-data-portal/issues/6146#issuecomment-1868030271) suggest that [mountpoint-s3](https://github.com/awslabs/mountpoint-s3) gives performance close enough to local disk caching and gives transparent access to the tiledb datafiles on S3 via a mount point on the local filesystem and thereby removes a lot of application complexity due to copying (see below).

What caching means depends on the compute infrastructure the application is running. Currently, we are using [FARGATE](https://docs.aws.amazon.com/AmazonECS/latest/userguide/what-is-fargate.html). This means that the EC2 instances that runs the web app containers are hidden from us.

Currently, the solution involves the following considerations:

1. With FARGATE the host machines that run the application container can change and has less privilege than using raw EC2 instances. Moreover, with FARGATE launch type, we cannot specify the [docker volumes on the ec2 local disk as is the case with ec2 launch type](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_data_volumes.html). This means that we would need to store the tiledb datafiles on [EFS](https://docs.aws.amazon.com/AmazonECS/latest/userguide/using_data_volumes.html). Note that [EFS is a Network File System](https://aws.amazon.com/efs/faq/), which is different from the local disk of an EC2 instance. However, we expect EFS to perform better than S3.
2. Assuming EFS volume to store the datafiles, is leveraging [ mountpoint-s3](https://github.com/awslabs/mountpoint-s3) more robust than copying the data onto EFS?
3. If mountpoint-s3 is not an option, at what point in the initialization phase of the application do we copy the tiledb data files to an EFS mount point? Specify a script that does the copy in the ECS task definition? Have the application code copy the data files in a separate thread?
4. Given that WMG reader is pinned to a data schema version should we only copy the tiledb datafiles pinned by the reader? If so, it might make more sense to spawn a separate thread to copy the data to EFS since the ECS task definition would not know which version of the data to copy. If copying all the version of the data from s3, then it might be possible to have the copying specified in the ECS task definition.
[...]

PR #6447, opened 2024-01-09
feat: Download WMG snapshot to Fargate ephemeral storage on task startup

prathapsridharan:
## Reason for Change

To implement local disk caching of WMG snapshot (#6403), this PR tackles the part of downloading the WMG data snapshot from S3 to the [FARGATE ephemeral storage](https://docs.aws.amazon.com/AmazonECS/latest/userguide/using_data_volumes.html). The data is currently about 27 GB.

## Changes

- Modify ECS task definition to specify ephemeral storage size and environment variables needed for access at container initialization time
- Modify container initialization script to start a download the WMG data snapshot.

## Testing steps

- Manual testing on rdev

## Checklist 🛎️

## Notes for Reviewer

Currently the download is done serially (that is there is no parallel download). Once the serial download has been shown to work, we will optimize by implementing parallel download to speed up download time


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -154,6 +161,10 @@ module backend_service {
   dataset_submissions_bucket = local.dataset_submissions_bucket
   datasets_bucket            = local.datasets_bucket
   execution_role             = local.ecs_execution_role
+
+  # Bump health_check_interval from 15 seconds to 30 seconds so that WMG snapshot download,
+  # which at the time of this writing is around 27GB, has time to complete.
+  health_check_interval      = 30
```
atarashansky, line 167: is 30s enough to download 27GBs?

Should we bump this to 60s to be safe? Having a long interval doesn't seem like it should matter much. Although if the task dies, it means that it would need to wait a little bit longer before restarting the task, because i think a failing health check is what triggers that.. Not sure - I'd like @ebezzi's input here.

Alternatively, is there a parameter that specifies a delay before the health check starts pinging in 15s intervals? That way we can add a one minute delay before health check starts and then ping at the standard 15s afterwards.

backend/wmg/api/v2.py
```diff
@@ -173,6 +184,7 @@ def markers():
     snapshot: WmgSnapshot = load_snapshot(
         snapshot_schema_version=WMG_API_SNAPSHOT_SCHEMA_VERSION,
         explicit_snapshot_id_to_load=WMG_API_FORCE_LOAD_SNAPSHOT_ID,
+        snapshot_local_disk_path=SNAPSHOT_LOCAL_DISK_PATH,
```
atarashansky, line 187: TBH I'm not a huge fan of the required arguments for load_snapshot. Would it be possible to put these defaults in the kwargs of the function definition? That way we could run `load_snapshot()` without needing all the extra verbosity? I only bring this up here because we're adding yet another required argument. To replicate deployment's behavior in a local kernel now requires importing or hardcoding all these extra constants and it's a bit annoying.

backend/wmg/data/snapshot.py
```diff
@@ -103,100 +103,103 @@ def load_snapshot(
...
-    cell_type_ancestors = _load_cell_type_ancestors(snapshot_dir_path)
-    dataset_metadata = _load_dataset_metadata(snapshot_dir_path)
+def _load_snapshot(
+    *, snapshot_schema_version: str, snapshot_id: str, snapshot_local_disk_path: Optional[str] = None
+) -> WmgSnapshot:
+    snapshot_rel_path = _get_wmg_snapshot_rel_path(snapshot_schema_version, snapshot_id)
 
-    snapshot_base_uri = _get_wmg_snapshot_dir_s3_uri(snapshot_dir_path)
-    logger.info(f"Loading WMG snapshot from directory path {snapshot_dir_path} into URI {snapshot_base_uri}")
+    cell_type_orderings = _load_cell_type_order(snapshot_rel_path, snapshot_local_disk_path)
```
atarashansky, line 185: Can we add some more comments explaining _rel_path vs local_disk_path vs full_path and how these values logically interact with each other? If this already exists, apologies in advance.

backend/wmg/data/snapshot.py
```diff
@@ -259,20 +303,31 @@ def _should_reload_snapshot(
...
     """
     snapshot_id = explicit_snapshot_id_to_load or _get_latest_snapshot_id(
-        snapshot_schema_version=snapshot_schema_version, read_versioned_snapshot=read_versioned_snapshot
+        snapshot_schema_version, snapshot_local_disk_path
     )
 
     if cached_snapshot is None:
         logger.info(f"Loading snapshot id: {snapshot_id}")
         return (True, snapshot_id)
+    # IMPORTANT NOTE: Updating cached_snapshot when snapshot_identifer value of
```
atarashansky, line 320: - Under what conditions do we expect the deployment to fallback onto S3?
 - If we change S3's latest_snapshot_identifier, does the backend recognize that if we are still reading from local disk? Or does the backend only detect a change in S3's latest_snapshot_identifier if we are in fallback mode?

To rollback the snapshot when the deployment is in local disk mode, do we need to shell into the instance and edit the latest_snapshot_identifier file on disk?

atarashansky (changes requested):

.happy/terraform/modules/ecs-stack/main.tf
ebezzi: We tested this and this is a fairly conservative interval. Increasing this has the side effect that deployments will generally take longer, so I would only do that if we observe failures. We have enough logs to understand, for instance, if the download took longer on that specific instance (but it's an internal AWS download, so the speed should be somewhat predictable).

There is no parameter to specify a delay, as far as I know.

prathapsridharan: @atarashansky - 30 seconds was tested and has never failed so far. Health check fails if there are 10 consecutive pings to port 80 that fails where each ping is done 30 seconds apart. So as long as the server is can respond to the ping in 300 seconds (10 x 30 seconds), it should be good.

Meanwhile, observing the logs, I see that we are getting at a minimum 150 MiB/s and typically 220 MiB/s when downloading. So download of 27GB takes no longer than 180 seconds (27000/150).

I am not against incrementing the healthcheck interval if you and @ebezzi feel that is still cutting it too close

atarashansky: 300s is perfect - I wasn't aware of the 10 consecutive pings required. Resolving this thread!

backend/wmg/data/snapshot.py
prathapsridharan: @atarashansky - 

- The only time the app will fallback to S3 is if `_local_disk_snapshot_is_valid()` returns `False`
- If `_local_disk_snapshot_is_valid()` returns `True`, the backend _will always read from local disk_. However, changes to `latest_snapshot_identifier` on S3 could affect what snapshot_id the app reads from local disk, if circumstances are such that the `snapshots` directory is downloaded onto the local disk _after `latest_snapshot_identifier` is changed on S3_.

This might happen if we manually redeploy after changing `latest_snapshot_identifier` on S3 - That is expected.

But a re-download of the `snapshots` directory would also happen if Fargate preempts a task and reschedules it on a new node. If prior to this reschedule, we update the `latest_snapshot_identifer` on S3, then this is reflected on the local disk. This has to do with the concern about data consistency and the [comments in this ticket](https://github.com/chanzuckerberg/single-cell-data-portal/issues/6403#issuecomment-1879432598) talk about it.

- To rollback when app is in local disk mode, nothing needs to change. Since we download the _entire `snapshots` directory_, we simply rollback by changing the schema version and/or snapshot_id in [wmg_api_config.py](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/wmg/api/wmg_api_config.py)

backend/wmg/data/snapshot.py
prathapsridharan: @atarashansky - I will add that and change the naming to be clearer. I agree that this can be confusing

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -141,10 +144,14 @@ module backend_service {
...
   task_role_arn              = local.ecs_role_arn
   service_port               = 5000
+
+  # 100gb of disk storage allocated for the task running backend container 
+  task_storage_size_gb       = 100
   memory                     = var.backend_memory
   cpu                        = var.backend_cpus * 1024
   cmd                        = local.backend_cmd
   deployment_stage           = local.deployment_stage
+  wmg_disk_cache_path        = "/tmp/wmg_disk_cache"
```
ebezzi, line 154: I'm not sure this is relevant in Docker, but Linux generally has cleanup processes that can remove files in `/tmp/`. I suggest another location (especially since the cache here isn't really temporary).

.happy/terraform/modules/ecs-stack/variables.tf
```diff
@@ -35,6 +35,18 @@ variable happy_config_secret {
   description = "Happy Path configuration secret name"
 }
 
+variable "wmg_disk_cache_path" {
+  type        = string 
+  description = "Absolute path of directory on the local disk where WMG data snapshot is cached"
+  default     = "/tmp/wmg_disk_cache"
```
ebezzi, line 41: Doesn't seem necessary - no one is calling this right now.

.happy/terraform/modules/service/main.tf
```diff
@@ -179,6 +182,10 @@ resource aws_ecs_task_definition task_definition {
         "name": "DATASETS_BUCKET",
         "value": "${var.datasets_bucket}"
       },
+      {
+        "name": "WMG_DISK_CACHE_PATH",
```
ebezzi, line 186: This is a bit messy - you're adding a variable to a module that is not specifically tied to wmg (in fact, it is also used by the frontend). Unless you have specific reasons for keeping this parametric, I wouldn't mind hardcoding this in the dockerfile.

backend/wmg/api/v2.py
prathapsridharan: @atarashansky - Remember that `load_snapshot()` is used by the pipeline too. Setting `snapshot_local_disk_path=SNAPSHOT_LOCAL_DISK_PATH` as the in `def load_snapshot(...)` wouldn't make sense since we wouldn't want the pipeline run to be reading from local disk.

_We want to keep the API decoupled from the pipeline (decoupling the reader from the writer)_.

That is why the function declaration in `backend/wmg/data` looks like this:

```
def load_snapshot(
    *,
    snapshot_schema_version: str,
    explicit_snapshot_id_to_load: Optional[str] = None,
    snapshot_local_disk_path: Optional[str] = None,
) -> WmgSnapshot
```

I think specifying defaults in the function declaration that pin it to values for WMG API violates the decoupling between the API and the pipeline.

I understand your point about having to specify these values being annoying but there is a straightforward fix for this:

Make wrapper functions for the reader and writer that wrap `load_snapshot(...)`. These wrapper functions would have sensible defaults for the API and pipeline and call pass them the call to `load_snapshot(...)`.

```
# WMG API

def api_load_snapshot(
    snapshot_schema_version=WMG_API_SNAPSHOT_SCHEMA_VERSION,
    explicit_snapshot_id_to_load=WMG_API_FORCE_LOAD_SNAPSHOT_ID,
    snapshot_local_disk_path=SNAPSHOT_LOCAL_DISK_PATH):

    return load_snapshot(...)
```

```
# WMG Pipeline

def pipeline_load_snapshot(snapshot_schema_version=WMG_DATA_SCHEMA_VERSION):
    return load_snapshot(...)
```

This way, API component callers can call `api_load_snapshot()` without ever having to supply arguments since there will always sensible defaults going forward. Similarly, pipeline component callers can call `pipeline_load_snapshot()` without supplying arguments 

If you buy my reasoning and this approach, I propose we do this in another PR rather than in this PR because several call sites would need to change.

atarashansky: We can add this in a separate PR. I'm onboard with the proposed solution!

.happy/terraform/modules/ecs-stack/main.tf
prathapsridharan: @ebezzi - will fix

.happy/terraform/modules/service/main.tf
prathapsridharan: @ebezzi - lol! That is what I had originally and changed to passing an environment variable. I had the same qualms as you but now that you mention it I will remove it entirely

.happy/terraform/modules/ecs-stack/variables.tf
prathapsridharan: @ebezzi - I removed this

ebezzi (approved):

atarashansky (approved):

PR #6425, opened 2024-01-05
chore: add python to dependabot tracker

Bento007:
## Reason for Change

- So keeping are dependencies up-to-date is easier.

## Changes

- add dependabot configuration to track python requirements.
- move dependencies to their own directory to be compatible with dependabot
- update dockerfiles to use new requirements path.

## Notes for Reviewer
- depends on #6424

PR #6481, opened 2024-01-09
fix: db/remote_migration_init

Bento007:
## Reason for Change

- fixing the make recipe db/remote_migration_init
- the break was caused by https://github.com/chanzuckerberg/single-cell-data-portal/pull/6425


prathapsridharan (approved):
lgtm

Issue #6169, opened 2023-11-06
CI/CD Workflow v2


atolopko-czi:
TODO: Write proposal to explain the proposed new workflow, as depicted in this [diagram](https://lucid.app/lucidchart/ca8f0bb7-f0d8-4fd0-88ab-21cb54dcc05a/edit?viewport_loc=-776%2C-108%2C2465%2C1289%2C0_0&invitationId=inv_515f1f03-f63f-4098-9630-f615cecc719b)

Bento007:
We will likely want to make use of [GHA reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)

PR #6490, opened 2024-01-10
chore: require push-test to pass before running e2e tests, 3 of 3

Bento007:
## Reason for Change

- To speed up the testing cycle. E2E tests are skipped if unit tests don't pass.
- related to https://github.com/chanzuckerberg/single-cell-data-portal/issues/6169

## Changes

- modify push-test so it's callable from other workflows
- call push-test GHA from update-rdev-for-PR GHA before running E2E tests.
- fix build-images failing randomly by moving seed-wmg-cellguide-rdev to depend on build-images

## Notes for Reviewer
new workflow
Screenshot 2024-02-16 at 2 08 05 PM


.github/workflows/push-tests.yml
```diff
@@ -1,12 +1,12 @@
 name: Push Tests
 
 on:
-  pull_request:
-    branches:
-      - "*"
   push:
     branches:
       - "main"
+      - "prod"
```
nayib-jose-gloria, line 7: do we want push tests to run in prod? we only promote if push tests pass on that commit in main, so it seems redundant

Bento007: I included this originally because we were making pull requests from main->prod which would have triggered the push test. So in order to keep this running as it was before I should add 
```
on:
   pull-requests:
      branches:
        - "prod"
```
However the new workflow for deploying prod uses a GHA called via workflow which would set off the pull-request trigger. Adding "prod" to push would make sure the push-tests run on prod at least one. But as you said this is probably redundant. I'm happy to remove if that is the consensus.

nayib-jose-gloria: my vote is to remove, but I don't mind if you want to ask around and get a consensus. I think a lot of people are out today, though

Bento007, line 7:
```suggestion
```

danieljhegeman (commented):
First concern: do we want to wait on rdev update completion _before_ running push tests?

Bento007:
> First concern: do we want to wait on rdev update completion _before_ running push tests?

We want push tests to run early because they are quick and indicate functionality.

What is your concern?

.github/workflows/push-tests.yml
```diff
@@ -1,13 +1,12 @@
 name: Push Tests
 
 on:
-  pull_request:
-    branches:
-      - "*"
```
danieljhegeman, line 6: this caused push tests to be run previously. Now we're only triggering on push to main and in the rdev update script. So these push tests will not be run _as quickly as possible_.

Bento007: When else should it running?

danieljhegeman: Why would we move to running push tests _later_, in the rdev, as opposed to what we're doing now? Can't we push a result to some location/env var that the rdev update job can read and use to determine its flow?

Bento007: It's not later. It runs at the exact same time. The only difference is it's now part of the rdev workflow and deploying rdev.
Screenshot 2024-02-16 at 2 03 07 PM

danieljhegeman: I see what you're saying. Great!

danieljhegeman (approved):
Cool! I think it's time for an update to the name of the rdev workflow, though. ✅

Issue #5921, opened 2023-10-05
feat( ): Filter out ribosomal and mitochondrial genes from returned marker gene lists


sidneymbell:
## Need

Our marker gene lists often contain many ribosomal and mitochondrial genes, which are less informative. Sometimes there are so many of them that it's hard to see what "real" marker genes are similar / different for each cell type. 

Big thank you to @jychien for surfacing this issue and providing context:
> Ribo and mito genes are usually used already to filter these cells that make it into the corpus. So, ideally, that means mito and ribo genes are moreso indicative of cell state than really marker genes. This would probably apply to general cell cycle genes as well. Maybe if there was a way to toggle those genes on/off.

>Traditionally, mito and ribo genes are being used as a [QC filter](https://satijalab.org/seurat/articles/pbmc3k_tutorial#qc-and-selecting-cells-for-further-analysis) upstream of when the data gets to the corpus. Especially with these microfluidics assays, there is a lot of background that not everyone has uniformly subtracted out, and these genes are not informative and not typically seen as marker genes and moreso reflect cell state or cycle.  So, it’s extra noise that would be nice to toggle on/off, as there is always exceptions to the rule.

## Approach

**Describe the solution you'd like**
Curate a list of mitochondrial and ribosomal genes, for human and mouse. Filter these out of the marker gene list before returning the top N genes.

## Definition of Done

    • Mitochondrial and ribosomal genes are no longer returned as "marker genes" in the UI.
- @sidneymbell has validated the updated JSON of marker genes before it hits prod

sidneymbell:
Timing note: if the JSONs are available to validate by Tues 12/19, then I can validate them and we'll get this out before the holidays. Otherwise, let's move it to the sprint of Jan 9-22 once I'm back in the office. Thanks!

PR #6504, opened 2024-01-12
feat: blacklist mitochondrial and ribosomal proteins from computational marker genes

atarashansky:
## Reason for Change

- #5921 

## Changes

- added a fixture containing Ensembl v104 gene IDs annotated by ribosomal and mitochondrial GO terms (see comment in code).
- remove genes from the unfiltered marker gene dataframe matching any of these genes

## Testing steps

- Unit test fixtures updated
- RDEV runs will be triggered to generate new snapshots for inspection in rdev.

## Notes for Reviewer
Merging BLOCKED until @sidneymbell approves.


backend/cellguide/pipeline/computational_marker_genes/computational_markers.py
```diff
@@ -42,6 +43,12 @@
 
 MARKER_SCORE_THRESHOLD = 0.5
 
+# This file contains blacklisted marker genes that match the following criteria:
+# Ensembl 104 gene IDs that are annotated with the GO terms: ['GO:0005840', 'GO:0005739']
+# These GO terms correspond to the cellular components "ribosome" and "mitochondrion".
+file_dir = os.path.dirname(os.path.realpath(__file__))
+MARKER_GENE_BLACKLIST_FILENAME = os.path.join(file_dir, "marker_gene_blacklist.txt")
```
joyceyan, line 50: i assume this file was written using a script. do you have the script you used to generate this? might make sense to put it in the `scripts/` directory so it can be used to regenerate the marker gene blacklist if we decide we want to blacklist more types of marker genes

atarashansky: It wasn't written by a script as I was not aware of an easy way to get these genes through an API, though I'm sure one does exist.

joyceyan: gotcha - how did you generate the file then?

atarashansky: Biomart downloads the results as `mart_export.txt` files, one per organism, which contain newline-separated gene IDs and a header line. I skipped the first line for human, read the entire file, split by newline, unique'd the resulting list, did the same thing for mouse, concatenated the lists, joined by commas and dumped to the file.

atarashansky: Not sure if it merits extra tooling. If we ever need to do it again, then it would make more sense to make a script for it the second time around.

joyceyan: thanks for the explanation! agreed, i don't think it merits extra tooling. but it's useful to have the explanation on the PR somewhere or ideally as a code comment

joyceyan (approved):

PR #6567, opened 2024-01-22
feat: Update copy for scale constraints

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/3685

## Changes

Adds the proposed copy changes, with a few minor tweaks:
- Sets the maximum dataset size to 30GB instead of 50GB, since that is what the [code](https://github.com/chanzuckerberg/single-cell-data-portal/blob/75c648e8b277d23c4ff0fe198754e660dfd981d9/frontend/src/components/DropboxChooser/index.tsx#L29) appears to say
- Removes "in 2023" since we're doing this after 2023

## Testing steps

image

## Checklist 🛎️

- [x] Add product, design, and eng as reviewers for rdev review
- [x] For UI changes, add screenshots/videos, so the reviewers know what you expect them to see
- [n/a] For UI changes, add e2e tests to prevent regressions

### DIFF
--- a/frontend/doc-site/032__Contribute and Publish Data.mdx
+++ b/frontend/doc-site/032__Contribute and Publish Data.mdx
@@ -25,7 +25,7 @@ CELLxGENE is focused on supporting the global community attempting to create ref
 
 ### Scale Constraints
 
-CELLxGENE Discover sets the maximum dataset size for uploads to 50GB. This is enforced in the UX by a property in the [Dropbox Chooser](https://github.com/chanzuckerberg/single-cell-data-portal/blob/75c648e8b277d23c4ff0fe198754e660dfd981d9/frontend/src/components/DropboxChooser/index.tsx#L29). There is pending discovery to support larger dataset sizes. Additionally, CELLxGENE Explorer sets the maximum number of cells to 4.6 million for exploration of datasets. See [Disable exploration of datasets larger than 2 million cells](https://github.com/chanzuckerberg/single-cell-data-portal/issues/1538).
+CELLxGENE Discover sets the maximum dataset size for submissions to 50GB. Additionally, CELLxGENE Explorer sets the maximum number of cells to 4.6 million for exploration of datasets.
 
 ### Formatting Requirements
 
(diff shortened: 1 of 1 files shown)


frontend/doc-site/032__Contribute and Publish Data.mdx
```diff
@@ -23,6 +23,11 @@ CELLxGENE is focused on supporting the global community attempting to create ref
 - cell lines
 - organisms other than mouse or human
 
+### Scale Requirements
+
+- CELLxGENE Discover sets the maximum dataset size for uploads to 30GB. This is enforced in the UX by a property in the [Dropbox Chooser](https://github.com/chanzuckerberg/single-cell-data-portal/blob/75c648e8b277d23c4ff0fe198754e660dfd981d9/frontend/src/components/DropboxChooser/index.tsx#L29). There is pending discovery to support larger dataset sizes.
```
brianraymor: I would rewrite to:

> CELLxGENE Discover sets the maximum dataset size to 50GB.

There is a 30 GB limitation in the Curation UX (enforced by the dropbox chooser) and a 50 GB limitation in the Discover API, but both cases are implementation details that apply to curators and should not be generally documented. @Bento007 is a good contact about the details if you're curious. 

Also, here's what Lattice shares with contributors in their "first contact" emails:

**Scale limitations**

H5AD files must be less than 50 GB for submission

Datasets with more than 4.3M observations will not have visualizations available, but data will be accessible via download and in CELLxGENE features, such as Census

Datasets with more than 2^31 -1 non-zero elements in a single data layer will not have RDS files available (per R dgCMatrix sparse matrix class limitations), but data will be accessible via visualization, H5AD download and in CELLxGENE features, such as Census

joyceyan: Thanks for the context! Updated.

brianraymor (approved):

Issue #6416, opened 2024-01-04
[Analytics] Test app <> Plausible connection


tihuan:
1. Ticket 1: Stop gap for next sprint -
    1. Check with KNI and Joyce on how the analytics broke
    2. Write a test to cover the root cause

Write an e2e test for:
1. Trigger an event through user action
1. Verify the payload(?) that is sent to Plausible
1. Check for response code for the network call
1. Condition to run rdev/staging

tihuan:
@joyceyan I probably need some help on getting the context on how the Plausible connection broke while KNI worked on the homepage 🙏

I think the goal of this ticket is for me to see if I could write test or add a checklist item to ensure Plausible works every PR!

Thanks so much!

joyceyan:
@tihuan Originally, the plausible proxy was set up in Cloudfront manually. Sometimes, if we're deploying an infra change, we may apply a Terraform change with the prod deploy, which will reapply Cloudfront changes based on what we have in `single-cell-infra/terraform/modules/hosted-cellxgene/cloudfront.tf`. We manually resolved this by reapplying the Cloudfront config manually again. [This PR](https://github.com/chanzuckerberg/single-cell-infra/pull/961/files) from Emmanuele updated Terraform so that if we reapply Cloudfront configs through Terraform, it will include the Plausible proxy.

I don't expect this to generally be a problem that's "our fault" in the future, but if Cloudfront or something goes down which causes the proxy to go down, we now have a Datadog alert [here](https://app.datadoghq.com/monitors/135807689?view=spans) that will alert us if the proxy is down (aka if https://cellxgene.cziscience.com/js/script.js 404's). I think this alert is sufficient for informing us of outages and I don't think we need any additional work.

tihuan:
Will close this issue now, since the Datadog test seems sufficient 👍 Thanks so much, @joyceyan !

tihuan:
Actually I'll probably add an e2e test to ensure a Plausible event does register a successful API call to Plausible in staging (since the proxy script is not downloaded locally nor rdev)

PR #6568, opened 2024-01-22
test: Test analytics event

tihuan:
## Reason for Change

- #6416

## Changes

1. Add a **STAGING ONLY** test to assert that analytics event is indeed sent out with expected payload
2. TIL: Playwright driven browser will automatically set `navigator.webdriver` to `true`, which Plausible script uses to turn off  event logging. So in this particular test, I temporarily change `navigator.webdriver` to `false`, so the API call will go out

## Testing steps

1. The new e2e test should pass

## Checklist 🛎️


frontend/tests/features/home/homePage.test.ts
```diff
@@ -131,6 +132,59 @@ describe("Homepage", () => {
...
+        request.url().includes("api/event") && request.method() === "POST";
+
+        const payload = JSON.parse(request?.postData() || "{}");
+
+        const { n, p } = payload;
+
+        return n === "WMG_CLICKED" && p.button === "See how it works";
+      }),
+      page.getByText(WMG_CTA_TEXT).click(),
+    ]);
```
kaloster, line 186: 👏🏽  nice!

kaloster (approved):
LGTM 🚀

Issue #6484, opened 2024-01-10
Create set collection dataset order API endpoint


MillenniumFalconMechanic:
#### Need
Custom ordering of collection datasets must be maintained on the backend.

#### Approach
See [tech design](https://docs.google.com/document/d/1hUeXtmiITKU7kCBZPYjsHYy4CQdzTsNe2Ltg8ojgQ7U/edit#heading=h.g46vnzhs4nlc).

PR #6579, opened 2024-01-25
feat: order-datasets DP API endpoint

MillenniumFalconMechanic:
## Reason for Change

- #6484

## Changes

- Added database migration to add `datasets_custom_ordered` column.
- Added `datasets_custom_ordered` to CollectionVersion
- Added `order-datasets` endpoint.

## Testing steps

- Added tests to [test_portal_api.py](https://github.com/chanzuckerberg/single-cell-data-portal/compare/mim/6484-save-dataset-order?expand=1#diff-09a58764042645262735c766e7d4656104dcab6eda35bc3d82f641fe42efe581) and [test_business.py](https://github.com/chanzuckerberg/single-cell-data-portal/compare/mim/6484-save-dataset-order?expand=1#diff-98efcfe189db84e8fdd45833310cbcf9bcd4debb00f0e1aa2e89581034980fed).
- Tested in rdev
  - **Valid reorder:** `curl -X PUT --cookie _oauth2_proxy=${oauth2_proxy} --cookie cxguser=${cxguser} -H 'Content-Type: application/json' -d '{"datasets":["30362c1c-96e1-44ef-b0dd-6a783ae9a40d", "6a7a0fad-101f-468e-9938-f446b743d113"]}' https://pr-6579-backend.rdev.single-cell.czi.technology/dp/v1/collections/1b3d1248-5c47-45f3-8295-0d98dbf0af8f/order-datasets`
  - **Invalid dataset length:** `curl -X PUT --cookie _oauth2_proxy=${oauth2_proxy} --cookie cxguser=${cxguser} -H 'Content-Type: application/json' -d '{"datasets":["30362c1c-96e1-44ef-b0dd-6a783ae9a40d"]}' https://pr-6579-backend.rdev.single-cell.czi.technology/dp/v1/collections/1b3d1248-5c47-45f3-8295-0d98dbf0af8f/order-datasets`
  - **Invalid dataset ID:** `curl -X PUT --cookie _oauth2_proxy=${oauth2_proxy} --cookie cxguser=${cxguser} -H 'Content-Type: application/json' -d '{"datasets":["30362c1c-96e1-44ef-b0dd-6a783ae9a40d", "invalid_id"]}' https://pr-6579-backend.rdev.single-cell.czi.technology/dp/v1/collections/1b3d1248-5c47-45f3-8295-0d98dbf0af8f/order-datasets`

### DIFF
--- a/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
+++ b/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
@@ -5,6 +5,7 @@ Revises: 04_2f30f3bcc9aa
 Create Date: 2024-01-25 11:42:46.126701
 
 """
+
 import sqlalchemy as sa
 from alembic import op
 
@@ -18,7 +19,7 @@ depends_on = None
 def upgrade():
     op.add_column(
         "CollectionVersion",
-        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=False, server_default=sa.sql.expression.false()),
+        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=True, server_default=sa.sql.expression.false()),
         schema="persistence_schema",
     )
 
--- a/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
+++ b/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
@@ -1,4 +1,4 @@
-"""add datasets custom ordered to collection version
+"""add custom dataset order
 
 Revision ID: 05_67c4015bce73
 Revises: 04_2f30f3bcc9aa
@@ -18,10 +18,10 @@ depends_on = None
 def upgrade():
     op.add_column(
         "CollectionVersion",
-        sa.Column("datasets_custom_ordered", sa.Boolean(), nullable=False, server_default=sa.sql.expression.false()),
+        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=False, server_default=sa.sql.expression.false()),
         schema="persistence_schema",
     )
 
 
 def downgrade():
-    op.drop_column("CollectionVersion", "datasets_custom_ordered", schema="persistence_schema")
+    op.drop_column("CollectionVersion", "custom_dataset_order", schema="persistence_schema")
--- a/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
+++ b/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
@@ -19,7 +19,7 @@ depends_on = None
 def upgrade():
     op.add_column(
         "CollectionVersion",
-        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=True, server_default=sa.sql.expression.false()),
+        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=True),
         schema="persistence_schema",
     )
 
--- a/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
+++ b/backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
@@ -19,10 +19,10 @@ depends_on = None
 def upgrade():
     op.add_column(
         "CollectionVersion",
-        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=True),
+        sa.Column("has_custom_dataset_order", sa.BOOLEAN(), nullable=True),
         schema="persistence_schema",
     )
 
 
 def downgrade():
-    op.drop_column("CollectionVersion", "custom_dataset_order", schema="persistence_schema")
+    op.drop_column("CollectionVersion", "has_custom_dataset_order", schema="persistence_schema")
--- a/backend/database/versions/05_67c4015bce73_add_datasets_custom_ordered_to_.py
+++ b/backend/database/versions/05_67c4015bce73_add_datasets_custom_ordered_to_.py
@@ -1,0 +1,27 @@
+"""add datasets custom ordered to collection version
+
+Revision ID: 05_67c4015bce73
+Revises: 04_2f30f3bcc9aa
+Create Date: 2024-01-25 11:42:46.126701
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "05_67c4015bce73"
+down_revision = "04_2f30f3bcc9aa"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "CollectionVersion",
+        sa.Column("datasets_custom_ordered", sa.Boolean(), nullable=False, server_default=sa.sql.expression.false()),
+        schema="persistence_schema",
+    )
+
+
+def downgrade():
+    op.drop_column("CollectionVersion", "datasets_custom_ordered", schema="persistence_schema")
--- a/backend/layers/persistence/orm.py
+++ b/backend/layers/persistence/orm.py
@@ -12,7 +12,6 @@ mapper_registry = registry(metadata=metadata)
 
 @mapper_registry.mapped
 class CollectionTable:
-
     __tablename__ = "Collection"
 
     id = Column(UUID(as_uuid=True), primary_key=True)
@@ -24,7 +23,6 @@ class CollectionTable:
 
 @mapper_registry.mapped
 class CollectionVersionTable:
-
     __tablename__ = "CollectionVersion"
 
     id = Column(UUID(as_uuid=True), primary_key=True)
@@ -37,11 +35,11 @@ class CollectionVersionTable:
     created_at = Column(DateTime)
     schema_version = Column(String)
     datasets = Column(ARRAY(UUID(as_uuid=True)))
+    datasets_custom_ordered = Column(BOOLEAN)
 
 
 @mapper_registry.mapped
 class DatasetTable:
-
     __tablename__ = "Dataset"
 
     id = Column(UUID(as_uuid=True), primary_key=True)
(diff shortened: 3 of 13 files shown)


backend/layers/persistence/persistence.py
```diff
@@ -114,23 +114,26 @@ def _row_to_collection_version(self, row: Any, canonical_collection: CanonicalCo
             created_at=row.created_at,
             schema_version=row.schema_version,
             canonical_collection=canonical_collection,
+            datasets_custom_ordered=row.datasets_custom_ordered,
         )
 
     def _row_to_collection_version_with_datasets(
         self, row: Any, canonical_collection: CanonicalCollection, datasets: List[DatasetVersion]
     ) -> CollectionVersionWithDatasets:
+        sorted_datasets = self._sort_datasets(row, datasets)
```
MillenniumFalconMechanic, line 123: Adding sort here in the persistence layer before dataset version IDs (i.e. dataset order) are replaced with dataset versions.

backend/database/versions/05_67c4015bce73_add_datasets_custom_ordered_to_.py
```diff
@@ -0,0 +1,27 @@
...
+revision = "05_67c4015bce73"
+down_revision = "04_2f30f3bcc9aa"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "CollectionVersion",
+        sa.Column("datasets_custom_ordered", sa.Boolean(), nullable=False, server_default=sa.sql.expression.false()),
```
danieljhegeman, line 21: We're using `sa.BOOLEAN()` in all other places (a total of 2) in the migrations where we need a default bool. It's slightly different based on documentation. It's not totally clear to me which choice is best; I'd have to read more. But let's change to `sa.BOOLEAN()` for consistency (also: we know it works already 😅 ).

danieljhegeman: Actually, I'm noticing that in the other two instances we allow `nullable=True` and we do not set a `server_default` value. I'm thinking we should do this.

backend/layers/persistence/orm.py
```diff
@@ -37,11 +35,11 @@ class CollectionVersionTable:
     created_at = Column(DateTime)
     schema_version = Column(String)
     datasets = Column(ARRAY(UUID(as_uuid=True)))
+    datasets_custom_ordered = Column(BOOLEAN)
```
danieljhegeman, line 38: can we change this to `custom_dataset_order`? As it is, two minor problems:
- sounds like it is itself an array of datasets that have a custom order applied
- annoyance for tab-complete as it masks `datasets` 😏 (kidding but also not kidding)

Actually, what about `use_dataset_order` or `use_custom_dataset_order`?

backend/layers/persistence/persistence.py
```diff
@@ -168,6 +171,18 @@ def _row_to_dataset_version(self, row: Any, canonical_dataset: CanonicalDataset,
...
+    def _sort_datasets(self, row: Any, datasets: List[DatasetVersion]) -> List[DatasetVersion]:
+        """
+        Sorts datasets by the order they appear in the collection version's datasets array if custom
+        order is enabled, otherwise sorts datasets by cell count.
+        """
+        if row.datasets_custom_ordered:
+            datasets_order = [DatasetVersionId(str(id)) for id in row.datasets]
+            return sorted(datasets, key=lambda d: datasets_order.index(d.version_id))
+        return sorted(
+            datasets, key=lambda d: 0 if d is None or d.metadata is None else d.metadata.cell_count, reverse=True
```
danieljhegeman, line 183: in what scenario is `d` going to evaluate to `None`?

backend/layers/persistence/persistence.py
```diff
@@ -951,6 +968,32 @@ def replace_dataset_in_collection_version(
...
+        """
+        Updates the datasets in a collection version to match the given order.
+        """
+        with self._manage_session() as session:
+            collection_version = session.query(CollectionVersionTable).filter_by(id=collection_version_id.id).one()
+
+            # Confirm collection version datasets length matches given dataset version IDs length.
+            if len(collection_version.datasets) != len(dataset_version_ids):
+                raise ValueError(
+                    f"Dataset version IDs length does not match collection version {collection_version_id} datasets length"
```
danieljhegeman, line 985: Let's capitalize Dataset Version, Collection Version, Datasets...  same thing in the error message text immediately below 👇

backend/layers/persistence/persistence_mock.py
```diff
@@ -103,6 +104,11 @@ def _update_version_with_canonical(
                     datasets_to_include.append(dataset_version)
             # Replace 'datasets' array of Dataset version ids with 'datasets' array of actual Dataset versions
             copied_version.datasets = datasets_to_include
+            # Order by cell count if not custom ordered.
+            if not copied_version.datasets_custom_ordered:
+                copied_version.datasets.sort(
+                    key=lambda d: 0 if d is None or d.metadata is None else d.metadata.cell_count, reverse=True
+                )
```
danieljhegeman, line 111: since we have this in at least two places now, let's consolidate into a common utils function, e.g. `sort_datasets_by_cell_count`

danieljhegeman (changes requested):
I've reviewed everything except the tests. Some changes 🙏

backend/database/versions/05_67c4015bce73_add_datasets_custom_ordered_to_.py
MillenniumFalconMechanic: Thanks for noticing `Boolean` vs `BOOLEAN`! I had a look at the docs: `BOOLEAN` uses the [SQL type](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sql-standard-and-multiple-vendor-uppercase-types); `Boolean` uses a [generic type](https://docs.sqlalchemy.org/en/20/core/type_basics.html#generic-camelcase-types). 

> Actually, I'm noticing that in the other two instances we allow nullable=True and we do not set a server_default value. I'm thinking we should do this.

Regarding making `datasets_custom_ordered` `nullable=True`, wouldn't we want the database to protect against NULL values being inserted? Happy to move it to `nullable=True` but want to double-check this is the desired approach.

backend/layers/persistence/persistence.py
MillenniumFalconMechanic: Looks like this is no longer possible; I have removed `is None`.

backend/database/versions/05_67c4015bce73_add_datasets_custom_ordered_to_.py
danieljhegeman: I say this because at present the other boolean values are configured this way (even though we otherwise enforce a True or False value for all rows). I know there is some tangentially-related issue with the way we treat boolean values when we extract rows from psql (we have some code somewhere that tests for `is not False` or something like that, rather than `is True`...)

cc @nayib-jose-gloria @Bento007 for comment on these patterns before I get a chance to look more into it

backend/layers/persistence/persistence.py
ebezzi: If we go by the original design, I would argue that this belongs to the business logic.

ebezzi: That said, I'm not suggesting that we change it now, since it will be quite a few changes - just a reminder that we diverged quite significantly from the original design and the difference between the two layers is now less marked.

ebezzi (approved):

backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
```diff
@@ -0,0 +1,28 @@
...
+revision = "05_67c4015bce73"
+down_revision = "04_2f30f3bcc9aa"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "CollectionVersion",
+        sa.Column("custom_dataset_order", sa.BOOLEAN(), nullable=True, server_default=sa.sql.expression.false()),
```
danieljhegeman, line 22: I'm slightly confused. I thought we had agreed to instantiate this / treat this in the same fashion as the [boolean tombstone](https://github.com/chanzuckerberg/single-cell-data-portal/blob/317080ab15531c043f326574ea5c2c0bf409e031/backend/database/versions/03_98cf7564214f_add_tombstone_to_datasets.py)

danieljhegeman: Per our conversations in Slack

danieljhegeman: Also this column really should be given an assertive statement as a variable name, in keeping with conventions for boolean values (and yes I realize that `tombstone` should be `is_tombstoned` or something like that! 😭 ). Maybe `use_custom_dataset_order` or `has_custom_dataset_order`? Was this overruled somewhere in discussions? Can't remember if I suggested/mentioned this.

danieljhegeman (commented):
Commented on column name. Separately -- and I apologize if we went over this already in discussions elsewhere -- what is the plan for how we treat a scenario where a curator has organized a collection in a custom fashion, (re)publishes it, and then decides they actually want to revert to default (cell count) order? Do they have to manually arrange the Datasets according to their cell count? It would seem that way since we've only introduced one endpoint, and not another way to do a reset.

brianraymor:
> what is the plan for how we treat a scenario where a curator has organized a collection in a custom fashion, (re)publishes it, and then decides they actually want to revert to default (cell count) order? Do they have to manually arrange the Datasets according to their cell count?

Yes. We discussed and rejected `revert` UX in favor of manual reordering. It's an edge case.

backend/database/versions/05_67c4015bce73_add_custom_dataset_order.py
MillenniumFalconMechanic: Hey Dan! I updated the column as per your initial suggestion [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6579#discussion_r1470536403) but can make another update: what is your preference between `use_custom_dataset_order` and `has_custom_dataset_order`?

danieljhegeman (approved):

Issue #6626, opened 2024-02-06
WMG Pipeline run failed due to awscli version


prathapsridharan:
## Problem

1. On Feb 4 WMG Pipeline run, the logs show the following traceback (see end of ticket for the traceback) indicating that `awscli` was not able to import `is_s3express_bucket` module
2. This led to the cube data not being copied into `cellxgene-wmg-prod` bucket on S3
3. However, the `latest_snapshot_identifier` was copied.
4. This led to failures on various `wmg/v2` api endpoints that were attempting to load the snapshot from S3 because `latest_snapshot_identifier` had the latest snapshot_id value but the data folder associated with the snapshot_id value was not there due to (1) and (2).

The import was introduced in [this commit](https://github.com/aws/aws-cli/commit/99b79a014ed62e9130250fbef57d535c0388efea) by awscli maintainers which was [released as 1.32.25 in PyPI](https://pypi.org/project/awscli/#history)

## Potential Solutions

1. Pin `awscli` version to be `< 1.32.25`. _This might not be advisable if awscli updates are required from a security standpoint_
2. Upgrade `awscli` to `v2` - [installation instructions](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
3. Fix forward by debugging why the import failed and ensure that it imports and works. There has been [some discussion in the web about this problem](https://repost.aws/it/questions/QU2l9P6ZEhSkSL0lFQonOHfQ/a-lambda-failed-to-start-with-the-error-cannot-import-name-is-s3express-bucket-from-botocore-utils)

(1) could be temporary fix (if it works) to unblock WMG pipeline while we work on (2)

### Traceback

```
INFO:backend.wmg.pipeline.load_cube:Writing snapshot data to s3://cellxgene-wmg-prod/snapshots/v3/1707069806
Traceback (most recent call last):
  File "/usr/local/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/usr/local/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/usr/local/lib/python3.10/site-packages/awscli/clidriver.py", line 69, in main
    driver = create_clidriver()
  File "/usr/local/lib/python3.10/site-packages/awscli/clidriver.py", line 78, in create_clidriver
    load_plugins(session.full_config.get('plugins', 
{}
),
  File "/usr/local/lib/python3.10/site-packages/awscli/plugin.py", line 44, in load_plugins
[...]

PR #6627, opened 2024-02-06
fix: Update awscli library for wmg pipeline

prathapsridharan:
## Reason for Change

#6626 

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️

### DIFF
--- a/Dockerfile.wmg_pipeline
+++ b/Dockerfile.wmg_pipeline
@@ -4,7 +4,11 @@ FROM python:3.10
 RUN /usr/local/bin/python -m pip install --upgrade pip && \
     apt update && apt -y install graphviz graphviz-dev && \
     rm -rf /var/lib/apt/lists/* && \
-    pip3 install awscli==1.32.24
+    apt-get install unzip && \
+    apt-get install curl && \
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+    unzip awscliv2.zip && \
+    ./aws/install
 
 # For lighter weight Docker images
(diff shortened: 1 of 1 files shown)


Dockerfile.wmg_pipeline
```diff
@@ -4,7 +4,7 @@ FROM python:3.10
 RUN /usr/local/bin/python -m pip install --upgrade pip && \
     apt update && apt -y install graphviz graphviz-dev && \
     rm -rf /var/lib/apt/lists/* && \
-    pip3 install awscli
+    pip3 install awscli==1.32.24
```
Bento007, line 7: if this will be pinned then we should pin it in the requirements.txt file, that way dependabot can check it for us.

ebezzi: This is a CLI command and not a library, so I think this is the right place. In fact, I would recommend to install it with the method suggested [here](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html).

prathapsridharan:
@Bento007 @ebezzi - We have reason to believe that the following exception might be temporary glitch:

```
File "/usr/local/lib/python3.10/site-packages/awscli/customizations/s3/subcommands.py", line 18, in <module>
    from botocore.utils import is_s3express_bucket
ImportError: cannot import name 'is_s3express_bucket' from 'botocore.utils' (/usr/local/lib/python3.10/site-packages/botocore/utils.py)
```

`staging` snapshot, which is only 1 second older, ran fine. I will re-run `prod` snapshot and see if the error happens again before attempting a fix. Keeping this PR open until then. Stay tuned....

ebezzi (approved):

Issue #6688, opened 2024-02-21
Update seed test data to include four collections three datasets each


MillenniumFalconMechanic:
#### Need
E2E tests covering the reordering of datasets require four collections with at least three datasets each in rdev, dev and staging. 

The three datasets is a new requirement from the reorder tests, the minimum of four collections is an existing requirement from [startRevision](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/frontend/tests/features/collection/revision.test.ts#L220).

#### Background
The reordering E2E tests [reuse](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6687/files#diff-ec34758f85529418ae063bdd0d1a5948fbae53c6c311be97a5498557ed2b4362R731) [revision test code](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/frontend/tests/features/collection/revision.test.ts#L219) that requires four writable collections. 

#### Tech Spec
[Tech Spec: Update seed test data to include four collections with  three datasets per collection](https://docs.google.com/document/d/1ndu59LCyL9G8WGTBpu-JvdkkzGKbh-7b3VXXmQ_o2pM/edit#heading=h.jcf27gbiblyg)

PR #6696, opened 2024-02-22
test: seed collections contain three datasets each

MillenniumFalconMechanic:
## Reason for Change

- #6688 

## Changes

- Updated seed collections to contain three datasets each.

## Testing steps

- Verify PR tests run clean.
- Verify dev and staging merge tests run clean.

### DIFF
--- a/scripts/smoke_tests/setup.py
+++ b/scripts/smoke_tests/setup.py
@@ -29,20 +29,8 @@ class SmokeTestsInitializer(BaseFunctionalTestCase):
                 return num_collections
         return num_collections
 
-    def multi_upload_and_wait(self, collection_id: str, dropbox_url: str, upload_count: int) -> None:
-        threads = []
-        for _ in range(upload_count):
-            thread = threading.Thread(
-                target=self.upload_and_wait, args=(collection_id, dropbox_url), kwargs={"cleanup": False}
-            )
-            threads.append(thread)
-            thread.start()
-        for thread in threads:
-            thread.join()
-
     def create_and_publish_collection(self, dropbox_url):
         collection_id = self.create_collection()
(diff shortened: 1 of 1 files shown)


tihuan:
I see 2 "test collection"'s in the rdev instead of 10, but maybe we only seed 2 test collections in rdev? I'm not sure lol

https://pr-6696-frontend.rdev.single-cell.czi.technology/collections

Screenshot 2024-02-22 at 12 01 27 PM

MillenniumFalconMechanic:
> I see 2 "test collection"'s in the rdev instead of 10

Hey @tihuan! The setup script should create [10 collections](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/scripts/smoke_tests/setup.py#L9). I think what might be going on here is that the [multi-threaded](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6696/commits/ce78e02153e3012096d5c5a0856c080d0248d91f#diff-f9a8778214272c2fa77ebc11012bf2e3e3614711a13aecf2b09b0d522350d591R46) dataset creation [timed out](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/tests/functional/backend/common.py#L168) when I attempted to spawn a new thread for each dataset (three datasets per collection); 10 collections were created but only two "made it" to [publish](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/scripts/smoke_tests/setup.py#L34). From your screenshot, it looks like you're not authenticated so you are only seeing the two public collections? We won't be able to very my theory however (!) as I have since cleared out the test data so I could run the dataset creation [single-threaded](https://github.com/chanzuckerberg/single-cell-data-portal/pull/6696/files#diff-f9a8778214272c2fa77ebc11012bf2e3e3614711a13aecf2b09b0d522350d591R34).

joyceyan (approved):

PR #6697, opened 2024-02-22
fix: update RDS citation to use .rds instead of .h5ad

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/6671

## Changes

- modifies the anndata of the `local.h5ad` that we pull down from S3 before doing the seurat RDS conversion
- i chose this approach so that i don't have to learn R lol

#### To test that the citation has been updated in the RDS artifact:

1. i downloaded an RDS dataset from the rdev generated by this PR
2. i sent the `.rds` file to @brianraymor, who has RStudio up and running, and verified the citation field has `.rds` instead of `.h5ad`

image

#### To test that the regex for replacing the citation works:

i isolated this logic to `utils.rds_citation_from_h5ad` and added some tests in `test_utils.py`

## Checklist 🛎️

### DIFF
--- a/backend/layers/processing/process_seurat.py
+++ b/backend/layers/processing/process_seurat.py
@@ -2,6 +2,8 @@ import logging
 import os
 import subprocess
 
+import anndata
+
 from backend.layers.business.business_interface import BusinessLogicInterface
 from backend.layers.common.entities import (
     DatasetArtifactType,
@@ -61,12 +63,19 @@ class ProcessSeurat(ProcessingLogic):
             self.logger.info("Skipping Seurat conversion")
             return
 
+        # Download h5ad locally
         labeled_h5ad_filename = "local.h5ad"
-
         key_prefix = self.get_key_prefix(dataset_version_id.id)
         object_key = f"{key_prefix}/{labeled_h5ad_filename}"
         self.download_from_s3(artifact_bucket, object_key, labeled_h5ad_filename)
 
+        # Convert the citation from h5ad to RDS
+        adata = anndata.read_h5ad(labeled_h5ad_filename)
+        if "citation" in adata.uns:
+            adata.uns["citation"] = adata.uns["citation"].replace(".h5ad", ".rds")
+        adata.write_h5ad(labeled_h5ad_filename)
+
+        # Use Seurat to convert to RDS
         seurat_filename = self.convert_file(
             self.make_seurat,
             labeled_h5ad_filename,
--- a/backend/layers/processing/process_seurat.py
+++ b/backend/layers/processing/process_seurat.py
@@ -13,6 +13,7 @@ from backend.layers.common.entities import (
 )
 from backend.layers.processing.logger import logit
 from backend.layers.processing.process_logic import ProcessingLogic
+from backend.layers.processing.utils import rds_citation_from_h5ad
 from backend.layers.thirdparty.s3_provider import S3ProviderInterface
 from backend.layers.thirdparty.uri_provider import UriProviderInterface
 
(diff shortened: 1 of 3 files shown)


backend/layers/processing/process_seurat.py
```diff
@@ -61,12 +63,19 @@ def process(self, dataset_version_id: DatasetVersionId, artifact_bucket: str, da
...
         labeled_h5ad_filename = "local.h5ad"
-
         key_prefix = self.get_key_prefix(dataset_version_id.id)
         object_key = f"{key_prefix}/{labeled_h5ad_filename}"
         self.download_from_s3(artifact_bucket, object_key, labeled_h5ad_filename)
 
+        # Convert the citation from h5ad to RDS
+        adata = anndata.read_h5ad(labeled_h5ad_filename)
+        if "citation" in adata.uns:
+            adata.uns["citation"] = adata.uns["citation"].replace(".h5ad", ".rds")
```
danieljhegeman, line 75: let's make use of the [dataset id regex](https://github.com/chanzuckerberg/single-cell-data-portal/blob/2f655e203430d69634f8b9885d6780a747c8e4ae/backend/common/utils/regex.py#L5) so that we don't get caught by a doi that has `.h5ad` in it.

tests/unit/processing/test_processing.py
```diff
@@ -13,28 +13,32 @@
...
-            artifact.type = "RDS"
-            artifact.uri = f"s3://fake_bucket_name/{dataset_version_id.id}/local.rds"
+        self.assertTrue(self.s3_provider.uri_exists(f"s3://fake_bucket_name/{dataset_version_id.id}/local.rds"))
+        self.assertTrue(self.s3_provider.uri_exists(f"s3://fake_datasets_bucket/{dataset_version_id.id}.rds"))
+
+        artifacts = list(self.business_logic.get_dataset_artifacts(dataset_version_id))
+        self.assertEqual(1, len(artifacts))
+        artifact = artifacts[0]
+        artifact.type = "RDS"
+        artifact.uri = f"s3://fake_bucket_name/{dataset_version_id.id}/local.rds"
```
danieljhegeman, line 41: Did you mean to have another assertion here? What are you arranging on these three lines?

danieljhegeman (changes requested):
Can you point me to where the citation is being tested? I'm assuming that's happening in the creation of the artifact in the test but it's not crystal-clear to me...

tests/unit/processing/test_processing.py
joyceyan: sorry i should have clarified what i was doing here. all i've updated for this test is add a `mock_seurat` and mocked the return value for that as `local.rds`. since this test now has two mocks, i changed the test to use the `@patch` decorator instead of using `with patch(...) as`. removing the `with patch` alters some of the whitespace, which are the changes that you see here. none of the actual logic of this test was changed.

danieljhegeman: Ah, right, thanks. Ok. I'll check with Nayib later.

backend/layers/processing/utils.py
```diff
@@ -0,0 +1,7 @@
+import re
+
+from backend.common.utils.regex import DATASET_ID_REGEX
+
+
+def rds_citation_from_h5ad(citation: str) -> str:
+    return re.sub(f"{DATASET_ID_REGEX}\.h5ad", r"\g<dataset_id>.rds", citation)
```
danieljhegeman, line 7: Thanks, just learned of the `\g` usage for group capture reference 🎉

danieljhegeman (approved):
Since we're only using the `rds_citation_from_h5ad` helper function in one spot, I would normally say let's inline it rather than factor out into a (new) utils file, but I'll leave that final decision to you. I suppose we may end up using it once we come back to handle tech debt related to unit tests for rds artifacts.

Issue #6202, opened 2023-11-09
Data Submission Policy was regressed and is returning incorrect version


brianraymor:
# Context

The `data_submission_policy_version` was previously stored for each collection to support:
1. Tracking which collections have been updated to a new policy when approval by existing submitters or others is required
2. Allowing CELLxGENE Discover to support multiple submission policies when existing submitters do not agree to an updated policy 

The value is also returned by DP API. The current version is 2.0, but the implementation was regressed (eliminated?) and 1.0 is being returned as a hard coded value. See the [code](
https://github.com/chanzuckerberg/single-cell-data-portal/blob/655f391d81c89b403422ea81e0d5d69cf3427aa5/backend/portal/api/portal_api.py#L270):

` "data_submission_policy_version": "1.0",  # TODO`

#### Tech Spec
[Tech Spec: The Data Submission Policy was regressed and is returning an incorrect version](https://docs.google.com/document/d/1kTsaW-Paf5yHxzR3kK5RfGBd1gjY8hpm__tgojmv9ZU/edit#heading=h.jcf27gbiblyg)

metakuni:
Fix notes:
1. For each collection, find the most recent `data_submission_policy_version` that predates the hardcoded 1.0 value. (from before the Portal redesign)
2. Set that value in the current collection version via `cxg_admin.py`.
3. Fix the [API](https://github.com/chanzuckerberg/single-cell-data-portal/blob/655f391d81c89b403422ea81e0d5d69cf3427aa5/backend/portal/api/portal_api.py#L270) to return `collection.data_submission_policy_version`.

MillenniumFalconMechanic:
4. Save current data submission policy version on publish.

PR #6712, opened 2024-02-27
fix: data submission policy version

MillenniumFalconMechanic:
## Reason for Change

- #6202

## Changes

- Updated publish of collection to save data submission policy version to DB.
- Updated read of collection to return saved data submission policy version. 

## Testing steps

- Updated controller and business-layer tests.
- Tested in rdev.
 
## Checklist 🛎️

### DIFF
--- a/backend/portal/api/portal_api.py
+++ b/backend/portal/api/portal_api.py
@@ -263,7 +263,7 @@ def _collection_to_response(collection: CollectionVersionWithDatasets, access_ty
             "contact_name": collection.metadata.contact_name,
             "created_at": collection.created_at,
             "curator_name": collection.curator_name,
-            "data_submission_policy_version": "1.0",  # TODO
+            "data_submission_policy_version": collection.data_submission_policy_version,
             "datasets": [
                 _dataset_to_response(
                     ds,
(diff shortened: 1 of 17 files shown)


nayib-jose-gloria (commented):
looks good, thank you! one comment about gating against input submissions policies < our current submission policy

ebezzi (approved):

backend/portal/api/portal_api.py
```diff
@@ -68,18 +70,18 @@ def get_collections_list(from_date: int = None, to_date: int = None, token_info:
     user_private_collections = get_user_private_collections(token_info)
 
     collections = []
-    for version in itertools.chain(all_published_collections, user_private_collections):
+    for collection_version in itertools.chain(all_published_collections, user_private_collections):
```
MillenniumFalconMechanic, line 73: `version` here shadowed imported `version`; updated to `collection_version`.

backend/portal/api/portal_api.py
```diff
@@ -5,7 +5,9 @@
 from urllib.parse import urlparse
 
 from flask import Response, jsonify, make_response
+from packaging import version
```
ebezzi, line 8: I'm not sure introducing a dependency for something this simple is a good idea. Any reason for which you can't just compare the two strings (or even two parsed numbers)?

MillenniumFalconMechanic: The original requirement was `>=` but I think now it makes sense to just do a string comparison.

ebezzi (approved):

PR #6713, opened 2024-02-27
feat: update cellguide pipeline to calculate ontology trees per organism

atarashansky:
## Reason for Change
As part of the work to add the explorer instances to CellGuide, I noticed that there are some inconsistencies with the behavior of the organism and the tissue filter.
 - The organism/tissue filters are calculated based on the computational marker genes, but if a cell type has no markers, then the filters will be empty, even if that cell type is present in the corpus. For example, "Cell" should have all organisms and it should have all tissues, because it is the root of the ontology, but it currently has nothing because "Cell" has no marker genes.
 - Also note that the ontology tree cell counts do not respond to the organism filters.

To address this, I'm adding a new data artifact to the cellguide pipeline and updating the way it generates the ontology tree and tree states.
 - Now the pipeline will generate a separate tree PER ORGANISM.
 - It will also create a new data artifact which maps the tissues and organisms available for every cell type. This will be used to populate the filters in the CellGuide pages.

## Changes

- added new data artifact containing valid tissues for every cell type
- ontology tree states are split by organism

## Testing steps

- Unit tests updated
- Rdev run succeeded


Bento007:
Can you add a description please

prathapsridharan (approved):
lgtm

PR #5965, opened 2023-10-12
chore: Disable redeploying the dev environment for pentest

Bento007:
## Reason for Change
To freeze the dev environment while pentesting happens.

## Changes

- only deploy staging and prod in GHA

## Notes for Reviewer
This will be reverted after pentesting concludes

PR #6727, opened 2024-02-28
fix: dev freeze

joyceyan:
## Reason for Change

this PR: https://github.com/chanzuckerberg/single-cell-data-portal/pull/6715/files copied and pasted what we had previously done for a dev freeze: https://github.com/chanzuckerberg/single-cell-data-portal/pull/5965

however, since then we've changed how we do staging/prod deploys to rely on branches

## Changes

changes to use branch rather than github ref

### DIFF
--- a/.github/workflows/build-images-and-create-deployment.yml
+++ b/.github/workflows/build-images-and-create-deployment.yml
@@ -43,7 +43,7 @@ jobs:
           fi
       - uses: avakar/create-deployment@v1
         #  TODO remove if conidition after we're done with testing schema 5.0.0 in dev
-        if: github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/prod'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images-and-create-deployment.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           fi
       - uses: avakar/create-deployment@v1
         #  TODO remove if conidition after we're done with testing schema 5.0.0 in dev
-        if: github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/prod'
+        if: ${{ env.GITHUB_BRANCH }}" == "refs/heads/prod || ${{ env.GITHUB_BRANCH }}" == "refs/heads/staging
```
joyceyan, line 46: basically this check should be the same as the check we use on lines 37 and 39

ebezzi (approved):

PR #6732, opened 2024-02-29
chore: build rdev immediately without waiting for push tests

danieljhegeman:
unrelated fix: do not copy cellguide cxgs when performing data mirror

## Changes

- remove push tests from the rdev-update workflow
  - builds images and deploys rdev immediately
  - split across multiple workflows means lower likelihood of cancellation of tasks due to excessive demand within a single workflow. Instead we expect to see delayed/queued GHAs, but this is more acceptable because they will still be run.

## Testing steps

## Checklist 🛎️


backend/scripts/mirror_s3_data.sh
```diff
@@ -126,7 +126,7 @@ else  # i.e., if DEST_ENV != 'rdev'
   set -x
   echo -e "\n\n*** The following parallel aws s3 sync commands may appear to hang, but they are working. Patience is a virtue. ***\n\n"
   $PARALLEL_CMD $S3_SYNC_CMD s3://corpora-data-${SRC_ENV}/ s3://corpora-data-${DEST_ENV}/ --exclude "'*'" --include "'{}*'" --exclude "'*loom'" ::: $hex_chars
-  $PARALLEL_CMD $S3_SYNC_CMD s3://hosted-cellxgene-${SRC_ENV}/ s3://hosted-cellxgene-${DEST_ENV}/ --exclude "'*'" --include "'{}*'" ::: $hex_chars
+  $PARALLEL_CMD $S3_SYNC_CMD s3://hosted-cellxgene-${SRC_ENV}/ s3://hosted-cellxgene-${DEST_ENV}/ --exclude "'*'" --include "'{}*'" --exclude "'cellguide-cxgs*'" ::: $hex_chars
```
danieljhegeman, line 129: unrelated data mirror fix. These files are redundant because everything reads from prod for cellguide cxgs.

.github/workflows/rdev-update-for-pr.yml
```diff
@@ -24,14 +24,6 @@ permissions:
   contents: write
 
 jobs:
```
joyceyan, line 26: hm i think we should still run the push tests when you update the PR, but i agree we don't need the push tests to succeed to start building images.

regardless removing this job definition entirely i'm pretty sure will break since the `rdev-tests` at the bottom here requires `push-tests`. i think if we just remove the requirement of needing `push-tests` on `build_images` and keep the job definition + requirement on `rdev-tests` that would work well?

joyceyan (changes requested):
i don't think we should remove the push tests job definition

.github/workflows/rdev-update-for-pr.yml
Bento007: > regardless removing this job definition entirely i'm pretty sure will break since the rdev-tests at the bottom here requires push-tests. 

This would break the workflow if we don't remove the requirement from rdev-tests.

> i think if we just remove the requirement of needing push-tests on build_images and keep the job definition + requirement on rdev-tests that would work well?

That's wont fix the problem of workflows canceling randomly.

danieljhegeman: Thanks for catching the fact that `push-tests` got left in as needed for `rdev-tests`. I've removed it now.

PR #6749, opened 2024-03-04
feat: persist distributed mode in schema migration script

joyceyan:
## Reason for Change

so that we don't have to change this manually with each run

## Changes

changes to `DISTRIBUTED` mode


danieljhegeman (approved):
Thanks @joyceyan !! 🙏

PR #6757, opened 2024-03-05
feat: add canonical collection url to data wrangling reports

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/6392

## Changes

adds the canonical collection url to data wrangling error reports in slack as well as to the schema migration report

## Testing steps

updated tests for `test_collection_migrate.py` which verifies we send the collection URL in the schema migration report, and `test_handle_error.py` which verifies we send the collection URL in the data wrangling error reports


backend/layers/processing/upload_failures/app.py
```diff
@@ -104,6 +114,11 @@ def get_failure_slack_notification_message(
     batch_url = aws_batch_job_url_fmt_str.format(aws_region=aws_region, job_id=job_id)
     step_function_url = aws_sfn_url_fmt_str.format(aws_region=aws_region, execution_arn=execution_arn)
     collection_version_url = f"https://cellxgene.cziscience.com/collections/{collection_version_id}"
+
+    # Generate canonical collection URL
+    version = get_business_logic().get_collection_version(CollectionVersionId(collection_version_id))
+    collection_url = f"https://cellxgene.cziscience.com/collections/{version.collection_id.id}"
```
Bento007, line 120: should this be using the new logic?
```suggestion
    collection_url = self.business_logic.get_collection_url(version.collection_id.id)
```

backend/layers/business/business.py
```diff
@@ -265,6 +265,9 @@ def get_unpublished_collection_version_from_canonical(
                 unpublished_collection = collection
         return unpublished_collection
 
+    def get_collection_url(self, collection_id: str) -> str:
```
Bento007, line 268: Add a unit test for this function.

Bento007 (commented):
Looks good in general. A few small things to address.

backend/layers/processing/upload_failures/app.py
joyceyan: I had considered that, but this line which generates the collection version url: https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/layers/processing/upload_failures/app.py#L106 uses the hard-coded prod URL. My understanding is that this is only run in prod which is why that's okay, but I'm not 100% sure.

I'm fine with updating it to use `self.business_logic.get_collection_url` but I think we'd want to update both the version URL and the canonical URL.

Bento007: It's fine as is for now.

Bento007 (approved):

PR #6758, opened 2024-03-05
feat: alter dockerfile processing to use different requirements for dev

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/678

## Changes

introduces a new file, `processing/requirements-dev.txt` and has `Dockerfile.processing` install from this list of requirements if we're in dev, otherwise it'll install from `processing/requirements.txt`

## Testing steps

rdev build pip installs `requirements.txt`


joyceyan:
In this rdev build, we pip install requirements.txt

image

Dockerfile.processing
```diff
@@ -16,7 +16,9 @@ RUN pip3 install python-igraph==0.8.3 louvain==0.7.0 packaging==21.0 awscli
 
 
 COPY /python_dependencies/processing/ .
-RUN pip3 install -r requirements.txt
+
+# Install the dev requirements, which may contain a pinned cellxgene-schema version
+RUN if [ "$HAPPY_BRANCH" = "dev" ] ; then pip3 install -r requirements-dev.txt ; else pip3 install -r requirements.txt ; fi
```
nayib-jose-gloria, line 21: so that we don't have to maintain two sets of mostly the same dependencies, I think we should keep `RUN pip3 install -r requirements.txt` and then, conditionally, run an extra `pip3 install -r requirements-dev.txt` if branch is dev but otherwise no additional install. 

I believe that is sufficient to uninstall the version from requirements.txt and then install the version from requirements-dev.txt (which would now contain only the github link to cellxgene-schema). 

What do you think of that?

joyceyan: Oh yeah I hadn't considered uninstalling the `requirements.txt` version. That makes more sense, I'll update to do that

nayib-jose-gloria (approved):

PR #6766, opened 2024-03-07
test: fix revision deletion

tihuan:
The fix is simply removing the `tryUntil()` block around the assertion:

```ts
expect(await collectionRows.count()).toBeGreaterThan(
  MIN_USABLE_COLLECTION_COUNT - 1
);
```

so the `catch()` block can delete revision and recheck the assertion!


MillenniumFalconMechanic (approved):
LGTM, thanks so much Timmy! 🚀

PR #6768, opened 2024-03-08
chore(cg): a couple bug fixes and tooltip update

atarashansky:
## Reason for Change / Changes

- Changing organisms on a cell type with no cells for the new organism crashed the page.
  - Allowed `tissues` to be undefined.
- Tooltip for open integrated embedding button was using an incorrect warning icon
  - Used design-approved icon
- Tooltip was placed in an orientation that was not compatible with mobile layouts
  - Placed the tooltip on top, instead
- "Open Integrated Embedding" text wrap on mobile layouts looked ugly
  - Turned off text wrap
image

## Testing steps

- Tested locally
- Added e2e test for the primary bug


tihuan (approved):
LGTM! Thanks, Alec 😄

PR #6786, opened 2024-03-12
chore: persist memory bump to schema migration jobs

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/686

## Changes

bumps memory up as we did when running locally

## Testing steps

tested a dev SFN push with these memory numbers


joyceyan:
closing the PR since we haven't decided if we want to do this or not

PR #6788, opened 2024-03-13
feat: Integrate FE Filters with cellxgene-ontology-guide

nayib-jose-gloria:
## Reason for Change

- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/cellxgene-ontology-guide/103
- Integrate FE Filters with cellxgene-ontology-guide to pull latest descendant mappings at image build time, to keep filters 1) up-to-date with new cell types and tissues in the corpus, and 2) keep them aligned to new schema releases / curated ontology term lists (i.e. cell class, cell subclass, organ tissues, system tissues) at all times

## Changes

- add steps in Dockerfile to pull latest descendant mappings from cellxgene-ontology-guide repo
- remove static descendant mappings in FE directory
- use cellxgene-ontology-guide to enrich dataset ancestors in the backend (dp datasets index endpoint / user-datasets endpoint)
- add filtering step in dataset enrichment commands to only populate dataset response object with ontology term ancestors that the FE uses--terms that are descendants from curated lists of high-level Tissue / Cell Type / UBERON-specific Development Stage terms (pulled from COG version) or any terms that are in datasets in the data-portal public corpus


backend/portal/api/portal_api.py
```diff
@@ -796,7 +795,7 @@ def get_user_datasets_index(token_info: dict):
     non_revisions = [c for c in private_collections if not c.is_unpublished_version()]
     private_datasets = get_business_logic().get_datasets_for_collections(non_revisions)
 
-    response = enrich_dataset_response(itertools.chain(public_datasets, private_datasets))
+    response = enrich_dataset_response(list(itertools.chain(public_datasets, private_datasets)))
```
Bento007, line 798: Remove chain. We aren't really using it.
```suggestion
    response = enrich_dataset_response(public_datasets + private_datasets)
```

frontend/src/components/common/Filter/descendant_mappings/tissue_descendants.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{}
```
Bento007, line 1: If these are empty, can they be removed

frontend/Makefile
```diff
@@ -37,6 +37,11 @@ retrieve-vars:
 upload-vars:
 	aws s3 cp .env.production $(S3_ENVIRONMENT_FILE)
 
+.PHONY: retrieve-descendants
+retrieve-descendants:
+	curl -o src/components/common/Filter/descendant_mappings/tissue_descendants.json https://raw.githubusercontent.com/chanzuckerberg/cellxgene-ontology-guide/main/ontology-assets/tissue_descendants.json
```
Bento007, line 42: I worry that pulling from main could introduce too much variability. Around the time of schema migration we would need to make sure it's only pulling the tissue descendants from the current schema.

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -1023,17 +1023,15 @@ def test_get_collection_version_ok(self):
                     "collection_version_id": f"{first_version.version_id.id}",
                     "cell_count": 10,
                     "primary_cell_count": 5,
-                    "cell_type": [{"label": "test_cell_type_label", "ontology_term_id": "test_cell_type_term_id"}],
+                    "cell_type": [{"label": "unknown", "ontology_term_id": "unknown"}],
```
Bento007, line 1026: Are these`unknown`s from `OntologyParser`?

backend/portal/api/enrichment.py
```diff
@@ -3,33 +3,68 @@
...
     """
     Tag dataset with ancestors for all values of the given key, if any.
     """
     if key not in dataset:
         return
-
-    terms = [e["ontology_term_id"] for e in dataset[key]]
-
     is_tissue = key == "tissue"
+    is_cell_type = key == "cell_type"
```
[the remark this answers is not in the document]
nayib-jose-gloria: new feature; necessary to support filtering for accepted ancestors depending on what the key type is (i.e. accept cell type ancestors vs tissue ancestors)

tests/unit/backend/layers/api/test_curation_api.py
nayib-jose-gloria: no, its just that the parser will throw exceptions if we pass in invalid term IDs like "test_term_id" but it is built to accept "unknown" as an exception case

frontend/src/components/common/Filter/descendant_mappings/tissue_descendants.json
nayib-jose-gloria: These are populated dynamically at build-time; if we reference them without the file existing, TS throws lint errors. We could suppress these errors, but I think it might be clearer to add a README in the directory explaining when they get populated and leaving them as-is, so future devs know this directory/files are meant to exist. I added a README, let me know what you think

Bento007 (approved):

PR #6789, opened 2024-03-13
Change models page to support new embeddings mode

ebezzi:
Collaboration models will now exist in contributions.json, so this PR retrieves them from there.


tihuan (approved):
Nice clean code! Thanks so much 👏!!

PR #6817, opened 2024-03-21
feat: implement differential expression backend

atarashansky:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️


atarashansky:
Note to self: It seems that we are getting mouse-only filters when human organism is selected (see tabula muris senis consoritum publication option). This can lead to 0 cells returned. Figure out why this is happening and fix.

backend/de/api/v1.py
```diff
@@ -0,0 +1,333 @@
...
+DEPLOYMENT_STAGE = os.environ.get("DEPLOYMENT_STAGE", "")
+SNAPSHOT_FS_ROOT_PATH = (
+    WMG_API_SNAPSHOT_FS_CACHE_ROOT_PATH if (WMG_API_READ_FS_CACHED_SNAPSHOT and DEPLOYMENT_STAGE != "test") else None
+)
+
+
+@tracer.wrap(name="filters", service="wmg-api", resource="filters", span_type="wmg-api")
+def filters():
+    request = connexion.request.json
+    sanitize_api_query_dict(request["filter"])
```
kaloster, line 40: Discussion: what are the pros of mutating the original request object (~immutability principle), vs assigning a sanitized return `sanitized_filter = sanitize_api_query_dict(request["filter"])` and working off of that?

kaloster (approved):
Looks solid to me! 🚀 🙌🏽 

Just one discussion comment is all I had to contribute

backend/de/api/v1.py
atarashansky: Great question, don't think there are any pros beyond not needing a new variable - this is copied from the WMG API. Future work could refactor this logic into a shared utilities module.

PR #6876, opened 2024-03-28
chore: switch mock census version parameter to stable

atarashansky:
## Reason for Change

- Tests are failing because the WMG pipeline is currently forbidden from reading the latest Census 2.0 build.
- This will be reverted once the pipeline has been updated


danieljhegeman (approved):
🚀 🍀

PR #6880, opened 2024-04-01
chore: bump cellxgene-schema to version using cellxgene-ontology-guide 

nayib-jose-gloria:
## Reason for Change

- Draft PR to test https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/cellxgene-ontology-guide/102 integration in an rdev

## Changes

- use RC release of cellxgene-schema, which now uses COG

## Testing steps

- E2E dataset upload in rdev


Dockerfile.processing
```diff
@@ -8,21 +8,22 @@ RUN apt-get update && \
 	apt-get upgrade -y && \
 	apt install software-properties-common -y && \
 	add-apt-repository ppa:deadsnakes/ppa && \
-	apt-get install -y python3.10 python3-pip
+	apt-get update && apt-get install -y python3.10 python3.10-dev python3.10-venv && \
+    wget https://bootstrap.pypa.io/get-pip.py && \
+    python3.10 get-pip.py
 
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1 \
```
Bento007, line 15: it sounds like this doesn't actually work. Id' try removing and see if everything still works.

Bento007: otherwise you should be able to use `python` instead of `python3.10` to start the interpreter

python_dependencies/processing/requirements.txt
```diff
@@ -1,21 +1,23 @@
...
 ddtrace==2.1.4
 furl
 jsonschema
-matplotlib==3.6.3 # 3.7.0 isn't compatible with scanpy: https://github.com/scverse/scanpy/issues/2411
+matplotlib==3.7.3
 moto>=5.0.0
 numba==0.56.2
 numpy==1.23.2
 owlready2==0.40.0
+packaging==21.0
```
Bento007, line 20: packaging, what for?

Bento007: hopefully this can also be removed.

PR #6889, opened 2024-04-02
fix: add observed=True to pandas groupby

ebezzi:
## Reason for Change

Fixes a bug introduced by the categorical change in the Census.


ebezzi:
Replaced by https://github.com/chanzuckerberg/single-cell-data-portal/pull/6891

Issue #6852, opened 2024-03-25
bug(Gene Expression): No collection link for multiple datasets


niknak33:
## Description

Gene Expression is showing a blank reference for a collection under certain datasets.

## To Reproduce

1. Go to Gene Expression
2. Search for Native Cell
3. Click on Source Data
4. See error - Notice a blank collection row under 8 datasets

## Expected behavior

Collection name and link should be available for user to select

Screenshot attached. 
<img width="295" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMWdLQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--804e5182f9751df54465f7b9bed3c828721c4eb9/gene%20expression%20bug.JPG" alt="gene expression bug.JPG" />

seve:
This is a backend bug and not a frontend bug, it's due to a number of datasets not having metadata associated with them in the cube. This should be addressed with Joyce's PR that will migrate cellguide to schema 5.0 #6856 

@niknak33 should I introduce some failsafe in case this happens again some time in the future? Should I group all these datasets in a row that says: "Failed to get dataset metadata", or maybe just hide them?

PR #6890, opened 2024-04-03
fix: create render state for failed collection metadata fetching

seve:
#6852 

There's an edge case where collection metadata fails to load for a number of datasets. This adds a render state for that case instead of just rendering an empty row in the source data sidebar.

image


frontend/src/views/WheresMyGeneV2/components/InfoPanel/components/SourceData/style.ts
```diff
@@ -38,6 +38,12 @@ export const StyledLink = styled.a`
   font-weight: 400;
 `;
 
+export const StyledSpan = styled.span`
+  font-size: 14px;
+  font-weight: 400;
+  color: ${() => warning500};
```
tihuan, line 44: The following will do!

```suggestion
  color: ${warning500};
```

tihuan (approved):
Ooh I like it! Just one small code change otherwise LGTM thank you!

PR #6914, opened 2024-04-15
feat: wmg+cellguide cog integration

atarashansky:
## Reason for Change

- Integrate pipelines and backends with COG

## Testing steps

- Tests and fixtures updated
- QA'd locally
- Followup PR may be necessary if issues are observed in `staging`


python_dependencies/cellguide_pipeline/requirements.txt
nayib-jose-gloria, line 1: we should be able to get rid of the pronto and owlready2 reqs

python_dependencies/backend/requirements.txt
nayib-jose-gloria, line 1: we should be able to get rid of the owlready2 req

backend/wmg/pipeline/constants.py
```diff
@@ -35,3 +35,87 @@
 WMG_DATA_SCHEMA_VERSION = "v4"
 
 MAXIMUM_ADMISSIBLE_CENSUS_SCHEMA_MAJOR_VERSION = 2
+
+HIGH_LEVEL_TISSUES = [
```
nayib-jose-gloria, line 39: if this is the same list / order as this [one](https://github.com/chanzuckerberg/cellxgene-ontology-guide/blob/main/ontology-assets/tissue_general_list.json) used by census, you can fetch it via COG using [this](https://github.com/chanzuckerberg/cellxgene-ontology-guide/blob/main/api/python/src/cellxgene_ontology_guide/curated_ontology_term_lists.py) function, passing in `CuratedOntologyTermList.TISSUE_GENERAL` enum

tests/unit/cellguide_pipeline/fixtures/cell_metadata.json
```diff
@@ -1 +1 @@
...
... a memory phenotype including distinct surface markers and the ability to differentiate into an effector T cell upon antigen reexposure.", "synonyms": ["memory T lymphocyte", "memory T-cell", "memory T-lymphocyte"]}, "CL:0000898": {"name": "naive T cell", "id": "CL:0000898", "clDescription": "Mature T cell not yet exposed to antigen with the phenotype CCR7-positive, CD45RA-positive, and CD127-positive. This cell type is also described as being CD25-negative, CD62L-high and CD44-low.", "synonyms": ["naive T lymphocyte", "naive T-cell", "naive T-lymphocyte"]}, "CL:0001200": {"name": "lymphocyte of B lineage, CD19-positive", "id": "CL:0001200", "clDescription": "A lymphocyte of B lineage that is CD19-positive.", "synonyms": []}, "CL:1001599": {"name": "pancreas exocrine glandular cell", "id": "CL:1001599", "clDescription": "Glandular cell of exocrine pancreas epithelium. Example: pancreatic acinar cell, glandular cells in pancreatic canaliculi, glandular cells in pancreatic ducts.", "synonyms": []}, "CL:1001433": {"name": "epithelial cell of exocrine pancreas", "id": "CL:1001433", "clDescription": "An epithelial cell of the exocrine pancreas.", "synonyms": ["exocrine pancreas cell"]}}
```
nayib-jose-gloria, line 1: Noticed a diff here in synonyms. OWLs/OBOs have two synonym types--related Synonyms and exact Synonyms. Pronto does not distinguish between the two, while owlready2 does. For COG, it only looks at exact synonyms. However, I see that the previous version of this file also included related Synonyms. 

without further context, I think it makes sense to only count exact synonyms (for example: "CL:0000084" / T-Cell used to count both mature T Cell and immature T Cell as synonyms because they are "related synonyms", which seems misleading to me at a glance. It does not include these in the COG metadata file). But I can also patch COG to achieve parity by counting both if you think it does make sense to consolidate or don't want to make that call as part of this work.

nayib-jose-gloria (approved):
lgtm from cog integration perspective

backend/wmg/pipeline/primary_filter_dimensions.py
```diff
@@ -109,14 +109,13 @@ def list_grouped_primary_filter_dimensions_term_ids(
 
 def order_tissues(ontology_term_ids: list[str]) -> list[str]:
     """
-    Order tissues based on appearance in TissueMapper.HIGH_LEVEL_TISSUES. This will maintain the priority set in
+    Order tissues based on appearance in HIGH_LEVEL_TISSUES. This will maintain the priority set in
     that class which is intended to keep most relevant tissues on top and tissues that are related to be placed
     sequentially
     """
     ontology_term_ids = set(ontology_term_ids)
```
kaloster, line 116: nit - this is mutating the original input - it might be better practice to do something like:

`unique_ontology_term_ids = set(ontology_term_ids)` and work off of that, makes it slightly more readable and easier to debug if needs be

kaloster (approved):

PR #6933, opened 2024-04-23
chore: extract test + lint requirements into requirements-dev.txt

nayib-jose-gloria:
## Reason for Change

- Inflated docker image sizes due to installing dev-specific packages (i.e. pytest, coverage) in deployed images
- Easier to manage dev-specific dependencies if we consolidate versions + group them in one file

## Changes

- extract dev-specific dependencies into python_dependencies/common/requirements-dev.txt, removing them from other requirements.txt
- add new build arg INSTALL_DEV to Dockerfiles for backend, processing, wmg-processing, and cellguide that only installs dev-specific packages if docker build is incurred with INSTALL_DEV=true 
- only call INSTALL_DEV=true to build images during push-tests.yml
- update push-tests.yml to check for whether to rebuild images for unit-test steps that were checking for changed files but not using that value anywhere

## Testing steps

- unit tests + inspecting docker image build step in rdev deploy

### DIFF
--- a/python_dependencies/backend/requirements.txt
+++ b/python_dependencies/backend/requirements.txt
@@ -1,10 +1,8 @@
 alembic>=1.9, <2
 anndata==0.8.0
-allure-pytest<3
 Authlib==0.14.3
 boto3>=1.11.17
 botocore>=1.14.17
-click==8.1.7
 connexion[swagger-ui]==2.14.2
 coverage>=6.5.0
 dataclasses-json==0.6.4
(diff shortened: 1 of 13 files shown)


python_dependencies/backend/requirements.txt
```diff
@@ -1,10 +1,8 @@
 alembic>=1.9, <2
 anndata==0.8.0
-allure-pytest<3
 Authlib==0.14.3
 boto3>=1.11.17
 botocore>=1.14.17
-click==8.1.7
 connexion[swagger-ui]==2.14.2
 coverage>=6.5.0
```
Bento007, line 7: coverage can also be removed

Bento007 (changes requested):
Thank you for doing this. This has been a thorn in my side for while. Can you also add this new requirements file to be tracked by dependabot.

Bento007 (approved):

Issue #6930, opened 2024-04-22
[Implementation] Curators and Data Submitters do not want to cancel and restore long-term in-progress revisions during a schema migration


nayib-jose-gloria:
See background [here](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/6317)

Estimate: 3-4 weeks x 2 engineers + testing for regression 

Proposed Approach (Align with a Tech Spec):

- Add “is_migration_revision” flag to CollectionVersionTable
- During migration, make “is_migration_revision” True for all revisions opened by the schema_migrate.py script
-  If a collection has an open revision during a migration—Update schema_migration.py to both migrate the revision AND open a new revision with “is_migration_revision” is True, and migrate that as well 
- Update create_collection_version in business.py to allow creating second revision, ONLY IF coming from migrate script (pass flag?)
- Update publish_revisions to only publish revisions with “is_migration_revision” == True

We will NOT require changes to the Frontend. Two simultaneous revisions are a migration-exclusive event, is expected to be transient, and curation will not occur during migration. (per conversation with Brian/Jason)

PR #6943, opened 2024-04-24
feat: support additional "migration revision" if collection already has an open revision during a migration

nayib-jose-gloria:
## Reason for Change

- #6930

## Changes

- Add `is_migration_revision` flag to CollectionVersion Table in DB + CollectionVersion entity dataclass

- Updated persistence layer `add_collection_version` to support `is_migration_revision`

- Updated `create_collection_version` in `business.py` to allow creating second revision, ONLY IF `is_migration_revision` arg is True AND the canonical collection does not already have a revision with `is_migration_revision` set to True 
        - otherwise, throw errors
        - default arg for `is_migration_revision` is False, such that multiple revisions are not allowed by default
        - only pass `is_migration_revision` is True when `create_collection_version` is called from `schema_migrate.py` in the `"collection_migrate"` step. `create_collection_version` is called when processing a published collection passed into the `"collection_migrate"` step  from the `"gather_collections"` step. 

- Update `"gather collection"` logic in migration to:
          - only pass on published collections to next step ("`collection_migrate`") IF published collection does not have a revision open OR if it has an open revision with `is_migration_revision == False`. Otherwise, the published collection is filtered out of the output (i.e. we do not process collections that already have revisions-in-migration in "`collection_migrate`")
          -pass pre-existing revisions with `is_migration_revision == False` into "collection_migrate", with auto_publish set to False. This allows pre-existing revisions to be migrated without being published.
          - pass pre-existing revisions with `is_migration_revision == True` into "collection_migrate", with auto_publish set to True. This scenario occurs if we are running the same migration more than once (e.g. due to an error or cancelled run). This allows us to continue migrating revisions that were opened during previous migration runs. 

- Update publish_revisions batch job to only publish revisions with `is_migration_revision == True`. This avoids having to pass in a denylist, and it will only publish revisions created during the migration run.
- Update finalize_collection_version method in persistence (publish method) to set is_migration_revision to False as soon as a collection version is published.

Proposed Flow using is_migration_revision: 

image

## Testing steps

TODO: integration test in rdev


backend/layers/processing/schema_migration.py
```diff
@@ -60,20 +60,22 @@ def gather_collections(self, auto_publish: bool) -> List[Dict[str, str]]:
         """
         response = []
 
-        has_revision = set()
+        has_migration_revision = set()
```
Bento007, line 63: If this is a fresh migration, nothing should have a migration revision at this point since this is the gather step, right?

backend/layers/processing/schema_migration.py
```diff
@@ -60,20 +60,22 @@ def gather_collections(self, auto_publish: bool) -> List[Dict[str, str]]:
...
                 continue
 
             if collection.is_published():
                 # published collection without an active revision
                 _resp["can_publish"] = str(True)
             elif collection.is_unpublished_version():
                 # active revision of a published collection.
-                has_revision.add(collection.collection_id.id)  # revision found, skip published version
                 _resp["can_publish"] = str(False)
+                if collection.is_migration_revision:
```
Bento007, line 76: I don't think this flag would have been set yet since it isn't add until `collection_migrate`. Oh is this incase we rerun migration a second time? Add a comment to reflect the specific cases this is intended to cover.

Bento007 (commented):
The code looks fine, but for documentation sake, could you add a flow chart or something detailing the lifecycle of a revision, migration_revision, and published collection.

Shouldn't we set `is_migration_revision` to false or None after publishing?

backend/layers/processing/schema_migration.py
nayib-jose-gloria: correct

nayib-jose-gloria: this is for cases where we run the migration more than once (due to memory errors, cancelled run, etc.)

backend/layers/processing/schema_migration.py
nayib-jose-gloria: Yup, thats the case we want to cover with this. I can add a comment

nayib-jose-gloria:
> The code looks fine, but for documentation sake, could you add a flow chart or something detailing the lifecycle of a revision, migration_revision, and published collection.

Can do

> Shouldn't we set `is_migration_revision` to false or None after publishing?

I don't think so, it'd be potentially useful metadata to track whether a version was created by migration. Even if its not, once its got a published_at, it won't interfere with any checks involving is_migration_revision, since all of those only interact with unpublished versions. So it'd be neutral at worst to keep it.

Bento007 (commented):
When we open a revision on a collection that has the `is_migrated_revision` flag, will that flag be copied as well, and should it continue to be True?

nayib-jose-gloria:
> When we open a revision on a collection that has the `is_migrated_revision` flag, will that flag be copied as well, and should it continue to be True?

I think no, it should only be True if the revision is opened as part of a migration.

Now that you bring up that scenario though, it does make me reconsider whether we should just re-set it to False once a migration revision is published. I could go either way

Bento007:
to me it makes more sense to set the `is_migration_revision` to false in the context of a a migration as opposed to when a curator is start a revision outside of a migration senario.

nayib-jose-gloria:
> to me it makes more sense to set the `is_migration_revision` to false in the context of a a migration as opposed to when a curator is start a revision outside of a migration senario.

when a curator is starting a revision outside a migration scenario, we create a new collection version. So we're not re-assigning any values, its basically just metadata indicating whether a particular collection version was first created by a migration run or not.

Bento007:
[add_collection_version](https://github.com/chanzuckerberg/single-cell-data-portal/blob/234b2f6214a0865fb88fef1a2de63b76e9556cd0/backend/layers/persistence/persistence.py#L517-L518) does not copy over all of the fields. A new collection version will have a default `is_migration_revision` of `False` 👍.

Bento007:
The `is_migration_revision` flag is intended to be a temporary flag used during migrations, yet it will persist even after the migration is complete. Outside of a migration the flag being true could be confusing since the published collection is no longer considered a revision. False could also be potentially confusing. Can we set it to none after publish?

backend/database/versions/07_e31a29561f38_add_is_migration_revision_flag.py
```diff
@@ -0,0 +1,27 @@
...
+revision = "07_e31a29561f38"
+down_revision = "06_8bced1b1470b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "CollectionVersion",
+        sa.Column("is_migration_revision", sa.BOOLEAN(), nullable=True),
```
nayib-jose-gloria, line 21: note from Emanuele: rename to something that does not expose business logic terms ("migration" and "revision") to DB. Suggested: `is_auto_version`

tests/unit/backend/layers/business/test_business.py
Bento007, line 1: I think you are missing tests where:

Given a canonical collection with a published version and an unpublished version where `is_auto_version` is **False**, creating another unpublished collection version where `is_auto_verison` is **True** will succeed.

Given a canonical collection with a published version and an unpublished version where `is_auto_version` is **True**, creating another unpublished collection version where `is_auto_version` is **False** will succeed.

tests/unit/processing/schema_migration/test_collection_migrate.py
```diff
@@ -164,3 +164,23 @@ def test_no_datasets(self, schema_migrate_and_collections):
...
+        schema_migrate.schema_version = "0.0.0"
+        private = collections["private"][0]
+        published, revision = collections["revision"]
+        schema_migrate.business_logic.create_collection_version = Mock(
+            return_value=Mock(version_id=CollectionVersionId())
+        )
+
+        # only call create_collection_version if the collection is published
+        schema_migrate.collection_migrate(private.collection_id.id, private.version_id.id, False)
+        assert not schema_migrate.business_logic.create_collection_version.called
```
Bento007, line 180:
```suggestion
        schema_migrate.business_logic.create_collection_version.assert_not_called()
```

tests/unit/processing/schema_migration/test_collection_migrate.py
```diff
@@ -164,3 +164,23 @@ def test_no_datasets(self, schema_migrate_and_collections):
...
+
+        schema_migrate.collection_migrate(revision.collection_id.id, revision.version_id.id, False)
+        assert not schema_migrate.business_logic.create_collection_version.called
```
Bento007, line 183:
```suggestion
        schema_migrate.business_logic.create_collection_version.assert_not_called()
```

tests/unit/processing/schema_migration/test_collection_migrate.py
```diff
@@ -164,3 +164,23 @@ def test_no_datasets(self, schema_migrate_and_collections):
         assert "key_name" not in response
         assert response["collection_version_id"] == published.version_id.id
         assert response["execution_id"] == "test-execution-arn"
+
+    def test_create_migration_revision(self, schema_migrate_and_collections):
```
Bento007, line 168: This is three different independent tests. Split them up please

tests/unit/processing/schema_migration/test_gather_collections.py
```diff
@@ -6,19 +6,39 @@ def test_with_revision(self, schema_migrate_and_collections):
         # get_collections is called twice
         schema_migrate.business_logic.get_collections.side_effect = [[published], [revision]]
         _, response = schema_migrate.gather_collections(auto_publish=True)
+        assert len(response) == 2
         assert {
             "can_publish": "True",
             "collection_id": published.collection_id.id,
-            "collection_version_id": published.collection_id.id,
+            "collection_version_id": published.version_id.id,
```
Bento007, line 13: Thank you

tests/unit/processing/schema_migration/test_gather_collections.py
```diff
@@ -52,13 +72,15 @@ def test_with_auto_publish_false(self, schema_migrate_and_collections):
         schema_migrate, collections = schema_migrate_and_collections
         published_no_revision = collections["published"][0]
         private = collections["private"][0]
-        published_with_revision, revision = collections["revision"]
-        # get_collections is called twice
+        published_with_non_migration_revision, revision = collections["revision"]
+        published_with_migration_revision, migration_revision = collections["migration_revision"]
+        # get_collections is called thrice
```
Bento007, line 77: if `get_collections` is called thrice, then the side_effect, needs to be a list of len(3).
```suggestion
        # get_collections is called twice
```

Bento007 (changes requested):

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1769,19 +1791,47 @@ def test_create_collection_version_with_migration_revision_flag_succeeds(self):
...
+        existing_revision = self.business_logic.create_collection_version(
+            published_collection.collection_id, is_auto_version=False
+        )
+        auto_revision = self.business_logic.create_collection_version(
+            published_collection.collection_id, is_auto_version=True
+        )
 
-        with self.assertRaises(CollectionVersionException):
-            self.business_logic.create_collection_version(published_collection.collection_id, is_auto_version=True)
+        # The new version has a different version_id
```
Bento007, line 1817: nit: Use auto revision in the comments to keeps naming consistent. This is to avoid confusing it with the existing_revision.

tests/unit/backend/layers/business/test_business.py
```diff
@@ -1737,21 +1737,43 @@ def test_create_collection_version_ok(self):
...
         with self.assertRaises(CollectionVersionException):
-            self.business_logic.create_collection_version(published_collection.collection_id)
+            self.business_logic.create_collection_version(published_collection.collection_id, is_auto_version=False)
 
-    def test_create_collection_version_with_migration_revision_flag_succeeds(self):
+    def test_create_collection_version_multiple_auto_versions_fails(self):
         """
-        If is_auto_version is True, a collection version can be created if there is another
-        unpublished version with is_auto_version == False for that specific canonical collection
+        Given a canonical collection with a published version and an unpublished version with is_auto_version == True,
```
Bento007, line 1753: I agree with this behavior.

Bento007 (approved):
LGTM

PR #6971, opened 2024-04-29
chore: update workflows to run only when relevant changes have been detected

atarashansky:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️

### DIFF
--- a/.github/workflows/push-tests.yml
+++ b/.github/workflows/push-tests.yml
@@ -67,6 +67,7 @@ jobs:
   backend-unit-test:
     runs-on: ubuntu-22.04
     timeout-minutes: 30
+    if: contains(github.event.head_commit.modified, 'backend/')
     steps:
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/push-tests.yml
```diff
@@ -67,6 +67,7 @@ jobs:
   backend-unit-test:
     runs-on: ubuntu-22.04
     timeout-minutes: 30
+    if: contains(github.event.head_commit.modified, 'backend/')
```
nayib-jose-gloria, line 70: I like the idea--though we should also run tests if a dependency has been bumped for this container (same goes for all unit tests and their pertinent dependencies)

atarashansky: We might need to create a new workflow file to avoid skipping this test in main

Bento007 (changes requested):
This should be two separate PRs. One for the reassignment and the other for changing the workflow.

Are the backend tests so slow that this is problem that needs fixing now?

PR #6975, opened 2024-05-02
feat: cellguide marker gene api

atarashansky:
## Reason for Change

- We have had numerous users request the ability to bulk download all the computational marker genes available in CellGuide.

## Changes

- added `/cellguide/v1/marker_genes` `GET` endpoint.
  - Required parameter
    - Organism
  - Optional parameters
    - Tissue
    - Cell type
 - If only organism is provided, return a nested dictionary of tissues, cell types, and their genes.
 - If organism and tissue is provided, return a dictionary of cell types and their marker genes
 - If organism and tissue and cell type is provided, return a list of marker genes for that cell type in that tissue
 - If organism and cell type is provided, return a list of marker genes for that cell type agnostic to tissues

## Testing steps

- Added unit tests for the new API


backend/cellguide/api/v1/marker_genes/actions.py
```diff
@@ -0,0 +1,121 @@
...
+        .strip()
+    )
+    compressed_data = s3_provider.download_file(
+        bucket_name=bucket,
+        object_key=get_object_key(
+            object=f"{latest_snapshot_identifier}/{COMPUTATIONAL_MARKER_GENES_FOLDERNAME}/{MARKER_GENE_PRESENCE_FILENAME}"
+        ),
+    )
+    marker_gene_data = json.loads(gzip.decompress(compressed_data).decode("utf-8"))
+    data = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
```
kaloster, line 40: this is some inception type stuff right here 😁

kaloster (approved):
LGTM! 🚀

PR #6993, opened 2024-05-06
feat: CellGuide marker gene api refinements

atarashansky:
## Changes

- Updated `tissue` to `tissue_general` to indicate parity with Census `tissue_general`
- Augmented OpenAPI documentation
- Added CellGuide API documentation


frontend/doc-site/04__Analyze Public Data/4_3__CellGuide Documentation/4_3_1__CellGuide API.mdx
```diff
@@ -0,0 +1,34 @@
+# CellGuide API
+
+CellGuide currently has an experimental API that allows users to programmatically fetch the computational marker genes data underlying the marker gene tables in each CellGuide page.
```
ambrosejcarr, line 3: Given conversations with @pablo-gar and @niknak33 , I understand we believe it is likely that this API will change in the future as part of a larger initiative. If that is correct, I suggest we message the probability of change explicitly in addition to the experimental label.

atarashansky: >>> Please note that the CellGuide API is currently in an experimental phase and is subject to change as we continue to expand its offerings. Future versions of the API may introduce new endpoints or modify existing ones to enhance functionality and user experience. However, to ensure a seamless integration for our users, all previous versions of the API will remain operational even after new versions are released. This means that no immediate updates will be required from your side. Nonetheless, we encourage users to stay informed about new releases and improvements.

backend/cellguide/api/v1/marker_genes/actions.py
```diff
@@ -1,121 +1,23 @@
...
-
-    Parameters:
-        organism (str): The ontology ID or name of the organism.
-        tissue (Optional[str]): The ontology ID or name of the tissue. Default is None.
-        cell_type (Optional[str]): The ontology ID or name of the cell type. Default is None.
+    This function handles the retrieval of all CellGuide marker gene data.
 
     Returns:
-        Flask Response: JSON data of the marker genes for the requested parameters or a 404 error if not found.
+        Flask Response: JSON data of the marker genes for the requested parameters.
```
kaloster, line 13: That's not the case anymore, since we've removed the parameters?

backend/cellguide/api/common/data.py
```diff
@@ -0,0 +1,68 @@
...
+    else:
+        return d
+
+
+def _initialize_cellguide_marker_gene_dict():
+    bucket = CellGuideConfig().bucket
+    s3_provider = S3Provider()
+
+    latest_snapshot_identifier = (
+        s3_provider.download_file(bucket_name=bucket, object_key=get_object_key(object="latest_snapshot_identifier"))
```
kaloster, line 25: Discussion - does it make sense to add some error handling in case the snapshot download fails? I know there's some handling by the provider, but is it worth it to put this behind a try-catch and have some more graceful handling in your endpoint?

kaloster (commented):
Jus a couple of comments

backend/cellguide/api/common/data.py
atarashansky: We don't actually want graceful handling of the error - the endpoint should throw to indicate that it must be retried. Once the marker gene data cache is warmed up, then this entire codepath is skipped anyway.

atarashansky: https://github.com/chanzuckerberg/single-cell-data-portal/blob/b3ffb32476502277bee243e64e74f949fcfb62f9/backend/cellguide/common/providers/s3_provider.py#L67

This will raise an error in the endpoint

kaloster (approved):

Issue #6753, opened 2024-03-04
Add uns['spatial'] metadata to the Discover API


brianraymor:
# Design

**DRAFTING**

```JSON
"spatial": {
    "is_single": boolean
    "has_fullres": boolean,
}
```

See references:
[`is_single`](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#is_single)
[`has_fullres`](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#spatiallibrary_idimagesfullres) - `true` if the optional `uns['spatial'][library_id]['images']['fullres']` is present and `false` if not. 

**[GET /collections/{collection_id}](https://api.cellxgene.cziscience.com/curation/ui/#/Collection/backend.curation.api.v1.curation.collections.collection_id.actions.get)**

MUST be added to each dataset in `datasets`

**[GET /collections/{collection_id}/datasets/{dataset_id}](https://api.cellxgene.cziscience.com/curation/ui/#/Collection/backend.curation.api.v1.curation.collections.collection_id.datasets.dataset_id.actions.get)**

MUST be added

Bento007:
This will require updating the portal database with the new fields and updating the processing container to extract the spatial fields from the dataset.

nayib-jose-gloria:
@brianraymor For non-spatial datasets, should this field be excluded from the response? Or should we always include, and add an 'is_spatial' flag to this set of values? i.e. 
```
"spatial": {
    "is_spatial": boolean,
    "is_single": boolean
    "has_fullres": boolean,
}
```

atolopko-czi:
Primarily, I would stick with whatever pattern is currently in use wrt indicating optional data. 

But without that as a consideration, I would go with:
```
"spatial": {
    "is_single": boolean
    "has_fullres": boolean,
}
```
and
```
"spatial": null
```

PR #6997, opened 2024-05-07
feat: add spatial metadata to Discover API

nayib-jose-gloria:
## Reason for Change

- #6753 

## Changes

- add SpatialMetadata dataclass to entities.py, and add as an optional attribute to DatasetMetadata entity 
- add logic to determine is_single and has_fullres when processing datasets during ingestion, or set spatial to None if not spatial
- add new fields to curation API reshape functions so it surfaces in [GET /collections/{collection_id}](https://api.cellxgene.cziscience.com/curation/ui/#/Collection/backend.curation.api.v1.curation.collections.collection_id.actions.get) and [GET /collections/{collection_id}/datasets/{dataset_id}](https://api.cellxgene.cziscience.com/curation/ui/#/Collection/backend.curation.api.v1.curation.collections.collection_id.datasets.dataset_id.actions.get)
- document new fields in curation-api.yml

## Testing steps

- unit tests + rdev tests
- Example rdev collection with spatial properties: https://pr-6997-backend.rdev.single-cell.czi.technology/curation/v1/collections/7ebeaadd-1607-441a-a624-8576a024b150


MillenniumFalconMechanic (approved):

backend/layers/processing/process_validate.py
```diff
@@ -107,6 +108,23 @@ def populate_dataset_citation(
...
+
+        :param spatial_dict: the value of the 'spatial' key from the uns dict of an AnnData object
+        :return: SpatialMetadata object
+        """
+        is_single = spatial_dict.get("is_single")
+        has_fullres = False
+        # schema validation ensures nested 'fullres' key is only included when is_single is True
+        if is_single:
+            # schema validation ensures there can only be one other key in uns["spatial"] if "is_single" is True
+            library_id = [key for key in spatial_dict if key != "is_single"][0]
```
Bento007, line 123: This seems like a check that should be done by the validator and surface to it's consumers.  I know it's more work, but it is a good idea to avoid duplicating validation logic from one repo to another.

Bento007: True, this isn't really a check. My concern was introducing check logic into the ingest pipeline. Carry on.

Bento007 (approved):

PR #7020, opened 2024-05-10
fix: case where uns['spatial']['is_single'] is True and no library_id key

nayib-jose-gloria:
## Reason for Change

- Bug caught when processing a Slide-Seq-v2 dataset that has is_single: True and no library_id in uns['spatial']. Expected behavior: Should be valid, Actual Behavior: IndexError due to missing a library_id

## Changes

- modify check in extract_metadata step of DownloadValidate to account for cases where there are no library_ids, removed 1 bit of logic that was making (inaccurate) assumption about the uns shape post-validation
- added unit test to cover case

### DIFF
--- a/backend/layers/processing/process_validate.py
+++ b/backend/layers/processing/process_validate.py
@@ -118,7 +118,7 @@ class ProcessValidate(ProcessingLogic):
         is_single = spatial_dict.get("is_single")
         has_fullres = False
         spatial_library_ids = [key for key in spatial_dict if key != "is_single"]
-        # schema validation ensures there can only be max. one other key in uns["spatial"] if "is_single" is True
+        # schema validation ensures there can only be at max, one other key in uns["spatial"] if "is_single" is True
         library_id = spatial_library_ids.pop() if spatial_library_ids else None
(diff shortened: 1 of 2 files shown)


backend/layers/processing/process_validate.py
```diff
@@ -117,12 +117,11 @@ def get_spatial_metadata(self, spatial_dict: Dict[str, Any]) -> Optional[Spatial
...
         is_single = spatial_dict.get("is_single")
         has_fullres = False
-        # schema validation ensures nested 'fullres' key is only included when is_single is True
-        if is_single:
-            # schema validation ensures there can only be one other key in uns["spatial"] if "is_single" is True
-            library_id = [key for key in spatial_dict if key != "is_single"][0]
-            if "fullres" in spatial_dict[library_id]["images"]:
-                has_fullres = True
+        spatial_library_ids = [key for key in spatial_dict if key != "is_single"]
+        # schema validation ensures there can only be max. one other key in uns["spatial"] if "is_single" is True
```
Bento007, line 121:
```suggestion
        # schema validation ensures there can only be at max, one other key in uns["spatial"] if "is_single" is True
```

Bento007 (changes requested):
typo in comment

Bento007 (approved):

PR #7023, opened 2024-05-10
feat: cxg conversion script updates for uns["spatial"]

kaloster:
## Changes

- added `convert_uns_to_cxg_group()`  handler for all uns["spatial"] actions as an additional step in `make_cxg()` script
- created `SpatialDataProcessor` class to handle `cxg` conversion for `uns` and  spatial deep zoom asset generation and upload
- added unit test coverage

## Testing steps

- using a jupyter notebook - run `make_cxg()` against a local `h5ad` - verify generated assets are uploaded to s3 and that the generated cxg includes `uns` group.
- verified with different use cases with:
-- datasets with fullres spatial image
-- datasets with hires only spatial image
-- datasets without spatial embedding

## Notes for Reviewer
- the package `libvips` had to be added to backed and processing Docker in order to get `pyvips` library (that generates the deep zoom tiles) to work properly
- the need to store uns["spatial"] in the cxg is to support the mvp implementation of spatial in explorer, where that image is generated on the fly - once deepzoom is launched - the step can be removed from the cxg conversion script


backend/common/utils/cxg_generation_utils.py
```diff
@@ -31,6 +34,31 @@ def convert_dictionary_to_cxg_group(cxg_container, metadata_dict, group_metadata
...
+
+    with tiledb.open(array_name, mode="w", ctx=ctx) as metadata_array:
+        for key, value in metadata_dict.items():
+            if key == UNS_SPATIAL_KEY:
+                for object_id, content in value.items():
+                    object_filtered = spatial_processor.filter_spatial_data(content, object_id)
+                    spatial_processor.create_deep_zoom_assets(cxg_container, content)
+            else:
+                continue
+
```
kaloster, line 58: only store _spatial_ object in `cxg/uns`

backend/common/utils/cxg_generation_utils.py
```diff
@@ -52,7 +80,12 @@ def convert_dataframe_to_cxg_array(cxg_container, dataframe_name, dataframe, ind
     tdb_attrs = []
 
     for column_name, column_values in dataframe.items():
-        dtype, hints = get_dtype_and_schema_of_array(column_values)
+        # Cast 'in_tissue' column values as boolean to make it categorical
+        # https://github.com/chanzuckerberg/single-cell-explorer/issues/841
+        if column_name == "in_tissue":
+            dtype, hints = get_dtype_and_schema_of_array(column_values.astype(bool))
```
kaloster, line 86: this is to treat the [reported issue](https://github.com/chanzuckerberg/single-cell-explorer/issues/841) for `in_tissue` boolean column that sometimes stored as int. After casting it as bool it will appear as categorical in explorer

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+                    img = img.convert("RGB")
+                cropped_img = img.crop(self._calculate_aspect_ratio_crop(img.size))  # Crop the image
+                cropped_img.save(io.BytesIO(), format="JPEG", quality=90)  # Save or manipulate as needed
+                # Flip the image vertically due to explorer client rendering images upside down
+                flipped_img = cropped_img.transpose(Image.FLIP_TOP_BOTTOM)
+                return np.array(flipped_img)
+        except Exception as e:
+            logger.error(f"Error processing image: {e}")
+            raise Exception(f"Error processing image: {e}") from e
+
```
kaloster, line 95: @tihuan pleas take a look here to make sure it's all there 🙇🏽‍♂️

backend/layers/processing/h5ad_data_file.py
```diff
@@ -112,7 +116,11 @@ def is_valid_embedding(adata, embedding_name, embedding_array):
                 * with all values finite or NaN (no +Inf or -Inf)
             """
 
-            is_valid = isinstance(embedding_name, str) and embedding_name.startswith("X_") and len(embedding_name) > 2
+            is_valid = (
+                isinstance(embedding_name, str)
+                and (embedding_name.startswith("X_") or embedding_name == "spatial")
+                and len(embedding_name) > 2
```
kaloster, line 122: valid embedding either starts with _X__ or is _spatial_

Dockerfile.backend
```diff
@@ -6,7 +6,7 @@ ENV LC_ALL=C.UTF-8
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y python3 libhdf5-dev python3-h5py gettext moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm jq && \
+    apt-get install -y python3 libhdf5-dev python3-h5py gettext moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm jq libvips && \
```
Bento007, line 9: are you sure libvips is needed for the backend. The "backend" container refers to the the server hosting the API, I don't see any API calls or share business logic that uses pylibvips. I even think the `backend/common/utils/spatial.py` could live with in `backend/layers/processing/spatial.py`

Bento007: NVM, this isn't your fault. There is a lot of stuff in the `backend/common/utils` that needs to be moved elsewhere. If you could please move all code that depends on libvips and pylibvips under `backend/layers/processing/`. We should avoid adding unnecessary dependencies to containers if we can avoid it.

backend/common/utils/cxg_generation_utils.py
```diff
@@ -31,6 +34,31 @@ def convert_dictionary_to_cxg_group(cxg_container, metadata_dict, group_metadata
...
+    with tiledb.open(array_name, mode="w", ctx=ctx) as metadata_array:
+        for key, value in metadata_dict.items():
+            if key == UNS_SPATIAL_KEY:
+                for object_id, content in value.items():
+                    object_filtered = spatial_processor.filter_spatial_data(content, object_id)
+                    spatial_processor.create_deep_zoom_assets(cxg_container, content)
+            else:
+                continue
+
+            metadata_array.meta[key] = pickle.dumps(object_filtered)
```
Bento007, line 59: nit: drop the `else: continue` and move `metadata_array.meta[key] = pickle.dumps(object_filtered)` in side `if key == UNS_SPATIAL_KEY:`

Dockerfile.backend
kaloster: for some reason I needed it inside the backend container when running unit tests locally - when I pushed the code I had to add it to the processing container in order to get unit tests to pass

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+    def _calculate_aspect_ratio_crop(self, image_size):
+        """
+        Calculate the aspect ratio crop coordinates for an image.
+
+        Args:
+            image_size (int, int): The size of the image (width, height).
+
+        Returns:
+            tuple: The crop coordinates (left, upper, right, lower).
```
Bento007, line 46: what is the expected type int, float, or both?

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+                # Convert image to RGB mode if it's not already
+                if img.mode != "RGB":
+                    img = img.convert("RGB")
+                cropped_img = img.crop(self._calculate_aspect_ratio_crop(img.size))  # Crop the image
+                cropped_img.save(io.BytesIO(), format="JPEG", quality=90)  # Save or manipulate as needed
+                # Flip the image vertically due to explorer client rendering images upside down
+                flipped_img = cropped_img.transpose(Image.FLIP_TOP_BOTTOM)
+                return np.array(flipped_img)
+        except Exception as e:
+            logger.error(f"Error processing image: {e}")
```
Bento007, line 93:
```suggestion
            logger.exception(f"Error processing image")
```

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+            raise Exception(f"Error processing image: {e}") from e
```
Bento007, line 94: Is there a reason we are raising a generic Exception from the previous exception?
```suggestion
            raise
```

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+    def _upload_assets(self, assets_folder):
+        """
+        Upload the deep zoom assets to the S3 bucket.
+
+        Args:
+            assets_folder (str): The folder containing the assets.
+        """
+        s3_uri = f"s3://{self.bucket_name}/{self.asset_directory}/{assets_folder}"
+        self.s3_provider.upload_directory(assets_folder, s3_uri)
+        shutil.rmtree(assets_folder)
```
Bento007, line 118: suggestion: rather than call this explicitly, use a [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) context manager. It will cleanup the files when exiting the context. I find to be cleaner.

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+
+        """
+        try:
+            assets_folder = container_name.replace(".cxg", "") + "/"
+            image_array = self._prepare_image(content)
+            processed_image = self._process_and_flip_image(image_array)
+            self._generate_deep_zoom_assets(processed_image, assets_folder)
+            self._upload_assets(assets_folder)
+        except Exception as e:
+            logging.error(f"Failed to create and upload deep zoom assets: {e}")
```
Bento007, line 136:
```suggestion
            logging.exception(f"Failed to create and upload deep zoom assets")
```

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+            raise Exception(f"An error occurred while creating and uploading deep zoom assets: {e}") from e
```
Bento007, line 137:
```suggestion
            raise
```

Bento007 (changes requested):

tests/unit/backend/layers/utils/test_spatial_assets_utils.py
Bento007, line 1: I recommend moving these tests under *tests/unit/processing*, that way they run on the processing container which will have the pyvips requirement

python_dependencies/backend/requirements.txt
```diff
@@ -28,6 +28,7 @@ PyMySQL==1.1.0
 pydantic>=1.10.7, <2
 python-jose[cryptography]>=3.1.0
 python-json-logger
+pyvips==2.2.2
```
Bento007, line 31: is this only added for the tests?

tests/unit/backend/layers/utils/test_spatial_assets_utils.py
```diff
@@ -0,0 +1,251 @@
...
+        expected_flipped_array = np.flipud(test_image_array)
+
+        flipped_image_array = self.spatial_processor._process_and_flip_image(test_image_array)
+        assert np.array_equal(flipped_image_array, expected_flipped_array)
+
+    def test__crop_to_aspect_ratio(self):
+        """
+        Test the cropping method to ensure the image is cropped to a square
+        """
+        width, height = 20, 10
```
Bento007, line 89: does it also work with odd numbers?

Bento007 (changes requested):

backend/common/utils/spatial.py
kaloster: image_size will always be int because the image.shape will return (int, int)

the return value should be int, I've pushed an update to cast the returned tuple as int

tests/unit/backend/layers/utils/test_spatial_assets_utils.py
kaloster: Ok, will do. The reason it's here is because other cxg conversion tests reside under backend but it does make sense to move it

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+        Generate deep zoom assets from the image array.
+
+        Args:
+            image_array (np.ndarray): The image array.
+            folder_name (str): The name of the folder to save the assets.
+        """
+        h, w, bands = image_array.shape
+        linear = image_array.reshape(w * h * bands)
+        image = pyvips.Image.new_from_memory(linear.data, w, h, bands, "uchar")
+        image.dzsave(folder_name + "spatial", suffix=".jpeg")
```
tihuan, line 107: can we create a P1 performance ticket to see if we could use webp instead, so the tiles we store and download will be smaller 🙏 ?

backend/common/utils/spatial.py
tihuan: I see both cropping with 1:1 aspect ratio and flipping the image 👍 LGTM thanks so much!

tihuan (commented):
Thanks so much, Ronen!!

tests/unit/backend/layers/utils/test_spatial_assets_utils.py
kaloster: yes, I've revised the test to demonstrate that

backend/common/utils/spatial.py
kaloster: I've updated to save as WEBP

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+    def _generate_deep_zoom_assets(self, image_array, folder_name):
+        """
+        Generate deep zoom assets from the image array.
+
+        Args:
+            image_array (np.ndarray): The image array.
+            folder_name (str): The name of the folder to save the assets.
+        """
+        h, w, bands = image_array.shape
+        linear = image_array.reshape(w * h * bands)
```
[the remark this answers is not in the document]
kaloster: `image_array.reshape()` takes a 3D image_array and turns it to a 1D array (that contains all the pixel values in a linear sequence - hence `linear`) that is then used by `pyvips.Image.new_from_memory(linear.data, w, h, bands, "uchar")` which expects a linear memory buffer of pixel data and creates a new image from the reshaped image_array

backend/common/utils/spatial.py
```diff
@@ -0,0 +1,158 @@
...
+
+        Args:
+            content (dict): The content dictionary containing the image array.
+
+        Returns:
+            np.ndarray: The prepared image array.
+        """
+        resolution = "fullres" if "fullres" in content["images"] else "hires"
+        image_array = content["images"][resolution]
+        image_array_uint8 = np.uint8(image_array * 255 if resolution == "hires" else image_array)
```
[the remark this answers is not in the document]
kaloster: `hires` image array pixel intensity values are commonly normalized by floating point value range [0, 1] 
`fullres` values range [0, 255] - 0 representing black and 255 white or max intensity of a color channel
- since we are passing uint8 (8-bit unsigned integer) array to PIL, this line upscales `hires` data to full range between 0 and 255, to ensure that `hires` array fits the data type of uint8
'fullres` is already in the desired range so there's no need to upscale.

for example, if a `hires` image array has a value of 0.5:

0.5 * 255 = 127.5

Converting to uint8: np.uint8(127.5) = 127

Ensuring that values are in the range [0, 255] for uint8.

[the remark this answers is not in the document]
brianraymor: @kaloster - both [fullres](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#spatiallibrary_idimagesfullres) and [hires](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/5.1.0/schema.md#spatiallibrary_idimageshires) are _normalized_  per:
`The numpy.ndarray.dtype MUST be numpy.uint8
`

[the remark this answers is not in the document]
tihuan: Thanks, @brianraymor ! Does that mean both hires and fullres need `* 255`?

backend/common/utils/spatial.py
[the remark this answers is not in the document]
tihuan: ah brilliant! Thanks so much for the explanation 🙏

backend/common/utils/spatial.py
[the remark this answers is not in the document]
brianraymor: > Does that mean both hires and fullres need * 255?

No, it means that neither need `* 255` because both must be unsigned ints rather than floats.

backend/common/utils/spatial.py
tihuan: Thanks so much! I just realized that we will also need to update the generated dzi file to specify Format "webp" instead of "jpeg"

For example, right now a dzi file looks like:

```dzi
<Image xmlns="http://schemas.microsoft.com/deepzoom/2008"
  Format="jpeg"  // <--- this needs to be webp
  Overlap="1"
  TileSize="254"
  >
  <Size 
    Height="1956"
    Width="1955"
  />
</Image>
````

File: python_dependencies/common/requirements-dev.txt [hunk 18]
```diff
@@ -4,5 +4,6 @@ click
 coverage
 parameterized
 pytest
+pytest-mock
```
Bento007, line 7: cool, i didn't know this existed.

backend/common/utils/spatial.py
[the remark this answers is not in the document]
kaloster: Ok thanks - this was based on the current data I was testing this with - will remove the `_prepare_image()` method

tests/unit/processing/test_spatial_assets_utils.py
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    result = spatial_processor.filter_spatial_data(valid_spatial_data, library_id)
+
+    assert result == expected_output, f"Expected {expected_output}, but got {result}"
+
+
+def test__invalid_input_metadata_copy(setup_spatial_processor):
+    """
+    Test case to verify that a KeyError is raised when passing input metadata
```
Bento007, line 87:
```suggestion
    Test case to verify that a KeyError is raised when passing invalid input metadata
```

tests/unit/processing/test_spatial_assets_utils.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    spatial_processor = setup_spatial_processor["spatial_processor"]
+    invalid_uns = setup_spatial_processor["invalid_uns"]
+    library_id = setup_spatial_processor["library_id"]
+    with pytest.raises(KeyError):
+        spatial_processor.filter_spatial_data(invalid_uns, library_id)
+
+
+def test__full_high_res_image_present(setup_spatial_processor):
```
Bento007, line 96: split into at least two separate tests. One for full res and the other hires.

backend/common/utils/spatial.py
kaloster: this is how it looks after the recent update:

```
<?xml version="1.0" encoding="UTF-8"?>
<Image xmlns="http://schemas.microsoft.com/deepzoom/2008"
  Format="webp"
  Overlap="1"
  TileSize="254"
  >
  <Size 
    Height="19720"
    Width="19720"
  />
</Image>
```

tests/unit/processing/test_spatial_assets_utils.py
```diff
@@ -0,0 +1,315 @@
...
+    with pytest.raises(KeyError):
+        image_array_uint8 = spatial_processor._prepare_image(non_valid_uns)
+
+
+def test__process_and_flip_image(setup_spatial_processor):
+    """
+    Test the image processing method
+    """
+    spatial_processor = setup_spatial_processor["spatial_processor"]
+    test_image_array = (np.random.rand(10, 10, 3) * 255).astype(np.uint8)
```
Bento007, line 122: nit: It helps to have non-random values for reproducible tests. Is there a specific reason we are using random? Maybe you could create a pytest fixture with one large ndarray, and take slices of that for testing.

backend/common/utils/spatial.py
tihuan: Oh right, it generates the image files AND the manifest, so it makes sense that it'll update both 😂 Thanks for sharing!

kaloster: https://cellxgene.dev.single-cell.czi.technology/spatial-deep-zoom/0bb15784-1cea-47e1-9a00-57dcd127746c/spatial.dzi

tests/unit/processing/test_spatial_assets_utils.py
```diff
@@ -0,0 +1,315 @@
...
+    flipped_image_array = spatial_processor._process_and_flip_image(test_image_array)
+    assert np.array_equal(flipped_image_array, expected_flipped_array)
+
+
+def test__crop_to_aspect_ratio(setup_spatial_processor):
+    """
+    Test the cropping method to ensure the image is cropped to a square
+    """
+    spatial_processor = setup_spatial_processor["spatial_processor"]
+    width, height = 23, 11
```
Bento007, line 134: Does this also work with even numbers?

tests/unit/processing/test_spatial_assets_utils.py
```diff
@@ -0,0 +1,315 @@
...
+
+    output_folder = "test_output"
+    cxg_container = "test_container"
+    group_metadata_name = "uns"
+    ctx = None
+
+    mock_spatial_processor = mocker.MagicMock(spec=SpatialDataProcessor)
+    mock_spatial_processor.filter_spatial_data.return_value = valid_uns["spatial"]
+
+    return {
```
Bento007, line 45: Is all of this setup needed for each tests? For reusability could you break this down into small fixture: 1 to reduce the setup overhead for each test, and to make it clear exactly what fixture are need for each test to run successfully.

Bento007 (changes requested):

tests/unit/processing/test_spatial_assets_utils.py
kaloster: I agree, but for now I will leave it with random values - array for full res is huge and can go up to 2 GB in size.

tests/unit/processing/test_spatial_assets_utils.py
Bento007: you can use [@pytest.mark.parameterize](https://docs.pytest.org/en/7.4.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) to test that both work using the same test.

kaloster: added!

backend/common/utils/cxg_generation_utils.py
```diff
@@ -31,6 +34,29 @@ def convert_dictionary_to_cxg_group(cxg_container, metadata_dict, group_metadata
...
+    """
+
+    spatial_processor = SpatialDataProcessor()
+
+    array_name = f"{cxg_container}/{group_metadata_name}"
+    object_filtered = {}
+
+    tiledb.from_numpy(array_name, np.zeros((1,)))
+
+    with tiledb.open(array_name, mode="w", ctx=ctx) as metadata_array:
```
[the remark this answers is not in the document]
kaloster: @tihuan I believe the costly part mentioned is opening the h5ad, since were already doing it as part of the cxg conversion, we're just piggy backing/adding a step here for the `uns` conversion and generating deep zoom assets - @atarashansky

Bento007 (approved):

backend/layers/processing/h5ad_data_file.py
```diff
@@ -122,8 +131,12 @@ def is_valid_embedding(adata, embedding_name, embedding_array):
         tiledb.group_create(embedding_container, ctx=ctx)
 
         for embedding_name, embedding_values in self.anndata.obsm.items():
+            print(embedding_name)
             if is_valid_embedding(self.anndata, embedding_name, embedding_values):
-                embedding_name = f"{embedding_container}/{embedding_name[2:]}"
+                if embedding_name == "spatial":
+                    embedding_name = f"{embedding_container}/{embedding_name}"
+                else:
+                    embedding_name = f"{embedding_container}/{embedding_name[2:]}"
```
kaloster, line 139: if "spatial" no need to strip the first 2 characters

Issue #6877, opened 2024-03-28
Review and update schema references (5.0.0 -> 5.1.0)


brianraymor:
There are many references to the schema specification and/or its sections that must be reviewed for potential updates. In most cases, this will be a simple change - replacing "5.0.0" with "5.1.0".

For example, [search for 5.0.0 references](https://github.com/search?q=repo%3Achanzuckerberg%2Fsingle-cell-data-portal%205.0.0&type=code).

Ensure that the following are updated:

1. [Discover API](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/curation/api/curation-api.yml)
2. [Contribute and Publish](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/frontend/doc-site/032__Contribute%20and%20Publish%20Data.mdx)
3. [Download UX Modal](https://github.com/chanzuckerberg/single-cell-data-portal/blob/83383dea8103b2d9c33b6d8ca4c42cde15ea8d27/frontend/src/components/Collections/components/Dataset/components/DownloadDataset/components/Content/components/DownloadLink/components/CopyCaption/index.tsx#L6-L8)

Bento007:
complete

PR #7024, opened 2024-05-10
feat: 5.1.0 references

MillenniumFalconMechanic:
## Reason for Change

- #6877 

## Changes

- Updated schema references from 5.0.0 to 5.1.1 (excluding apps-specific constants; see [PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/7078))

## Testing steps

- Tested on local.

## Checklist 🛎️


backend/cellguide/pipeline/constants.py
```diff
@@ -15,4 +15,4 @@
 
 CELL_GUIDE_DATA_BUCKET_PATH_PREFIX = "s3://cellguide-data-public-"
 
-CELL_GUIDE_PINNED_SCHEMA_VERSION = "5.0.0"
+CELL_GUIDE_PINNED_SCHEMA_VERSION = "5.1.0"
```
nayib-jose-gloria, line 18: I remember you asking about this previously--after some more thought I think we should leave it at 5.0.0 and let Apps team handle when to bump it. 

We do want them to update as soon as possible but there may be a reason they want to delay

backend/wmg/data/constants.py
```diff
@@ -14,4 +14,4 @@
     "EFO:0010010": "CEL-seq2",
 }
 
-WMG_PINNED_SCHEMA_VERSION = "5.0.0"
+WMG_PINNED_SCHEMA_VERSION = "5.1.0"
```
nayib-jose-gloria, line 17: Same as Cell Guide pinned version--after some more thought I think we should leave it at 5.0.0 and let Apps team handle when to bump it.

nayib-jose-gloria (commented):
open thought, not a new requirement: I wonder if we should consider maintaining a "latest" branch / tag in single-cell-curation that points to the latest schema reference or something rather than having to bump these links every time. Not a huge priority but if we have time to think on that cc: @Bento007

Bento007:
refresh bot

Bento007:
refresh PR

nayib-jose-gloria (approved):

Issue #6924, opened 2024-04-18
Upon Publishing a Collection, Pull Any Updates to DOI metadata from Crossref API 


nayib-jose-gloria:
See Discovery [Issue](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell/642) for Context + Motivation

Estimate: 1-2 weeks x 1 engineer + 1 reviewer

* On publish of a collection, call Crossref API and:
  * if the DOI has changed (from pre-print to published) trigger an update of collection artifacts and raise an exception to force curators to re-publish once the artifact update is complete. ([See artifact update how-to](https://github.com/chanzuckerberg/single-cell-data-portal/pull/7108#discussion_r1621457858).)
  
  * if the DOI has not changed,  check the “deposited” date returned from Crossref against the canonical collection's  “revised_at” date (or the collection version's “published_at” date if "revised_at" is null) if publishing a revision, or against the collection's "created_at" date if publishing a private collection. If “deposited” is more recent, update collection metadata pulled from crossref (i.e. get published DOI if preprint and its been published, get updated author info). 

* Can reuse existing functions for extracting collection metadata from Crossref.
* As a result, any DOI updates will regularly be picked up during migrations

PR #7108, opened 2024-05-24
feat: update publisher metadata on publish revision

MillenniumFalconMechanic:
## Reason for Change

- #6924 

## Changes

- Updated publish of revision to check Crossref for updates in metadata.

## Testing steps

- Unit tests.
- rdev.


backend/layers/business/business.py
```diff
@@ -972,6 +972,18 @@ def publish_collection_version(
             if canonical_datasets != version_datasets:
                 has_dataset_revisions = True
 
+            # Update publisher metadata if there are updates in Crossref that have occurred since last publish.
+            doi_link = next((link for link in version.metadata.links if link.type == "DOI"), None)
```
Bento007, line 976: For my own understanding, this returns the first value in the iterator that matches `if link.type == "DOI"` or `None`?

Bento007 (approved):
Looks great. I especially appreciate the comments explaining the tests.

backend/layers/business/business.py
MillenniumFalconMechanic: That's right!

backend/layers/business/business.py
```diff
@@ -972,6 +972,18 @@ def publish_collection_version(
...
+            if doi_link:
+                # Don't track/raise errors here; just keep the existing publisher metadata.
+                publisher_metadata, _, deposited_at_timestamp = self._get_publisher_metadata(doi_link.uri, [])
```
nayib-jose-gloria, line 979: to update DOI if there is a new DOI, get that 2nd return value (doi_curie_from_crossref)

only if `f"https://doi.org/{doi_curie_from_crossref}" !=  doi_link.uri` (that is, we passed in a pre-print DOI but there is now a published DOI available that we've fetched in _get_publisher_metadata) then--build CollectionMetadataUpdate object with just the 'links' field (pass in entire list of links, replacing `doi_link.uri` in the list with `f"https://doi.org/{doi_curie_from_crossref}`). Pass into update_collection_version with `apply_doi_links=True. `. Do NOT run
self.database_provider.finalize_collection_version if this path is invoked, instead raise a CollectionPublishException explaining that the DOI was updated and artifact updates are required and to retry after updates. User will need to attempt re-publish once the artifact updates invoked by update_collection_version complete.
 
Otherwise, if doi_curie_from_crossref is not new, continue with the logic that's been added here (looks good!)

nayib-jose-gloria: Oh, and if DOI is updated, update_collection_version will also take care of updating the publisher_metadata. But the logic here looks good for the scenario where DOI is not updated but there is new publisher metadata

backend/layers/business/business.py
```diff
@@ -972,17 +972,8 @@ def publish_collection_version(
...
-                # Don't track/raise errors here; just keep the existing publisher metadata.
-                publisher_metadata, _, deposited_at_timestamp = self._get_publisher_metadata(doi_link.uri, [])
-                if (
-                    publisher_metadata
-                    and deposited_at_timestamp
-                    and datetime.fromtimestamp(deposited_at_timestamp) > date_of_last_publish
-                ):
-                    self.database_provider.save_collection_publisher_metadata(version_id, publisher_metadata)
+            # Check Crossref for updates in publisher metadata since last publish.
+            self._update_revision_publisher_metadata(version, date_of_last_publish)
```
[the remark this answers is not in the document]
nayib-jose-gloria: or update_crossref_metadata to be more generic

backend/layers/business/business.py
```diff
@@ -1099,6 +1090,57 @@ def delete_artifacts(self, artifacts: List[DatasetArtifact]) -> None:
...
+                description=None,
+                contact_name=None,
+                contact_email=None,
+                links=updated_links,
+                consortia=None,
+            )
+            self.update_collection_version(version.version_id, update, True)
+
+            # Curators will need to re-publish once artifact updates are complete.
+            raise CollectionPublishException(
```
nayib-jose-gloria, line 1132: nit: since this function is extracted from the publish function now, we probably shouldn't raise a publish-specific exception directly in it. That way it can be cleanly re-used in other contexts. 

What do you think of returning a bool like "has_updated_doi", where True means we throw the exception in publish_collection_version

backend/layers/business/business.py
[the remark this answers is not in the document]
MillenniumFalconMechanic: Hey @nayib-jose-gloria, double-checking the requirement here as I was under the impression this check is specific to revisions. The [Upon Publishing a Revision, Pull Any Updates to DOI metadata from Crossref API](https://github.com/chanzuckerberg/single-cell-data-portal/issues/6924) ticket states:

> If a collection version is a revision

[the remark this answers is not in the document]
nayib-jose-gloria: it's more pertinent for revisions because those are the ones likely to have a DOI change over time (vs. a short-lived initial private collection). But `publish_collection_version` is run on both initial private collections and revisions, so it seems reasonable to cover both cases since it has to go through this code path regardless. I don't believe it would require any logical changes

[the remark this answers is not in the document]
MillenniumFalconMechanic: Got it. The call to `_update_revision_publisher_metadata` is currently nested in an "if revision" block so I'll pull it out of there so it is always executed. Two follow-on questions:

1. Should I compare the Crossref `deposited` date against the collection version `created_at` for private collections?
2. Can I go ahead and update the ticket with these updated requirements?

backend/layers/business/business.py
MillenniumFalconMechanic: I'll need the new DOI returned from Crossref in the calling function for the error message. The function could return `Optional[Tuple[str, str]]` so `existing_doi, crossref_doi` if there is a change in DOI, otherwise `None`. Thoughts?

backend/layers/business/business.py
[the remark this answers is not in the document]
nayib-jose-gloria: 1. Not required, but a good idea if you have time to go ahead and implement that. 
2. Sure

nayib-jose-gloria (approved):
looks great, thank you! have you had a chance to test in rdev? 

I think if we find a recent preprint DOI (that has since been published) and add it to one of the pre-populated private rdev collections on purpose, then publish--it should trigger the DOI update workflow. 

To test the publisher-metadata only workflow, I think we'll need to mock some data

MillenniumFalconMechanic:
Yes, I tested the following on rdev:

- Publish private collection with no DOI
- Publish private collection with unchanged DOI
- Publish private collection with DOI moved from pre-print to published
- Publish private collection with DOI that returns a deposited date after created date of collection (required DB update)
- Publish revision with no DOI
- Publish revision with unchanged DOI
- Publish revision with DOI moved from pre-print to published
- Published revision with DOI that returns a deposited date after published date collection (required DB update)

PR #7114, opened 2024-05-27
chore(tech): refactor census snapshot into common to disentangle DE, WMG, and CG

atarashansky:
## Reason for Change

- #7115 

## Changes

- Update dockerfiles to remove overlapping code dependencies
- Remove WMG v1 as the code refactoring introduced breaking changes there that are not necessary to maintain
- Move contents of `backend/wmg/data/` to `backend/common/census_cube/data/`
- Rename all instances of "WmgSnapshot" to "CensusCubeSnapshot", "WmgQuery" to "CensusCubeQuery", etc.
- Move the computational marker genes pipeline from CellGuide pipeline into a shared library
- For CellGuide, remove the dependency on `OntologyTreeBuilder` to get all the cell type IDs in the corpus, which simplifies the pipeline step interfaces. A new function "get_all_cell_type_ids_in_corpus" is added to a common utils file, which is used by the CellGuide pipeline and the now-common marker genes pipeline.
- Separate WMG and DE API configs
- Simplify QueryCriteria pydantic models - there is now a `BaseQueryCriteria` used for all filters and DE endpoints, a subclass CensusCubeQueryCriteria model used for expression summary cubes in WMG, and a MarkerGeneQueryCriteria used for the WMG FMG endpoint.
- And more...

## Testing steps

- Unit tests have only been updated with the refactored import statements. None of them fail, indicating a successful refactoring.
- Test pipeline runs using the fixture generation scripts for WMG and CellGuide also succeeded (though the updated fixtures were not committed to ensure bugs were not masked).


prathapsridharan:
@atarashansky - Thanks for putting this up. I haven't reviewed the code yet. I might suggest the following to add to this PR:

1. Since this is a big change already, update the packages for WMG as notified by dependabot
2. Add yourself as the owner for dependabot for WMG since my only involvement with this repo will be as reviewer of PR and not an active maintainer.

If you think the above is best handled in a separate PR then that is fine too. Just calling it out here.

prathapsridharan (approved):
LGTM

PR #7123, opened 2024-05-29
chore(tech): split out backends

atarashansky:
## Reason for Change
We have decided to split the data visualization (data viz) backends from the data portal (DP) backends to enhance the modularity and performance of our system. This restructuring has been a topic of discussion within the team for some time, and it offers several key advantages.

 - Optimized resource allocation: DP is primarily read-heavy; WMG is memory-intensive and has medium compute requirements; DE is compute-heavy and has medium memory requirements.
 - Limited blast radii: By isolating each backend, any performance degradation or server bottlenecking will be confined to the offending application, minimizing the impact on the overall system.
 - Modular dependency management: The dependencies for data visualization are quite different from those required by the data portal. By keeping these components separate, we can manage dependencies more effectively, leading to cleaner and more maintainable code.

## Changes

- added two new containers (`Dockerfile.backend_wmg` and `Dockerfile.backend_de`) and two new backends:
  - Rdev: `backend-de` and `backend-wmg` prefixes
  - Dev/Staging/Prod: `de-api` and `wmg-api` prefixes
- organized flask app configuration in `app.py` into a shared library. Each backend now gets its own (greatly simplified) `app.py` and `container_init.sh`
- Updated the infra to support the new backends.

## Testing steps

- All tests pass!
- Rdev was manually QA'd


python_dependencies/backend/de/requirements.txt
```diff
@@ -0,0 +1,8 @@
+pandas~=2.2.2
+scipy~=1.13.1
+pydantic~=1.10.7
+numba~=0.59.1
+cellxgene_ontology_guide~=1.0.0
+tiledb
```
ebezzi, line 6: I think those should be pinned, or you risk to run into a version incompatibility issue if the pipeline picks up a newer version.

.happy/terraform/envs/dev/main.tf
```diff
@@ -22,6 +24,14 @@ module stack {
   frontend_memory              = 4096
   backend_instance_count       = 4
   backend_workers              = 1
+  backend_de_instance_count     = 4
```
ebezzi, line 27: These are a bit verbose now - maybe grouping them into an object would make it cleaner.

ebezzi (approved):
LGTM.

PR #7131, opened 2024-05-30
fix: pass Sfn parameters using S3 during migration

Bento007:
## Reason for Change

- AWS Sfn impose a 256 KB limit on the output passed between steps. During schema migration this limit has been reached when spanning datasets and collections. To prevent this error the outputs are stored and read from S3.

## Changes

- do not pass any information between step that isn't actually used
  - Remove $.error from ResultPath. The error can still be found by looking at the Sfn in AWS, but they will not be included in the output passed between steps.
  - Remove $.result from ResultPath in SpanDatasets. The only value that was used from this output was CAN_PUBLISH. This value is now passed using S3 to the collection_publish step
- modify `SpanCollection` and `SpanDataset` in the AWS Sfn definition to read from a JSON file in S3. This it to prevent running into AWS's output limit.
- modified `gather_collections` and `collection_migrate` to store their output in S3 for the Sfn steps `SpanCollection` and `SpanDataset` to use. 
- remove `no_dataset` and instead check if there is a `key_name`. If no `key_name`, then no datasets can be implied and move on to publish
- return all of the responses in the migration steps to simply testing.
- log the data read from s3 before deleting the object. This is for debugging
- delete files in the schema execution id path after generating report. After a migration these artifact in S3 have no value.

## Testing steps

- run a migration in rdev an verify SFN runs succesfully.
- update unit tests


.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -353,10 +353,6 @@ resource aws_sfn_state_machine sfn_schema_migration {
                     "Name": "COLLECTION_VERSION_ID",
                     "Value.$": "$.collection_version_id"
                   },
-                  {
```
Bento007, line 356: This is passed using S3

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -512,7 +517,16 @@ resource aws_sfn_state_machine sfn_schema_migration {
           }
         }
       },
-      "ItemsPath": "$",
+      "ItemReader": {
```
Bento007, line 520: For SpanDatasets, get the list of datasets to process from S3.

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -484,14 +480,23 @@ resource aws_sfn_state_machine sfn_schema_migration {
...
+                      "ResultPath": null
                     }
                   ],
-                  "ResultPath": "$.result"
+                  "ResultPath": null
                 }
               }
             },
-            "ItemsPath": "$.datasets",
+            "ItemReader": {
```
Bento007, line 490: For SpanCollections, get the list of datasets to process from S3.

backend/layers/processing/schema_migration.py
```diff
@@ -278,6 +281,7 @@ def _store_sfn_response(self, directory: str, file_name: str, response: Dict[str
         self.logger.info(
             "Uploaded to S3", extra={"file_name": local_file, "bucket": self.artifact_bucket, "key": key_name}
         )
+        return key_name
```
Bento007, line 284: returning the key name of the s3 object stored.

backend/layers/processing/schema_migration.py
```diff
@@ -166,37 +169,37 @@ def collection_migrate(self, collection_id: str, collection_version_id: str, can
...
+                }
+                for dataset in datasets
+                if dataset.status.processing_status == DatasetProcessingStatus.SUCCESS
+                # Filter out datasets that are not successfully processed
+            ]
+            response_for_publish_and_cleanup = {
                 "can_publish": str(can_publish),
                 "collection_version_id": private_collection_version_id,
-                "collection_url": collection_url,
-                # ^^^ The top level fields are used for handling error cases in the AWS SFN.
```
Bento007, line 174: Now that we are using S3 this extra info does not need to be passed to `publish_and_cleanup`

nayib-jose-gloria (approved):

PR #7143, opened 2024-05-31
chore: add memory and timing details to response headers

atarashansky:
## Reason for Change

- Memory usage information is important to assess memory efficiency of requests and evaluate if our tasks have enough memory allocated to them

## Changes

- add memory usage and response time headers
- memory tracking is only enabled if the request header `"X-Enable-Memory-Tracking"` is set to `"true"`

## Testing steps

- Tested in rdev

### DIFF
--- a/backend/common/server/config.py
+++ b/backend/common/server/config.py
@@ -78,7 +78,7 @@ def register_routes(app, api_base_paths):
         g.track_memory = request.headers.get("X-Enable-Memory-Tracking", "false").lower() == "true"
         if g.track_memory:
             g.memory_usage = []
-            g.memory_thread = threading.Thread(target=_track_memory_usage)
+            g.memory_thread = threading.Thread(target=_track_memory_usage, args=(g,))
             g.memory_thread.start()
         app.logger.info(
             dict(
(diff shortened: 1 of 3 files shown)


Bento007 (commented):
Fill in the description. This should be behind a flag we can toggle.

atarashansky:
> This should be behind a flag we can toggle.

Why?

backend/common/server/config.py
```diff
@@ -71,8 +73,12 @@ def apis_landing_page() -> str:
         """
 
     def before_request():
-        g.start = time.time()
+        g.start_time = time.time()
+        g.memory_usage = []
         g.request_id = generate_request_id()
+        g.track_memory = True
+        g.memory_thread = threading.Thread(target=_track_memory_usage)
```
Bento007, line 80: Does this measure the memory usage of the threads or the processes. If it is measuring the thread memory is it measuring `g.memory_thread` or the memory usage of the request.

backend/common/server/config.py
```diff
@@ -85,19 +91,30 @@ def before_request():
...
         )
-        response.headers["X-Request-Id"] = get_request_id()
+        response.headers.add("X-Request-Id", get_request_id())
+        response.headers.add("X-Total-Process-Time", f"{total_time:.3f}s")
+        response.headers.add("X-Peak-Memory-Usage", f"{peak_memory:.3f} MiB")
         return response
 
+    def _track_memory_usage():
+        while g.track_memory:
+            g.memory_usage.append(memory_usage(-1, interval=0.25, timeout=0.25)[0])
```
Bento007, line 116: With the current interval, what happens if the request is faster than this sample rate?

atarashansky: If the request is faster than 250ms, then it almost certainly has a small memory footprint. As of the most recent commit, peak_memory would just be set to 0 if there were no samples:
```
peak_memory = max(g.memory_usage) if g.memory_usage else 0
```

backend/common/server/config.py
atarashansky: `memory_usage(-1,...)` tracks the memory usage of the current process. Threads (request thread and memory tracking thread) share the same process.

Bento007: the process has multiple threads running, so wont this value be the sum of all of the thread. So if multiple requests were being handles the memory usage would be for both threads.

atarashansky: Each gunicorn worker is on its own process, but they are asynchronous gevent workers so it's true that memory usage will include other asynchronous requests happening at the same time. I don't think there's a way around this though

Bento007 (approved):

PR #7156, opened 2024-06-03
feat: remove auto_publish from migration

nayib-jose-gloria:
## Reason for Change

- There is a preference to publish all migration revisions at once, after manual review by curators. This change cleans up migration code by removing the option to publish revisions as soon as one finishes migration, which also involved some logic to determine which types of collections should be allowed to auto-publish.

## Changes

- renamed all infra and functions referring to CollectionPublish or publish_and_cleanup to CollectionCleanup and log_errors_and_cleanup
- removed can_publish / auto_publish parameters passed into SFN outputs or migration functions
- removed logic in gather_collections relating to determining which collections can be auto published
- renamed tests and test fixtures, and removed now unnecessary tests

## Testing steps

- unit tests + integration test


Bento007 (commented):
Looks good so far

Bento007:
What else do you plan on adding?

Bento007 (approved):

PR #7158, opened 2024-06-04
chore: drop banners

atarashansky:
## Reason for Change

- Census banner is droppin' cookies so we are removing the banner everywhere.

## Changes

- Remove cellguide, GE, and landing page banners.


Bento007:
Is there an open issue associated with this?

Bento007 (approved):

Bento007:
Are there any test that need to be updated associated with these changes.

Issue #7159, opened 2024-06-04
Add the DOI label to GET /datasets


brianraymor:
## Context

See [**cell-sci-eng**](https://czi-sci.slack.com/archives/C023Q1APASK/p1716245770988479). If Census supports the the human readable name for the DOI url anchor link in the collection UI - like "Ahern et al. (2022) Cell" for https://doi.org/10.1016/j.cell.2022.01.012, then GE can stop using the Discover API and depend on the Census API which will simplify post-migration. 

## Design

Unfortunately, the name is being [calculated on the FE](https://github.com/chanzuckerberg/single-cell-data-portal/blob/f0fa3b8054fa2bb4a185b5b50a5e5d419a9f722b/frontend/src/common/queries/filter.ts#L564) from the collection `publisher_metadata` rather than being returned  in the **DP** API, so there's no BE code to reuse.

The **Discover API** must add information to [GET /datasets](https://api.cellxgene.cziscience.com/curation/ui/#/Dataset/backend.curation.api.v1.curation.datasets.actions.get) -  either the calculated name or the `publisher_metadata` (which would force Census to re-create the name). 

There's definitely an opportunity to refactor and add the calculated name to both the **DP** and **Discover API** and remove the FE code?

nayib-jose-gloria:
My design preference is to calculate the short_citation in the backend and return in DP & Discover API, then remove the calculated citation from the FE.

brianraymor:
To align with the [census schema](https://github.com/chanzuckerberg/cellxgene-census/issues/1179#issuecomment-2148093202), I'd recommend using the name - `collection_doi_label`.

Issue #6166, opened 2023-11-06
Cleanup crossref code


atolopko-czi:
## Motivation

Code hygiene makes for healthy engineering. By DRY. Delete dead code.

## Definition of Done

- All CrossRef code is in encapsulated in a single service

## Tasks

- [ ] Refactor this [code](https://github.com/chanzuckerberg/single-cell-data-portal/blob/bc584a92c7585b511e70dc8822a7836b9a54a0b0/backend/cellguide/pipeline/canonical_marker_genes/utils.py#L39), moving it [here](backend/layers/thirdparty/crossref_provider.py)
- [x] Delete this dead code implementation of [CrossrefProvider](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/backend/layers/thirdparty/crossref_provider.py) and its tests.

seve:
We'll soon be making crossref calls from the frontend, too. If this is a service with an API, that'd be fantastic

atolopko-czi:
Currently, it's just a library-level API (in the "business layer") that is used privately by the backend code, so it's not exposed via an HTTP API. But do add a new issue for that if needed! (maybe as part of upcoming epic that requires it)

brianraymor:
CC: @metakuni ^^^

danieljhegeman:
Handling the second task in #6414

danieljhegeman:
Moved back to Disorder after completing subtask named above. Refactor for Cellguide (first item) still remains to be completed.

brianraymor:
I'm adding to a larger crossref epic so this doesn't get lost.

PR #7166, opened 2024-06-06
feat: add summary citation to discover API + DP API

nayib-jose-gloria:
## Reason for Change

- #7159
- #6166

## Changes

- Calculate publication citation in the backend, in portal_api.py and in curation API common under "reshape_for_dataset_index", leveraging the publisher_metadata fetched from the DB in both cases.
       - Reuse code that the cellguide already had for doing this calculation, moving it to backend.common 
- Remove code that calculates publication citation in the frontend (buildSummaryCitation)
- Update the FE to support fetching the summary_citation from Collection object in the DP API response 
- Return the publication citation as "collection_doi_label" in Discover API datasets index endpoint (name suggested by Brian / Census team for their integration ease) 
- Return the publication citation as "summary_citation" in DP API (matches the name used in the DP frontend for this value) in both the collection index endpoint and single collection fetch endpoint (since 
- Refactor + move various util functions and the crossref provider into backend.common, so it can be used by both cellguide + data-portal (until now, both were using their own custom utils to do the same things)

## Testing steps

- unit test + integration tests in rdev


backend/common/citation.py
```diff
@@ -0,0 +1,52 @@
+def format_citation_dp(message: dict) -> str:
+    """
+    Formats the citation message.
+
+    Parameters
```
Bento007, line 5: what style docstring format is this?

Bento007 (approved):
Thank you for cleaning up the code in the process,

PR #7168, opened 2024-06-06
fix: reduce schema-migration span concurrency

Bento007:
## Reason for Change

- this avoid us spamming the batch queue with jobs that will state in runnable stable for days.
- It also reduces the chance of us running into rate limiting from AWS on batch API requests. 
- adjust the compute resources for dataset_migration to make better use compute.

## Changes

- modify SpanCollection Concurrency from 40 to 10
- modify SpanDataset Concurrency from 32 to 10
- reduce the memory and VCPUs used by dataset_migrate to 64000MB and 8 VCPUs.
- Add 300000MB of swap to dataset_migrate.

### DIFF
--- a/.happy/terraform/modules/schema_migration/main.tf
+++ b/.happy/terraform/modules/schema_migration/main.tf
@@ -44,15 +44,15 @@ resource aws_batch_job_definition schema_migrations_swap {
     resourceRequirements = [
       {
         type= "VCPU",
-        Value="8"
+        Value="32"
       },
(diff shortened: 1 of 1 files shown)


nayib-jose-gloria (approved):

nayib-jose-gloria (commented):
reverting memory requirements for now, I'd rather take the hit for this migration and keep things consistent for the prod run and revisit before the next one

PR #7185, opened 2024-06-12
fix: accept NoneType do_not_publish_list

nayib-jose-gloria:
## Reason for Change

- Small bug where passing in do_not_publish_list as None for a publish-revisions job caused an error; workaround is to pass in an empty string but we want to support None.

## Changes

- support publish-revision batch job submission without a DO_NOT_PUBLISH_LIST env var being set.

### DIFF
--- a/backend/layers/processing/publish_revisions.py
+++ b/backend/layers/processing/publish_revisions.py
@@ -92,8 +92,8 @@ class PublishRevisions(ProcessingLogic):
                     )
                     try:
                         self.business_logic.publish_collection_version(collection_version.version_id)
-                    except Exception:
-                        logging.exception("Failed to publish collection version.")
+                    except Exception as e:
+                        logging.exception(f"Failed to publish collection version. {e}")
 
 
 if __name__ == "__main__":
@@ -105,5 +105,7 @@ if __name__ == "__main__":
         S3Provider(),
         UriProvider(),
     )
-    do_not_publish_list = os.environ.get("DO_NOT_PUBLISH_LIST", None).split(",")
+    do_not_publish_list = os.environ.get("DO_NOT_PUBLISH_LIST", None)
+    if do_not_publish_list is not None:
+        do_not_publish_list = do_not_publish_list.split(",")
     PublishRevisions(business_logic, do_not_publish_list).run()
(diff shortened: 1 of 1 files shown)


backend/layers/processing/publish_revisions.py
```diff
@@ -92,8 +92,8 @@ def run(self):
                     )
                     try:
                         self.business_logic.publish_collection_version(collection_version.version_id)
-                    except Exception:
-                        logging.exception("Failed to publish collection version.")
+                    except Exception as e:
```
Bento007, line 95: This is redudant. [Logging.exception](https://docs.python.org/3/library/logging.html#logging.Logger.exception) includes the exception.

backend/layers/processing/publish_revisions.py
```diff
@@ -105,5 +105,7 @@ def run(self):
         S3Provider(),
         UriProvider(),
     )
-    do_not_publish_list = os.environ.get("DO_NOT_PUBLISH_LIST", None).split(",")
+    do_not_publish_list = os.environ.get("DO_NOT_PUBLISH_LIST", None)
```
Bento007, line 108: nit: your's works, just an alternative suggestion.
```suggestion
    do_not_publish_list = os.environ.get("DO_NOT_PUBLISH_LIST", "").split(",")
```

Bento007 (changes requested):
One thing to change.

backend/layers/processing/publish_revisions.py
```diff
@@ -92,8 +92,8 @@ def run(self):
                     )
                     try:
                         self.business_logic.publish_collection_version(collection_version.version_id)
-                    except Exception:
-                        logging.exception("Failed to publish collection version.")
+                    except Exception as e:
+                        logging.exception(f"Failed to publish collection version. {e}")
```
nayib-jose-gloria, line 96:
```suggestion
                    except Exception:
                        logging.exception("Failed to publish collection version.")
```

backend/layers/processing/publish_revisions.py
nayib-jose-gloria: This would result in a list of len 1 `[""]`. Which would work, but I just prefer having an explicit None when we don't set it

Bento007 (approved):

Issue #7190, opened 2024-06-13
bug(curation api): 504 returned from GET collection and datasets


Bento007:
## Describe the bug

When making requests to GET /v1/collection with and without an API Key, a 504 is returned about 50% of the time.

### Error
```FAILED

Gateway Time-out

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<HTML><HEAD><META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=iso-8859-1">
<TITLE>ERROR: The request could not be satisfied</TITLE>
</HEAD><BODY>
<H1>504 ERROR</H1>
<H2>The request could not be satisfied.</H2>
<HR noshade size="1px">
CloudFront attempted to establish a connection with the origin, but either the attempt failed or the origin closed the connection.
We can't connect to the server for this app or website at this time. There might be too much traffic or a configuration error. Try again later, or contact the app or website owner.
<BR clear="all">
If you provide content to customers through CloudFront, you can find steps to troubleshoot and help prevent this error by reviewing the CloudFront documentation.
<BR clear="all">
<HR noshade size="1px">
<PRE>
Generated by cloudfront (CloudFront)
Request ID: 2yXpvcNQKTOWmetxS34Ul-Vj8QK1J6TvOSvaVOx3THJOvMP2Y9ovzw==
</PRE>
<ADDRESS>
</ADDRESS>
</BODY></HTML>

x-request-id: None
```

## To Reproduce

`curl https://api.cellxgene.cziscience.com/curation/v1/collections`

or
`curl https://api.cellxgene.cziscience.com/curation/v1/datasets`

## Expected behavior

the endpoints should never return a 504 under proper working conditions.

## Environment

Curation API prod

## Additional context

- the cloudwatch request timeout is set to 60s
- Looking at the logs in cloud watch, request to GET collection are taking up 78 second which is greater than the cloud watch timeout.
- There is a [keep-alive](https://docs.gunicorn.org/en/stable/settings.html#keepalive) configuration in for the server set for [61](https://github.com/chanzuckerberg/single-cell-data-portal/commit/abcbcbc5bdd159a3fd412f457c0e536059efcc4d#diff-609995c52e95e7c7d346ed100be4c52ef17f853ada8931e97db7c1c061807442R34) second which could be related 
[...]

PR #7191, opened 2024-06-14
fix: optimize database queries

Bento007:
## Reason for Change

- Avoid calling database queries in for loops. This will reduce the amount of time spent making network calls from O(n) to constant time.
- https://github.com/chanzuckerberg/single-cell-data-portal/issues/7190

## Changes
- Modify persistance layer calls to avoid database calls within a for loop. The queries are instead batched together into a single call. In the future we may need to split up the calls into batches to avoid waiting for large queries to finish.  
- change print statements to logging statements

## Testing steps
- All unit tests pass.

## Note
- These change would benefit from load testing but we currently lack the tooling. This would help determine when or if we should start batch the requests to the database.


backend/curation/api/v1/curation/collections/common.py
```diff
@@ -195,11 +195,19 @@ def reshape_datasets_for_curation_api(
...
     active_datasets = []
+    dataset_version_ids = []
+    dataset_versions = []
     for dv in datasets:
-        dataset_version = get_business_logic().get_dataset_version(dv) if isinstance(dv, DatasetVersionId) else dv
+        if isinstance(dv, DatasetVersion):
+            dataset_versions.append(dv)
+        else:
+            dataset_version_ids.append(dv)
+    dataset_versions.extend(business_logic.database_provider.get_dataset_versions_by_id(dataset_version_ids))
```
nayib-jose-gloria, line 207: we should add an empty list check, to avoid opening a session / querying for nothing in case all are DatasetVersion instances

backend/layers/persistence/persistence.py
```diff
@@ -281,16 +281,10 @@ def get_dataset_versions_by_id(
...
-            if get_tombstoned:
-                canonical_dataset_query = session.query(DatasetTable).filter(DatasetTable.id.in_(canonical_ids))
-            else:
-                canonical_dataset_query = (
-                    session.query(DatasetTable)
-                    .filter(DatasetTable.id.in_(canonical_ids))
-                    .filter(DatasetTable.tombstone.is_(False))
-                )
+            canonical_dataset_query = session.query(DatasetTable).filter(DatasetTable.id.in_(canonical_ids))
+            if not get_tombstoned:
```
nayib-jose-gloria, line 285: nit: the vast majority of cases get_tombstoned will be False (maybe even every case in practice?) so I am not convinced we should break this into two queries

backend/layers/persistence/persistence.py
```diff
@@ -679,11 +686,16 @@ def finalize_collection_version(
...
-                datasets = session.query(DatasetTable).filter(DatasetTable.id.in_(dataset_ids_to_tombstone)).all()
-                for dataset in datasets:
-                    dataset.tombstone = True
-                    dataset_all_versions = session.query(DatasetVersionTable).filter_by(dataset_id=dataset.id).all()
-                    dataset_version_ids_to_delete_from_s3.extend([dv.id for dv in dataset_all_versions])
+                tombstone_dataset_statement = (
+                    update(DatasetTable).where(DatasetTable.id.in_(dataset_ids_to_tombstone)).values(tombstone=True)
+                )
+                session.execute(tombstone_dataset_statement)
+                dataset_all_version_ids = (
```
nayib-jose-gloria, line 693: the tests are passing so maybe i'm misunderstanding, but this query looks like we're getting List[DatasetVersion] rather than List[DatasetVersionId], no? And dataset_version_ids_to_delete_from_s3 expects Ids

backend/layers/persistence/persistence.py
Bento007: it's not a new query, it's just adding an additional filter to the existing query

nayib-jose-gloria: ah got it

backend/layers/persistence/persistence.py
Bento007: You're totally right, but that is how the original code was written. I change it to query for DatasetVersion.id and it works. Not sure how it managed to work before, because it was returning a row not a string.

nayib-jose-gloria (approved):
test in rdev and ensure no regressions but looks good to me!

Bento007:
I added some timing metric to the queries and the two slowed queries are in `get_all_mapped_collection_versions`. Fetching all of the canonical collections take about seconds at worst. Fetching all current collection versions to 20 seconds at worst.
https://github.com/chanzuckerberg/single-cell-data-portal/blob/0f8483789eb44c00fb30b31248102990a2ac9809/backend/layers/persistence/persistence.py#L475-L475

I also did some performance and found the endpoint on average took over 60 seconds to return

PR #7214, opened 2024-06-18
feat: update dataset title DP endpoint

MillenniumFalconMechanic:
## Reason for Change

- #7063 

## Changes

- Added update dataset logic to business layer.
- Added update dataset endpoint to DP API.

## Testing steps

- DP API unit tests 
- Business layer unit tests
- rdev

### DIFF
--- a/backend/layers/common/validation.py
+++ b/backend/layers/common/validation.py
@@ -4,7 +4,7 @@ from urllib.parse import urlparse
 
 from backend.layers.business.entities import CollectionMetadataUpdate
 from backend.layers.business.exceptions import InvalidMetadataException
-from backend.layers.common.entities import CollectionMetadata, Link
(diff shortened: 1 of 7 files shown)


backend/layers/common/validation.py
```diff
@@ -29,6 +29,33 @@
 }
 
 
+def _verify_field(
```
MillenniumFalconMechanic, line 32: Moved [existing](https://github.com/chanzuckerberg/single-cell-data-portal/pull/7214/files#diff-fbff6622cc471e7a74db034a50d63b58f01e8b524aee67833ade64bb8fa53a6fL42) `check` method here for reuse, updated to include `or key == "title"`.

Bento007 (approved):

PR #7233, opened 2024-06-27
feat: update dataset title Curation endpoint

MillenniumFalconMechanic:
## Reason for Change

- #7062 

## Changes

- Added update dataset endpoint to Discover API.

## Testing steps

* Discover API unit tests
* rdev


nayib-jose-gloria (approved):

Bento007 (approved):
Manually verified that patching the datasets does not affect the published dataset

PR #7250, opened 2024-07-02
fix: update raw h5ad when editing dataset title

nayib-jose-gloria:
## Reason for Change

- Currently, the raw.h5ad is not updated on a dataset_metadata_update batch job. This is reasonable for updates to citation or schema_version, which are not fields on the raw H5AD (only labeled), but now that we support updates to dataset title (which is on the raw.h5ad as well), we need to update the dataset title on the raw.h5ad. This is because we use the raw.h5ad as the object to migrate during a migration; in the current state, we would accidentally roll back dataset title changes done since the last schema version bump.

## Changes

- add raw_h5ad_update function 
- only trigger raw_h5ad_update if a field being updated is contained in the raw h5ad (currently, of fields on the raw h5ad, we only support updating "title")
- otherwise, retain previous behavior (that is, copy existing raw_h5ad_uri to new dataset version id directory in the artifact bucket)

## Testing steps

- unit tests

### DIFF
--- a/backend/layers/processing/dataset_metadata_update.py
+++ b/backend/layers/processing/dataset_metadata_update.py
@@ -20,6 +20,7 @@ from backend.layers.common.entities import (
     DatasetConversionStatus,
     DatasetProcessingStatus,
     DatasetStatusKey,
+    DatasetUploadStatus,
     DatasetValidationStatus,
     DatasetVersion,
     DatasetVersionId,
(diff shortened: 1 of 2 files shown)


backend/layers/processing/dataset_metadata_update.py
```diff
@@ -56,6 +58,37 @@ def __init__(self, artifact_bucket: str, datasets_bucket: str) -> None:
...
+        self.create_artifact(
+            raw_h5ad_filename,
+            DatasetArtifactType.RAW_H5AD,
+            new_key_prefix,
+            new_dataset_version_id,
+            self.artifact_bucket,
+            DatasetStatusKey.H5AD,
+        )
+        self.update_processing_status(new_dataset_version_id, DatasetStatusKey.UPLOAD, DatasetUploadStatus.UPLOADED)
+        os.remove(raw_h5ad_filename)
```
Bento007, line 90: I'd add this to a try: finally: block to make sure it happens. A Context manager would also work.

Bento007 (changes requested):
one suggested change

Bento007 (approved):

PR #7259, opened 2024-07-05
chore(de+cellguide): add share url feature to DE and add link button to cellguide

atarashansky:
## Reason for Change

- Required feature for launching DE

## Changes

- Updated DE cross-filtering behavior to rollup the cell type values. For example, whereas before "pancreatic acinar cell" was a valid cell type when selecting mouse pancreas and "acinar cell" was not, now both are valid cell types.
- Added URL query parameters and filter initialization to DE
- Added the Open in Differential Expression button to CellGuide

## Testing steps

- e2e test added
- Tested locally

### DIFF
--- a/frontend/src/views/DifferentialExpression/components/Main/connect.ts
+++ b/frontend/src/views/DifferentialExpression/components/Main/connect.ts
@@ -8,14 +8,30 @@ import {
   submitQueryGroups,
   clearQueryGroup1Filters,
   clearQueryGroup2Filters,
+  selectQueryGroup1Filters,
+  selectOrganism,
 } from "src/views/DifferentialExpression/common/store/actions";
 import useProcessedQueryGroupFilterDimensions from "./components/common/query_group_filter_dimensions";
 
 import { track } from "src/common/analytics";
 import { EVENTS } from "src/common/analytics/events";
 import { craftPayloadWithQueryGroups } from "./utils";
+import { removeParams } from "src/common/utils/removeParams";
+import { useRouter } from "next/router";
+import {
+  useAvailableOrganisms,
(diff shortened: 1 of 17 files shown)


frontend/src/views/DifferentialExpression/components/Main/connect.ts
```diff
@@ -58,6 +76,130 @@ export const useConnect = () => {
...
+      isLoadingFilterOptions ||
+      queryGroupInitializedFromShareURL
+    )
+      return;
+
+    const { search } = window.location;
+    const params = new URLSearchParams(search);
+
+    const paramsToRemove: string[] = [];
+    const cellTypes = params.get("celltypes")?.split(",");
```
kaloster, line 126: suggestion - instead of repeating this and paramsToRemove do something like this:

```
    const getParamsAndRemove = (paramName) => {
      const paramValues = params.get(paramName)?.split(",");
      if (paramValues) {
        paramsToRemove.push(paramName);
      }
      return paramValues;
    };
    
    
  const filterOptions = (paramValues, filterTerms) => {
      return filterTerms.filter((term) => paramValues.includes(term.id));
    };
```

and then do:
```
    const cellTypes = getParamsAndRemove("celltypes");
```
etc... for the rest


and then below do:

```
    if (tissues) {
      const tissuesFiltered = filterOptions(tissues, allFilterOptions.tissue_terms);
      dispatch(selectQueryGroup1Filters("tissues", tissuesFiltered));
    }
```
etc.. for the rest

that'll make it more readable and less repetitive imo

kaloster (commented):
Just one improvement suggestion - looks good otherwise!

kaloster (approved):

PR #7266, opened 2024-07-08
fix: update models page for the new LTS version

ebezzi:
Update the Census Models page to support multiple LTS versions.


tihuan (approved):

hthomas-czi (approved):
Looks good!

PR #7281, opened 2024-07-15
chore(wmg+cellguide): remove calls to datasets and collections in discover api

atarashansky:
## Reason for Change

- Removed dependency of data viz apps on discover API. 

## Changes

- Updated source data generator in cellguide and dataset metadata generator in WMG along with tests

## Testing steps

- Tested with rdev pipeline runs
- Tests pass


backend/common/census_cube/data/constants.py
```diff
@@ -14,7 +14,7 @@
     "EFO:0010010": "CEL-seq2",
 }
 
-CENSUS_CUBE_PINNED_SCHEMA_VERSION = "5.0.0"
+CENSUS_CUBE_PINNED_SCHEMA_VERSION = "5.1.0"
```
Bento007, line 17: why is this value hardcoded, shouldn’t it use the implied schema version from census?

backend/cellguide/pipeline/source_collections/source_collections_generator.py
```diff
@@ -25,57 +46,35 @@ def generate_source_collections_data(all_cell_type_ids_in_corpus: list[str]) ->
...
         # Generate a mapping from collection id to SourceCollectionsData
         collections_to_source_data = {}
-        for dataset_id in dataset_ids:
-            dataset = datasets_dict[dataset_id]
-            collection_id = dataset["collection_id"]
 
-            # If we don't have any source data on this collection yet, create a new SourceCollectionsData object
+        for i in range(len(datasets_metadata_df)):
+            dataset_id = datasets_metadata_df.iloc[i]["dataset_id"]
+            if dataset_id not in dataset_ids:
```
Bento007, line 61: why would this happen? Add a comment please.

backend/wmg/pipeline/constants.py
```diff
@@ -119,3 +119,23 @@
...
+ORGANISM_INFO = [
+    {"label": "homo_sapiens", "id": "NCBITaxon:9606"},
+    {"label": "mus_musculus", "id": "NCBITaxon:10090"},
+]
+
+
+class CensusParameters:
+    census_version = "latest"
+
+    def value_filter(organism: str) -> str:
```
Bento007, line 133:
```suggestion
    @staticmethod
    def value_filter(organism: str) -> str:
```

backend/wmg/pipeline/constants.py
```diff
@@ -119,3 +119,23 @@
...
 ]
+
+
+ORGANISM_INFO = [
+    {"label": "homo_sapiens", "id": "NCBITaxon:9606"},
+    {"label": "mus_musculus", "id": "NCBITaxon:10090"},
+]
+
+
+class CensusParameters:
```
Bento007, line 130: What was your reasoning for making this a class?

Bento007 (changes requested):
I see you are using asserts in the code. If this code is run with python optimizations the assert will be ignored.

backend/wmg/pipeline/constants.py
atarashansky: Easy to patch for testing.

Bento007 (approved):

Issue #7290, opened 2024-07-18
Update Migration SFN to rollback upon error


nayib-jose-gloria:
In DatasetError, if the migrated dataset does NOT have a successful processing status:

- log the error returned by StepFunctionsStartExecution
- remove the failed DatasetVersion from the CollectionVersions’ datasets list
- re-add the most recently successful DatasetVersion (the one we attempted to migrate) back into the CollectionVersions’ datasets list. 
- Schedule deletion of DatasetArtifacts + DatasetVersion for the errored DatasetVersionId by triggering a clean-up batch job (see work below)

PR #7295, opened 2024-07-18
feat: rollback errored migration DatasetVersions during a migration run

nayib-jose-gloria:
## Reason for Change

- #7290

## Changes

-at Collection-level error round-up step in migration run, restore previous dataset version for any processed datasets that failed post-migration processing for any reason
         - this allows future run to roll-forward with a solution for this dataset
- log this rollback state
- if rolled back, delete orphaned DatasetVersion from DB and associated artifacts in DB and S3
- consolidate steps for handling deleted DatasetVersions + artifacts from DB and S3 into a business function call

## Testing steps

- unit tests, TBD rdev testing

### DIFF
--- a/backend/layers/business/business.py
+++ b/backend/layers/business/business.py
@@ -943,6 +943,14 @@ class BusinessLogic(BusinessLogicInterface):
             )
         )
 
+    def delete_dataset_versions(self, dataset_versions: List[DatasetVersion]) -> None:
+        """
+        Deletes a list of dataset versions and associated dataset artifact rows from the database, as well
+        as kicking off deletion of their corresponding assets from S3
+        """
+        self.database_provider.delete_dataset_versions(dataset_versions)
+        self.delete_dataset_version_assets(dataset_versions)
+
     def delete_dataset_version_assets(self, dataset_versions: List[DatasetVersion]) -> None:
         self.delete_dataset_versions_from_public_bucket([dv.version_id.id for dv in dataset_versions])
         self.delete_artifacts(reduce(lambda artifacts, dv: artifacts + dv.artifacts, dataset_versions, []))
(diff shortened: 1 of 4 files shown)


backend/layers/business/business.py
```diff
@@ -943,6 +943,14 @@ def get_unpublished_dataset_versions(self, dataset_id: DatasetId) -> List[Datase
             )
         )
 
+    def delete_dataset_versions(self, dataset_versions: List[DatasetVersion]) -> None:
+        """
+        Deletes a list of dataset versions and associated dataset artifact rows from the database, as well
+        as kicking off deletion of their corresponding assets from S3
+        """
+        self.database_provider.delete_dataset_versions(dataset_versions)
```
Bento007, line 951: We should delete the s3 assets first. It's easier to find stray database rows than it is to find stray s3 assets. enforcing the asset deletion before the database row ensures we have a pointer to the assets incase assets deletion fails for some reason.

backend/layers/processing/schema_migration.py
```diff
@@ -263,6 +269,10 @@ def log_errors_and_cleanup(self, collection_version_id: str) -> list:
         self.s3_provider.delete_files(self.artifact_bucket, object_keys_to_delete)
         if errors:
             self._store_sfn_response("report/errors", collection_version_id, errors)
+            # clean up artifacts for any now-orphaned, rolled back datasets
+            if rolled_back_datasets:
+                # TODO: replace with async batch job to delete orphaned dataset version DB rows + artifacts
```
Bento007, line 274: A batch job seems heavy handed to make API calls. I think you should do it as a lambda, or as a subprocess or thread of this batch job.

Bento007 (commented):
Please change the order your are deleing data.

backend/layers/processing/schema_migration.py
nayib-jose-gloria: I'm fine with the idea of using a lambda (or another alternative) instead. Updated the comment to be agnostic to infra type

Bento007 (approved):

lvreynoso (approved):
I took a look at `restore_previous_dataset_version()` and it seems sane to me. 👍

Issue #7292, opened 2024-07-18
Create a standalone, manual batch job that runs bulk collection and dataset rollbacks


nayib-jose-gloria:
- For public collections–rolls back the CollectionVersion associated with canonical Collection to the previously published version with the previous schema version
              - only if the published collection is actually on the most recent schema version per cellxgene-schema CLI
- Delete erroneous CollectionVersion from DB
               - Queue clean-up batch jobs for each DatasetVersion associated with the CollectionVersion, once available
- For private collections–1) removes the DatasetVersions from the CollectionVersions’s datasets list that are of the latest schema version then 2) re-adds the most recently successful previous schema version DatasetVersion back into the CollectionVersions’ datasets list.
               - Queue clean-up DatasetVersion batch job once available 
- Accept an allowlist of specific Collections / Datasets to rollback (and queue clean-up DatasetVersion batch jobs as necessary)

PR #7317, opened 2024-07-30
feat: Create a standalone, manual batch job that runs bulk collection and dataset rollbacks for migration error recovery

nayib-jose-gloria:
## Reason for Change

- #7292

## Changes

- add logic for rollback batch job
- functions to support 5 kinds of rollbacks: 
        1) Rolling back all Published Collections to their previous CollectionVersion
        2) Rolling back all Private Collections to their pre-migration state by rolling back their Datasets to their previous DatasetVersion
        3) Rolling back allowlist of specific Datasets (in private Collections only)
        4) Rolling back allowlist of specific Private Collections
        5) Rolling back allowlist of specific Publishing Collections
- add business layer functions to fetch all unpublished collection versions associated with a collection ID (max of 2, one is_auto_version = False revision and one is_auto_version = True)
- add business layer function to restore a collection's the previous, most recently published collectionversion as the canonical collection + misc. input sanitation and error handling
- add persistence layer function to replace the collection_version_id that a canonical Collection points to, to support rollback

## Testing steps

- unit testing + rdev integration testing
- tested all 5 rollback types in rdev--ran a dummy 5.2 migration using a cellxgene-schema-cli branch, then rolled back + cleaned up assets successfully.


backend/layers/processing/rollback.py
```diff
@@ -0,0 +1,301 @@
...
+        raise ValueError("ROLLBACK_TYPE is required")
+
+    rollback_type = RollbackType(rollback_type_str)
+    if rollback_type in (
+        RollbackType.PUBLIC_COLLECTION_LIST,
+        RollbackType.PRIVATE_COLLECTION_LIST,
+        RollbackType.PRIVATE_DATASET_LIST,
+    ):
+        entities_to_rollback = os.environ.get("ENTITY_LIST", None)
+        entities_to_rollback_list = entities_to_rollback.split(",") if entities_to_rollback else None
```
Bento007, line 298: Is there a use case to provide no Entities? If ti is a no op, I think it would be a better signal to throw an error. That should make it clear to the user that they forgot to add entities.

backend/layers/processing/rollback.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        Rolls back the Datasets in the CollectionVersions associated with each DatasetVersionId passed in, to their
+        respective previous, most recently created DatasetVersion. Then, triggers deletion of the DB references and S3
+        assets for the rolled back DatasetVersions.
+        """
+        rolled_back_datasets = []
+        for dataset_version_id in self.entity_id_list:
+            dataset_version_id = DatasetVersionId(dataset_version_id)
+            try:
+                rolled_back_dataset = self.rollback_private_dataset(dataset_version_id)
```
Bento007, line 168: I'm finding it confusing to distingush between `rollback_{public,private}_{collection, dataset}_list`. For readability consider reodering the function name to include the entity first ex: `private_dataset_rollback`, `private_dataset_list_rollback`. Alternatively `dataset_private_rollback`, `dataset_private_list_rollback`.

backend/layers/processing/rollback.py
```diff
@@ -0,0 +1,301 @@
...
+        """
+        cv_id = collection_version_id
+        dataset_version = self.business_logic.get_dataset_version(dataset_version_id)
+        if cv_id is None:
+            # account for collection potentially having a migration revision and a non-migration revision
+            collection_versions = self.business_logic.get_unpublished_collection_versions_from_canonical(
+                dataset_version.collection_id
+            )
+            for cv in collection_versions:
+                if cv_id is not None:
```
Bento007, line 195: Is this condition ever met? It looks like the `break` on line 200 should make this conditional superfluous.

Adding a comment would help clarify.

backend/layers/processing/rollback.py
```diff
@@ -0,0 +1,301 @@
...
+        references and S3 assets for the rolled back CollectionVersions and their associated DatasetVersions.
+        """
+        filter = CollectionQueryFilter(is_published=True)
+        collections = self.business_logic.get_collections(filter)
+        rolled_back_collection_versions = []
+        for collection in collections:
+            try:
+                rolled_back_collection_version = self.rollback_public_collection(collection.collection_id)
+            except Exception as e:
+                logger.info(f"Failed to rollback Collection {collection.collection_id}, {e}")
```
Bento007, line 221: - change `info` to `exception` and remove `{e}`. 
- add that it was a public collection to error message

backend/layers/processing/rollback.py
```diff
@@ -0,0 +1,301 @@
...
+        references and S3 assets for the rolled back DatasetVersions.
+        """
+        rolled_back_datasets = []
+        for collection_version_id in self.entity_id_list:
+            try:
+                collection_dataset_versions = self.rollback_private_collection(
+                    CollectionVersionId(collection_version_id)
+                )
+            except Exception as e:
+                logger.info(f"Failed to rollback CollectionVersion {collection_version_id}, {e}")
```
Bento007, line 137: - change `info` to `exception` and remove `{e}`. 
- add that it was a private collection to error message

backend/layers/processing/rollback.py
```diff
@@ -0,0 +1,301 @@
...
+                        cv_id = cv.version_id
+                        break
+        if cv_id is None:
+            raise CollectionNotFoundException(
+                f"An Associated CollectionVersion not found for DatasetVersion {dataset_version_id}"
+            )
+        self.business_logic.restore_previous_dataset_version(cv_id, dataset_version.dataset_id)
+        return dataset_version
+
+    def rollback_public_collections(self) -> None:
```
Bento007, line 208: It looks like this could use `rollback_public_collection_list` and remove some duplicated logic.

Bento007 (changes requested):
A few thing to look into.

Great job with the testing.

.happy/terraform/modules/batch/main.tf
```diff
@@ -123,6 +123,68 @@ resource aws_batch_job_definition dataset_metadata_update {
...
+    {
+      "name": "AWS_DEFAULT_REGION",
+      "value": "${data.aws_region.current.name}"
+    },
+    {
+      "name": "REMOTE_DEV_PREFIX",
+      "value": "${var.remote_dev_prefix}"
+    }
+  ],
+  "vcpus": 1,
```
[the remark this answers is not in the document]
nayib-jose-gloria: @Bento007 why that amount? We're only dealing with dataset metadata, so I figured we could keep it pretty small

.happy/terraform/modules/batch/main.tf
```diff
@@ -123,6 +123,68 @@ resource aws_batch_job_definition dataset_metadata_update {
...
 
+resource aws_batch_job_definition rollback {
+  type = "container"
+  name = "dp-${var.deployment_stage}-${var.custom_stack_name}-rollback"
+  container_properties = jsonencode({
+  "command": ["python3", "-m", "backend.layers.processing.rollback"],
+  "jobRoleArn": "${var.batch_role_arn}",
+  "image": "${var.image}",
+  "memory": var.batch_container_memory_limit,
+  "environment": [
```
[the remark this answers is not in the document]
nayib-jose-gloria: hm good question--we don't need to, but it might be useful to clean-up. Actually, I imagine we already don't clean these up when we delete datasets in other scenarios. So this might be a general follow-up issue

backend/layers/processing/rollback.py
nayib-jose-gloria: I believe breaking out a nested for loop does not break you out of the outer for loop

lvreynoso (approved):
+1 to Trent's suggestions, looks great

Bento007 (approved):

Issue #7309, opened 2024-07-25
Enforce canonical for seurat conversion.


Bento007:
## Motivation

To convert an anndata dataset to seurat we need to enforce [canonical form](https://docs.scipy.org/doc//scipy/tutorial/sparse.html#canonical-formats) of the sparse matrixes. Because of how the anndata library read and write the sparse format, the whole dataset must be read into memory. This is an expensive operation that is slow and memory Intensive. Since this enforcement is only required for seurat conversion, the step of enforcing canonical should be move to the seurat conversion container. This will allow us to speed up validation.

## Definition of Done
- canonical form is enforced only for datasets that support seurat conversion
- canonical form is enforced in the same AWS batch job the seurat conversion happens.

## Tasks

- [remove canonical enforcement from the cli](https://github.com/chanzuckerberg/single-cell-curation/pull/965)
- enforce canonical form in the seurat container before converting to seurat.

nayib-jose-gloria:
Estimate (incl testing): 2-4 days

@Bento007 if you agree

nayib-jose-gloria:
From @ivirshup: 

> > Enforcing canonical format is the biggest consumer of memory because it requires reading in the whole X and raw.X matrix.

> Btw, this should be possible without reading the whole matrix, since canonicalization should only remove entries. That part shouldn't matter if the canonicalization isn't happening in place though. I can advise on this if needed.


Recommend contacting him for more details to see whether we can implement canonicalization w/o reading in the whole matrix even in the seurat conversion

PR #7326, opened 2024-08-09
feat: enforce canonical format of anndata for seurat conversion

nayib-jose-gloria:
## Reason for Change

- #7309

## Changes

- move code to enforce canonical format on anndata to process_seurat, because it is a seurat-exclusive requirement. It + its test code has been moved from cellxgene-schema CLI

## Notes to Reviewers 
- chose to call sum_duplicates() for the matrix entirely read in memory rather than batched calls, because we already require reading the matrix into memory for the seurat conversion and there is no way around that

## Testing steps

- rdev + unit tests


python_dependencies/processing/requirements.txt
```diff
@@ -2,7 +2,7 @@ anndata==0.8.0
 awscli
 boto3>=1.11.17
 botocore>=1.14.17
-cellxgene-schema
+git+https://github.com/chanzuckerberg/single-cell-curation/@nayib/remove-canonical#subdirectory=cellxgene_schema_cli
```
nayib-jose-gloria, line 5: todo: revert

Bento007 (approved):

lvreynoso (approved):

PR #7328, opened 2024-08-16
feat: bastion replacement

ebezzi:
Replace makefile rules for connecting to Postgres through the new SSM-enabled machines.

Note: due to how awscli works, it is not possible to launch the tunnel in the background without resorting to bash shenanigans (forking the process via `&`, killing the process via `kill`). Therefore, I made the following changes:

1. `make db/tunnel/up` now needs to be launched on a separate shell, and will not fork the process
1. `make db/connect` will need to be called after the tunnel is up
1. Instead of running `make db/tunnel/down`, it is sufficient to shut down the tunnel via CTRL-C.


nayib-jose-gloria (approved):
mind adding a comment reflecting that db/tunnel/up needs to be run first in a diff shell before invoking db/connect? 

we'll also need to update the mirroring script, but that can be a separate PR

PR #7336, opened 2024-09-12
chore: Fix data mirroring script and instructions

nayib-jose-gloria:
## Reason for Change

- Infra team recently deprecated bastions, breaking our data mirroring and PostgreSQL ssh connection make commands
- ebezzi replaced the bastions with accessing our DBs via SSM-enabled [machines](https://github.com/chanzuckerberg/single-cell-data-portal/pull/7328)
- This PR uses those changes to fix our data mirroring script and updating its instructions, so we can continue to mirror prod data to dev/staging for testing purposes.

## Changes

- require running  `make db/tunnel/up` for both SRC and DEST env, with a different local port number forwarding to the SSM-enabled machine for each DB env.
- add port number as argument for mirroring script and tunneling command
- update 'make db/dump' to also take PORT as an arg and remove `make db/connect` call, which is now interactive-only and replaced by the new `make db/tunnel/up`.

## Testing steps

- Ran prod -> dev mirroring

### DIFF
--- a/.github/workflows/rdev-tests.yml
+++ b/.github/workflows/rdev-tests.yml
@@ -226,7 +226,7 @@ jobs:
       - name: Run e2e Logged In tests
         run: |
           DEBUG=pw:api RDEV_LINK=https://${{ env.STACK_NAME }}-frontend.rdev.single-cell.czi.technology npm run e2e-rdev-logged-in-ci
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         if: always()
         with:
           name: logged-in-test-results
--- a/python_dependencies/wmg_processing/requirements.txt
+++ b/python_dependencies/wmg_processing/requirements.txt
@@ -7,7 +7,7 @@ cellxgene-ontology-guide~=1.0.0
 dataclasses-json==0.5.7
 ddtrace==2.1.4
 numba>=0.58.0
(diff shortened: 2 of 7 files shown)


.github/workflows/rdev-tests.yml
```diff
@@ -226,7 +226,7 @@ jobs:
       - name: Run e2e Logged In tests
         run: |
           DEBUG=pw:api RDEV_LINK=https://${{ env.STACK_NAME }}-frontend.rdev.single-cell.czi.technology npm run e2e-rdev-logged-in-ci
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
```
nayib-jose-gloria, line 229: replacing deprecated version of upload-artifact action

python_dependencies/wmg_processing/requirements.txt
```diff
@@ -7,7 +7,7 @@ cellxgene-ontology-guide~=1.0.0
 dataclasses-json==0.5.7
 ddtrace==2.1.4
 numba>=0.58.0
-numpy==1.23.5
+numpy>=1.24.0,<2.1.0
```
nayib-jose-gloria, line 10: unrelated fix for wmg processing unit tests

ebezzi (approved):

Issue #7339, opened 2024-09-17
bug(codecoverage): broken due to updated GHA


Bento007:
## Describe the bug

code coverage is broken due to change in how files are uploaded and downloaded using `actions/upload-artifact@v3` and actions/download-artifact@v3

PR #7341, opened 2024-09-18
fix: codecoverage

nayib-jose-gloria:
## Reason for Change

- #7339
- The way our CI/CD is set-up, failures in coverage are blocking staging/prod deploys.

## Changes

- upload-artifact v3 now needs an arg to explicitly allow for uploading hidden files (i.e. ".coverage-<sha>")

## Testing steps

- run code coverage GHAs in this pr


Bento007 (approved):
they clearly aren't following conventional commit. This might be broken in other repos we manage.

PR #7346, opened 2024-09-27
fix: missing server_metadata

Bento007:
## Reason for Change
rolling back to previously working version of authlib. 
Previously it was upgraded here https://github.com/chanzuckerberg/single-cell-data-portal/commit/a12556f997d9caa1cfbd52453446c5c4af460df3m which caused 502 on the frontend.

## Changes

- add server_metadata to oath config
- roll back to previous authlib version.

## Testing steps
- verify you can login on frontend
- all automated tests pass.

### DIFF
--- a/backend/portal/api/app/v1/authentication.py
+++ b/backend/portal/api/app/v1/authentication.py
@@ -8,7 +8,6 @@ from urllib.parse import urlencode
 
 import requests
 from authlib.integrations.flask_client import OAuth
-from authlib.integrations.flask_client.remote_app import FlaskRemoteApp
 from flask import Response, after_this_request, current_app, g, jsonify, make_response, redirect, request, session
 
 from backend.common.authorizer import assert_authorized_token, get_userinfo_from_auth0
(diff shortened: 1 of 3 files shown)


nayib-jose-gloria (commented):
If we're rolling back the authlib change, why do we still need to add in server metadata when its working without it?

Bento007:
it will be needed in the future anyways according to auth0 documentation.

nayib-jose-gloria:
> it will be needed in the future anyways according to auth0 documentation.

I'm still getting a 502 gateway when i try to log in

backend/portal/api/app/v1/authentication.py
```diff
@@ -50,6 +50,7 @@ def get_oauth_client(config: CorporaAuthConfig) -> FlaskRemoteApp:
         authorize_url=config.api_authorize_url,
         client_kwargs={"scope": "openid profile email offline_access write:collections"},
         authorize_params={"audience": config.api_audience},
+        server_metadata_url=config.server_metadata_url,
```
Bento007, line 53:
```suggestion
```

Bento007:
Still looking for where that 502 is coming from.
- rdev does not use cloudfront 
- I did not find any 502 in the logs for the contianer dataportal-rdev-happy-proxy

PR #7275, opened 2024-07-10
chore: resolve deprecation warning and update requirements

Bento007:
## Reason for Change

- resolving deprecation errors.
- keeping requirements up to date.

## Changes

- renamed `assertEquals` to `assertEqual` to resolve deprecated assertEquals
- rename `method_whitelist` to `allowed_methods` to resolve DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
- use backend/server/requirements in backend portal container. This means we can remove the duplicate requirements in backend/requirements.txt
- Update requirements for the backend portal api.
- migrate to authlib 1.x.x

## Testing steps

- verify tests pass

PR #7350, opened 2024-10-03
fix: Revert "chore: resolve deprecation warning and update requirements (#7275)"

nayib-jose-gloria:
This reverts commit a12556f997d9caa1cfbd52453446c5c4af460df3.

## Reason for Change

- commit being reverted has a bug causing log-ins to fail

## Changes

- reverted previous commit w/ bug while its investigated

## Testing steps

- was able to log-in as normal after deploying an rdev with the commit reverted

### DIFF
--- a/tests/unit/backend/layers/api/test_portal_api.py
+++ b/tests/unit/backend/layers/api/test_portal_api.py
@@ -462,7 +462,7 @@ class TestCollection(BaseAPIPortalTest):
         # TODO: this endpoint should also return `version_id`
         collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
         doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
-        self.assertEqual(doi, "https://doi.org/10.1016/foo")
+        self.assertEquals(doi, "https://doi.org/10.1016/foo")
 
     # ✅
     def test__post_collection_rejects_two_dois(self):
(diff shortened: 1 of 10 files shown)


tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -462,7 +462,7 @@ def test__post_collection_normalizes_doi(self):
         # TODO: this endpoint should also return `version_id`
         collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
         doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
-        self.assertEqual(doi, "https://doi.org/10.1016/foo")
+        self.assertEquals(doi, "https://doi.org/10.1016/foo")
```
nayib-jose-gloria, line 465:
```suggestion
        self.assertEqual(doi, "https://doi.org/10.1016/foo")
```

tests/unit/backend/wmg/test_query.py
```diff
@@ -158,7 +158,7 @@ def test__single_dimension__returns_all_dimension_and_terms(self):
         with create_temp_wmg_snapshot(dim_size=dim_size) as snapshot:
             q = CensusCubeQuery(snapshot, self.cube_query_params)
             result = q.list_primary_filter_dimension_term_ids("tissue_ontology_term_id")
-            self.assertEqual(
+            self.assertEquals(
```
nayib-jose-gloria, line 161:
```suggestion
            self.assertEqual(
```

tests/unit/backend/wmg/test_query.py
```diff
@@ -177,7 +177,7 @@ def exclude_one_tissue_per_organism(logical_coord: NamedTuple) -> bool:
             result = q.list_grouped_primary_filter_dimensions_term_ids(
                 "tissue_ontology_term_id", "organism_ontology_term_id"
             )
-            self.assertEqual(
+            self.assertEquals(
```
nayib-jose-gloria, line 180:
```suggestion
            self.assertEqual(
```

PR #7351, opened 2024-10-03
fix: anndata 0.10 matrix slicing bug

nayib-jose-gloria:
## Reason for Change

- dev schema 5.2 migration revealed a bug in anndata 0.10 that accidentally loads the full matrix into memory every time we chunked the matrix. this caused, in cases of datasets with very large matrices, much longer wait times (and in one case, a timeout) when traversing the matrix in chunks to calculate the mean_gene_per_cell metadata field
- because we allow collection publishing without a seurat conversion, we have chosen to not rollback datasets that have only skipped seurat conversion in a schema migration

## Changes

- update our matrix slicing to avoid anndata bug that slices inefficiently given both row and col bounds in one slice operation, by explicitly slicing by row first then col 
- condition to not auto-rollback datasets that have successfully converted all other artifacts to a new schema, while processing errors

## Testing steps

- migrated affected datasets in dev env

### DIFF
--- a/tests/unit/backend/layers/api/test_portal_api.py
+++ b/tests/unit/backend/layers/api/test_portal_api.py
@@ -462,7 +462,7 @@ class TestCollection(BaseAPIPortalTest):
         # TODO: this endpoint should also return `version_id`
         collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
         doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
-        self.assertEqual(doi, "https://doi.org/10.1016/foo")
+        self.assertEquals(doi, "https://doi.org/10.1016/foo")
 
     # ✅
     def test__post_collection_rejects_two_dois(self):
--- a/tests/unit/backend/layers/api/test_portal_api.py
+++ b/tests/unit/backend/layers/api/test_portal_api.py
@@ -462,7 +462,7 @@ class TestCollection(BaseAPIPortalTest):
         # TODO: this endpoint should also return `version_id`
         collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
         doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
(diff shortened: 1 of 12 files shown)


tests/unit/backend/wmg/test_query.py
```diff
@@ -177,7 +177,7 @@ def exclude_one_tissue_per_organism(logical_coord: NamedTuple) -> bool:
             result = q.list_grouped_primary_filter_dimensions_term_ids(
                 "tissue_ontology_term_id", "organism_ontology_term_id"
             )
-            self.assertEqual(
+            self.assertEquals(
```
nayib-jose-gloria, line 180:
```suggestion
            self.assertEqual(
```

tests/unit/backend/wmg/test_query.py
```diff
@@ -158,7 +158,7 @@ def test__single_dimension__returns_all_dimension_and_terms(self):
         with create_temp_wmg_snapshot(dim_size=dim_size) as snapshot:
             q = CensusCubeQuery(snapshot, self.cube_query_params)
             result = q.list_primary_filter_dimension_term_ids("tissue_ontology_term_id")
-            self.assertEqual(
+            self.assertEquals(
```
nayib-jose-gloria, line 161:
```suggestion
            self.assertEqual(
```

tests/unit/backend/layers/api/test_portal_api.py
```diff
@@ -462,7 +462,7 @@ def test__post_collection_normalizes_doi(self):
         # TODO: this endpoint should also return `version_id`
         collection = self.business_logic.get_collection_version_from_canonical(CollectionId(collection_id))
         doi = next(link.uri for link in collection.metadata.links if link.type == "DOI")  # TODO: careful
-        self.assertEqual(doi, "https://doi.org/10.1016/foo")
+        self.assertEquals(doi, "https://doi.org/10.1016/foo")
```
nayib-jose-gloria, line 465:
```suggestion
        self.assertEqual(doi, "https://doi.org/10.1016/foo")
```

PR #7363, opened 2024-10-16
fix: add default values for cell guide description (SCE-22)

kaloster:
## Reason for Change

- When a user tries to click on a cell type in GE that doesn't exist in CG, the app would crash for missing `clDescription`
- Context: [here](https://czi.atlassian.net/jira/software/c/projects/SCE/list?selectedIssue=SCE-22)

## Changes

- Added default values for missing `clDescription` and `rawDescriptionGpt`
- Hide `Open in Cell Guide `link when a cell type is missing in Cell Guide

## Testing steps

- In GE, filter by `chondroblast` - click on info icon


tihuan (approved):
Ahh brilliant solution! Love it 💡 ❤️ Thank you!!

PR #7364, opened 2024-10-16
fix: use dask to speed up cxg conversion and run in constant memory

Bento007:
[co-authored by @nayib-jose-gloria]

## Reason for Change
- https://czi.atlassian.net/browse/VC-1342

## Changes

- modify cxg conversion code to use dask to write tiledb arrays.
- update CXG (and CXG remaster) batch job to request constant 2 VCPU, 16 GB memory
- remove swap memory infra (no longer required)
- define separate CXG and H5AD processing batch jobs, for distinct configurability
- in tiledb config context, set max and min for concurrent fragments held in memory during tiledb.consolidate, as recommended by tiledb docs, to control memory usage
- in `h5ad_data_file`, use cellxgene-schema's read_h5ad call to read anndata (uses dask for matrix read)
- merged in changes from branch to skip seurat conversion, as it is no longer supported for schema 5.3.0 (same release that will include anndata 0.11 and dask improvements)
- update schema_migration to use collection version ID as a key for error logs in s3 rather than canonical collection ID, since there could be a key conflict using canonical collection ID (i.e. we migrate both the public version and an open revision of the same collection)
- use count_matrix_nonzero from cellxgene-schema to calculate matrix sparsity, as its already dask enabled and does similar work as code that was written here
- update data env mirror script, which was missing a call to update env-specific references in the mirrored env (i.e. prod domains in URL fields --> dev domains)

## Testing steps

- Ran test dev migration, ran in under 2 days on all datasets at a fraction of the cost of previous migration runs (1/10th the cost, at least)

## Notes for Reviewer
- **Do not merge until the offical release of cellxgene-schema 5.3.0**


kaloster (approved):

Bento007:
### Memory test result
|dataset | main | dask |
|-----|-----| ------|
|[10x](https://datasets.cellxgene.cziscience.com/f50deffa-43ae-4f12-85ed-33e45040a1fa.h5ad) | plot_10x| plot_10x|
|[slide seq](https://datasets.cellxgene.cziscience.com/9c889e34-c6b7-4493-affa-be30ca85a299.h5ad)|plot_seq | plot_seq|
|[visium](https://datasets.cellxgene.cziscience.com/f9dc9230-cd78-4cc4-8b05-b47341e6c977.h5ad)| plot_vis | plot_vis|

Overall the speed improved across the board. 

### Peak Memory
| dataset | main | dask |
|10x | 9021.359 MiB | 8481.504 MiB |
|seq | 8847.473 MiB | 8811.594 MiB |
|vis | 8642.473 MiB | 9412.105 MiB |

The memory usage was stable across the datasets

Bento007:
I ran the 10x dataset with different cores to see if speeds would improve:

| 4 cores | 8 cores|
|---------|-------|
|plot_10x|plot_10x|
The speed increased without changing the max memory usage.

backend/layers/processing/utils/dask_utils.py
```diff
@@ -0,0 +1,25 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TileDBSparseArrayWriteWrapper:
+    def __init__(self, uri, *, ctx=None):
+        self.uri = uri
+        self.ctx = {
+            "sm.io_concurrency_level": 1,
+            "sm.compute_concurrency_level": 1,
+        }
```
ivirshup, line 15: It's possible that these parameters are worth tuning. I had forced these to 1 in my testing so that I could make sure I wouldn't have oversubscription problems.

Bento007: Looking at the tiledb documentation the default setting should be sufficient.

backend/layers/processing/h5ad_data_file.py
```diff
@@ -95,19 +98,23 @@ def to_cxg(
...
         matrix_container = f"{output_cxg_directory}/X"
-
         x_matrix_data = self.anndata.X
-        is_sparse = is_matrix_sparse(x_matrix_data, sparse_threshold)  # big memory usage
-        logging.info(f"is_sparse: {is_sparse}")
-
-        convert_matrices_to_cxg_arrays(matrix_container, x_matrix_data, is_sparse, ctx)  # big memory usage
+        with dask.config.set(
+            {
+                "num_workers": 2,  # match the number of workers to the number of vCPUs
```
Bento007, line 104: Use the default value for `num_works`, `threads_per_worker`, and `distributed.worker.memory.limit`use. The default values are calculated based on the available resources. it will be easier to maintain since we might decide to change the resource allocation to the job in TF.

backend/layers/processing/h5ad_data_file.py
```diff
@@ -95,19 +98,23 @@ def to_cxg(
...
+                "threads_per_worker": 1,
+                "distributed.worker.memory.limit": "6GB",
+                "scheduler": "threads",
```
Bento007, line 107: Change to "processes" to make use of multiple cores when reading the chunked h5ad.

backend/layers/processing/process_validate.py
```diff
@@ -125,8 +125,8 @@ def validate_h5ad_file_and_add_labels(
         output_filename = CorporaConstants.LABELED_H5AD_ARTIFACT_FILENAME
         try:
             is_valid, errors, can_convert_to_seurat = self.schema_validator.validate_and_save_labels(
-                local_filename, output_filename
-            )
+                local_filename, output_filename, n_workers=1
```
Bento007, line 128: in our CLI code we should check if `dask.config` is already set and leave it alone if it is, then we don't need to pass n_workers to this function. That way we don't need to worry about keeping this value synced with the TF.

backend/layers/processing/utils/cxg_generation_utils.py
```diff
@@ -2,14 +2,20 @@
...
 import pandas as pd
 import tiledb
+from cellxgene_schema.utils import get_matrix_format
 
 from backend.common.constants import SPATIAL_KEYS_EXCLUDE, UNS_SPATIAL_KEY
+from backend.layers.processing.utils.dask_utils import TileDBSparseArrayWriteWrapper
 from backend.layers.processing.utils.spatial import SpatialDataProcessor
 from backend.layers.processing.utils.type_conversion_utils import get_dtype_and_schema_of_array
 
+dask.config.set(scheduler="single-threaded")
```
Bento007, line 17: I assume this was for debugging? It should be removed.

backend/layers/processing/utils/dask_utils.py
```diff
@@ -0,0 +1,22 @@
+import logging
+
+import tiledb
+from scipy import sparse
+
+logger = logging.getLogger(__name__)
+
+
+class TileDBSparseArrayWriteWrapper:
```
Bento007, line 9: Nice abstraction

backend/layers/processing/utils/matrix_utils.py
```diff
@@ -1,63 +1,27 @@
...
+def is_matrix_sparse(matrix: da.Array, sparse_threshold: float) -> bool:
     """
     Returns whether `matrix` is sparse or not (i.e. dense). This is determined by figuring out whether the matrix has
     a sparsity percentage below the sparse_threshold, returning the number of non-zeros encountered and number of
-    elements evaluated.  This function may return before evaluating the whole matrix if it can be determined that matrix
-    is not sparse enough.
-    """
+    elements evaluated.
 
+    TODO move to dask_utils.py in a separate PR
```
Bento007, line 15: Remove, or is there more to come?

backend/layers/thirdparty/schema_validator_provider.py
```diff
@@ -23,17 +23,18 @@ def get_current_schema_version(self) -> str:
 
 
 class SchemaValidatorProvider(SchemaValidatorProviderInterface):
-    def validate_and_save_labels(self, input_file: str, output_file: str) -> Tuple[bool, list, bool]:
+    def validate_and_save_labels(self, input_file: str, output_file: str, n_workers: int) -> Tuple[bool, list, bool]:
```
Bento007, line 26: make `n_workers` Optional.
```suggestion
    def validate_and_save_labels(self, input_file: str, output_file: str, **kwargs) -> Tuple[bool, list, bool]:
```

Also update `SchemaValidatorProviderInterface(Protocol)` to match

backend/layers/processing/h5ad_data_file.py
nayib-jose-gloria: I tried to at first, but kept running into misc. memory issues on some datasets. Most common was oversplitting and not having enough memory allocated on a thread to load the chunk it was processing. Did you ever run into issues like that in your testing?

backend/layers/processing/h5ad_data_file.py
nayib-jose-gloria: based on this dask documentation, I think threads scheduler makes more sense for our use-case (dealing w/ numpy/pandas, tiledb objects) https://docs.dask.org/en/stable/best-practices.html#processes-and-threads

what do you think?

backend/layers/processing/utils/dask_utils.py
nayib-jose-gloria: i'm glad you think so, you wrote it LOL

backend/layers/processing/h5ad_data_file.py
Bento007: I would run a bench mark to see. The chunked reads from the anndata could thread bottle neck.

backend/layers/processing/utils/dask_utils.py
Bento007: LOL, let me just pat myself on the back.

PR #7380, opened 2024-11-05
feat: remove seurat [do not merge yet]

joyceyan:
## Reason for Change

https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/7366

## Changes

removes all code relevant to seurat. for dataset metadata updates, i'm just setting the dataset RDS version to be skipped always.

## Testing steps

i updated the unit tests, and uploaded a dataset to the rdev to verify it went through the step function okay. it's available here: https://pr-7380-frontend.rdev.single-cell.czi.technology/collections/06b40cf0-f576-4e91-aeba-4f06925ae36d

## Checklist 🛎️

## Notes for Reviewer

🚨 we don't want to deploy this until we're ready to start the schema 5.3.0 migration


backend/layers/processing/dataset_metadata_update.py
```diff
@@ -353,28 +296,8 @@ def update_metadata(
...
-            artifact_jobs.append(rds_job)
-            rds_job.start()
-        elif current_dataset_version.status.rds_status == DatasetConversionStatus.SKIPPED:
-            self.update_processing_status(new_dataset_version_id, DatasetStatusKey.RDS, DatasetConversionStatus.SKIPPED)
-        else:
-            self.logger.error(
-                f"Cannot find RDS artifact uri for {current_dataset_version_id}, and Conversion Status is not SKIPPED."
-            )
-            self.update_processing_status(new_dataset_version_id, DatasetStatusKey.RDS, DatasetConversionStatus.FAILED)
+        # Mark all RDS conversions as skipped
```
joyceyan, line 299: this seemed like the simplest backwards compatible solution

nayib-jose-gloria:
> i updated the unit tests, but i'm not sure what the best way is to test the terraform changes and verify it's correct. 

easiest would be to test an upload in the rdev and make sure it skips seurat and all other actions are behaving normally (can publish, FE looks normal, can update a collection DOI, etc.). Migration is basically just an orchestrator that runs the 'migrate' script then kicks off the regular ingestion pipeline so if regular ingestion works we can expect migration to work as well. 

> i assume i should also be making a corresponding change to single-cell-infra based on this comment: https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/.happy/terraform/modules/sfn/main.tf#L1 ?

I think this might be outdated? Link is dead and I don't recall this code being duplicated in infra 


> if we merged this change and deployed it, this would mean we would no longer generate seurat conversions for any new uploads in addition to any schema version migrations we run, right? so we could theoretically launch this change without it being directly tied to a schema release?

Correct. We'll want to remove the seurat-specific code in the validator in the next release, which will require an update here as well since currently the 'validate' command returns a tuple of (is_valid, errors, is_seurat_convertible) and we'll want to change that to just (is_valid, errors). But it won't block launch if we just ignore the third return value for now

backend/layers/processing/dataset_metadata_update.py
```diff
@@ -34,7 +33,6 @@
 from backend.layers.thirdparty.uri_provider import UriProvider
 
 base = importr("base")
```
nayib-jose-gloria, line 35: r-specific, can remove this!

backend/layers/processing/dataset_metadata_update.py
nayib-jose-gloria: this approach is consistent since we treat metadata updates as new versions, but we should double-check with Lattice/Brian that they're aware + cool with it

nayib-jose-gloria (commented):
I think it makes sense to set RDS status as skipped for everything for now, since the FE changes are pending and this lets us decouple without blocking. But I think it'd be useful to write a follow-up maintenance issue to deprecate RDS status (while still maintaining the assets in s3/surfacing them in the API)

joyceyan:
> I think it makes sense to set RDS status as skipped for everything for now, since the FE changes are pending and this lets us decouple without blocking. But I think it'd be useful to write a follow-up maintenance issue to deprecate RDS status (while still maintaining the assets in s3/surfacing them in the API)

i'm actually inclined to say that we should just keep rds_status indefinitely. there isn't any meaningful benefit to removing it since there's really only one place we're reading it from, and removing it seems like it would just cause more headache than it's worth, especially if it's a required field on the db level.

nayib-jose-gloria:
> i'm actually inclined to say that we should just keep rds_status indefinitely. there isn't any meaningful benefit to removing it since there's really only one place we're reading it from, and removing it seems like it would just cause more headache than it's worth, especially if it's a required field on the db level.

the DB stores a dataset status JSON, but doesn't codify what its composed of. So we could get rid of rds status without having to update the DB. Agreed it opens us up to more regression risk to remove it, but it seems like for long-term code readability it would be confusing to keep a status indicator for an artifact type we don't plan on processing anymore. I think its not p0 but worth at least tracking as something to clean-up at some point

backend/layers/processing/dataset_metadata_update.py
joyceyan: https://czi-sci.slack.com/archives/C07AV4NU9D2/p1731535747847619

nayib-jose-gloria (changes requested):
I'm not exactly sure what the syntax error is, but there seems to be one still in the ingestion SFN. Successful jobs are being marked as failed in AWS due to one. See https://us-west-2.console.aws.amazon.com/states/home?region=us-west-2#/v2/executions/details/arn:aws:states:us-west-2:699936264352:execution:dp-rdev-pr-7380-sfn:4d88ae99-a598-4933-8cb2-c8eda43b606b_1731535270

nayib-jose-gloria (approved):
Note: there's some additional seurat related downloads in the processing base [image](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/Dockerfile.processing_base) that we should remove as well. but I think that can safely be follow-on work and we can move ahead with this PR for the functional change. thank you!

joyceyan:
@nayib-jose-gloria Good catch, I created a ticket for that here: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/7390

joyceyan:
Closing this PR since the work for this has been folded into the 5.3 feature branch: https://github.com/chanzuckerberg/single-cell-data-portal/compare/main...feature/schema-5-3

PR #7398, opened 2024-12-11
feat: constant memory dataset DownloadValidate

nayib-jose-gloria:
## Reason for Change

- https://czi.atlassian.net/browse/VC-1338

## Changes

- replace chunked matrix traversal in extract_metadata with call to count_matrix_nonzero in cellxgene_schema that does the same thing
- When populating dataset citation, update dataset.uns in-place using h5py rather than re-opening and re-writing the entire anndata object.
- Use memory efficient read_h5ad call from cellxgene_schema in extract_metadata
- update Validate batch jobs to use constant memory 
- delete ProcessDownload (more complex logic no longer required if we are using a constant memory allocation) and instead incorporate the URI -> S3 upload into the ProcessValidate step
- Update SFN infra to remove obviated steps (Download, RegisterJobDefinition, DeregisterJobDefinition)

## Testing steps
- local testing with large datasets
- TODO: rdev testing,

## Notes for Reviewer
- TODO: determine what constant memory should be set at, based on testing
- TODO: fix tests as needed


Bento007 (approved):

nayib-jose-gloria:
added to feature/schema-5-3 branch

PR #7405, opened 2025-01-08
test cli 5.2.3 release

Bento007:


Bento007:
This currently fails because we have change in 5.2.2 that deprecate seurat and that work has not made it's way into the portal yet.

PR #7414, opened 2025-01-24
chore: pin COG to < 2

joyceyan:
## Reason for Change

updates all of the COG versions in data portal to <2, and specifies the schema version in the OntologyParser as 5.2.0. this means we'll fetch the latest version of COG that's compliant with 5.2.0.

when we bump the schema version in the future, we'll need to bump the version specified in OntologyParser. i've added this to step 10 in the [confluence page](https://czi.atlassian.net/wiki/spaces/SCI/pages/2967797841/How+to+run+a+schema+migration)

## Testing steps

let's see what happens with the unit tests

## Checklist 🛎️

not applicable here

### DIFF
--- a/backend/common/census_cube/utils.py
+++ b/backend/common/census_cube/utils.py
@@ -12,7 +12,7 @@ from urllib3.util import Retry
 from backend.common.census_cube.data.snapshot import CensusCubeSnapshot
 
 # exported and used by all modules related to the census cube
-ontology_parser = OntologyParser()
+ontology_parser = OntologyParser(schema_version="v5.2.0")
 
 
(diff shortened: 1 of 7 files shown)


nayib-jose-gloria (changes requested):
looks like the ontologyparser is initiated in various spots w/o a version arg https://github.com/search?q=repo%3Achanzuckerberg%2Fsingle-cell-data-portal%20ontologyparser&type=code

this means they'll default to the latest schema version; in 1.3.1, that's 5.3.0 (which is not released yet). You should pass in  5.2.0 as an arg for all the OntologyParsers and then just shift all these to cellxgene-ontology-guide<2

Bento007 (approved):

PR #7426, opened 2025-02-04
feat(multiome): add manifest validator and converter

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/738
- Need a way to validate the manifest and convert existing data_uri to a manifest to work with the new ingest pipeline.

## Changes

- Add manifest schema in a pydantic model, 
- Add manifest schema validation 
- Add function to easily convert a uri into a manifest. This is a stop gap function until integration is complete. It also helps simplify some tests.
- Update ingestion Sfn caller to pass manifest to the Sfn. This is done by converting the url into a manifest.
- Add error handling for invalid manifest before calling ingestion Sfn
- Update ingestion Sfn to parse values from manifest
- Add unit test for the schema
- Update unit tests related to ingestions to support the new manifest.

## Testing steps

- added additional unit tests
- Updated tests


ivirshup (approved):
Looks mostly good, just a couple comments we talked about in person.

I would maybe think about doing validation with pydantic, since it's already a dependency, but not a requirement.

Bento007:
To avoid adding additional dependencies I'm happy to switch to pydantic.

PR #7431, opened 2025-02-11
feat: PUT for manifest files

ivirshup:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/738

## TODO

- [x] Tests
  - [x] Test submission of dataset with ATAC fragments
  - [x] Fill out tests for errors at new endpoint
- [x] Features
  - ~[ ] Submit ATAC to SFN~
- [x] Open questions
  - [x] Do we need to handle file sizes for ATAC fragments? **No**

I think it would be better if `ingest_dataset` ONLY took a manifest, and that the routing layer above it had to account for transforming the JSON. But the method is called in really a lot of tests, so changing it would touch a lot of code.

## Changes

- Added a PUT endpoint for manifest files

### DIFF
--- a/backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/manifest/actions.py
+++ b/backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/manifest/actions.py
@@ -1,0 +1,58 @@
+from flask import Response
+
+from backend.common.utils.exceptions import MaxFileSizeExceededException
+from backend.common.utils.http_exceptions import (
+    ForbiddenHTTPException,
+    InvalidParametersHTTPException,
+    MethodNotAllowedException,
+    NotFoundHTTPException,
+    TooLargeHTTPException,
+)
+from backend.curation.api.v1.curation.collections.common import (
+    _get_collection_and_dataset,
+)
+from backend.layers.auth.user_info import UserInfo
+from backend.layers.business.exceptions import (
+    CollectionIsPublishedException,
+    CollectionNotFoundException,
+    DatasetInWrongStatusException,
+    DatasetNotFoundException,
+    InvalidURIException,
+)
+from backend.portal.api.providers import get_business_logic
+
+
+def put(collection_id: str, dataset_id: str, body: dict, token_info: dict):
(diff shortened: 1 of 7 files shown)


backend/layers/business/business.py
```diff
@@ -555,32 +555,38 @@ def ingest_dataset(
...
+        except ValidationError as e:
+            raise InvalidIngestionManifestException("Ingestion manifest is invalid.", errors=e.errors()) from e
+        urls = [str(_url) for _url in manifest.dict(exclude_unset=True).values()]
+
         logger.info(
             {
                 "message": "ingesting dataset",
                 "collection_version_id": collection_version_id,
-                "url": url,  # TODO: change to manifest
+                "url": manifest,  # TODO: what should the behavior here be for manifest vs url?
```
Bento007, line 569:
```suggestion
                "manifest": manifest,
```

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/manifest/actions.py
```diff
@@ -0,0 +1,58 @@
...
+
+    if not UserInfo(token_info).is_user_owner_or_allowed(collection_version.owner):
+        raise ForbiddenHTTPException()
+
+    try:
+        business_logic.ingest_dataset(
+            collection_version.version_id,
+            body,
+            None,
+            None if dataset_id is None else dataset_version.version_id,
```
Bento007, line 39: when is this none?

Bento007 (approved):
One nit, otherwise it looks good.

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/manifest/actions.py
ivirshup: I don't know. I copied this over from the other `put` implementation https://github.com/chanzuckerberg/single-cell-data-portal/blob/3652b718bfad452f730407f7e677854c5ffe1874/backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/actions.py#L126

To me, `_get_collection_and_dataset` looks like it shouldn't return a `None` here.

PR #7434, opened 2025-02-12
feat: GET for manifest

ivirshup:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️


backend/layers/common/entities.py
```diff
@@ -80,6 +80,7 @@ class DatasetArtifactType(str, Enum):
     H5AD = "h5ad"
     RDS = "rds"
     CXG = "cxg"
+    ATAC_FRAGMENT = "bgz"  # TODO: This is not great, extension doesn't exactly map to "artifact type"
```
ivirshup, line 83: @Bento007 I don't love this, but AFAICT this is what the value *should* be.

WDYT?

backend/curation/api/v1/curation/collections/collection_id/datasets/dataset_id/manifest/actions.py
```diff
@@ -12,16 +12,50 @@
...
+    artifacts = dataset_version.artifacts
+    matches = [a for a in artifacts if a.type == artifact_type]
+
+    if len(matches) > 1:
+        raise ValueError(f"Multiple '{artifact_type}' artifacts found.")
+
+    if not matches and required:
+        raise ValueError(f"No '{artifact_type}' artifact found.")
+
+    return BusinessLogic.generate_permanent_url(dataset_version.version_id, artifact_type) if matches else None
```
Bento007, line 42: The atac fragment will not be stored on the `dataset_version.version_id` but instead under the artifact_id. We will need to handle that case when returning The permanent url.

backend/layers/common/entities.py
Bento007: I think having another table to translate the identifier we use and the actual extension would be helpful

tests/unit/backend/layers/api/test_curation_api.py
```diff
@@ -2640,6 +2641,111 @@ def test_get_dataset_id_version_4xx(self):
...
+                    collection_version=collection,
+                    artifacts=list(case["artifacts"].values()),
+                )
+
+                test_url = f"/curation/v1/collections/{dataset.collection_id}/datasets/{dataset.dataset_id}/manifest"
+                response = self.app.get(test_url)
+                self.assertEqual(200, response.status_code)
+
+                expected = {
+                    k: f"http://domain/{dataset.dataset_version_id}.{v.type}" for k, v in case["artifacts"].items()
```
Bento007, line 2709: Since we are using the artifact_id for the fragment file, this test will need to change to assert the fragments is named with the artifact_id.

Bento007 (changes requested):
One key feature we need to include is that fragment files will be prefix with the artifact_id not the dataset_version_id.

backend/layers/common/entities.py
ivirshup: That makes sense, but I have two concerns:

* Making it so the DatasetArtifactType value is not used as the extension will touch a lot of code
* The only case where the value is different (AFAIK) doesn't actually exist in the system/ database yet so I don't think I'm going to find all the cases

My personal preference for implementation would be to just no longer use an Enum here, and instead use classes so we can extend when needed. But I think that's a much larger code change, so probably won't go that direction.

Bento007 (approved):
Awesome, that works. Though I actually don't know why we started generating the permanent_urls rather than just using the URI in the artifact table.

PR #7436, opened 2025-02-14
fix: Census Models last-updated dates

mlin:
Several of the "Last Updated" dates on the [Census Models](https://cellxgene.cziscience.com/census-models) page show as December 31, 1969. I tracked it down to this comparison.


mlin:
> Yes, you should apply this correction. Comparing the time with getTime() instead of getDate() provides a more accurate comparison because getTime() returns the number of milliseconds since the Unix Epoch, whereas getDate() returns the day of the month. This change ensures that the comparison is performed correctly.

PR #7443, opened 2025-02-20
feat: support reingestion of a published dataset

Bento007:
## Reason for Change

- Support re-ingesting a published dataset artifacts a private revision if the publish datasets artifacts are part of one of the dataset versions of the canonical dataset.

## Changes

- add support for ingesting a publish dataset from the public dataset URL

## Testing steps

- unit tests have been updated to reflect the change

### DIFF
--- a/backend/layers/business/business.py
+++ b/backend/layers/business/business.py
@@ -547,7 +547,7 @@ class BusinessLogic(BusinessLogicInterface):
 
         return new_dataset_version
 
-    def is_public_uri(self, uri):
+    def is_already_ingested(self, uri):
         return str(uri).startswith(CorporaConfig().dataset_assets_base_url)
 
     # TODO: Alternatives: 1) return DatasetVersion 2) Return a new class
@@ -583,22 +583,24 @@ class BusinessLogic(BusinessLogicInterface):
         for key, _url in manifest.model_dump(exclude_unset=True).items():
             if not self.uri_provider.validate(_url):
                 raise InvalidURIException(f"Trying to upload invalid URI: {_url}")
-            if not self.is_public_uri(_url):
+            if not self.is_already_ingested(_url):
                 continue
             if not current_dataset_version_id:
                 raise InvalidIngestionManifestException(
-                    "Cannot ingest public datasets without a current dataset version"
+                    message="Cannot ingest public datasets without a current dataset version"
                 )
             if key == "anndata":
-                dataset_version_id, extension = url.split("/")[-1].split(".")
+                dataset_version_id, extension = url.split("/")[-1].split(".", maxsplit=1)
                 if extension != ARTIFACT_TO_EXTENSION[DatasetArtifactType.H5AD]:
-                    raise InvalidIngestionManifestException(f"{_url} is not an h5ad file")
+                    raise InvalidIngestionManifestException(message=f"{_url} is not an h5ad file")
                 previous_dv = self.database_provider.get_dataset_version(DatasetVersionId(dataset_version_id))
                 if previous_dv is None:
-                    raise InvalidIngestionManifestException(f"{_url} anndata file not found")
+                    raise InvalidIngestionManifestException(
+                        message=f"{_url} refers to existing dataset, but that dataset could not be found."
+                    )
                 all_dvs = self.database_provider.get_all_versions_for_dataset(previous_dv.dataset_id)
                 if current_dataset_version_id not in [dv.version_id for dv in all_dvs]:
-                    raise InvalidIngestionManifestException(f"{_url} is not apart of the canonical dataset")
+                    raise InvalidIngestionManifestException(message=f"{_url} is not a part of the canonical dataset")
                 manifest.anndata = [a for a in previous_dv.artifacts if a.type == DatasetArtifactType.RAW_H5AD][0].uri
 
         if file_size is None:
--- a/backend/layers/business/business.py
+++ b/backend/layers/business/business.py
@@ -590,7 +590,7 @@ class BusinessLogic(BusinessLogicInterface):
                     message="Cannot ingest public datasets without a current dataset version"
                 )
             if key == "anndata":
-                dataset_version_id, extension = url.split("/")[-1].split(".", maxsplit=1)
+                dataset_version_id, extension = _url.split("/")[-1].split(".", maxsplit=1)
                 if extension != ARTIFACT_TO_EXTENSION[DatasetArtifactType.H5AD]:
                     raise InvalidIngestionManifestException(message=f"{_url} is not an h5ad file")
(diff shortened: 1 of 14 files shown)


backend/layers/business/business.py
```diff
@@ -576,11 +578,28 @@ def ingest_dataset(
...
+            if key == "anndata":
+                dataset_version_id, extension = url.split("/")[-1].split(".")
+                if extension != ARTIFACT_TO_EXTENSION[DatasetArtifactType.H5AD]:
+                    raise InvalidIngestionManifestException(f"{_url} is not an h5ad file")
+                previous_dv = self.database_provider.get_dataset_version(DatasetVersionId(dataset_version_id))
+                if previous_dv is None:
+                    raise InvalidIngestionManifestException(f"{_url} anndata file not found")
+                all_dvs = self.database_provider.get_all_versions_for_dataset(previous_dv.dataset_id)
+                if current_dataset_version_id not in [dv.version_id for dv in all_dvs]:
+                    raise InvalidIngestionManifestException(f"{_url} is not apart of the canonical dataset")
```
ivirshup, line 601:
```suggestion
                    raise InvalidIngestionManifestException(f"{_url} is not a part of the canonical dataset")
```

backend/layers/persistence/persistence_mock.py
```diff
@@ -452,6 +452,15 @@ def get_all_versions_for_dataset(self, dataset_id: DatasetId) -> List[DatasetVer
...
 
+    def check_artifact_is_part_of_dataset(self, dataset_version_id: DatasetVersionId, artifact_id: DatasetArtifactId):
+        dataset_version = self.datasets_versions[dataset_version_id.id]
+        return any(artifact.id == artifact_id for artifact in dataset_version.artifacts)
+
+    def get_artifact_by_uri_suffix(self, uri_suffix: str) -> Optional[DatasetArtifact]:
+        for artifact in self.dataset_artifacts.values():
+            if artifact.uri.endswith(uri_suffix):
+                return artifact
+
```
ivirshup, line 463: I don't think either of these methods are called anywhere

tests/unit/backend/layers/business/test_business.py
```diff
@@ -856,6 +859,61 @@ def test_add_dataset_to_unpublished_collection_ok(self):
...
+            f'{{"anndata":"s3://artifacts/{dataset_version.version_id}/raw.h5ad","atac_fragment":null}}',
+        )
+
+    def test_reingest_published_dataset__not_h5ad(self):
+        """A cellxgene public dataset url used for reingesting an h5ad must be an h5ad"""
+        collection, revision = self.initialize_collection_with_an_unpublished_revision(num_datasets=1)
+        dataset_version = revision.datasets[0]
+        url = f"https://dataset_assets_domain/{dataset_version.version_id}.rds"
+        with self.assertRaises(InvalidIngestionManifestException):
+            self.business_logic.ingest_dataset(revision.version_id, url, None, dataset_version.version_id)
```
ivirshup, line 890: This is optional, but I think two meaningful improvements that could be made to these tests are:

* Checking text of the error to make sure the correct helpful message is given
* Ideally I would check that this message is actually returned in the response, e.g. make sure the curator can actually see the helpful error message

backend/layers/business/business.py
```diff
@@ -576,11 +578,28 @@ def ingest_dataset(
...
             if not self.uri_provider.validate(_url):
                 raise InvalidURIException(f"Trying to upload invalid URI: {_url}")
+            if not self.is_public_uri(_url):
+                continue
+            if not current_dataset_version_id:
+                raise InvalidIngestionManifestException(
+                    "Cannot ingest public datasets without a current dataset version"
+                )
+            if key == "anndata":
+                dataset_version_id, extension = url.split("/")[-1].split(".")
```
ivirshup, line 593: This will error (and throw a 500) if a fragment file has been passed (since `.split(".")` would return 3 items).

```suggestion
                dataset_version_id, extension = url.split("/")[-1].split(".", maxsplit=1)
```

This should allow something like `.tsv.gz` for extension

ivirshup (commented):
Main thing is that there are a couple methods which are not used anywhere, other than that minor naming/ error message things.

backend/layers/business/business.py
```diff
@@ -547,6 +547,9 @@ def create_empty_dataset_version_for_current_dataset(
 
         return new_dataset_version
 
+    def is_public_uri(self, uri):
+        return str(uri).startswith(CorporaConfig().dataset_assets_base_url)
```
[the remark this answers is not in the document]
Bento007: I can go for that.

backend/layers/persistence/persistence_mock.py
ivirshup: Same goes for the other definitions of these methods too

Bento007: they are called when we run the backend tests not in integration mode

ivirshup: Can you link to that? When I search the your branch for usage I don't see it:

image

image

Bento007: there are a few places where the unit tests fork and either use `DatabaseProviderMock`, or `DatabaseProvider`,  for example here https://github.com/chanzuckerberg/single-cell-data-portal/blob/7cfda1a5a0f5bf272cdd0862e220041e57cdd6e7/tests/unit/backend/layers/common/base_test.py#L108-L108.

ivirshup: Oh sorry for the confusion. 

My point here is not about the class. It is that I do not see these methods (the implementation here, or from any other class) being called anywhere. Like, if I do a search of the full codebase for the string `"check_artifact_is_part_of_dataset"` only the definitions show up. So these methods are currently dead code branches.

backend/layers/business/exceptions.py
```diff
@@ -91,7 +91,7 @@ class InvalidIngestionManifestException(DatasetIngestException):
     """
 
     def __init__(self, message: str, errors: Optional[List[str]] = None) -> None:
-        self.errors: Optional[List[dict]] = errors
+        self.errors: Optional[List[dict]] = errors if errors else []
```
ivirshup, line 94:
```suggestion
        self.errors: List[dict] = errors if errors else []
```

ivirshup (commented):
Main thing is the methods which I believe are not being called.

I think what happened here is that these were meant to go with another batch of changes, and were accidentally included in this PR.

backend/layers/persistence/persistence_mock.py
Bento007: Oh, you're right, they are part of my next PR

ivirshup (approved):
lgtm!

PR #7446, opened 2025-02-21
test artifact retention after dataset deletion

ivirshup:
MAIN FINDINGS:

* Curators do not have access to this endpoint
* delete_published=True is required if any version of a dataset was published. E.g. if an unpublished revision of a collection is created, you cannot delete a dataset from the new revision without passing in delete_published=True (I think this is dangerous)

## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️


ivirshup:
From our investigation: it turns out this approach will not work

Issue #7438, opened 2025-02-14
Flaky CellGuide test `test__source_collections_generator`


ivirshup:
[This cellguide test](https://github.com/chanzuckerberg/single-cell-data-portal/blob/5b4d0bf929004c795fea2d3ab86523faf418c4a1/tests/unit/backend/cellguide/pipeline/test_source_collections_generator.py#L19) is being flaky (e.g. I had to re-run CI 3 times [here](https://github.com/chanzuckerberg/single-cell-data-portal/actions/runs/13317149249/job/37198143378) and previous times). It would be great if it wasn't.

Otherwise, I would suggest we limit the impact of flaky tests on the rest of test infrastructure. In particular I would still like coverage reports to be uploaded even if some tests fail. Right now if any test fails in any suite we don't create a coverage report.

PR #7451, opened 2025-02-21
fix: CellGuide flakey test and coverage

kaloster:
## Reason for Change

- #7438 

## Changes

- Updated Coverage to always upload a report - `if: always()`
- Updated `compare_dict()` to handle lists of dictionaries (sort by keys before comparing) to ensure tests are deterministic

### DIFF
--- a/tests/test_utils/__init__.py
+++ b/tests/test_utils/__init__.py
@@ -26,7 +26,7 @@ def compare_dicts(dict1, dict2):
             if all(isinstance(v, dict) for v in value1) and all(isinstance(v, dict) for v in value2):
                 sorted_value1 = sorted(value1, key=lambda d: json.dumps(d, sort_keys=True))
                 sorted_value2 = sorted(value2, key=lambda d: json.dumps(d, sort_keys=True))
-                if not all(compare_dicts(d1, d2) for d1, d2 in zip(sorted_value1, sorted_value2, strict=False)):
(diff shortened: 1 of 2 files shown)


tests/test_utils/__init__.py
```diff
@@ -19,12 +21,16 @@ def compare_dicts(dict1, dict2):
...
-                for i in range(len(value1)):
-                    if not compare_dicts(value1[i], value2[i]):
-                        return False
-            elif sorted(value1) != sorted(value2):
+
+            # Handle lists of dictionaries (sort by keys before comparing)
+            if all(isinstance(v, dict) for v in value1) and all(isinstance(v, dict) for v in value2):
+                sorted_value1 = sorted(value1, key=lambda d: json.dumps(d, sort_keys=True))
+                sorted_value2 = sorted(value2, key=lambda d: json.dumps(d, sort_keys=True))
+                if not all(compare_dicts(d1, d2) for d1, d2 in zip(sorted_value1, sorted_value2, strict=False)):
```
ivirshup, line 29:
```suggestion
                if not all(compare_dicts(d1, d2) for d1, d2 in zip(sorted_value1, sorted_value2, strict=True)):
```

Only thing that's a little weird, since these should definitely be the same length

ivirshup (approved):
I think this looks good, but @atarashansky could review the test changes better.

The GitHub actions look good to me.

Thanks for the fix!

tests/test_utils/__init__.py
kaloster: Oops - agreed!

PR #7454, opened 2025-02-21
chore: Ivirshup/sync main w schema 5.3

ivirshup:


Bento007 (approved):

ivirshup:
@Bento007, I think the functional test failure here may be real. I'll investigate a little but will probably just scope this down to the one pr that fixed the flaky test.

ivirshup:
Now it looks like this is a pre-existing problem on the schema-5.3 branch so I am okay with merging this

PR #7457, opened 2025-02-25
feat: support processing atac seq fragments during ingestion

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/729
- https://github.com/chanzuckerberg/single-cell/issues/736
- This adds atac seq validation of fragments to the processing container.

### Features
- add `ProcessValidateATAC` to process atac seq data. First is is check to see if validation is needed. Otherwise it's skipped. Then validation happens. Followed by generating and uploading artifacts. If any errors occur, they are raise as `ValidationAtacFailed`.
- Update `ProcessMain` to call `ProcessValidateATAC`, and handle `ValidationAtacFailed` errors.
- add `DatasetStatusKey.ATAC_FRAGMENT` status to track the processing status of an atac dataset]. It is updated when processing an atac dataset. It will either be Uploaded, Skipped, or Failed by the end of processing depending on the `IngestionManifeset` provided.
- add `Persistance.get_artifact_by_uri_suffix` to find the artifact row associated with an URI for a published dataset. This is during reingestion of existing dataset. 
- add `BusinessLogic.add_artifact_to_dataset_version` so sharing an artifact across multiple dataset versions can be done easily. This is part of an optimization to avoid make excess copies of fragment files when no changes are detected.
- Modify delete datasets to also delete fragments files which follow a different naming format.
- Update `Persistance.update_validate_message` has been updated to append messages if messages exist.

### Fixs:
- changed the order of `ProcessLogic.create_artifact` to make more sense. Moved the bucket names to the end of the function definition.
- add `BusinessLogicInterface.update_dataset_artifact`. This was already defined, but was missing from the `BusinessLogicInterface`
- Use `'/'.join()` to create s3 uris, instead of `os.path.join`. This is less dependent on the OS.
- remove `ProcessLogic.make_s3_uri`, because it was causes us to generate uris with the `dataset_version_id` doubled. For example a s3 URI which should have been "s3://bucket/1234-1234.h5ad" was being stored as  "s3://bucket/1234-1234/1234-1234.h5ad" in the database. This only affects the uri in the database and not the actual location in s3.

## Testing steps

- tests have been updated to reflect the changes in the code.

## Checklist 🛎️
- [x] A follow up PR will be created to [add the atac seq processing to the ingestion pipeline](https://github.com/chanzuckerberg/single-cell-data-portal/pull/7458)
- [ ] The database will need to be check and updated for artifacts with incorrect URIs. I'll create a follow up PR for this work.


backend/layers/business/business.py
```diff
@@ -931,6 +938,14 @@ def update_dataset_artifact(self, artifact_id: DatasetArtifactId, artifact_uri:
...
         self.database_provider.update_dataset_artifact(artifact_id, artifact_uri)
 
+    def add_artifact_to_dataset_version(
+        self, dataset_version_id: DatasetVersionId, artifact_id: DatasetArtifactId
+    ) -> None:
+        """
+        Adds an artifact to a dataset version
+        """
+        self.database_provider.add_artifact_to_dataset_version(dataset_version_id, artifact_id)
+
```
ivirshup, line 948: I personally would not define this method and would instead just call `.database_provider.add_artifact_to_dataset_version`

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,219 @@
...
+        return False
+
+    def hash_file(self, file_name: str) -> str:
+        """
+        Hash the file
+        :param file_name: the file to hash
+        :return: the hash
+        """
+        with open(file_name, "rb") as f:
+            return hashlib.sha256(f.read()).hexdigest()
```
ivirshup, line 130: This will read the entire file into memory, which would be bad for the files that are being hashed.

In python 3.11, you can do [`hashlib.file_digest`](https://docs.python.org/3/library/hashlib.html#hashlib.file_digest):

```suggestion
        with open(file_name, "rb") as f:
            return hashlib.file_digest(f, "sha256").hexdigest()
```

But I think you are on 3.10 now so that'd look more like:

```suggestion
        # TODO: use hashlib.file_digest when we upgrade to python>=3.11
        hashobj = hashlib.sha256()
        buf = bytearray(2**18)
        view = memoryview(buf)

        with open(file_name, "rb") as f:
            while True:
                size = f.readinto(buf)
                if size == 0:
                    break  # EOF
                hashobj.update(view[:size])
        return hashobj.hexdigest()
```

I also don't know that this needs to be a method, and could be a utility.

backend/layers/thirdparty/schema_validator_provider.py
```diff
@@ -65,3 +77,16 @@ def add_labels(self, input_file: str, output_file: str) -> None:
...
     def count_matrix_nonzero(self, matrix):
         return validate.Validator.count_matrix_nonzero(matrix)
+
+    def validate_atac(self, fragment_file, anndata_file) -> Optional[List[str]]:
+        """
+        Validates an ATAC fragment file against an anndata file.
+        """
+        import cellxgene_schema.atac_seq as atac_seq
+
+        return atac_seq.validate_fragment(fragment_file, anndata_file)
```
ivirshup, line 87: This is installing from `main` right?

I do not see a function called `validate_fragment` in this module: https://github.com/chanzuckerberg/single-cell-curation/blob/main/cellxgene_schema_cli/cellxgene_schema/atac_seq.py

Not totally why coverage isn't being complained about here.

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,219 @@
...
+            # if the validation fails, update the processing status and raise a ValidationAtacFailed exception
+            self.update_processing_status(
+                dataset_version_id, DatasetStatusKey.ATAC_FRAGMENT, DatasetConversionStatus.FAILED
+            )
+            raise ValidationAtacFailed(errors=errors)
+
+        # check to see if the new fragments is the same as the old fragment
+        # if it is the same, skip the upload and use link the old fragment to the new dataset
+        # This case is for when the dataset is reprocessed, or updated
+        if original_fragment_hash is not None and original_fragment_hash == self.hash_file(local_fragment_filename):
```
ivirshup, line 190: I don't understand the logic here. On line 164 above:

```python
original_fragment_hash = self.hash_file(local_fragment_filename)
```

But it's not immediately obvious to me that that path would be overwritten by now.

I am also not able to figure out what `self.schema_validator.validate_atac` does at the moment (see other comment)

backend/layers/processing/process_validate_atac.py
ivirshup: I would probably also go for a faster hash like md5 here
```suggestion
        # TODO: use hashlib.file_digest when we upgrade to python>=3.11
        hashobj = hashlib.md5()
        buf = bytearray(2**18)
        view = memoryview(buf)
        with open(file_name, "rb") as f:
            while True:
                size = f.readinto(buf)
                if size == 0:
                    break  # EOF
                hashobj.update(view[:size])
        return hashobj.hexdigest()
```

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,219 @@
...
+            source_uri=str(manifest.atac_fragment), local_path=CorporaConstants.ORIGINAL_ATAC_FRAGMENT_FILENAME
+        )
+
+        # Hash the original fragment file. This is used to check if the new fragment file is the same as the old
+        # fragment file to avoid uploading the same file multiple times
+        if self.business_logic.is_already_ingested(manifest.atac_fragment):
+            original_fragment_hash = self.hash_file(local_fragment_filename)
+        else:
+            # new fragment file is from a private source, so we must upload it
+            original_fragment_hash = None
```
ivirshup, line 167: I think it would make sense to store the hash in the database.

backend/layers/processing/process_validate_atac.py
ivirshup: Depending on the definition of `validate_atac`, do we actually expect it to return exactly the same result every time? I wouldn't expect deterministic hash values from the processing because you were appending chromosomes of data to the fragment file based on when they were done, not with a global order.

backend/layers/persistence/persistence.py
```diff
@@ -840,6 +842,16 @@ def get_dataset_artifacts_by_version_id(self, dataset_version_id: DatasetVersion
...
 
+    def get_artifact_by_uri_suffix(self, uri_suffix: str) -> Optional[DatasetArtifact]:
+        """
+        Returns the artifact with the given uri suffix
+        """
+        with self._manage_session() as session:
+            artifact = session.query(DatasetArtifactTable).filter(DatasetArtifactTable.uri.endswith(uri_suffix)).first()
+            if artifact:
+                return self._row_to_dataset_artifact(artifact)
+            return None
```
ivirshup, line 853: This seems like a very permissive approach to me. What is the use-case where there is more than one entry that matches the query, but it's fine to return whichever?

ivirshup (commented):
I've gotten started, but will probably need another pass.

backend/layers/processing/process_validate_atac.py
Bento007: why do you have 
```
        buf = bytearray(2**18)
        view = memoryview(buf)
```

backend/layers/thirdparty/schema_validator_provider.py
Bento007: whoops, should be `process_fragment`

backend/layers/processing/process_validate_atac.py
ivirshup: So that you reuse the buffer instead of allocating new ones.

Totally forgot to mention this is based off the definition of `hashlib.file_digest` in python 3.11

backend/layers/processing/process_validate_atac.py
Bento007: True, it may not be consistent. Have any ideas how to get a consistent hash?

backend/layers/processing/process_validate_atac.py
Bento007: in what table and field?

backend/layers/processing/process_validate_atac.py
Bento007: if could calculate the hash in the cli, we could hash the separate chromosomes and combine those hashes in order to achieve a consistent hash. I don't think it's possible for us to do in the 1 day we have remaining for this work.

ivirshup: I would make the append step serial so you can control the order

ivirshup: I'm also not actually sure if gzip is deterministic by default. A quick google says you might need to set some flags. Either that or hash the decompressed data.

Bento007: We can create a ticket to do more validation in this area.

backend/layers/processing/process_validate_atac.py
Bento007: That's something we can do as a future optimzation

Bento007:
This ticket was broken up into

https://github.com/chanzuckerberg/single-cell-data-portal/pull/7466
https://github.com/chanzuckerberg/single-cell-data-portal/pull/7467
https://github.com/chanzuckerberg/single-cell-data-portal/pull/7468
https://github.com/chanzuckerberg/single-cell-data-portal/pull/7469

PR #7458, opened 2025-02-26
feat: update sfn with validate atac

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/745
- https://github.com/chanzuckerberg/single-cell/issues/734
- Add functional tests for atac_seq
- add validate-atac as a step in the ingestion pipeline. It will run in parallel with validate anndata.

## Changes

- Add functional test for atac seq
- Add atac as a step in the ingestion step function
- modify

## Testing steps

## Checklist 🛎️


.happy/terraform/modules/sfn/main.tf
```diff
@@ -12,196 +13,283 @@ resource "aws_sfn_state_machine" "state_machine" {
...
-                "Name": "DATASET_VERSION_ID",
-                "Value.$": "$.dataset_version_id"
+        {
+          "StartAt": "ValidateAtac",
+          "States": {
+            "ValidateAtac": {
+              "Type": "Task",
+              "Resource": "arn:aws:states:::batch:submitJob.sync",
+              "Parameters": {
+                "JobDefinition": "${var.job_definition_arn}",
```
nayib-jose-gloria, line 95: Do you think it'll be necessary to create a separate batch job def for validateAtac? reminder that, with the dask changes, we changed this batch default (used for validate and add labels) to use 8 GB 1 vcpu. CXG has its own definition using 16 GB and 2 vcpu. Will validate atac need more? Is it chunked / parallelized similarly?

nayib-jose-gloria (commented):
1 question re: resource allocation requirements

.happy/terraform/modules/sfn/main.tf
nayib-jose-gloria: yeah, I think we can track that in a follow-up issue. When I tried, I was finding that the speed-up was not as much of a performance increase as I expected (almost negligible even on larger datasets) but that might've been an implementation error on my part.

.happy/terraform/modules/sfn/main.tf
```diff
@@ -92,7 +92,7 @@ resource "aws_sfn_state_machine" "state_machine" {
               "Type": "Task",
               "Resource": "arn:aws:states:::batch:submitJob.sync",
               "Parameters": {
-                "JobDefinition": "${var.job_definition_arn}",
+                "JobDefinition": "${var.cxg_definition_arn}",
```
nayib-jose-gloria, line 95: not blocking, but we should either make it its own definition or rename the cxg definition to something more generic like 2_core_definition

nayib-jose-gloria (approved):
approved with 1 suggestion re: cxg definition name

PR #7467, opened 2025-02-27
fix: cleanup processing code and tests

Bento007:
## Changes

- Use '/'.join() to create s3 uris, instead of os.path.join. This is less dependent on the OS.
- remove ProcessLogic.make_s3_uri, it's redundant to ProcessLogic.upload_artifact
- changed the order of ProcessLogic.create_artifact to make more sense. Moved the bucket names to the end of the function definition.
- add BusinessLogicInterface.update_dataset_artifact. This was already defined, but was missing from the BusinessLogicInterface
- made variable naming more exact to make distinguishing ATAC relate code easier

## Testing steps

- made tests

## Checklist 🛎️


backend/layers/processing/process_logic.py
```diff
@@ -145,7 +131,7 @@ def get_key_prefix(self, identifier: str) -> str:
```
nayib-jose-gloria, line 131: can remove "import os"

nayib-jose-gloria (approved):
1 nit to remove an import statement, otherwise looks good

PR #7468, opened 2025-02-27
feat: Add ValidateATAC

Bento007:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell/issues/729

## Changes

- add `ProcessValidateATAC` to process atac seq data. First is is check to see if validation is needed. Otherwise it's skipped. Then validation happens. Followed by generating and uploading artifacts. If any errors occur, they are raise as `ValidationAtacFailed`.
- `DatasetStatusKey.ATAC_FRAGMENT` status is updated when processing an atac dataset. It will either be Uploaded, Skipped, or Failed by the end of processing depending on the `IngestionManifeset` provided.
- Update `ProcessMain` to call `ProcessValidateATAC`, and handle `ValidationAtacFailed` errors.
- add `Persistance.get_artifact_by_uri_suffix` to find the artifact row associated with an URI for a published dataset. This is during reingestion of existing dataset. 
- add `BusinessLogic.add_artifact_to_dataset_version` so sharing an artifact across multiple dataset versions can be done easily. This is part of an optimization to avoid make excess copies of fragment files when no changes are detected.


backend/layers/persistence/persistence.py
```diff
@@ -892,6 +900,18 @@ def update_dataset_artifact(self, artifact_id: DatasetArtifactId, artifact_uri:
             artifact = session.query(DatasetArtifactTable).filter_by(id=artifact_id.id).one()
             artifact.uri = artifact_uri
 
+    def add_artifact_to_dataset_version(
```
nayib-jose-gloria, line 903: can we rename this to better differentiate it from [add_dataset_artifact](https://github.com/chanzuckerberg/single-cell-data-portal/blob/0d7c89d23b2097e5ca41076022af71665efa2520/backend/layers/persistence/persistence.py#L879)? and update description(s) accordingly?

maybe add_dataset_artifact_by_id? or something?

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+        :param local_anndata_filename: the local anndata file
+        :param manifest: the manifest
+        :param dataset_version_id: the dataset version id
+        :return: True if the validation should be skipped, False otherwise
+        """
+        # check if the anndata should have a fragment file
+        try:
+            result = self.schema_validator.check_anndata_requires_fragment(local_anndata_filename)
+        except ValueError as e:  # fragment file forbidden
+            self.logger.warning(f"Anndata does not support atac fragment files for the follow reason: {e}")
```
nayib-jose-gloria, line 89: nit: typo
```suggestion
            self.logger.warning(f"Anndata does not support atac fragment files for the following reason: {e}")
```

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+                    DatasetValidationStatus.INVALID,
+                )
+                raise ValidationAtacFailed(errors=[str(e), "Fragment file not allowed for non atac anndata."]) from None
+            self.logger.warning("Skipping fragment validation")
+            self.update_processing_status(
+                dataset_version_id,
+                DatasetStatusKey.ATAC_FRAGMENT,
+                DatasetConversionStatus.SKIPPED,
+                validation_errors=[str(e)],
+            )
```
nayib-jose-gloria, line 103: missing context for me--why do we need to set ATAC_STATUS as skipped during dataset initialization if we set it here?

nayib-jose-gloria: Also, advocating that if it truly doesn't requite fragment validation and one wasn't submitted, this should be NA rather than SKIPPED. since it was only skipped bc its not applicable to this dataset.

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+                    dataset_version_id, DatasetStatusKey.VALIDATION, DatasetValidationStatus.INVALID
+                )
+                raise ValidationAtacFailed(errors=["Anndata requires fragment file"])
+            else:  # fragment file optional
+                self.logger.info("Fragment is optional and not present. Skipping fragment validation.")
+                self.update_processing_status(
+                    dataset_version_id,
+                    DatasetStatusKey.ATAC_FRAGMENT,
+                    DatasetConversionStatus.SKIPPED,
+                    validation_errors=["Fragment is optional and not present."],
```
nayib-jose-gloria, line 118: conversely, this seems to me like the correct usage of SKIPPED. atac IS applicable, but was skipped since it was optional

backend/layers/processing/process_validate_atac.py
Bento007: The skipped is temporarily since ATAC is not part of the pipeline yet. It's removed [here](https://github.com/chanzuckerberg/single-cell-data-portal/pull/7458/files).

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+            original_fragment_hash = None
+
+        # Validate the fragment with anndata file
+        try:
+            errors = self.schema_validator.validate_atac(local_fragment_filename, local_anndata_filename)
+        except Exception as e:
+            # for unexpected errors, log the exception and raise a ValidationAtacFailed exception
+            self.logger.exception("validation failed")
+            self.update_processing_status(
+                dataset_version_id, DatasetStatusKey.ATAC_FRAGMENT, DatasetConversionStatus.FAILED
```
nayib-jose-gloria, line 181: we should make ValidationStatus FAILED as well, right?

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+            )
+            raise ValidationAtacFailed(errors=[str(e)]) from None
+
+        if errors:
+            # if the validation fails, update the processing status and raise a ValidationAtacFailed exception
+            self.update_processing_status(
+                dataset_version_id, DatasetStatusKey.ATAC_FRAGMENT, DatasetConversionStatus.FAILED
```
nayib-jose-gloria, line 188: same as above

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+            )
+            raise ValidationAtacFailed(errors=errors)
+
+        # check to see if the new fragments is the same as the old fragment
+        # if it is the same, skip the upload and use link the old fragment to the new dataset
+        # This case is for when the dataset is reprocessed, or updated
+        if original_fragment_hash is not None and original_fragment_hash == self.hash_file(local_fragment_filename):
```
nayib-jose-gloria, line 195: Why do we need to take the hashes and compare them at all if original_fragment_hash is only populated if we confirm that the given uri is already ingested?

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+        :param artifact_type: the type of artifact to upload
+        :param dataset_version_id: the dataset version id
+        :param processing_status_key: the key to update the processing status
+        :param datasets_bucket: the bucket to upload the dataset to
+        :param fragment_artifact_id: the artifact id of the fragment file to be use in the fragment index file.
+        :return:
+        """
+        self.update_processing_status(dataset_version_id, processing_status_key, DatasetConversionStatus.UPLOADING)
+        try:
+            artifact_id = self.business_logic.add_dataset_artifact(dataset_version_id, artifact_type, "dummy")
```
nayib-jose-gloria, line 59: why are we passing in a dummy value?

nayib-jose-gloria: I'd rather add a new function that generates an artifact ID, and create (or reuse if there is one) a function that lets us insert with our preferred artifact ID instead.

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+        self.uri_provider = uri_provider
+        self.s3_provider = s3_provider
+        self.schema_validator = schema_validator
+
+    def create_atac_artifact(
+        self,
+        file_name: str,
+        artifact_type: DatasetArtifactType,
+        dataset_version_id: DatasetVersionId,
+        processing_status_key: DatasetStatusKey,
```
nayib-jose-gloria, line 42: why is this an arg if there's only one ATAC_STATUS?

backend/layers/processing/process_validate_atac.py
Bento007: That will happen here
https://github.com/chanzuckerberg/single-cell-data-portal/blob/460bf988b5206454c17e22d672ed49ed9fe8c6c9/backend/layers/processing/process.py#L146-L146

backend/layers/processing/process_validate_atac.py
```diff
@@ -0,0 +1,224 @@
...
+        else:
+            fragment_artifact_id = self.create_atac_artifact(
+                local_fragment_filename,
+                DatasetArtifactType.ATAC_FRAGMENT,
+                dataset_version_id,
+                DatasetStatusKey.ATAC_FRAGMENT,
+                datasets_bucket,
+            )
+            self.create_atac_artifact(
+                local_fragment_filename + ".tbi",
```
nayib-jose-gloria, line 216: is this file created as part of `self.schema_validator.validate_atac(local_fragment_filename, local_anndata_filename)`? if so, could we make that more clear somehow? otherwise it seems like we're passing in a filepath to nothing

backend/layers/processing/process_validate_atac.py
Bento007: https://github.com/chanzuckerberg/single-cell-data-portal/pull/7468#discussion_r1974414041

backend/layers/processing/process_validate_atac.py
Bento007: because the CLI might change which may process the data differently.

backend/layers/processing/process_validate_atac.py
nayib-jose-gloria: got it, thanks

nayib-jose-gloria: in that case, def think this should be NA to distinguish from the SKIPPED case below

backend/layers/processing/process_validate_atac.py
nayib-jose-gloria: I see. In that case, shouldn't we be comparing the TBI hashes rather than the TSV? since the TSV is pointing to the same file, its the TBI that could be generated differently.

nayib-jose-gloria: could be misunderstanding something here

nayib-jose-gloria: also, to consider--for very large files, reading from disk again would make this a pretty slow operation and cost us in AWS batch

backend/layers/processing/process_validate_atac.py
Bento007: i'll remove and inline it.

backend/layers/thirdparty/schema_validator_provider.py
```diff
@@ -65,3 +77,16 @@ def add_labels(self, input_file: str, output_file: str) -> None:
...
     def count_matrix_nonzero(self, matrix):
         return validate.Validator.count_matrix_nonzero(matrix)
+
+    def validate_atac(self, fragment_file, anndata_file) -> Optional[List[str]]:
+        """
+        Validates an ATAC fragment file against an anndata file.
+        """
+        import cellxgene_schema.atac_seq as atac_seq
+
+        return atac_seq.process_fragment(fragment_file, anndata_file)
```
Bento007, line 87:
```suggestion
        return atac_seq.process_fragment(fragment_file, anndata_file), file_name, file_name.tbi
```

backend/layers/processing/process_validate_atac.py
Bento007: No the files will only be hashed during a migration.

backend/layers/processing/process_validate_atac.py
```diff
@@ -49,28 +49,29 @@ def create_atac_artifact(
         :param file_name: the local file to upload
         :param artifact_type: the type of artifact to upload
         :param dataset_version_id: the dataset version id
-        :param processing_status_key: the key to update the processing status
         :param datasets_bucket: the bucket to upload the dataset to
         :param fragment_artifact_id: the artifact id of the fragment file to be use in the fragment index file.
```
nayib-jose-gloria, line 53:
```suggestion
        :param fragment_artifact_id: the artifact id of the fragment file, to be used in the fragment index filepath for storage
```

nayib-jose-gloria: reads slightly clearer for me, i figured it out with context but was confused at first at why there were two identifiers

backend/layers/processing/process_validate_atac.py
```diff
@@ -97,8 +98,8 @@ def skip_atac_validation(
             self.logger.warning("Skipping fragment validation")
```
nayib-jose-gloria, line 98:
```suggestion
            self.logger.warning("Fragment validation not applicable for dataset assay type.")
```

backend/layers/processing/process_validate_atac.py
```diff
@@ -163,60 +164,52 @@ def process(
...
-        if self.business_logic.is_already_ingested(manifest.atac_fragment):
-            original_fragment_hash = self.hash_file(local_fragment_filename)
-        else:
-            # new fragment file is from a private source, so we must upload it
-            original_fragment_hash = None
-
         # Validate the fragment with anndata file
         try:
-            errors = self.schema_validator.validate_atac(local_fragment_filename, local_anndata_filename)
+            errors, fragment_file, fragment_index_file = self.schema_validator.validate_atac(
```
nayib-jose-gloria, line 169: wrong return order I think; it's returning (errors, index, fragment) currently

backend/layers/processing/process_validate_atac.py
```diff
@@ -163,60 +164,52 @@ def process(
...
-        # if it is the same, skip the upload and use link the old fragment to the new dataset
-        # This case is for when the dataset is reprocessed, or updated
-        if original_fragment_hash is not None and original_fragment_hash == self.hash_file(local_fragment_filename):
+        # Changes to processing only happen during a migration. Only hash the files if the migration is set to true
+        in_migration = os.environ.get("MIGRATION", "").lower() == "true"
+        if in_migration:
+            # check if the new fragment is the same as the old fragment
+            fragment_unchaged = self.hash_file(local_fragment_filename) == self.hash_file(fragment_file)
+        else:
+            fragment_unchaged = False
```
nayib-jose-gloria, line 189: typo: fragment_unchanged

backend/layers/processing/process_validate_atac.py
```diff
@@ -163,60 +164,52 @@ def process(
...
+
+        # fragment file to avoid uploading the same file multiple times
+        # if the frament file is unchanged from a migration or the fragment file is already ingested, use the old fragment.
```
nayib-jose-gloria, line 192:
```suggestion
        # if the fragment file is unchanged from a migration or the fragment file is already ingested, use the old fragment.
```

PR #7479, opened 2025-03-06
feat: Change DatasetArtifact.type from enum to String

Bento007:
## Reason for Change

- Changing `DatasetArtifact.type` from enum to String will make it easier to add new artifact types in the future without doing a database migration. This is safe because only the backend code has access to the database and will handle enforment. 

## Changes

- add database migration to change `DatasetArtifact.type` from an enum to a string
- update `orm.py` to make `DatasetArtifact.type`

## Testing steps

- tested the migration locally
- Ran the migration on dev and verified existing datasets could be retrieved and that new datasets can be created.

### DIFF
--- a/backend/layers/persistence/orm.py
+++ b/backend/layers/persistence/orm.py
@@ -1,9 +1,8 @@
-from sqlalchemy import Column, DateTime, Enum, ForeignKey, String
+from sqlalchemy import Column, DateTime, ForeignKey, String
 from sqlalchemy.dialects.postgresql import ARRAY, BOOLEAN, JSON, UUID
 from sqlalchemy.orm import registry
 from sqlalchemy.schema import MetaData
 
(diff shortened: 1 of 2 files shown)


backend/layers/persistence/orm.py
```diff
@@ -68,5 +67,5 @@ class DatasetArtifactTable:
     __tablename__ = "DatasetArtifact"
 
     id = Column(UUID(as_uuid=True), primary_key=True)
-    type = Column(Enum(DatasetArtifactType))
+    type = Column(String)
```
joyceyan, line 70: we can do this in a follow up PR, but it might make sense to use string literals that way you still have some type enforcement but passing strings makes it more flexible for db / API layer.

https://mypy.readthedocs.io/en/stable/literal_types.html

joyceyan (approved):

PR #7484, opened 2025-03-10
fix:  support uploading files from a cellxgene public url

Bento007:
## Reason for Change

- support uploading file using a public cellxgene url

## Changes

- add CXGPublicURL to parse public cellxgene urls
- remove import side effects when importing `backend.common.utils.dl_sources.uri` by moving `RegisteredSources.add` calls to inside `from_uri` if `RegisteredSources` has no sources.
- renamed `_registered` to `_sources` in `RegisteredSources`.
- To help with testing the `dataset_assets_base_url` is defined in `CorporaConfig` when `DEPLOYMENT_STAGE == test`

## Testing steps

- Added unit tests for `S3URL` and `CXGPublicURL`

### DIFF
--- a/backend/common/utils/dl_sources/uri.py
+++ b/backend/common/utils/dl_sources/uri.py
@@ -6,6 +6,7 @@ from urllib.parse import ParseResult, urlparse
 import boto3
 import requests
 
+from backend.common.corpora_config import CorporaConfig
 from backend.common.utils import downloader
 from backend.layers.thirdparty.s3_provider import S3Provider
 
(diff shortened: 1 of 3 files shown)


backend/common/utils/dl_sources/uri.py
```diff
@@ -160,6 +161,20 @@ def download(self, local_file_name: str):
         downloader.download(self.uri, local_file_name)
 
 
+class CXGPublicURL(S3URL):
```
nayib-jose-gloria, line 164: missing a validate abstractmethod implementation. I'm assuming we'll want to make sure we only accept datasets from the expected domain / format

nayib-jose-gloria: NVM, just noticed it extends S3URL.

PR #7493, opened 2025-03-18
fix: type hint on DatasetArtifactType + persist using DatasetArtifactType.name for consistency

nayib-jose-gloria:
## Reason for Change

- feature/schema-5-3 includes an alembic upgrade to the DatasetArtifact type column in the db--from enum to str. This is causing there to be a difference between how type was represented in the db before vs. how it is now. New uploads are represented in the db as the enum value (“raw_h5ad”) whereas previous uploads are represented as the enum name (“RAW_H5AD”). This is breaking some code that is trying to do comparisons against the DatasetArtifactType enum, including get_ingestion_manifest.
- Led to issues during 5.3.0 schema migration testing 

## Changes

- fix type hints to create_dataset_artifact and add_dataset_artifact, which hinted str but were getting passed in DatasetArtifactType enum
- Create dataset artifacts using DatasetArtifactType.name, to enforce consistency with previous db rows and maintain codebase expectations
- Also enforce type as an enum when translating between persistence object DatasetArtifact and the business layer object DatasetArtifact (uses type as an enum)

## Testing steps

- rdev migration + unit tests


backend/layers/business/business.py
```diff
@@ -606,7 +606,7 @@ def ingest_dataset(
 
         # Validate the URIs
         # TODO: This should be done in the IngestionManifest class
-        for key, _url in manifest.model_dump(exclude_unset=True).items():
+        for key, _url in manifest.model_dump(exclude_none=True).items():
```
ejmolinelli, line 609: is this config change to avoid the case when the `atac_fragment` is set to None directly?

nayib-jose-gloria: Yup, which is the case in `get_ingestion_manifest` invoked by the schema migration.

ejmolinelli (approved):

PR #7494, opened 2025-03-19
fix: update rdev db seed to include visium and atac datasets

Bento007:
## Reason for Change

- Updating the rdev database seed to include visium and atac datasets. This is to allow frontend developer to test their changes in an rdev without uploading additional files. It also improves our ability to test changes as the types of data we ingest continues to evolve.

## Changes
- Use the curation API to upload datasets using the manifest endpoint to the rdev database.
- make `create_and_publish_collection` reusuable
- seed db creates a collection with a visium datasets if it does not exist
- seed db creates a collection with a atac dataset if it does not exist.

## Testing steps
verified the datasets exist in the rdev environment
Screenshot 2025-03-20 at 10 43 34 AM
Screenshot 2025-03-20 at 10 43 55 AM

### DIFF
--- a/scripts/smoke_tests/setup.py
+++ b/scripts/smoke_tests/setup.py
@@ -6,17 +6,17 @@ import threading
 
 from backend.common.constants import DATA_SUBMISSION_POLICY_VERSION
 from backend.common.corpora_config import CorporaAuthConfig
-from tests.functional.backend.constants import API_URL, DATASET_URI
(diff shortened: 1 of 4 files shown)


ejmolinelli (approved):
I'm very new to the underlying business logic being addressed here. But it looks fine to me. It appears to cover the stated changes in the PR description.

PR #7496, opened 2025-03-20
feat: cleanup up artifacts from containers during ingestion

Bento007:
## Reason for Change

- During ingestion and migration, files used for processing were not stored in a location that 

## Changes

- add `TemporaryDirectory` to `ProcesssLogic`. All downloaded and generated artifact are now store in this directory and cleanup. `TemporaryDirectory.cleanup()` is called explicitly in `ProcessMain.process`, but is not required. The `TemporaryDirectory` will cleanup the files when the object is destroyed. 
- Updated ingestion steps to store downloaded and generated artifact in the `TemporaryDirectory`.
- remove test h5ads when running processing test. This was polluting my local environment with tons of h5ad I had to manually delete. Now they are cleaned up after the test.

## Testing steps

- updated test to reflect the changes

## Checklist 🛎️


ivirshup:
@Bento007 do you think the test failures here are related to the changes?

I think the changes look good, but the failures look *maybe* related

PR #7500, opened 2025-03-24
fix: Next critical update

kaloster:
## Reason for Change

Hotfix for prod - updating `next` to the latest patch version


tihuan (approved):
Amazing thanks so much!

PR #7506, opened 2025-03-25
fix: bigger ATAC compute

Bento007:
## Reason for Change

- per @ivirshup suggestion based off the recent changes to ATAC validation, the compute resources need to validate ATAC datasets must be increased. 
 
## Changes

- add a new batch job compute definition for ATAC to happy infra. This will be used in the ingestion SFN to validate ATAC datasets.
- ATAC uses 8 vcpu and 64GB memory

## Testing steps

- verify ingestion pipeline run correctly for ATAC datasets
- verify the largest ATAC datasets runs through successfully.


.happy/terraform/modules/batch/main.tf
```diff
@@ -59,7 +59,7 @@ resource aws_batch_job_definition cxg_job_def {
   container_properties = jsonencode({
   "jobRoleArn": "${var.batch_role_arn}",
   "image": "${var.image}",
-  "memory": 16000,
+  "memory": 48000,
```
Bento007, line 62:
```suggestion
  "memory": 16000,
```

.happy/terraform/modules/batch/main.tf
```diff
@@ -90,7 +90,55 @@ resource aws_batch_job_definition cxg_job_def {
       "value": "${var.frontend_url}"
     }
   ],
-  "vcpus": 2,
+  "vcpus": 7,
```
Bento007, line 93:
```suggestion
  "vcpus": 8,
```

.happy/terraform/modules/batch/main.tf
```diff
@@ -90,7 +90,55 @@ resource aws_batch_job_definition cxg_job_def {
       "value": "${var.frontend_url}"
     }
   ],
-  "vcpus": 2,
+  "vcpus": 8,
```
Bento007, line 93:
```suggestion
  "vcpus": 2,
```

.happy/terraform/modules/batch/main.tf
```diff
@@ -101,6 +101,54 @@ resource aws_batch_job_definition cxg_job_def {
...
+    {
+      "name": "REMOTE_DEV_PREFIX",
+      "value": "${var.remote_dev_prefix}"
+    },
+    {
+      "name": "FRONTEND_URL",
+      "value": "${var.frontend_url}"
+    }
+  ],
+  "vcpus": 8,
```
valenzuelaomar, line 141: what was  the previous job definitions vCPU's & memory?

valenzuelaomar:
Couple of noob questions! 
1. What are the recent changes that @ivirshup introduced which require more compute? (Link to PR maybe?)
2. What was the previous job definition's vCPU's & memory?
3. Do you have a link to jobs that have run out of compute?

ivirshup (approved):
I think this looks good to me, but take that with a grain of salt since I'm not that experienced with Terraform. 

The only suggestion I would make is that you can probably throw more CPUs at this and they will be used. But I think it's also fine to go ahead as is.

ivirshup:
@valenzuelaomar, my relevant PRs are:

I'm not sure that these changes require more compute per se. We had a dataset that was requiring too much memory and using too much disk space. 

* https://github.com/chanzuckerberg/single-cell-curation/pull/1314

This PR reduced total disk space used by removing an uncompressed intermediate. This did also reduce memory usage for initial convert, but I forget if that had been limiting.

* https://github.com/chanzuckerberg/single-cell-curation/pull/1315

This PR got peak memory usage low enough that I could process the aforementioned dataset on a 128gb machine without running out of memory. I was seeing peek memory usage of about 40gb here.

.happy/terraform/modules/batch/main.tf
```diff
@@ -101,6 +101,54 @@ resource aws_batch_job_definition cxg_job_def {
 })
 }
 
+resource aws_batch_job_definition atac_job_def {
+  type = "container"
+  name = "dp-${var.deployment_stage}-${var.custom_stack_name}-atac"
+  container_properties = jsonencode({
+  "jobRoleArn": "${var.batch_role_arn}",
+  "image": "${var.image}",
+  "memory": 64000,
```
ebezzi, line 110: I recommend using less than this. If you use 64000, it might not fit in a 64G machine and will require a bigger type, due to the OS and the container overhead. Using 60k-62k is a safer option.

ebezzi:
> I think this looks good to me, but take that with a grain of salt since I'm not that experienced with Terraform.
> 
> The only suggestion I would make is that you can probably throw more CPUs at this and they will be used. But I think it's also fine to go ahead as is.

If we want to use more vCPUs, then we might have to switch to a different instance class. Right now I believe we're using `r` which are memory optimized. Problem is, these share the same compute environment as the regular ingestion pipeline which is historically memory hungry (mostly due to the validator/Seurat converter being quite inefficient), so that would likely require using a different compute environment. You don't have the same granularity in choosing compute resources as you would get with Argo or Kubernetes based orchestrators.

nayib-jose-gloria:
> > I think this looks good to me, but take that with a grain of salt since I'm not that experienced with Terraform.
> > The only suggestion I would make is that you can probably throw more CPUs at this and they will be used. But I think it's also fine to go ahead as is.
> 
> If we want to use more vCPUs, then we might have to switch to a different instance class. Right now I believe we're using `r` which are memory optimized. Problem is, these share the same compute environment as the regular ingestion pipeline which is historically memory hungry (mostly due to the validator/Seurat converter being quite inefficient), so that would likely require using a different compute environment. You don't have the same granularity in choosing compute resources as you would get with Argo or Kubernetes based orchestrators.

seurat converter is being deprecated and the validator has been made significantly more memory efficient with the introduction of dask and chunking--its running on 1 vcpu, 8 gb machines now for all datasets

PR #7510, opened 2025-03-28
fix: Collection Edit Details bug

kaloster:
## Reason for Change

An edge case was [reported](https://czi-sci.slack.com/archives/C07AV4NU9D2/p1742501442196159?thread_ts=1742501316.004119&cid=C07AV4NU9D2) where certain letters are blocked when typed as leading letters. 

This was due to the usage of StyledMenu component that was listening for these letters (`e`, `d`, `r`) as part of screen navigation and accessibility features. However since the menu was still mounted under the Modal form, those letters were  blocked inside the form input fields.

## Changes

- Refactored  `CreateCollection` and `Menu` components to allow unmounting of the styled menu when the modal is in open state.

## Testing steps

- While logged in as curator - create or click on an existing dataset with write permissions, trigger the "more" menu then click Edit Details - try `e`, `d` or `r` as the leading letters


tihuan (approved):
Great solution, @kaloster 💡 🚀  Thanks for separating the show/hide logic of the modal from the menu and fixing the bug 👏 🏅

PR #7513, opened 2025-04-02
fix: Newsletter Subscribe VC-2285

kaloster:
## Reason for Change

- Newsletter `Subscribe` was not displaying success message upon signup

## Changes

- The script was checking for a hidden success message returned from HubSpot - since the copy of that message has changed by the provider over time, the logic that displays the success message to the user was skipped
- Updated the logic to look for `submitted-message` class name in the hidden form node upon submission instead of looking for the hidden text, as it is less likely to change over time

## Testing steps

- Click subscribe from the bottom banner - and enter a valid email


tihuan (approved):
Amazing thanks for the fix, Ronen 💡 👏 !!

PR #7528, opened 2025-04-14
fix: e2e flakyness

kaloster:
## Reason for Change

- Address e2e failures in staging

## Changes

- Updated e2e runners to dedicated playwright instances
- Updated Playwright + package-lock
- Updated timeouts
- Updated DE e2e due to new PW version
- Updated filter test to escape panel


tihuan (approved):
Fanatic work, Ronen 🔥♥️ Thanks so much for resolving the issue ⚔️ ✨!!

PR #7537, opened 2025-04-17
fix: various e2e improvements

kaloster:
## Reason for Change

- Identify offending tests and de-flake

## Changes

- Seed db workflow - explicitly define python
- Collections
- CellGuide
- Logged in DOI tests
- Logged in Edit


rainandbare (approved):
Looks good when its green!

PR #2523, opened 2022-05-11
chore: Remove unused auth code

Bento007:
### Reviewers
**Functional:** @ebezzi  

**Readability:** @danieljhegeman 

---

## Changes
- remove unused auth code and related unit tests.

PR #7544, opened 2025-04-22
fix: Fix Developmental Filters and Download Modal

rainandbare:
## Reason for Change

- [#2523](https://czi.atlassian.net/browse/VC-2523)

## Changes

- download all links on initial load of download modal
- add curated parent developmental stage ancestors to all relevant datasets (not just empty ones)
- change label "RNA" to "Gene Matrix"
- Accounts for datasets with no organism key on Super curator


frontend/src/common/queries/filter.ts
```diff
@@ -1045,27 +1045,33 @@ function processUserCollectionResponse(
 function addDevelopmentalStageAncestors(
   datasetResponse: DatasetResponse
 ): string[] {
-  const hasNonUberonLifeStageOrganisms = datasetResponse.organism.some((obj) =>
-    NON_UBERON_LIFE_STAGE_ORGANISMS.includes(obj.ontology_term_id)
-  );
+  const hasNonUberonLifeStageOrganisms =
+    datasetResponse.organism &&
```
tihuan, line 1049: Super nit: maybe optional chaining here 😁? organism?.some()

tihuan (approved):
LGTM thanks so much! Just a non blocking feedback we can address later lol Thanks again!

PR #7566, opened 2025-05-15
feat: fiddle around with the schema migration + ingestion SFN's

joyceyan:
## Reason for Change

https://czi.atlassian.net/browse/VC-2376

## Changes

- i doubled `MaxConcurrency` from 10 to 20
- it looks like we're only implementing retry logic in the ingestion / migration SFN's when catching specific AWS errors: `"AWS.Batch.TooManyRequestsException", "Batch.BatchException", "Batch.AWSBatchException"` so these transient docker / container pull errors aren't being caught
- i updated everything in the schema migration SFN as well as the ingestion SFN to retry up to 3 times for any sort of error that occurs. this means that we might end up retrying for certain error types that we know shouldn't be transient. but i don't think our error reporting in these step functions is very well defined, so i don't think we could reasonably define a set of errors that we definitely wouldn't want to retry for.

## Testing steps

🚢 it and see what happens?

## Checklist 🛎️

n/a


Bento007 (changes requested):
For `ValidateAtac`, `CXG`, and `ValidateAnndata` they tend to run long if they run at all. For that reason we should avoid retrying if they run longer than a minute, because that most likely means it's an actual error, and not an AWS transient error. If that's not feasible then reduce the retry to 1 to avoid increasing costs on something that isn't going to pass.

Bento007 (commented):
Thinking about this a bit more, we should have a catch statement at the top that will not retry if we get a Cellxgene specific error or maybe `States.TaskFailed` will work. You should be able to catch any exceptions we throw from the container in the AWS SFN.

joyceyan:
@Bento007 Do we know what type of error `State` is reported for the AWS / Docker transient errors? I can't find the Batch logs from the last schema migration, so I'm not sure exactly how these types of errors get reported.

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -224,11 +224,15 @@ resource aws_sfn_state_machine sfn_schema_migration {
       },
       "Next": "SpanCollections",
       "Retry": [ {
-          "ErrorEquals": ["AWS.Batch.TooManyRequestsException", "Batch.BatchException", "Batch.AWSBatchException"],
+          "ErrorEquals": ["States.TaskFailed"],
+          "IntervalSeconds": 30,
+          "MaxAttempts": 1
+      }, {
+          "ErrorEquals": ["States.ALL"],
```
Bento007, line 231: The only error this will catch is `State.Timeout` according to the AWS docs. Is that what is intended?

Bento007 (commented):
The errors defined in catch are resolved in order. Given that, TaskFailed error should go at the bottom of the Catches, and continue to catch `"AWS.Batch.TooManyRequestsException", "Batch.BatchException", "Batch.AWSBatchException"` before the TaskFailed. If you still want to catch the `Timeouts`, then add those after `TaskFailed`

.happy/terraform/modules/schema_migration/main.tf
joyceyan: Where are you seeing that this would only catch `States.Timeout`? I do see from the [AWS docs](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html) this note:
```
The States.ALL error type can't catch the States.DataLimitExceeded terminal error type and runtime error types. For more information about these error types, see [States.DataLimitExceeded](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html#error-data-limit-exceed) and [States.Runtime](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html#states-runtime-error).
```

Regardless, I think your suggestion to explicitly catch the AWS batch errors we were previously catching is a good one, so I'll update to do that.

Bento007: Because `States.TaskFailed` would catch every thing else and except for `States.Timeout`.

From [AWS docs](https://docs.aws.amazon.com/step-functions/latest/dg/concepts-error-handling.html)
>States.TaskFailed
A Task state failed during the execution. When used in a retry or catch, States.TaskFailed acts as a wildcard that matches any known error name except for States.Timeout.

.happy/terraform/modules/schema_migration/main.tf
```diff
@@ -227,6 +227,11 @@ resource aws_sfn_state_machine sfn_schema_migration {
           "ErrorEquals": ["States.TaskFailed"],
           "IntervalSeconds": 30,
           "MaxAttempts": 1
+      }, {
+          "ErrorEquals": ["AWS.Batch.TooManyRequestsException", "Batch.BatchException", "Batch.AWSBatchException"],
```
Bento007: These should be above `States.TaskFailed` since the first match error used.

> When one of these states reports an error and there's a Retry field, Step Functions scans through the retriers in the order listed in the array. When the error name appears in the value of a retrier's ErrorEquals field, the state machine makes retry attempts as defined in the Retry field.

joyceyan: it is, it looks like you're commenting on an outdated version of the file!

Bento007 (approved):

PR #7569, opened 2025-05-16
feat: ATAC processor

kaloster:
## Reason for Change

  This PR adds ATAC-seq fragment file processing capabilities to the CxG processing pipeline. The
  main goal is to add a coverage array in TileDB for Explorer's ATAC-seq datasets by implementing an
  `ATACDataProcessor` that handles fragment file processing and read normalization.

## Changes

  - New ATAC processing step in CxG conversion workflow
  - Fragment file processing utilities with unit tests coverage
  - Modified core processing files (process_cxg.py, h5ad_data_file.py, cxg_generation_utils.py,
  process.py)

  The feature enables the Multiome Coverage plot functionality in Explorer and has been tested on multiple
  conversion cases.

## Testing steps

- Processing unit tests should all pass
- Ran full conversion for several cases and tested in Explorer against Multiome (Coverage plot) feature

PR #7619, opened 2025-07-18
fix: ATAC Process updates

kaloster:
## Reason for Change

- https://github.com/chanzuckerberg/single-cell-data-portal/pull/7569
- This iterates on the previous implementation where `fragment_artifact_id` is being pulled from the database/artifacts

## Changes

-  CXG step now queries database directly for fragment artifacts instead of relying on parameter passing
-  Eliminated fragment_artifact_id from process() method signatures and environment variables
-  Updated ATAC validation: No longer returns fragment_artifact_id since it's not needed
-  Cleaned up infrastructure: Removed unused FRAGMENT_ARTIFACT_ID from Terraform Step Function config

### DIFF
--- a/backend/layers/processing/process_cxg.py
+++ b/backend/layers/processing/process_cxg.py
@@ -41,7 +41,6 @@ class ProcessCxg(ProcessingLogic):
         dataset_version_id: DatasetVersionId,
         artifact_bucket: str,
         cellxgene_bucket: str,
-        fragment_artifact_id: str = None,
         is_reprocess=False,
     ):
         """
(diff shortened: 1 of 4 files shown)


backend/layers/processing/process_cxg.py
```diff
@@ -70,48 +68,21 @@ def process(
...
         fragment_file_path = None
-        self.logger.info(f"Fragment artifact ID: {fragment_artifact_id}")
-        if fragment_artifact_id:
-            fragment_file_path = self.get_fragment_file_s3_url(fragment_artifact_id, artifact_bucket, current_artifacts)
+        fragment_artifacts = [
+            artifact for artifact in current_artifacts if artifact.type == DatasetArtifactType.ATAC_FRAGMENT
+        ]
+        if fragment_artifacts:
+            fragment_artifact_id = fragment_artifacts[0].id
+            fragment_file_path = f"s3://{artifact_bucket}/{fragment_artifact_id}"
```
Bento007, line 78: if you are pulling from the DB then you can use the `DatasetArtifactType.uri`. That creates one less place we are building the uri string.

Bento007 (commented):
one

Bento007 (approved):

PR #7621, opened 2025-07-22
feat: Memory optimization for CXG Conversion

kaloster:
## Reason for Change

- This is to resolve OOM reported in tiledb consolidate, and ATAC concatenation 

## Changes

  1. ATAC Processing Memory Optimization
    - Replaced memory-intensive pd.concat() with hybrid streaming approach using pre-allocated numpy arrays
    - Reduced peak memory from 4.7GB to 2.8GB for large datasets (35.5M records)
  2. TileDB Consolidation Memory Optimization
    - Added memory-optimized consolidation with dynamic buffer sizing based on available memory
    - Implemented chunked consolidation for low-memory environments (<8GB)
    - Reduced consolidation memory usage from 5.7GB to 2.1GB (63% reduction)
  3. Code Cleanup
    - Removed unused methods: _compute_cell_type_totals() and _write_coverage_to_tiledb()
    - Updated corresponding tests
    - Cleaned up redundant code paths


Bento007 (changes requested):
This is great work. Thank you.

Bento007 (approved):

PR #7623, opened 2025-07-28
fix: use chunked write to tiledb to avoid OOM

Bento007:
## Reason for Change

- vc-3630
- fixes an OOM error when processing ATAC fragments into CXG using a function generator instead of preallocating an array. This is slower but will be less memory intensive.

### Refactoring for memory efficiency:

* [`backend/layers/processing/utils/atac.py`](diffhunk://#diff-76e55c365137fb5f4b085debb6488c1800836a335a3becc92444b273973c1caeL389-R408): Replaced pre-allocated arrays for storing intermediate data with a generator function (`generate_coverage_records`) that dynamically yields coverage records. This eliminates the need for large in-memory arrays.

* [`backend/layers/processing/utils/atac.py`](diffhunk://#diff-76e55c365137fb5f4b085debb6488c1800836a335a3becc92444b273973c1caeL389-R408): Introduced `itertools.islice` to process records in chunks, reducing memory usage by writing smaller fragments to TileDB iteratively.

### Code improvements:

* [`backend/layers/processing/utils/atac.py`](diffhunk://#diff-76e55c365137fb5f4b085debb6488c1800836a335a3becc92444b273973c1caeR1): Added the `itertools` module to support the new generator-based chunking logic.

### DIFF
--- a/backend/layers/processing/utils/atac.py
+++ b/backend/layers/processing/utils/atac.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


tihuan (approved):
Ooh awesome! Generator functions FTW 👏 Thanks for cutting the memory usage significantly!

PR #7625, opened 2025-08-06
fix: ATAC processing memory optimization VC-3630

kaloster:
## Reason for Change

- VC-3630

## Changes

-   Dynamically limits process count based on fragment file size to prevent OOM
-   Eliminated redundant double-processing of fragment data, reducing peak memory
-   Stream 500K record chunks on-demand instead of pre-allocating arrays for 200M+ records, preventing memory exhaustion
-   Added missing TileDB consolidation + vacuum for ATAC coverage arrays, reducing multiple fragments to single fragment for optimal read performance

### DIFF
--- a/backend/layers/processing/utils/atac.py
+++ b/backend/layers/processing/utils/atac.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import os
 import tempfile
@@ -9,6 +10,7 @@ import boto3
 import cellxgene_schema.atac_seq as atac_seq
 import numpy as np
 import pandas as pd
+import psutil
 import pysam
 import tiledb
 from tqdm import tqdm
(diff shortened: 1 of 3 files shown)


backend/layers/processing/utils/atac.py
```diff
@@ -339,38 +340,20 @@ def _process_chromosomes_in_batches(
             all_found_cells.update(batch_found_cells)
 
             if batch_coverage_aggregator:
+                # Compute totals incrementally to avoid double-pass
                 for key, count in batch_coverage_aggregator.items():
+                    chrom, bin_id, cell_type = key
```
Bento007, line 345: change to `_, _, cell_type = key`

backend/layers/processing/utils/atac.py
```diff
@@ -381,53 +364,77 @@ def _stream_coverage_chunks_to_tiledb(
         coverage_aggregator: defaultdict,
         global_cell_type_totals: Dict[str, int],
         array_name: str,
-        chunk_size: int = 10_000_000,
+        chunk_size: int = 500000,
```
Bento007, line 367: The curators were saying the cxg was taking longer after I made revert some of your changes, and switched back to 50_000. So I increased it to 10_000_000, to reduce the number of calls to `_write_chunks_generator_to_tiledb`. The goal was to speed up cxg conversion.

I estimated 10_000_000 by estimating the size of each record:  
- chrom (int, 4 bytes)
- bin_id (int, 4 bytes)
- cell_type (str, variable, estimate avg 20 bytes)
- count (int, 4 bytes)
- total_coverage (int, 4 bytes)
- normalized_coverage (float, 4 bytes)

Estimated per-record size: 4 + 4 + 20 + 4 + 4 + 4 = 40 bytes 

For a chunk of N records: 
- Size in bytes: N * 40
This will output approximately 0.00186 GB for 50,000 records.

I felt like there was enough space to increase this number.

backend/layers/processing/utils/atac.py
```diff
@@ -381,53 +364,77 @@ def _stream_coverage_chunks_to_tiledb(
...
+                    "bin_id": bin_id,
+                    "cell_type": cell_type,
+                    "coverage": count,
+                    "total_coverage": total_coverage,
+                    "normalized_coverage": normalized_coverage,
+                }
+            )
+
+            if len(current_chunk) >= chunk_size:
+                yield current_chunk
```
Bento007, line 402: nit: I'm curious why you prefer this approach rather than using `itertools.islice`

Bento007 (changes requested):
These memory sizes ex: 28GB, we should define that based on the memory availible to the machine. I'm worried hardcoding will result in ,matinance issues down the line if the machine size changes.

backend/layers/processing/utils/atac.py
kaloster: Added dynamic chunk size calculation

backend/layers/processing/utils/atac.py
kaloster: Not so much about performance, it was more about my own readability and following existing patterns - reverted back to itertools.isslice

PR #7630, opened 2025-08-13
feat: remove parallelization of ValidateAtac and ValidateAnndata

joyceyan:
## Reason for Change

- one of the outputs of this ticket https://czi.atlassian.net/browse/VC-3817

## Changes

removes the parallelization of ValidateAnndata and ValidateAtac, and instead runs these sequentially. we think that there could be some memory resource contention when running these two jobs in parallel, and the increased memory allocation to ValidateAtac could be causing OOM issues with ValidateAnndata.

i looked more into seeing if we could skip the ValidateAtac job and allocating the resources for that based on some output from ValidateAnndata, but it seems like that requires deploying some sort of lambda that i think is maybe not worth it at the time. hopefully with running these jobs serially the unnecessary allocation of resources to ValidateAtac won't pose an issue for us.

i also increased the f-test timeout to 30 mins

## Testing steps

i uploaded two test datasets (one with atac, one without) to rdev. both show this graph:
image

`example_valid.h5ad` is at `dp-rdev-pr-7630-sfn/67df266f-3dea-4dc8-8d45-0ff4bed0b9da_1755194921`
`atac.h5ad` is at `dp-rdev-pr-7630-sfn/ec7de2ad-c708-4caa-a72b-0a9632226dfa_1755194973`

## Checklist 🛎️

### DIFF
--- a/backend/layers/processing/process_cxg.py
+++ b/backend/layers/processing/process_cxg.py
@@ -79,7 +79,12 @@ class ProcessCxg(ProcessingLogic):
 
         # Convert the labeled dataset to CXG and upload it to the cellxgene bucket
         self.process_cxg(
-            labeled_h5ad_filename, dataset_version_id, cellxgene_bucket, fragment_file_path, current_artifacts
+            labeled_h5ad_filename,
+            dataset_version_id,
+            cellxgene_bucket,
+            fragment_file_path,
+            current_artifacts,
+            is_reprocess,
         )
 
     @logit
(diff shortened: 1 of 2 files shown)


.happy/terraform/modules/sfn/main.tf
```diff
@@ -233,8 +232,7 @@ resource "aws_sfn_state_machine" "state_machine" {
           "BackoffRate": 5
       }, {
           "ErrorEquals": ["States.TaskFailed"],
-          "IntervalSeconds": 30,
-          "MaxAttempts": 1
+          "MaxAttempts": 0
```
nayib-jose-gloria, line 235: I was thinking we should leave the auto-retry until we go a while without issues or until it becomes a problem that we're retrying mostly jobs that keep failing

joyceyan: good catch - this wasn't intentional and was a mistake made while rebasing 😅

nayib-jose-gloria (commented):
i think we may need to increase the timeout on the functional tests now that this isn't happening in parallel. the failures are timeout errors (10+ min) but looking in aws console the uploads actually are succeeding, they're just taking longer now that the steps aren't in parallel

nayib-jose-gloria (approved):

PR #7634, opened 2025-08-15
fix: Revise delete_objects

kaloster:
## Reason for Change

- Changes to S3 delete_files Method

## Changes

- Blocks deletion of root-level files or files with prefixes <8 characters
- **Added large batch warnings**: Logs warning for deletions >10,000 files
- All deletion operations now logged with bucket name, key count, and sample keys
- Raises IllegalS3RecursiveDelete for dangerous key patterns
- Requires proper / separated paths with sufficient prefixes
- Added tests for dangerous key blocking and safe key allowance

## Testing steps

- Unit tests should all pass for processing
- Test sfn in rdev and staging prior to deploying


Bento007 (approved):
thank you for adding those guard rails

PR #7648, opened 2025-09-17
fix: improve error handling and add publishing tests for collections

Bento007:
## Reason for Change

- #TICKET_NUMBER

## Changes

- add
- remove
- modify

## Testing steps

## Checklist 🛎️

### DIFF
--- a/backend/portal/api/portal_api.py
+++ b/backend/portal/api/portal_api.py
@@ -601,8 +601,8 @@ def publish_post(collection_id: str, body: object, token_info: dict):
 
     try:
         get_business_logic().publish_collection_version(version.version_id, data_submission_policy_version)
-    except CollectionPublishException:
-        raise ConflictException(detail="The collection must have a least one dataset.") from None
+    except CollectionPublishException as e:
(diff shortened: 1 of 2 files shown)


nayib-jose-gloria (commented):
One of the error checks for CollectionPublishException is that all datasets must be of the same schema version in order to publish (this is a requirement from product) https://github.com/chanzuckerberg/single-cell-data-portal/blob/7eaca2cf901da49fd28d88d938ec1fa6497faa4b/backend/layers/business/business.py#L1171

cellxgene-schema package with the atac fixes was released as version 6.2.0. We determine what schema_version is stamped based on the major and minor package versions, so new uploads are being minted as schema 6.2.0. Since there was no migration associated with this package update, we have mixed datasets, some still 6.0.0, potentially some with 6.1.0. We can observe these mixed sets of dataset versions on the collection that is failing to publish. 

assuming product does not want to change the requirement re: mixed versions co-existing, the simplest resolution is probably to release another patch of cellxgene-schema that does not couple it with the package version, and bring it back to 6.0.0. The current "6.2.0" uploads can be re-triggered.

PR #7667, opened 2025-10-17
fix: flip the logic so changed fragment takes precedence

Bento007:
## Reason for Change

We were copying the fragments when we should have been replacing it with a new one. The old logic was confusing and also wrong. The old logic was copy the artifact if:
1. the fragment has not changed
2. the fragment is already ingested and we aren’t in a migration.
Otherwise we create a new fragment.

The part that was causing issues is the already ingested part(ignore the not `in_migration` because it always true in this scenario). Now when we are trying to run deduplication, we always want to create a new artifact, so we set `fragment_changed = True` . With that set you’d assume we’d create a new artifact, but that’s not what was happening. `fragment_changed is False`  would resolve to `false `as we expect, but `self.business_logic.is_already_ingested(manifest.atac_fragment)` would resolve to true which would result in use copying the original fragment file again. The new logic simplifies things; Upload the fragment if:
1. The fragment has changed (new content), OR
2. The fragment is not already ingested (we don’t have it yet)
Otherwise, copy the existing artifact reference (unchanged AND already ingested)

## Changes
- the the original fragment was still copied. Flipping the logic so `if fragment_changed is True`, we upload a new fragment artifact.

### DIFF
--- a/backend/layers/processing/process_validate_atac.py
+++ b/backend/layers/processing/process_validate_atac.py
@@ -203,20 +203,11 @@ class ProcessValidateATAC(ProcessingLogic):
             else:
                 fragment_changed = True
 
-        # fragment file to avoid uploading the same file multiple times
-        # if the fragment file is unchanged or the fragment file is already ingested, use the old fragment.
-        if fragment_changed is False or (
-            self.business_logic.is_already_ingested(manifest.atac_fragment) and not in_migration
-        ):
-            # get the artifact id of the old fragment, and add it to the new dataset
-            artifact_name = str(manifest.atac_fragment).split("/")[-1]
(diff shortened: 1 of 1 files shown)


tests/unit/processing/test_process_validate_atac.py
```diff
@@ -454,6 +455,9 @@ def test_deduplication_always_uploads_even_if_already_ingested(
         )
         process_validate_atac.schema_validator.deduplicate_fragments.return_value = "test_dedup.tsv.bgz"
         process_validate_atac.hash_file = Mock()
+        # Mock is_already_ingested to return True to properly test the scenario where
```
Bento007:
```suggestion
        # Mock is_already_ingested to return True to test the scenario where
```

PR #7670, opened 2025-10-17
fix: replace raise_for_status with custom assertStatusCode for API re…

Bento007:
…sponse validation

This makes the error response more informative and easier to debug.


tests/functional/backend/utils.py
```diff
@@ -271,5 +271,5 @@ def get_curation_api_access_token(session, api_url, config) -> str:
         f"{api_url}/curation/v1/auth/token",
         headers={"x-api-key": config.super_curator_api_key},
     )
-    response.raise_for_status()
+    assertStatusCode(requests.codes.ok, response)
```
Bento007:
```suggestion
    assertStatusCode(201, response)
```

PR #7687, opened 2025-11-14
fix: set Cache-Control for files larger than 50GB in S3 uploads

Bento007:
https://czi.atlassian.net/browse/VC-4788


kliu-czi (approved):

Bento007:
Test and this solution will not work for us. See the attach jira ticket in description

